PR #4, opened 2015-09-04
fixed some safety and potential perf issues

zeliard:
removed a type conversion warning to make buildable in MSVC
removed unused variables: stringstream's base c'tor is not cheap
fixed wrong usages on calling Windows APIs


JonathanHenson:
please include your version of visual studio if you had a compiler error so we know what to test against.

JonathanHenson:
Thank you for your generous work on helping to improve this for the community. Please see my posted comments to each diff. Feel free to reply directly to those if you have a better idea (especially #2). 

I will try and get the things that can be exported today out to you today. Be aware that we have an internal repository that we export, so we will not be merging this directly, but your work is very much appreciated.

zeliard:
Thank you for a quick and kind reply :+1:

JonathanHenson:
We've fixed the signed/unsigned comparison and pushed the change.

PR #11, opened 2015-09-08
Fixed build errors and failed unit tests with MSVC 2015

bmildner:
Tried to build the sdk on Win7 x64 with MSVC 2015 (x86 + x64) but hit several compilation errors, warnings and one broken unit test!

Please note that i did not try to build it with my changes on Linux/MacOS with gcc/clang!

### DIFF
--- a/aws-cpp-sdk-core/include/aws/core/utils/memory/stl/AWSFunction.h
+++ b/aws-cpp-sdk-core/include/aws/core/utils/memory/stl/AWSFunction.h
@@ -41,9 +41,9 @@ std::function< F > BuildFunction(const std::function< F > &f)
 template<typename F>
 std::function< F > BuildFunction(std::function< F > &&f) 
 {
-    // TODO: there seems to be no move c'tor that also takes an allocator !?
+    // TODO: there seems to be no move c'tor that also takes an allocator, do a copy instead of a move !
     //return std::function< F >( std::move(f), Aws::Allocator<void>() );
-    return std::function< F >(std::move(f));
+    return std::function< F >(std::allocator_arg_t(), Aws::Allocator<void>(), f);
 }
 
 } // namespace Aws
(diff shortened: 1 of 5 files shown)


bmildner:
You are right, bretambrose!
Unfortunately my change in ComputeIdentityFilePath() went sideways ... it breaks the test when built with MSVC 2013!
Apparently tmpnam_s() returns an absolute path (like "C:\Users\You\whatever\Temp\temp.file") when built with MSVC 2015 but only a filename with a "\" in front  (like "\temp.file") when built with MSVC 2013 :( 
At least it does in my environment ...

I use this command lines:
x86: 
cmake c:\Users\Berti\Documents\aws-sdk-cpp -G "Visual Studio 14 2015"
cmake c:\Users\Berti\Documents\aws-sdk-cpp -G "Visual Studio 12 2013"

x64: 
cmake c:\Users\Berti\Documents\aws-sdk-cpp -G "Visual Studio 14 2015 Win64"
cmake c:\Users\Berti\Documents\aws-sdk-cpp -G "Visual Studio 12 2013 Win64"

I just committed a runtime fix for this in my fork. 
The way i read the MSDN description of tmpnam_s() one has to be able to handle both although the documentation does not note any changes between MSVC 2013 and 2015 that i could spot ...

bmildner:
Well, then, size_t is the correct type, yes.

The two renamed local variables in jsoncpp.cpp and URI.cpp where warnings, but treat warnings as errors is enabled on the MSVC project files created by cmake!

JonathanHenson:
this change has been merged. Thank you for your help!

bmildner:
@bretambrose: 
good find on CRT changes in MSVC 2014 CTP1!

BTW: MSVC 2014 got renamed to 2015 before release, so only mentioning 2014 in the comment my be confusing.

One more thing: 
My cmake (3.3.1) help states that "Visual Studio 12 2013 [Win64]" is the generator name for MSVC 2013 but your README.md suggests "Visual Studio 12 [Win64]"!
Maybe not really an issue but the difference caught my eye!?

PR #13, opened 2015-09-17
Amazon is wrong namespace name, Replaced to Aws

d9magai:
I think `Amazon::` is wrong namespace name.Isnt it?


bretambrose:
You are correct!  The names have evolved over time and that example was written before things were finalized.  We will apply your change internally and it will appear in our next published version.  Thanks!

d9magai:
I'm waiting next published version.
Thank you!

PR #18, opened 2015-09-30
erase useless line

wates:
this variable is always null


JonathanHenson:
This was pushed a while back. Thanks for your contribution.

PR #29, opened 2015-10-06
Fix S3 metadata for PutObjectRequest

sdoh55:
stringstream wasn't cleared for subsequent metadata keys


JonathanHenson:
Good catch! We'll integrate this as soon as possible.

JonathanHenson:
I just pushed an update for this. Thanks for the pull request. Can you verify everything works for you?

sdoh55:
Yep, this fixed the s3 metadata for me.

PR #32, opened 2015-10-06
Back off the aggressive cmake version requirements

dhh1128:
The CMakeLists.txt file requires a very recent version of CMake. This is fine on Windows and Mac, but most linux distros are multiple revisions behind the specified version. For example, Ubuntu 14.04 LTS distributes CMake 2.8. Although it is possible to install a newer version of CMake there, there is really no benefit to doing so; nothing in our cmake files appears to need the newer build tool. The build files created by older cmakes work flawlessly for me on Linux.

One caveat with this pull request is that I have only tested the resulting build on Ubuntu 14. If we truly need a newer version of CMake on other platforms, my change might break something.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -13,7 +13,7 @@
 # permissions and limitations under the License.
 #
 
-cmake_minimum_required (VERSION 3.2.3)
+cmake_minimum_required (VERSION 2.8)
 
 # Cmake invocation variables:
 #   CUSTOM_MEMORY_MANAGEMENT - if set to 1, generates the sdk project files with custom memory management enabled, otherwise disables it


JonathanHenson:
This is here because of security requirements in our android build. I'm not
sure how long we'll  keep the android build in cmake in the long term
anyways, but feel free to reduce it on your builds.

bretambrose:
The higher Cmake version requirement is (indirectly) due to some changes strongly suggested by our security review team, in particular to use stronger verification on the ExternalProject_Adds of the android build.  Only the version bump has gotten pushed through to the main repo yet, so from your perspective it doesn't make sense right now, but eventually it will.  As a compromise solution, assuming it works, I will make the version requirements conditional on whether or not android is the target platform.

guillermomuntaner:
@bretambrose 
+1 to cmake conditional requirement. 
If it is only required by Android, you shouldn't force most of your developers to update Cmake. Updating Cmake is not trivial in some enterprise OS distributions.

JonathanHenson:
We backed off to 3.1.2. Can you give it a try?

bretambrose:
Hold off on that.  It looks like the new CMakeLists.txt did not make the export process.  We should be able to address that shortly.

vaintroub:
Affects me. I'm not building  for Android, but I need to build for a variety of Linuxes, FreeBSD, Windows etc

JonathanHenson:
This should have been resolved. Please reopen if you have ongoing issues.

vaintroub:
this is resolved where? https://github.com/aws/aws-sdk-cpp/blob/master/CMakeLists.txt#L19 still has 3.1.2

Bu11etmagnet:
Look at line 64.
Isn't line 19 redundant?

vaintroub:
It is, and it is what requires a very new cmake. i.e line 70
cmake_minimum_required (VERSION 2.8)

has no effect given 
cmake_minimum_required (VERSION 3.1.2)
elsewhere

PR #61, opened 2015-12-14
Changed the type of SQS message's attribute

skobaken7:
According to the AWS library of Java, SQS message's attributes should be String.
http://docs.aws.amazon.com/AWSJavaSDK/latest/javadoc/com/amazonaws/services/sqs/model/Message.html
For example, I cant' get "ApproximateReceiveCount" by the current implementation.


JonathanHenson:
The AWS library for Java doesn't use enums, which is why they specified a String. We need to look into why ApproximateReceiveCount is not a member of the enum.

JonathanHenson:
This was addressed in c00923192cc55a640394b8eebf08b0bb6c3a9a30

PR #62, opened 2015-12-15
Fix crash in CurlHttpClient when request times out.

pokey909:
Integration test was crashing with a segfault when the connection timed out.
Reason was that the response was nulled in this case but still dereferenced to flush it.


JonathanHenson:
This was pulled over in my latest update. Feel free to consider this pull request merged. Due to the export process that was in process (until I deprecated it a few minutes ago), I didn't actually pull it. I will next time though. Thanks for your work on this.

PR #68, opened 2016-01-07
AWS Seoul region support

zeliard:
[add] ap-northeast-2 region

### DIFF
--- a/README.md
+++ b/README.md
@@ -460,7 +460,25 @@ getObjectRequest.SetResponseStreamFactory([](){ return Aws::New<Aws::FStream>( A
 auto getObjectOutcome = s3Client->GetObject(getObjectRequest);
 ```
 
+###Contributing Back
+*Please Do!
 
+#####Guidlines
+* Don't make changes to generated clients directly. Make your changes in the generator. Changes to Core, Scripts, and High-Level interfaces are fine directly in the code.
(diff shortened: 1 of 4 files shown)


JonathanHenson:
Wow..... that was quick.... You need to update the Region.h and Region.cpp
file in core and then update the endpoint computation in the generator.
Most of those files you manually updated are generated, so the fix needs to
happen there.

Thanks for your work on this.

zeliard:
Yep! just fixed RegionEndpointMapper.java
Thanks for adding AWS Seoud region :+1:

JonathanHenson:
can you shoot me a pull request without the updates to the clients? I just need the client generator update and the region.h and region.cpp updates.

JonathanHenson:
b.t.w. you can use endpoint override and authenticationRegion in the client config anytime you need a region we haven't explicitly called out.

zeliard:
OK, I fixed this PR in the contribution guidelines.
3 files (region.h, region,cpp, and regionendpointmapper.java) are changed eventually.

JonathanHenson:
One last thing. Could you rebase this into one commit?

zeliard:
done. 
(but your last commit (contribution guidelines) has been included while squashing my commits)
anyway, thanks for your kindness.

Issue #70, opened 2016-01-07
EC2 DescribeInstances API doesn't work (and likely many others)


steveniemitz:
I was attempting to use EC2Client::DescribeInstances, but never got any reservations back.

Digging into the code, it looks like the generated parsers are incorrect.  DescribeInstancesResult looks for a "Reservations" XML node, however the actual XML response is "reservationsSet".  It looks like the generated code isn't looking for the locationName as specified in the API JSON descriptors, and instead only using the member key.

steveniemitz:
It looks like this patch to the code generator template was enough to fix most places:

```
diff --git a/code-generation/generator/src/main/resources/com/amazonaws/util/awsclientgenerator/velocity/cpp/xml/ModelClassMembersDeserializeXml.vm b/code-generation/generator/src/main/resources/com/amazonaws/util/awsclientgenerator/velocity/cpp/xml/ModelClassMembersDeserializeXml.vm
index 6208d31..b01a4c7 100644
--- a/code-generation/generator/src/main/resources/com/amazonaws/util/awsclientgenerator/velocity/cpp/xml/ModelClassMembersDeserializeXml.vm
+++ b/code-generation/generator/src/main/resources/com/amazonaws/util/awsclientgenerator/velocity/cpp/xml/ModelClassMembersDeserializeXml.vm
@@ -29,7 +29,11 @@
 #set($listVarName = $CppViewHelper.computeVariableName($memberName) + "Member")
 #set($listMemberName = "member")
 #end##location specified in model
+#if($member.locationName)##location specified
+    XmlNode ${lowerCaseVarName}Node = resultNode.FirstChild("${member.locationName}");
+#else##no location specified
     XmlNode ${lowerCaseVarName}Node = resultNode.FirstChild("${memberName}");
+#end##
```

I'll submit a PR soon.  Based on the diff, it looks like this mostly impacted EC2 APIs, but there were some changes in S3, SNS, iam, and email as well.

steveniemitz:
Looks like there's a similar problem on the request side as well.  For instance the InstanceIds filter is serialized incorrectly.

JonathanHenson:
thanks for your work on this, after you submit the pull request, I'll rerun the integration tests and make sure we didn't break anything. Ec2 is a special snowflake when it comes to protocol and is unfortunately an edge case we haven't tested too much.

steveniemitz:
[...]

PR #72, opened 2016-01-07
Fixed codegen for lists with locationNames.

steveniemitz:
This primarily impacts the EC2 APIs.
See issue #70

This fixes enough of the API to get it working.  There are still a few places I found that don't work, for example the Filter on DescribeInstances still doesn't serialize correctly, but InstanceIds does now.


JonathanHenson:
ran full test suite, everything looks good. merging now.

PR #76, opened 2016-01-11
Add ability to set metadata for object uploads in transfer library.

kahkeng:
Adding the ability for users of TransferClient to set metadata of objects they are uploading. Let me know if there's anything that needs to change with the new public methods I added here.

### DIFF
--- a/aws-cpp-sdk-transfer-tests/TransferTests.cpp
+++ b/aws-cpp-sdk-transfer-tests/TransferTests.cpp
@@ -1188,5 +1188,101 @@ TEST_F(TransferTests, ScopeTests)
     }
 }
 
+// Single part upload with metadata specified
+TEST_F(TransferTests, SinglePartUploadWithMetadataTest)
+{
+    if (EmptyBucket(GetTestBucketName()))
+    {
+        WaitForBucketToEmpty(GetTestBucketName());
+    }
+    GetObjectRequest getObjectRequest;
+    getObjectRequest.SetBucket(GetTestBucketName());
+    getObjectRequest.SetKey(TEST_FILE_NAME);
+
+    GetObjectOutcome getObjectOutcome = m_s3Client->GetObject(getObjectRequest);
+    EXPECT_FALSE(getObjectOutcome.IsSuccess());
+
+    Aws::Map<Aws::String, Aws::String> metadata;
+    metadata["key1"] = "val1";
+    metadata["key2"] = "val2";
+    const bool cCreateBucket = true;
+    const bool cConsistencyChecks = false;
+    std::shared_ptr<UploadFileRequest> requestPtr = m_transferClient->UploadFile(TEST_FILE_NAME, GetTestBucketName(), "", "", metadata, cCreateBucket, cConsistencyChecks);
+
+    ASSERT_FALSE(requestPtr->IsDone());
+
+    WaitForUploadAndUpdate(requestPtr, 100.0f);
+
+    ASSERT_TRUE(requestPtr->IsDone());
+
+    ASSERT_TRUE(requestPtr->CompletedSuccessfully());
+
+    WaitForObjectToPropagate(GetTestBucketName(), TEST_FILE_NAME);
+
+    // Check the metadata matches
+    HeadObjectRequest headObjectRequest;
+    headObjectRequest.SetBucket(GetTestBucketName());
+    headObjectRequest.SetKey(TEST_FILE_NAME);
+
+    HeadObjectOutcome headObjectOutcome = m_s3Client->HeadObject(headObjectRequest);
+    ASSERT_TRUE(headObjectOutcome.IsSuccess());
+
+    Aws::Map<Aws::String, Aws::String> headObjectMetadata = headObjectOutcome.GetResult().GetMetadata();
+    ASSERT_EQ(metadata.size(), headObjectMetadata.size());
+    ASSERT_EQ(metadata["key1"], headObjectMetadata["key1"]);
+    ASSERT_EQ(metadata["key2"], headObjectMetadata["key2"]);
+
+}
+
+// Multipart upload with metadata specified
+TEST_F(TransferTests, MultipartUploadWithMetadataTest)
+{
+    if (EmptyBucket(GetTestBucketName()))
+    {
+        WaitForBucketToEmpty(GetTestBucketName());
+    }
+
+    GetObjectRequest getObjectRequest;
+    getObjectRequest.SetBucket(GetTestBucketName());
+    getObjectRequest.SetKey(MEDIUM_FILE_KEY);
+
+    GetObjectOutcome getObjectOutcome = m_s3Client->GetObject(getObjectRequest);
+    EXPECT_FALSE(getObjectOutcome.IsSuccess());
+
+    ListMultipartUploadsRequest listMultipartRequest;
+
+    listMultipartRequest.SetBucket(GetTestBucketName());
+
+    Aws::Map<Aws::String, Aws::String> metadata;
+    metadata["key1"] = "val1";
+    metadata["key2"] = "val2";
+    const bool cCreateBucket = true;
+    const bool cConsistencyChecks = false;
+    std::shared_ptr<UploadFileRequest> requestPtr = m_transferClient->UploadFile(MEDIUM_TEST_FILE_NAME, GetTestBucketName(), MEDIUM_FILE_KEY, "", metadata, cCreateBucket, cConsistencyChecks);
+
+    ASSERT_FALSE(requestPtr->IsDone());
+
+    WaitForUploadAndUpdate(requestPtr, 100.0f);
+
+    ASSERT_TRUE(requestPtr->IsDone());
+    ASSERT_TRUE(requestPtr->CompletedSuccessfully());
+
+    WaitForObjectToPropagate(GetTestBucketName(), MEDIUM_FILE_KEY);
+
+    // Check the metadata matches
+    HeadObjectRequest headObjectRequest;
+    headObjectRequest.SetBucket(GetTestBucketName());
+    headObjectRequest.SetKey(MEDIUM_FILE_KEY);
+
+    HeadObjectOutcome headObjectOutcome = m_s3Client->HeadObject(headObjectRequest);
+    ASSERT_TRUE(headObjectOutcome.IsSuccess());
+
+    Aws::Map<Aws::String, Aws::String> headObjectMetadata = headObjectOutcome.GetResult().GetMetadata();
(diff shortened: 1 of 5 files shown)


aws-cpp-sdk-transfer/include/aws/transfer/UploadFileRequest.h
```diff
@@ -126,6 +126,11 @@ class AWS_TRANSFER_API UploadFileRequest : public S3FileRequest, public std::ena
     bool HasPassedGetObject() const { return m_getObjectPassed.load(); }
     bool HasPassedListObjects() const { return m_listObjectsPassed.load(); }
 
+    // A map of metadata to store with the object in S3.
```
JonathanHenson: I'd like for the map to be in the function call instead of in this object. This object does not contain any other s3 object specific data; it is a handle for callbacks and status checks etc...

Also, when you do so, be sure to handle r-values instead of just a copy.

JonathanHenson:
nice work on this, see my comment above.

aws-cpp-sdk-transfer/include/aws/transfer/UploadFileRequest.h
kahkeng: Ok that makes sense, thanks for the review. I was wondering if I should go that route but there are many other request properties available in a PutObjectRequest that are yet to be implemented here (cache control, content encoding, request payer, etc) and this would be an api breaking change each time unless we did function overloading for UploadFile. Would that be acceptable or we're ok changing the function signature each time to add a new property?

JonathanHenson: Two overloads should work. Leave the original, then add one with the map as
const ref, then another as an r-value.

kahkeng:
Ok I refactored to what I think you were suggesting, let me know if I misunderstood you or if anything needs to change. I understand that we don't want to use the UploadFileRequest object as a container for data. I'm just a little concerned that we will need to have many overloaded methods to transfer the various attributes to the underlying CreateMultipartUploadRequest and PutObjectRequest objects. I'm just doing it for metadata here but there are many more that aren't supported yet.

Thanks!

aws-cpp-sdk-transfer-tests/TransferTests.cpp
```diff
@@ -1188,5 +1188,101 @@ TEST_F(TransferTests, ScopeTests)
...
+    HeadObjectRequest headObjectRequest;
+    headObjectRequest.SetBucket(GetTestBucketName());
+    headObjectRequest.SetKey(TEST_FILE_NAME);
+
+    HeadObjectOutcome headObjectOutcome = m_s3Client->HeadObject(headObjectRequest);
+    ASSERT_TRUE(headObjectOutcome.IsSuccess());
+
+    Aws::Map<Aws::String, Aws::String> headObjectMetadata = headObjectOutcome.GetResult().GetMetadata();
+    ASSERT_EQ(2u, headObjectMetadata.size());
+    ASSERT_EQ("val1", headObjectMetadata["key1"]);
```
JonathanHenson: ASSERT_EQ(metadata["key1"], headObjectMetadata["key1"]);

aws-cpp-sdk-transfer-tests/TransferTests.cpp
```diff
@@ -1188,5 +1188,101 @@ TEST_F(TransferTests, ScopeTests)
...
+    ASSERT_EQ("val2", headObjectMetadata["key2"]);
```
JonathanHenson: ASSERT_EQ(metadata["key1"], headObjectMetadata["key1"]);

aws-cpp-sdk-transfer-tests/TransferTests.cpp
```diff
@@ -1188,5 +1188,101 @@ TEST_F(TransferTests, ScopeTests)
...
+    HeadObjectRequest headObjectRequest;
+    headObjectRequest.SetBucket(GetTestBucketName());
+    headObjectRequest.SetKey(MEDIUM_FILE_KEY);
+
+    HeadObjectOutcome headObjectOutcome = m_s3Client->HeadObject(headObjectRequest);
+    ASSERT_TRUE(headObjectOutcome.IsSuccess());
+
+    Aws::Map<Aws::String, Aws::String> headObjectMetadata = headObjectOutcome.GetResult().GetMetadata();
+    ASSERT_EQ(2u, headObjectMetadata.size());
+    ASSERT_EQ("val1", headObjectMetadata["key1"]);
```
JonathanHenson: ditto here

aws-cpp-sdk-transfer/include/aws/transfer/UploadFileRequest.h
```diff
@@ -83,6 +83,22 @@ class AWS_TRANSFER_API UploadFileRequest : public S3FileRequest, public std::ena
...
+                      bool doConsistencyChecks);
+    UploadFileRequest(const Aws::String& fileName,
+                      const Aws::String& bucketName,
+                      const Aws::String& keyName,
+                      const Aws::String& contentType,
+                      Aws::Map<Aws::String, Aws::String>&& metadata,
+                      const std::shared_ptr<Aws::S3::S3Client>& s3Client,
+                      bool createBucket,
+                      bool doConsistencyChecks);
     ~UploadFileRequest();
```
JonathanHenson: This one is not your fault, I must have let it slip through a code-review internally, but the destructor needs to be virtual.

aws-cpp-sdk-transfer/include/aws/transfer/UploadFileRequest.h
```diff
@@ -255,6 +271,8 @@ class AWS_TRANSFER_API UploadFileRequest : public S3FileRequest, public std::ena
     Aws::List<std::shared_ptr<UploadBuffer> > m_buffersReady;
 
     Aws::String m_contentType;
+    Aws::Map<Aws::String, Aws::String> m_metadata;
+    bool m_metadataHasBeenSet;
```
JonathanHenson: you don't need this, you have m_metadata.size() > 0

aws-cpp-sdk-transfer/source/transfer/UploadFileRequest.cpp
```diff
@@ -105,6 +106,34 @@ m_headBucketRetries(0)
...
+                                     const Aws::String& keyName,
+                                     const Aws::String& contentType,
+                                     Aws::Map<Aws::String, Aws::String>&& metadata,
+                                     const std::shared_ptr<Aws::S3::S3Client>& s3Client,
+                                     bool createBucket,
+                                     bool doConsistencyChecks) :
+UploadFileRequest(fileName, bucketName, keyName, contentType, s3Client, createBucket, doConsistencyChecks)
+{
+    m_metadataHasBeenSet = true;
+    m_metadata = metadata;
```
JonathanHenson: Move this to the initialization list, and add a std::forward around it.

aws-cpp-sdk-transfer/source/transfer/UploadFileRequest.cpp
```diff
@@ -105,6 +106,34 @@ m_headBucketRetries(0)
...
+                                     const Aws::String& keyName,
+                                     const Aws::String& contentType,
+                                     const Aws::Map<Aws::String, Aws::String>& metadata,
+                                     const std::shared_ptr<Aws::S3::S3Client>& s3Client,
+                                     bool createBucket,
+                                     bool doConsistencyChecks) :
+UploadFileRequest(fileName, bucketName, keyName, contentType, s3Client, createBucket, doConsistencyChecks)
+{
+    m_metadataHasBeenSet = true;
+    m_metadata = metadata;
```
JonathanHenson: initilalization list

aws-cpp-sdk-transfer/source/transfer/UploadFileRequest.cpp
```diff
@@ -202,6 +231,10 @@ bool UploadFileRequest::CreateMultipartUpload()
     {
         createMultipartUploadRequest.SetContentType(m_contentType);
     }
+    if (m_metadataHasBeenSet)
```
JonathanHenson: if(m_metadata.size() > 0)

JonathanHenson:
Thanks for your work on this! Just a few thoughts.

kahkeng:
Thanks for the feedback, I think I addressed them all in this new commit. Regarding putting metadata into the initialization list, I had to make one of the new constructors the default constructor, otherwise, I'd get a `mem-initializer for 'metadata' follows constructor delegation` error. Also, I tried to read up more about rvalues/std::forward, hope I got its use correct here but let me know if it isn't. Thanks!

JonathanHenson:
This looks good. I'm going to pull down and run through the test suite. If everything works correctly, I'll merge the request.

Thanks for your work on this.

aws-cpp-sdk-transfer/source/transfer/TransferClient.cpp
```diff
@@ -74,6 +74,24 @@ std::shared_ptr<UploadFileRequest> TransferClient::UploadFile(const Aws::String&
...
+    auto request = Aws::MakeShared<UploadFileRequest>(ALLOCATION_TAG, fileName, bucketName, keyName, contentType, metadata, m_s3Client, createBucket, doConsistencyChecks);
+
+    UploadFileInternal(request);
+
+    return request;
+}
+
+std::shared_ptr<UploadFileRequest> TransferClient::UploadFile(const Aws::String& fileName, const Aws::String& bucketName, const Aws::String& keyName, const Aws::String& contentType, Aws::Map<Aws::String, Aws::String>&& metadata, bool createBucket, bool doConsistencyChecks)
+{
+    auto request = Aws::MakeShared<UploadFileRequest>(ALLOCATION_TAG, fileName, bucketName, keyName, contentType, metadata, m_s3Client, createBucket, doConsistencyChecks);
```
kahkeng: Actually, to confirm, I think I also need to wrap metadata here with std::forward? If so, I can adjust this patch.

kahkeng:
I've added one more instance of `std::forward` in TransferClient when creating the UploadFileRequest object with rvalue metadata, and changed my tests to have `cCreateBucket = true` so they can run individually, and replaced the old commit with this new one. Hope it's all set now.

kahkeng:
Updated the diff to use std::move instead of std::forward since I've learnt the former is meant for rvalue references which is the case here, and the latter, although it works here too, seems meant more for universal references. I've also checked via print statements that the rvalue reference constructor of UploadFileRequest gets called when I pass in an rvalue metadata into the TransferClient uploadFile method. Please let me know if it's fine to merge now, thanks!

PR #85, opened 2016-01-21
Allow building EC2 API in linux

steveniemitz:
Open for comments, on clang with -j4 the build is pretty fast.


JonathanHenson:
You kept android in the exclusion, this should be fine.

JonathanHenson:
So we rely on jump-table optimization heavily in enum parsing. I'm a bit worried that -fPIC will end up turning off the optimization on a few platforms.

PR #114, opened 2016-02-22
Fix segfault in curl due to SIGALRM

edolstra:
The curl options set by CurlHandleContainer::CheckAndGrowPool() were
being clobbered by CurlHttpClient::MakeRequest()'s call to
curl_easy_reset(). In particular, this made CURLOPT_NOSIGNAL revert to
0, causing random crashes in multithreaded applications.

The fix is to have CurlHandleContainer::AcquireCurlHandle() do the
curl_easy_reset() and option setting, ensuring a consistent handle
state.

### DIFF
--- a/aws-cpp-sdk-core/source/http/curl/CurlHandleContainer.cpp
+++ b/aws-cpp-sdk-core/source/http/curl/CurlHandleContainer.cpp
@@ -131,7 +131,7 @@ CurlHandleContainer::CurlHandleContainer(unsigned maxSize, long requestTimeout, 
                 m_maxPoolSize(maxSize), m_requestTimeout(requestTimeout), m_connectTimeout(connectTimeout),
                 m_poolSize(0)
 {
-    AWS_LOGSTREAM_INFO(CurlTag, "Intializing CurlHandleContainer with size " << maxSize);
+    AWS_LOGSTREAM_INFO(CurlTag, "Initializing CurlHandleContainer with size " << maxSize);
     if (!isInit)
     {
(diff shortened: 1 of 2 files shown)


JonathanHenson:
good catch! I'll get this merged in ASAP.

aws-cpp-sdk-core/source/http/curl/CurlHandleContainer.cpp
```diff
@@ -205,12 +215,6 @@ bool CurlHandleContainer::CheckAndGrowPool()
 
             if (curlHandle)
             {
-                //for timeouts to work in a multi-threaded context,
-                //always turn signals off. This also forces dns queries to
-                //not be included in the timeout calculations.
-                curl_easy_setopt(curlHandle, CURLOPT_NOSIGNAL, 1L);
-                curl_easy_setopt(curlHandle, CURLOPT_TIMEOUT_MS, m_requestTimeout);
-                curl_easy_setopt(curlHandle, CURLOPT_CONNECTTIMEOUT_MS, m_connectTimeout);
```
JonathanHenson: wouldn't this prevent the options from being set on the first run? I'll put it in both places.

edolstra: No, because `CheckAndGrowPool()` is only called from `AcquireCurlHandle()`, so after the former has allocated a new handle, the latter will always apply the options.

JonathanHenson:
Have a look at my most recent push. Reopen if this isn't the spirit of what you were going for.

PR #138, opened 2016-03-27
Avoid warning as error for always true comparison on g++ 4.9.3

crawlins:
On platforms where char is unsigned, the expression (c >= 0)
will always be true and will fail to compile when -Werror
is specified.

Signed-off-by: Clark Rawlins <clark@bit63.org>


JonathanHenson:
Good catch, this has been updated in the internal release. It will be going out shortly. Thanks for your contribution.

Issue #139, opened 2016-03-28
CMakeLists.txt not generating vcxproj correctly for static libraries


mattman71:
When I use cmake to specify my desired environment for building, like this:
cmake . -DTARGET_ARCH="WINDOWS" -DSTATIC_LINKING=1 -G "Visual Studio 14 Win64"

I get .vcxproj files that have:
&lt;RuntimeLibrary&gt;MultiThreadedDebugDLL&lt;/RuntimeLibrary&gt;
and
&lt;RuntimeLibrary&gt;MultiThreadedDLL&lt;/RuntimeLibrary&gt;

instead of what they should be:
&lt;RuntimeLibrary&gt;MultiThreadedDebug&lt;/RuntimeLibrary&gt;
&lt;RuntimeLibrary&gt;MultiThreaded&lt;/RuntimeLibrary&gt;

I fixed this on my local copy, but I'm a noob, both to this project as well as to cmake, so I'm not naïve enough to try to push my change back without someone reviewing to either approve or do it in a more proper way.

I've attached a patch file of what I did to fix the problem, so hopefully someone who knows what they're doing can incorporate it into the solution so everyone (who is using Visual Studio, Windows, and wants Static libraries) can benefit.

[CMakeLists.patch.txt](https://github.com/aws/aws-sdk-cpp/files/192001/CMakeLists.patch.txt)

Thanks,
-Matt

JonathanHenson:
Can you submit a pull request?

mattman71:
Sorry, I don't understand. I just did a pull, and there were changes since my last pull, but the problem I mentioned still exists. There were no changes in the CMakeLists.txt that I could see that would effect my patch, thus my confusion about a pull request. Can you clarify? (remember, I already admitted I'm a noob, haha)

JonathanHenson:
Google "github create a pull request"

bretambrose:
Does this issue cause things not to build or run properly?  Or does it just result in a link/run set up that you don't want?

mattman71:
It causes a linker failure with my static project, so I'm unable to build successfully.

mattman71:
@JonathanHenson thanks, how embarrassing! (forgot to mention I'm a github noob as well, coming from subversion land)

bretambrose:
[...]

PR #140, opened 2016-03-28
CMakeLists.txt not generating vcxproj correctly for static libraries #139

mattman71:
https://github.com/aws/aws-sdk-cpp/issues/139

This is the attempt I made to fix issue 139. It works for me, but somebody who knows what they are doing should probably review and/or fix it properly. I try to provide an idea for a solution when I point out an issue, so hopefully this helps enough. I'm a noob to aws-sdk-cpp, cmake, and github, so with that disclaimer, here it goes...

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -944,6 +944,46 @@ macro(copyDlls exeName)
     endif()
 endmacro()
 
+# solution for static library taken and adapted from http://stackoverflow.com/questions/10113017/setting-the-msvc-runtime-in-cmake
+macro(configure_msvc_runtime)
+    if(MSVC AND NOT BUILD_SHARED_LIBS)
+        # Default to statically-linked runtime.
+        if("${MSVC_RUNTIME}" STREQUAL "")
+            set(MSVC_RUNTIME "static")
+        endif()
+        # Set compiler options.
+        set(variables
+            CMAKE_C_FLAGS_DEBUG
+            CMAKE_C_FLAGS_MINSIZEREL
+            CMAKE_C_FLAGS_RELEASE
+            CMAKE_C_FLAGS_RELWITHDEBINFO
+            CMAKE_CXX_FLAGS_DEBUG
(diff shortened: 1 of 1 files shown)


JonathanHenson:
We've recently updated this in the CMakeLists file. Have a look and reopen if this doesn't solve your issue.

mattman71:
Checking it out... will post back as soon as I validate. Thanks for the
notice!

On Mon, May 9, 2016 at 5:52 PM, Jonathan M. Henson <notifications@github.com

bretambrose:
The new cmake params you'll want to use (you can drop the -DSTATIC_LINKING=1) are -DBUILD_SHARED_LIBS=OFF -DFORCE_SHARED_CRT=OFF.  Both of these default to ON to maintain existing behavior.

mattman71:
I had only limited success with this. I can provide more console output or
logs as feedback if needed, but basically -DBUILD_SHARED_LIBS=OFF
-DFORCE_SHARED_CRT=OFF weren't enough. Still ended up with dynamic
libraries. I had to add -DMSVC=1 as well. However, this is where it gets
weird. Even that did not work, unless I ran the cmake (with those same
parameters) more than once. When running the cmake and it shows the CXX
flags and such that will be used, still came up with /MD, unless I ran with
all 3 of those -D parameters, and only if I ran the cmake 2 or more times
in a row, then I saw the proper /MTD type flags.

Should I open a new issue?

The good news is that at least I can use that work-around without having to
edit anything now, if I remember to do the hokie-pokie. :)

bretambrose:
Reopen this please.  It looks like I either didn't test the fix (think I did) or I moved it around after testing it.  If you move the block from "# Compiler recognition" through "set_msvc_crt_settings()" down below "project(aws-sdk-cpp-all)", it should work properly.  As you discovered, MSVC wasn't set, so nothing was happening since the macro was getting run before the toolchain settings were finalized.

Issue #135, opened 2016-03-23
Aws::EC2::Model::Snapshot: startTime


quatmax:
Hi!

First of all thanks a lot for this cool C++ toolkit!

I have a problem with `Aws::EC2::Model::Snapshot` in connection with the [DescribeSnapshots-API](http://docs.aws.amazon.com/AWSEC2/latest/APIReference/API_DescribeSnapshots.html).
The `startTime` property is parsed as string, trimmed and converted to double. 
On my configuration (Win10, VC2015, German locale) this results in a double containing only the year-part of the `startTime`. 

Wouldn't the raw string of the result be better (or a TimeStamp data type)?
Is this an error or am I missing something?

Thanks for your answer!

ps.: If this is an error, it affects `startTime` properties of all responses and requests.

JonathanHenson:
We've reworked the way date timestamps are handled in the SDK. Have a look at aws-cpp-sdk-core/include/aws/core/utils/DateTime.h. Let me know if you have any questions.

PR #142, opened 2016-03-30
Aws::Ec2::model::Snapshot: m_startTime double -> Aws::String

quatmax:
Hi!

Regarding #135, I patched Snapshot handling startTime as string instead of double.

I think, there is a general bug in handling timestamps within the sdk.

The string conversion to double produces wrong results.

I see three possible solutions for timestamp attributes:

1. Work with the raw string data instead of doubles (an example is this pull request)
2. Convert them to a correct double representation.
3. Convert them to std::time_t or a custom Aws::TimeStamp type 

BR, max

### DIFF
--- a/aws-cpp-sdk-ec2/include/aws/ec2/model/Snapshot.h
+++ b/aws-cpp-sdk-ec2/include/aws/ec2/model/Snapshot.h
@@ -216,17 +216,17 @@ namespace Model
     /**
      * <p>The time stamp when the snapshot was initiated.</p>
      */
-    inline double GetStartTime() const{ return m_startTime; }
+    inline const Aws::String& GetStartTime() const{ return m_startTime; }
 
     /**
      * <p>The time stamp when the snapshot was initiated.</p>
      */
-    inline void SetStartTime(double value) { m_startTimeHasBeenSet = true; m_startTime = value; }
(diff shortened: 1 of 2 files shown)


Bu11etmagnet:
My preference would be for time-related fields to be stored as std::chrono types (duration or time_point, as appropriate) and be converted to strings when communicating with Amazon services.

cbsmith:
It makes sense to create a custom Amazon Time type with automatic conversions from chrono, string, time_t.

bizzarri:
I think this problem exists with the LastModified date/time field in S3 – which returns a double even though it’s a string.  All I get is the year when I use this.

Maurice

JonathanHenson:
I'm about to try and knock this out today. I'll keep you posted.

quatmax:
Thanks! Looking forward to see your solution.

JonathanHenson:
Ok, I've reworked the way all date timestamps are handled in the SDK. Have a look and reopen an issue if the issue persists.

JonathanHenson:
In particular, have a look at aws-cpp-sdk-core/include/aws/core/utils/DateTime.h

quatmax:
Hi! Perfect solution. It seems that there is a bug in ISO 8601 parsing. I will open a new PR.

PR #147, opened 2016-04-05
bugfix: parse ISO_8601 timestamp as returned from amazon webservices

quatmax:
For me the format "2016-04-05T10:05:28**.000**Z" is returned from AWS...

Am I wrong?

### DIFF
--- a/aws-cpp-sdk-core-tests/utils/DateTimeTest.cpp
+++ b/aws-cpp-sdk-core-tests/utils/DateTimeTest.cpp
@@ -63,7 +63,7 @@ TEST(DateTimeTest, TestRFC822Parsing_WrongFormat)
 
 TEST(DateTimeTest, TestISO_8601Parsing)
 {
-    const char* gmtDateStr = "2002-10-02T08:05:09Z";
+    const char* gmtDateStr = "2002-10-02T08:05:09.000Z";
     DateTime gmtDate(gmtDateStr, DateFormat::ISO_8601);
     ASSERT_TRUE(gmtDate.WasParseSuccessful());
     ASSERT_EQ(DayOfWeek::Wednesday, gmtDate.GetDayOfWeek());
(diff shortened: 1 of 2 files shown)


JonathanHenson:
Nice work. Let me look around here real quick and confirm.

JonathanHenson:
http://docs.aws.amazon.com/AmazonCloudWatch/latest/APIReference/API_GetMetricStatistics.html

Here is an example from our docs that shows my original format.

aws-cpp-sdk-core-tests/utils/DateTimeTest.cpp
```diff
@@ -63,7 +63,7 @@ TEST(DateTimeTest, TestRFC822Parsing_WrongFormat)
 
 TEST(DateTimeTest, TestISO_8601Parsing)
 {
-    const char* gmtDateStr = "2002-10-02T08:05:09Z";
+    const char* gmtDateStr = "2002-10-02T08:05:09.000Z";
```
JonathanHenson: We don't actually want to send this format. We don't have millisecond precision in ISO Dates and it isn't actually used by any of the other SDKs.

aws-cpp-sdk-core/source/utils/DateTime.cpp
```diff
@@ -24,7 +24,7 @@
 static const char* CLASS_TAG = "DateTime";
 static const char* RFC822_DATE_FORMAT_STR_MINUS_Z = "%a, %d %b %Y %H:%M:%S";
 static const char* RFC822_DATE_FORMAT_STR_WITH_Z = "%a, %d %b %Y %H:%M:%S %Z";
-static const char* ISO_8601_LONG_DATE_FORMAT_STR = "%Y-%m-%dT%H:%M:%SZ";
+static const char* ISO_8601_LONG_DATE_FORMAT_STR = "%Y-%m-%dT%H:%M:%S.000Z";
```
JonathanHenson: I meant to put that previous comment here. This isn't the format we want to send.

aws-cpp-sdk-core/source/utils/DateTime.cpp
```diff
@@ -802,6 +802,21 @@ class ISO_8601DateParser : public DateParser
...
+                    else if (isdigit(c))
+                    {
+						//skip
+                    }
+                    else
+                    {
+                        m_error = true;
+                    }
+
+                    break;
```
JonathanHenson: This last state needs to be optional. Most services do not send the MS precision, however it they do, we need to parse correctly. I'll merge these two this morning and push it out.

Thanks for your work on this and for pointing out this edge case.

quatmax:
OK! I see. I thought this is the default precision.
I came around this while using Snapshot startTime. This seems to be sent with MS precision in my case.

If you want me to, I will make the parsing of the MS state optional. 
So that every precision should be ok...

JonathanHenson:
I just pushed an update. Let me know if that works for you.

quatmax:
Hi! 
Works perfect for me!
Thanks for the quick help.
:+1:

Issue #160, opened 2016-04-14
Compile error on Linux with `-DNO_HTTP_CLIENT=1`


zwass:
We need to build static libs on Linux without libcurl, and are running into this error:

```
$ cmake -DCMAKE_BUILD_TYPE=Release -DBUILD_ONLY="aws-cpp-sdk-kinesis;aws-cpp-sdk-firehose;aws-cpp-sdk-iam" -DSTATIC_LINKING=1 -DNO_HTTP_CLIENT=1 ..
-- TARGET_ARCH not specified; inferring host OS to be platform compilation target
-- Generating linux build config
-- Dynamic linking disabled
-- You will need to inject an http client implementation before making any http requests!
-- Encryption: Openssl
-- The C compiler identification is GNU 4.8.4
-- The CXX compiler identification is GNU 4.8.4
-- Check for working C compiler: /usr/bin/cc
-- Check for working C compiler: /usr/bin/cc -- works
-- Detecting C compiler ABI info
-- Detecting C compiler ABI info - done
-- Detecting C compile features
-- Detecting C compile features - done
-- Check for working CXX compiler: /usr/bin/c++
-- Check for working CXX compiler: /usr/bin/c++ -- works
-- Detecting CXX compiler ABI info
-- Detecting CXX compiler ABI info - done
-- Detecting CXX compile features
-- Detecting CXX compile features - done
-- Install binary prefix: linux/intel64
-- Found Git: /usr/bin/git (found version "1.9.1")
Updating version info to 0.9.6-90-g8581d6e
-- Custom memory management disabled
-- Configuring done
-- Generating done
-- Build files have been written to: /home/vagrant/aws-sdk-cpp/build
$ make
Scanning dependencies of target aws-cpp-sdk-core
[  0%] Building CXX object aws-cpp-sdk-core/CMakeFiles/aws-cpp-sdk-core.dir/source/Region.cpp.o
[  0%] Building CXX object aws-cpp-sdk-core/CMakeFiles/aws-cpp-sdk-core.dir/source/Version.cpp.o
[  0%] Building CXX object aws-cpp-sdk-core/CMakeFiles/aws-cpp-sdk-core.dir/source/AmazonSerializableWebServiceRequest.cpp.o
[  1%] Building CXX object aws-cpp-sdk-core/CMakeFiles/aws-cpp-sdk-core.dir/source/Globals.cpp.o
[...]

PR #161, opened 2016-04-14
Fix error when using -DNO_HTTP_CLIENT=1

zwass:
Fixes issue #160


zwass:
@JonathanHenson please take a look at this and merge when you can.

bretambrose:
Rather than multiple #ifdefs, my preference would be to just add an AWS_UNREFERENCED_PARAM(clientConfiguration) to the #else clause.

zwass:
Love it, updated, thanks @bretambrose.

JonathanHenson:
Integrating now. Once we've confirmed the builds, we'll push it out.

JonathanHenson:
This has been merged in.

PR #168, opened 2016-04-23
Add cmake variable to enable/disable the tests

svagionitis:
#### Changes
Introduce cmake variable `ENABLE_TESTING` which if set to 1 enables the build and run of the tests. If set to 0 no test is build or run. If the variable is not defined, then the default value is 1 which means that will build and run the tests.

#### Testing
Tested the following scenarios
* `cmake -DENABLE_TESTING=1 -DBUILD_ONLY="aws-cpp-sdk-s3"`: Build and run the tests for the core module. Build the integration tests for the S3 module.
* `cmake -DBUILD_ONLY="aws-cpp-sdk-s3"`: Same as above.
* `cmake -DENABLE_TESTING=0 -DBUILD_ONLY="aws-cpp-sdk-s3"`: Build only the core and the S3 library modules with no tests.
* `cmake -DENABLE_TESTING=0 -DBUILD_ONLY="aws-cpp-sdk-s3;aws-cpp-sdk-dynamodb;aws-cpp-sdk-cognito-identity"`: Build only the core, S3, dynamo DB and cognito identity library modules without any tests.

Please review my changes.


thedevilfish:
This change would be extremely helpful!

bretambrose:
I'll try and get to this today.

JonathanHenson:
this has been shipped. Thanks for your contribution!

Issue #83, opened 2016-01-19
SQS - Message Attributes


patilarpith:
Unable to fetch ```SentTimestamp``` message attribute.

ReceiveMessageRequest class functions like GetAttributeNames() & SetAttributeNames() deals with QueueAttributeName. 

How to set ReceiveMessageRequest to also fetch SentTimestamp during a ReceiveMessage() request?

patilarpith:
@JonathanHenson Could you please take a look at this?

JonathanHenson:
I think you need GetMessageAttributes() and SetMessageAttributes()

patilarpith:
@JonathanHenson Here is my confusion

SQS: Message.h has method
```text
inline const Aws::Map<QueueAttributeName, Aws::String>& GetAttributes()
```

Here is the description of this method
```xml
<p><code>SenderId</code>, <code>SentTimestamp</code>, <code>ApproximateReceiveCount</code>, and/or <code>ApproximateFirstReceiveTimestamp</code>. <code>SentTimestamp</code> and <code>ApproximateFirstReceiveTimestamp</code> are each returned as an integer representing the <a href="http://en.wikipedia.org/wiki/Unix_time">epoch time</a> in milliseconds.</p>
```

This is the QueueAttributeName enum 
```cpp
  enum class QueueAttributeName
  {
    NOT_SET,
    Policy,
    VisibilityTimeout,
    MaximumMessageSize,
    MessageRetentionPeriod,
    ApproximateNumberOfMessages,
    ApproximateNumberOfMessagesNotVisible,
    CreatedTimestamp,
    LastModifiedTimestamp,
    QueueArn,
    ApproximateNumberOfMessagesDelayed,
    DelaySeconds,
    ReceiveMessageWaitTimeSeconds,
    RedrivePolicy
  };
```

It looks like SetMessageAttributeNames() of ReceiveMessageRequest class deals with custom message attributes that can be added to SQS Messages.

It is still unclear how to fetch ```SentTimestamp``` attribute along with the SQS Message

JonathanHenson:
I think the documentation is probably wrong. Try setting SentTimestamp on message attributes and Getting it back out of the message attributes. I'll check around here today to confirm.

patilarpith:
Yes that would be one solution but I don't have access to the en-queuing service. Thanks for the assistance.

JonathanHenson:
I'm slightly confused. Did that work? Is your issue resolved?

[...]

PR #172, opened 2016-05-10
Modeling SQS message default attributes

patilarpith:
Modeled it very similar to Java SDK. 

Java SDK ReceiveMessageRequest has 2 methods
1. setMessageAttributeNames() - > For custom attributes 
2. setAttributeNames() - > For built-in message attributes

Attributes cannot be a union of both the built-in message attributes as well as custom attributes since:
1. built-in message attributes are attached with AttributeName.N={Value} to the request 
2. custom attributes are attached with MessageAttributeName.N={Value} to the request

Towards #83

### DIFF
--- a/aws-cpp-sdk-sqs/include/aws/sqs/model/AttributeName.h
+++ b/aws-cpp-sdk-sqs/include/aws/sqs/model/AttributeName.h
@@ -1,0 +1,42 @@
+/*
+* Copyright 2010-2015 Amazon.com, Inc. or its affiliates. All Rights Reserved.
+*
+* Licensed under the Apache License, Version 2.0 (the "License").
+* You may not use this file except in compliance with the License.
+* A copy of the License is located at
+*
+*  http://aws.amazon.com/apache2.0
+*
+* or in the "license" file accompanying this file. This file is distributed
+* on an "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either
+* express or implied. See the License for the specific language governing
+* permissions and limitations under the License.
+*/
+#pragma once
+#include <aws/sqs/SQS_EXPORTS.h>
+#include <aws/core/utils/memory/stl/AWSString.h>
+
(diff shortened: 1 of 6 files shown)


patilarpith:
cc @JonathanHenson @prestomation

patilarpith:
@JonathanHenson @prestomation Could someone please take a look at this. 

SentTimestamp of a message is the single more crucial metric that we use to measure the availability of our service. Unavailability of it is the single blocker that we are facing in order to move to aws-sdk-cpp and scrap the old version out.

If someone could take a look at it and either accept or reject with feedback it would be much appreciated

JonathanHenson:
I'll add the additional methods this week.

patilarpith:
@JonathanHenson Thank you

JonathanHenson:
Ok, so I think I understand what is happening here. SendMessage does not take the queue attributes field. Instead SQS inserts these values into the queue attributes for the message upon receiving it. You want to get this value back out on the receiving side, but since SQS hasn't actually modeled these members in their service model, it doesn't show up in the enum and you can't extract the value. Sound correct? 

Also "Modeled it very similar to Java SDK.

Java SDK ReceiveMessageRequest has 2 methods"

is very seldom something we are interested in. We aren't Java and don't have a design goal of doing things similarly to the Java SDK (in fact often times we purposely do the opposite).

We already expose MessageAttributes for custom attributes and those should be used for that purpose. The queue attributes seem to be the problem, and that boils down to the fact that the SQS model in https://github.com/aws/aws-sdk-cpp/blob/master/code-generation/api-descriptions/sqs-2012-11-05.normal.json is just plain wrong. I'll add all of the enums called out in the documentation to the QueueAttributeName enum. I think that should solve your problem. Then I'm going to bug the SQS team to fix their model.

JonathanHenson:
https://github.com/aws/aws-sdk-cpp/commit/c00923192cc55a640394b8eebf08b0bb6c3a9a30

PR #174, opened 2016-05-17
Fixed BUILD_ONLY example in README.md. Modified CMakeLists.txt for better integration with IDE.

JoyIfBam5:
Assigned BUILD_ONLY, STATIC_LINKING, and CUSTOM_MEMORY_MANAGEMENT to their default values.
Some IDEs, qtcreator in particular allow to edit cmake build options in a nice GUI form. However if an option is not set to some value in CMakeLists.txt then the nice GUI form does not recognize the option.

BUILD_ONLY accepts project names without "aws-cpp-sdk-" prefix.
But In the README example BUILD_ONLY uses the prefix.


bretambrose:
STATIC_LINKING only exists for backwards compatibility and should stay hidden; BUILD_SHARED_LIBS is the preferred option now.

Defaulting CUSTOM_MEMORY_MANAGEMENT to 0 is actually a breaking change given the way it's used in core (gives a different result if you build shared libraries and don't specify anything memory-wise on the command line).  This was a mistake on my part as I should have defaulted it to off unconditionally.  We may go ahead and make the change in the near future, but if we do it will be a transition to ON/OFF and not the awkward "0"/"1" that we used early on.

JoyIfBam5:
Removed  STATIC_LINKING and CUSTOM_MEMORY_MANAGEMENT default assignments.
Put BUILD_SHARED_LIBS instead of STATIC_LINKING in README.md.

PR #178, opened 2016-05-19
Fixed BUILD_ONLY example in README.md. Modified CMakeLists.txt for better integration with IDE.

JoyIfBam5:
This is the same request as  #174 but from separate branch

Assigned BUILD_ONLY to empty string in CMakeLists.txt
Fixed BUILD_ONLY example in README.md
Put BUILD_SHARED_LIBS instead of STATIC_LINKING in README.md because STATIC_LINKING is deprecated.


JonathanHenson:
Thanks for your contriubtion. parts of this have been pulled internally and will be going out shortly.

PR #184, opened 2016-05-24
Fix compile error when building core package independently with Clang

marcomagdy:
Building the core package on OS X using stock Clang reports 3 errors because of missing class definition.
error: member access into incomplete type 'element_type' (aka 'Aws::Utils::Crypto::HashFactory') 
error: member access into incomplete type 'element_type' (aka 'Aws::Utils::Crypto::HMACFactory')

Adding the necessary header files fixes the problem.


JonathanHenson:
this has been updated internally and will be going out shortly thanks for your contribution!

Issue #194, opened 2016-05-31
Build fails on Ubuntu 14.04 64-bit (gcc 4.8.4)


thedevilfish:
Hello,

Doing a build fails on Ubuntu 14.04 for v0.12.3 of the SDK:

Ran the following commands:
```
cmake -DBUILD_ONLY="s3"
make
```

Here is the output from the build:
```
[ 28%] Building CXX object aws-cpp-sdk-core-tests/CMakeFiles/runCoreUnitTests.dir/utils/crypto/SymmetricCiphersTest.cpp.o
/home/dev/Downloads/aws-sdk-cpp/aws-cpp-sdk-core-tests/utils/crypto/SymmetricCiphersTest.cpp: In function ‘void TestGCMMultipleBuffers(const String&, const String&, const String&, const String&, const String&)’:
/home/dev/Downloads/aws-sdk-cpp/aws-cpp-sdk-core-tests/utils/crypto/SymmetricCiphersTest.cpp:498:70: error: call of overloaded ‘CryptoBuffer(<brace-enclosed initializer list>)’ is ambiguous
     auto decryptResult = CryptoBuffer({ std::move(decryptedStreams) });
                                                                      ^
/home/dev/Downloads/aws-sdk-cpp/aws-cpp-sdk-core-tests/utils/crypto/SymmetricCiphersTest.cpp:498:70: note: candidates are:
In file included from /home/dev/Downloads/aws-sdk-cpp/aws-cpp-sdk-core/include/aws/core/utils/crypto/Factories.h:19:0,
                 from /home/dev/Downloads/aws-sdk-cpp/aws-cpp-sdk-core-tests/utils/crypto/SymmetricCiphersTest.cpp:17:
/home/dev/Downloads/aws-sdk-cpp/aws-cpp-sdk-core/include/aws/core/utils/Array.h:246:13: note: Aws::Utils::CryptoBuffer::CryptoBuffer(Aws::Utils::CryptoBuffer&&)
             CryptoBuffer(CryptoBuffer&& other) : ByteBuffer(std::move(other)) {}
             ^
/home/dev/Downloads/aws-sdk-cpp/aws-cpp-sdk-core/include/aws/core/utils/Array.h:245:13: note: Aws::Utils::CryptoBuffer::CryptoBuffer(const Aws::Utils::CryptoBuffer&)
             CryptoBuffer(const CryptoBuffer& other) : ByteBuffer(other) {}
             ^
[...]

PR #196, opened 2016-05-31
Fix gcc 4.8.5 build

JoyIfBam5:
Fix for build using g++ 4.8.5 and maybe 4.8.4 #194 

g++ 5.3.1 does not generate an error here.


JonathanHenson:
Good catch, I'll merge this over shortly.

JonathanHenson:
This has been updated internally and will be going out shortly. Thanks for your contribution!

PR #209, opened 2016-06-10
Updated aws-cpp-sdk-core's CMakeLists.txt

zeliard:
When I generate Visual Studio Solution (.sln) from cmake for SDK DLL, I'm in a spot of bother due to adding manually AWS_CUSTOM_MEMORY_MANAGEMENT into Visual Studio Project file.

So, how about adding AWS_CUSTOM_MEMORY_MANAGEMENT definition when USE_AWS_MEMORY_MANAGEMENT is ON?


bretambrose:
Recent versions of the SDK should no longer need any definitions like this.  SDKConfig.h will get generated at configuration time and contain your memory management choice.

zeliard:
AWS_CUSTOM_MEMORY_MANAGEMENT does not be defined in SDKConfig.h when doing cmake from the master branch of this repo.

bretambrose:
What's in your SDKConfig.h?

zeliard:
one line: 
#define USE_AWS_MEMORY_MANAGEMENT

but, I can't find the relation between USE_AWS_MEMORY_MANAGEMENT and AWS_CUSTOM_MEMORY_MANAGEMENT.

bretambrose:
Oops, every reference to AWS_CUSTOM_MEMORY_MANAGEMENT should be replaced by USE_AWS_MEMORY_MANAGEMENT.  We'll push a fix soon.

JonathanHenson:
this has been pushed. Thanks for your contriubtion.

PR #210, opened 2016-06-10
Fix typos in XML element names

marianobarrios:


JonathanHenson:
I've updated this in the code generator and pushed out the changes. Thanks for bringing this issue to our attention.

PR #212, opened 2016-06-14
Added const for DynamoDB::Model::AttributeValue::GetM()

ukjhsa:
Hello,
I found all of `DynamoDB::Model::AttributeValue::GetX()` are `const` member function,
and `AttributeValueValue::GetM()` is `const` too.
But `AttributeValue::GetM()` is non-const.
This helps users who just want to get entry inside M without letting parameter be non-const.


JonathanHenson:
merged internally, will be going out soon. Thanks for your contribution!

PR #223, opened 2016-07-01
*fixed AWSMemory.cpp for including SDKConfig.h

zeliard:
Added one include line to apply USE_AWS_MEMORY_MANAGEMENT 
The reason is that both AWSMemory.h and MemorySystemInterface.h don't include the header file, SDKConfig.h

### DIFF
--- a/aws-cpp-sdk-core/source/utils/memory/AWSMemory.cpp
+++ b/aws-cpp-sdk-core/source/utils/memory/AWSMemory.cpp
@@ -14,7 +14,7 @@
   */
 
 #include <aws/core/utils/memory/AWSMemory.h>
-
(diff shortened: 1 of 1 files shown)


bretambrose:
Thanks for finding this mistake.  I'm going to do something slightly different with the fix though, to try and make sure this kind of a mistake doesn't happen again even if SDKConfig.h gains additional definitions.

JonathanHenson:
This change should have been pushed out by now. Feel free to reopen if you still have issues.

zeliard:
That change hasn't affect yet in a current master branch.
USE_AWS_MEMORY_MANAGEMENT should be defined for AWSMemory.cpp currently.

cppsdk_pic

PR #233, opened 2016-07-11
Keep ANSI function and type usage consistent

yamashi:
This code failed to build when using unicode on windows


JonathanHenson:
I've merged this in, the change will propogate out shortly.

PR #236, opened 2016-07-13
change cmake_minimum_required to 3.0.0

yupeng820921:
this lib require target_compile_options, which was added to cmake 3.0.0,
a older version cmake will report this error:
CMake Error at aws-cpp-sdk-core/CMakeLists.txt:342 (target_compile_options):
  Unknown CMake command "target_compile_options".

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -16,7 +16,7 @@
 # minimum version of cmake that 
 #   (1) supports ExternalProject_Add URL_HASH
 #   (2) correctly extracts OPENSSL's version number from openssl/opensslv.h in version 1.0.2d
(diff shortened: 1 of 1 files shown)


vaintroub:
I think this should be handled other way around, not bumping the cmake version,  but replacing offending target_compile_options with some 2.x equivalent. CMake is 2.8 on a multitude of currently supported systems

JonathanHenson:
We set this to 2.8.12, since it was added there. 3.0 would break many users and make it difficult for them to get a working cmake installation.

kolbe:
A CMake 2.8.12 is also not great. This means it cannot be built using stock CMake on EL7.

PR #242, opened 2016-07-20
BlockingExecutor

a7c:
First attempt at BlockingExecutor implementation + unit test. Currently has a build error:

/Users/alacheng/Documents/aws-sdk-cpp/aws-cpp-sdk-core/source/utils/threading/Executor.cpp:133:17: note: in instantiation of function template specialization 'Aws::Utils::Threading::Executor::Submit<void
      (Aws::Utils::Threading::BlockingExecutor::*)(std::__1::function<void ()> &&), std::__1::function<void ()> &>' requested here
    m_executor->Submit(&BlockingExecutor::ExecuteTask, fn);

### DIFF
--- a/aws-cpp-sdk-core-tests/utils/threading/ExecutorTest.cpp
+++ b/aws-cpp-sdk-core-tests/utils/threading/ExecutorTest.cpp
@@ -1,5 +1,5 @@
 /*
- * Copyright 2010-2015 Amazon.com, Inc. or its affiliates. All Rights Reserved.
+ * Copyright 2010-2016 Amazon.com, Inc. or its affiliates. All Rights Reserved.
  *
  * Licensed under the Apache License, Version 2.0 (the "License").
  * You may not use this file except in compliance with the License.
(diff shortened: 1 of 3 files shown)


aws-cpp-sdk-core/include/aws/core/utils/threading/Executor.h
```diff
@@ -113,6 +113,44 @@ namespace Aws
...
+                 */
+                BlockingExecutor(const BlockingExecutor&) = delete;
+                BlockingExecutor& operator =(const BlockingExecutor&) = delete;
+                BlockingExecutor(BlockingExecutor&&) = delete;
+                BlockingExecutor& operator =(BlockingExecutor&&) = delete;
+                
+            protected:
+                bool SubmitToThread(std::function<void()>&&) override;
+                /** Runs fn and then informs the executor of task completion. */
+                void ExecuteTask(std::function<void()>&& fn);
```
JonathanHenson: I'd probably make these two methods private since they aren't virtual.

aws-cpp-sdk-core/include/aws/core/utils/threading/Executor.h
```diff
@@ -113,6 +113,44 @@ namespace Aws
...
+                void OnTaskComplete();
+                
+            private:
+                std::mutex m_syncPointLock;
+                std::condition_variable m_syncPoint;
+                
+                std::shared_ptr<Executor> m_executor;
+                size_t m_poolSize;
+                
+                size_t m_numTasksRunning;
```
JonathanHenson: This should be std::atomic<size_t>

aws-cpp-sdk-core/source/utils/threading/Executor.cpp
```diff
@@ -107,3 +107,43 @@ bool PooledThreadExecutor::HasTasks()
     std::lock_guard<std::mutex> locker(m_queueLock);
     return m_tasks.size() > 0;
 }
+
+BlockingExecutor::BlockingExecutor(std::shared_ptr<Executor> executor, size_t poolSize) :
+m_executor(executor), m_poolSize(poolSize)
```
JonathanHenson: Why not include this in the initializer list?

aws-cpp-sdk-core/source/utils/threading/Executor.cpp
```diff
@@ -107,3 +107,43 @@ bool PooledThreadExecutor::HasTasks()
...
+    if (m_numTasksRunning < m_poolSize)
+    {
+        m_numTasksRunning++;
+    }
+    else
+    {
+        m_syncPoint.wait(locker);
+    }
+    locker.release();
+    m_executor->Submit(&BlockingExecutor::ExecuteTask, fn);
```
JonathanHenson: fn needs to be cast to an r-value. std::move()

aws-cpp-sdk-core-tests/utils/threading/ExecutorTest.cpp
```diff
@@ -0,0 +1,96 @@
...
+
+/** Clears all entries in the task order tracker */
+void clearTaskOrder()
+{
+    std::lock_guard<std::mutex> locker(taskOrderLock);
+    shouldClearTaskOrder = true;
+    taskOrder.clear();
+}
+
+/** Trivial test to see if BlockingExecutor can run a single task. */
```
JonathanHenson: You are going to run in to ordering problems with std::thread::detach() so you won't get deterministic results. 

I would run two tests. In the first, the mock executor doesn't spin up any threads at all, it simply runs the function passed to it. You can pass a lambda with a capture to it and make sure the counts match up correctly and that ordering happens correctly etc...

The second, you can mock the executor and signaling to force thread contention. Then set the poolSize to something like two and make sure the thread contention stuff you want to test is correct.

aws-cpp-sdk-core/include/aws/core/utils/threading/Executor.h
a7c: If I use std::atomic, do I still need to lock and unlock for each update of the counter?

JonathanHenson: No

aws-cpp-sdk-core-tests/utils/threading/ExecutorTest.cpp
```diff
@@ -0,0 +1,275 @@
...
+    bool SubmitToThread(std::function<void()>&& fn)
+    {
+        fn();
+        return true;
+    }
+    
+};
+
+/** Trivial test to see if BlockingExecutor can run a single task. */
+TEST(BlockingExecutorTest, TestExecuteSingleTask)
```
JonathanHenson: This is a good test case. Well done!

aws-cpp-sdk-core-tests/utils/threading/ExecutorTest.cpp
```diff
@@ -0,0 +1,275 @@
...
+{
+    AWS_BEGIN_MEMORY_TEST(16, 10)
+    
+    std::shared_ptr<ThreadlessMockExecutor> mockExecutor =
+            Aws::MakeShared<ThreadlessMockExecutor>(ALLOCATION_TAG);
+    BlockingExecutor blockingExecutor(mockExecutor, 5);
+    
+    std::vector<u_int> taskOrder;
```
JonathanHenson: Remember to use the Aws aliased STL types. This should be Aws::Vector

aws-cpp-sdk-core-tests/utils/threading/ExecutorTest.cpp
```diff
@@ -0,0 +1,275 @@
...
+/** Test if BlockingExecutor can successfully run more tasks than its pool size. */
+TEST(BlockingExecutorTest, TestExecuteMoreTasksThanPoolSize)
+{
+    AWS_BEGIN_MEMORY_TEST(16, 10)
+    
+    std::shared_ptr<ThreadlessMockExecutor> mockExecutor =
+    Aws::MakeShared<ThreadlessMockExecutor>(ALLOCATION_TAG);
+    BlockingExecutor blockingExecutor(mockExecutor, 5);
+    
+    std::vector<u_long> taskOrder;
```
JonathanHenson: Aws::Vector

aws-cpp-sdk-core-tests/utils/threading/ExecutorTest.cpp
```diff
@@ -0,0 +1,275 @@
...
+{
+    AWS_BEGIN_MEMORY_TEST(16, 10)
+    
+//    std::shared_ptr<MockExecutor> mockExecutor = Aws::MakeShared<MockExecutor>(ALLOCATION_TAG);
+//    BlockingExecutor blockingExecutor(mockExecutor, 3);
+    
+    std::shared_ptr<DefaultExecutor> mockExecutor = Aws::MakeShared<DefaultExecutor>(ALLOCATION_TAG);
+    BlockingExecutor blockingExecutor(mockExecutor, 3);
+    
+    std::vector<u_long> taskOrder;
```
JonathanHenson: Aws::Vector

aws-cpp-sdk-core/source/utils/threading/Executor.cpp
```diff
@@ -107,3 +107,42 @@ bool PooledThreadExecutor::HasTasks()
...
+BlockingExecutor::~BlockingExecutor()
+{
+}
+
+bool BlockingExecutor::SubmitToThread(std::function<void()>&& fn)
+{
+    std::unique_lock<std::mutex> locker(m_syncPointLock);
+    if (m_numTasksRunning < m_poolSize)
+    {
+        m_numTasksRunning++;
```
JonathanHenson: If you are going to use a mutex surrounding the incrementing and decrementing of the task running count, it doesn't need to be atomic.

aws-cpp-sdk-core/include/aws/core/utils/threading/Executor.h
```diff
@@ -113,6 +113,43 @@ namespace Aws
...
+                BlockingExecutor& operator =(BlockingExecutor&&) = delete;
+                
+            private:
+                bool SubmitToThread(std::function<void()>&&) override;
+                /** Runs fn and then informs the executor of task completion. */
+                void ExecuteTask(std::function<void()>&& fn);
+                /** Handler for when a task finishes */
+                void OnTaskComplete();
+                
+                std::mutex m_syncPointLock;
```
JonathanHenson: We should probably make this reentrant due to the use case. See http://en.cppreference.com/w/cpp/thread/recursive_mutex

aws-cpp-sdk-core-tests/utils/threading/ExecutorTest.cpp
```diff
@@ -0,0 +1,275 @@
...
+                locker.unlock();
+                taskSignal.notify_all();
+                locker.lock();
+            }
+        }
+    };
+    
+    // Tasks 0 and 1 wait until they are signaled by another task
+    blockingExecutor.Submit(executeTask, 0, WAIT_FOR_TASK_SIGNAL);
+    blockingExecutor.Submit(executeTask, 1, WAIT_FOR_TASK_SIGNAL);
```
JonathanHenson: Ok, you've got the right idea here but it seems way too complicated:

Having both of the first two tasks wait on a semaphore is good.
The next problem is that if the third task is waiting on the blocked two threads, then it will never execute and according to the contract, it should block the calling thread.

Why not try instead of using the DefaultExecutor for the nested executor, actually mock out the executor as planned (you can certainly borrow the code from DefaultExecutor) moving the semaphore logic in there. Then you can actually control the flow of the test. Embeddeding the semaphore signalling in the tasks itself won't be of much help.

Blocking Executor
    Underlying Executor
           --> semaphore logic in here.

I think you just need one semaphore altogether.
When you call submit the second time, it should be stopped on the signal. so far so good.

now when you call it a third time, you need to call it in a separate thread. std::thread.
You should verify after spawning the thread that execution hasn't happened (a flag check would do the trick).
Then signal the semaphore, and make sure the thread does then execute.

aws-cpp-sdk-core-tests/utils/threading/ExecutorTest.cpp
```diff
@@ -0,0 +1,276 @@
+/*
+ * Copyright 2010-2015 Amazon.com, Inc. or its affiliates. All Rights Reserved.
```
JonathanHenson: Copyright date

aws-cpp-sdk-core/include/aws/core/utils/threading/Executor.h
```diff
@@ -113,6 +113,44 @@ namespace Aws
...
+                /**
+                 * Rule of 5 stuff.
+                 * Don't copy or move
+                 */
+                BlockingExecutor(const BlockingExecutor&) = delete;
+                BlockingExecutor& operator =(const BlockingExecutor&) = delete;
+                BlockingExecutor(BlockingExecutor&&) = delete;
+                BlockingExecutor& operator =(BlockingExecutor&&) = delete;
+                
+                bool SubmitToThread(std::function<void()>&&) override;
```
JonathanHenson: This should be protected, not public

aws-cpp-sdk-core/source/utils/threading/Executor.cpp
```diff
@@ -107,3 +107,48 @@ bool PooledThreadExecutor::HasTasks()
...
+    }
+    else
+    {
+        m_syncPoint.wait(locker);
+    }
+    locker.unlock();
+    
+    std::function<void()> executeTask = [&]() {
```
JonathanHenson: This isn't free. A normal bind call would be cheaper. 
m_executor->Submit(&BlockingExecutor::ExecuteTask, this, fn);
this is the exact code we do in the async client calls.

aws-cpp-sdk-core-tests/utils/threading/ExecutorTest.cpp
```diff
@@ -0,0 +1,249 @@
...
+    // Order should be deterministic because mock executor is not concurrent
+    for (size_t i = 0; i < FINAL_SIZE; i++)
+    {
+        ASSERT_TRUE(taskOrder.at(i) == i);
+    }
+    
+    AWS_END_MEMORY_TEST
+}
+
+// TODO: is it fine to have these variables outside like this?
```
JonathanHenson: Yeah it's fine. remove the TODO.

aws-cpp-sdk-core-tests/utils/threading/ExecutorTest.cpp
```diff
@@ -0,0 +1,249 @@
...
+public:
+    MockExecutor() /*: m_numReceivedTasks(0), m_numFinishedTasks(0)*/ { }
+    ~MockExecutor() { }
+    
+    MockExecutor(const MockExecutor&) = delete;
+    MockExecutor& operator =(const MockExecutor&) = delete;
+    MockExecutor(MockExecutor&&) = delete;
+    MockExecutor& operator =(MockExecutor&&) = delete;
+    
+//    std::atomic<std::size_t>* GetNumReceivedTasks() { return &m_numReceivedTasks; }
```
JonathanHenson: lose the comments.

aws-cpp-sdk-core-tests/utils/threading/ExecutorTest.cpp
```diff
@@ -0,0 +1,249 @@
...
+    }
+    locker.unlock();
+    
+    /* Additionally, the last task should now have been queued up (i.e. received by the mock executor
+     * , now that the others have finished. */
+//    ASSERT_TRUE(*mockExecutor->GetNumReceivedTasks() == FINAL_SIZE);
+    ASSERT_TRUE(numReceivedTasks == FINAL_SIZE);
+    
+    // Busy wait for the last task to let us know it's ready, then signal it.
+    while (!lastTaskWaiting);
```
pplachta: normally one would do a Yield() in the loop otherwise you'll tie up the CPU 100%. Of course, this is just a test so it's ok.

aws-cpp-sdk-core-tests/utils/threading/ExecutorTest.cpp
```diff
@@ -0,0 +1,249 @@
...
+    locker.lock();
+    doneSignal.wait(locker);
+    
+    // Check completion and ordering of tasks.
+    ASSERT_TRUE(taskOrder.size() == FINAL_SIZE);
+    ASSERT_TRUE(taskOrder.at(3) == 3);
+    locker.unlock();
+    
+    AWS_END_MEMORY_TEST
+}
```
pplachta: It would seem good to do a test where you run (say) 3x pool size tasks and ensure they are actually running concurrently. I'd set pool size = 10 and have a task that sleeps 100ms. Then, you can make sure the 30 tasks take < 10 * 100ms (actually it should be about 300 ms but you want a little insurance).
I am not sure if this will be deemed too ugly for a test but at least good for a unit test!

aws-cpp-sdk-core/include/aws/core/utils/threading/Executor.h
pplachta: you also use it for the counting semaphore so I wouldn't change it, not easy to get that right

aws-cpp-sdk-core/source/utils/threading/Executor.cpp
```diff
@@ -107,3 +107,40 @@ bool PooledThreadExecutor::HasTasks()
...
+void BlockingExecutor::ExecuteTask(std::function<void()> fn)
+{
+    fn();
+    OnTaskComplete();
+}
+
+void BlockingExecutor::OnTaskComplete()
+{
+    m_syncPoint.notify_one();
+    std::unique_lock<std::recursive_mutex> locker(m_syncPointLock);
```
pplachta: Upon thought: you need to hold the locker around the notify_one() call to be 100% correct wrt ordering of tasks.

Moving this around we fixed a race where a new task could be submitted when a whole bunch of other tasks were still waiting (and this would cause m_numTasksRunning to be > m_poolSIze).

As coded the situation is much more subtle: m_numTasksRunning can easily be > m_poolSize but it does not reflect the ACTUAL number of guys running. And no new guys will be allowed in.

Might be easiest to just move 142 below the locker and be done with it -- and then you can also assert that m_numTasksRunning is <= m_poolSize.

It's also good to write a test (a stress test) which just adds 100s of tasks with m_poolSize=10 just to make sure they all complete and there is no crash.

aws-cpp-sdk-core-tests/utils/threading/ExecutorTest.cpp
```diff
@@ -0,0 +1,373 @@
...
+    {
+        doneSignal.wait(locker);
+    }
+    endTime = high_resolution_clock::now();
+    auto sequentialDuration = duration_cast<milliseconds>(endTime - startTime).count();
+    
+    /* Check if the concurrent time was faster, with an adjustable tolerance for the strictness of
+     * the test */
+    const double TOLERANCE = 1.0;
+    ASSERT_LE(asyncDuration, sequentialDuration * TOLERANCE);
```
JonathanHenson: This seems non-deterministic to me. For example, a run loop with async IO will likely beat out multiple threads with synchronous IO. Also, who knows if the operating system thread creation time is faster than running 15 stack frames or not? I would certainly imagine that this test would fail on an embedded system.

aws-cpp-sdk-core/include/aws/core/utils/threading/Executor.h
```diff
@@ -113,6 +113,45 @@ namespace Aws
...
+             * Executor extension that limits number of simultaneously running tasks by forcing
+             * additional tasks to wait if the number of running tasks reaches the specified maximum.
+             */
+            class AWS_CORE_API BlockingExecutor : public Executor
+            {
+            public:
+                /** Constructor that takes an Executor to delegate thread execution to and the max
+                 *  number of tasks that can run at once */
+                BlockingExecutor(std::shared_ptr<Executor> executor, size_t poolSize);
+                ~BlockingExecutor();
```
JonathanHenson: since SubmitToThread is virtual, either mark it as final, or make the destructor virtual. My preference is for the virtual destructor.

aws-cpp-sdk-core/source/utils/threading/Executor.cpp
```diff
@@ -107,3 +107,58 @@ bool PooledThreadExecutor::HasTasks()
...
 }
+
+BlockingExecutor::BlockingExecutor(std::shared_ptr<Executor> executor, size_t poolSize) :
+m_executor(executor), m_poolSize(poolSize), m_numTasksRunning(0)
+{
+}
+
+BlockingExecutor::~BlockingExecutor()
+{
+    while (m_numTasksRunning != 0)
```
JonathanHenson: I typically prefer to have a public method on things like this to allow the user to call shutdown independently of the destruction of the object. The destructor should certainly call that shutdown function, but it gives the user more control over where they take the hit.

aws-cpp-sdk-core/source/utils/threading/Executor.cpp
```diff
@@ -107,3 +107,58 @@ bool PooledThreadExecutor::HasTasks()
...
+        });
+    }
+    m_numTasksRunning++;
+    locker.unlock();
+    
+    m_executor->Submit(&BlockingExecutor::ExecuteTask, this, fn);
+    return true;
+}
+
+void BlockingExecutor::ExecuteTask(std::function<void()> fn)
```
JonathanHenson: std::function<void>&& ?

aws-cpp-sdk-core/source/utils/threading/Executor.cpp
```diff
@@ -107,3 +107,58 @@ bool PooledThreadExecutor::HasTasks()
...
+{
+    fn();
+    OnTaskComplete();
+}
+
+void BlockingExecutor::OnTaskComplete()
+{
+    std::unique_lock<std::recursive_mutex> locker(m_syncPointLock);
+    assert(m_numTasksRunning <= m_poolSize);
```
JonathanHenson: we typically assert on contract violations (think runtime exceptions in Java) and not things like a bug in the code. In this case, aren't we just asserting on a bug?

pplachta: we were hitting this before when the queueing was incorrect, so having an assert seemed prudent.

pplachta: is there some other form of assert we could use?

aws-cpp-sdk-core-tests/utils/threading/ExecutorTest.cpp
pplachta: I was also concerned about this but didn't know how else to code this.
The test is supposed to assert that when you allow a lot of parallel threads, things actually run in parallel. The time for each task (10 ms -- we could make it 100) is a lot longer than anything else that runs on the CPU on any system I know. It's O(I/O time). So this should be perfectly safe. The only scary thing, I think, is the inaccuracy of the time measurement but again, it shouldn't be anything comparable to ms.
We run 15 tasks and I did assume we will have at least 2 cores... otherwise well obviously you're running serially. Maybe we can run this test only if we have >= 2 cores ?

I am also open to any other suggestions or just deleting this test. It did its job as a unit test.

aws-cpp-sdk-core/source/utils/threading/Executor.cpp
a7c: When I try to use an rvalue there, it causes the error that we were trying to debug before ("no matching function for call to BuildFunction")

aws-cpp-sdk-core/source/utils/threading/Executor.cpp
pplachta: RESOLUTION: remove assert

aws-cpp-sdk-core/source/utils/threading/Executor.cpp
pplachta: RESOLUTION: leave as is

aws-cpp-sdk-core-tests/utils/threading/ExecutorTest.cpp
pplachta: RESOLUTION: leave as is

a7c:
Work on the BlockingExecutor is complete! Merged into transfer-rework-client branch.

PR #244, opened 2016-07-22
Mark DateTime comparison operators as 'const'

marcomagdy:
Currently, you have to access the 'Millis()' member function to compare DateTime members on an object marked as const.

For example when using the Async version of the requests, the lambda declares the 'Outcome' parameter as const, which makes all members of that operation's result const. Therefore, you can't do something like this: 
`outcome.GetResult().GetItem().CreatedOn() >= myOtherDateObject.`


marcomagdy:
any feedback on this?

JonathanHenson:
It's looks good. I'll merge it shortly. Thanks for your contribution

Issue #276, opened 2016-08-17
Update CMakeLists to use GNUInstallDirs


ambasta:
Currently, installation targets are hardcoded to bin/lib etc... On PLATFORM_LINUX=1, GNUInstallDirs should be used instead.

bretambrose:
I've merged your changes (with some refactoring in) internally, but my refactoring (mostly pushing all the repetitive changes in per-project CMakeLists.txt files into macros) has broken something and I haven't had a chance to dig further into it yet.  It may be a few weeks before I finish it up.

JonathanHenson:
This has been merged in and it is working it's way through our CI.

PR #277, opened 2016-08-17
Linux stabilization

ambasta:
I am trying to move the project in a fashion that each module can be compiled and installed separately.

I'll add additional patches for PKGCONFIG pc files and CMake Modules so that module dependency can be appropriately detected and fallback can be to local subdir content.

i.e. kinesis dependency on core if available in system via pkg-config else fallback to local dir


JonathanHenson:
Thank you for this work, it's something I've been wanting to do for a while now. Im excited to have a look.

bretambrose:
Have you tried your changes on any other platforms?

ambasta:
Not yet. I am trying to set up MSVC to see how this works on Windows.

ambasta:
@bretambrose @JonathanHenson Hi,

Is there a way to run CI on the project?

JonathanHenson:
You mean our continuous integration pipeline? It runs whenever we push to master on our staging branch. Are you wanting to manually run it?

ambasta:
Indeed. Or a way I can set this up locally to ensure all tests are done

ambasta:
Hi,

I am practically done w/ this PR. Can someone review this and let me know if the approach seems okay? I have tested it on Linux but I am not sure how well this would work on other environments. If there's a way to do CI then please let me know as well so as I may find and resolve any issues generated by this PR.

Fixes #276

bretambrose:
Hello,

We have a high priority change pending where the main CMakeLists.txt file gets split into a number of modules.  This is going to create a huge conflict with your work.  We don't want to lose your work but we need to figure out a way to bring it in against that change.  Would it be possible to make a new pull request that does not include any generated files?  It is very difficult to review this with so many generated files changed.

delhivery-automagic:
Sure,

I have already updated the generator so I can probably cherry pick the single file and send a PR against it. However, there are other changes to the overall project CMakeLists.txt (and some other ones which don't seem to be autogenerated)

How would you want me to send those across? Or should I wait to sent the PR until after your CMakeLists changes are pushed

ambasta:
Sure,

I have already updated the generator so I can probably cherry pick the single file and send a PR against it. However, there are other changes to the overall project CMakeLists.txt (and some other ones which don't seem to be autogenerated)

How would you want me to send those across? Or should I wait to sent the PR until after your CMakeLists changes are pushed

bretambrose:
The CMake split should go out today.  For a PR, I think the generator (.vm) file plus all changes made to the top-level CMakeLists.txt as well as any changes to the CMakeLists.txt of {core, testing-resources, *-tests, access-management, identity-management, transfer, and queues}.

bretambrose:
Also, I will try and start doing a manual merge on our internal repository, but having a minimal reference to compare against would be very helpful.

bretambrose:
One more thing, your changes will be on by default (probably gated on SIMPLE_INSTALL=ON) but won't be unconditional since we have a critical user who needs custom linux build/installs that use the existing fixed directories with platform-specific qualifiers.

delhivery-automagic:
Hi @bretambrose 

I've cleaned this up significantly, and updated the PR to show up against the new CMake structure. In the process of testing this manually though, so any advise with CI would be really appreciated.

JonathanHenson:
It will run CI when we merge it internally, so there is little danger that we will accidentally push it out to the world if it is broken.

aws-cpp-sdk-core/CMakeLists.txt
```diff
@@ -402,6 +397,13 @@ else()
     endif()
 endif()
 
+configure_file("${CMAKE_SOURCE_DIR}/toolchains/pkg-config.pc.in" "${PROJECT_NAME}.pc" @ONLY)
```
bretambrose: Why was this moved here?  Core's config and install  is now slightly different than all the other libraries.

ambasta: Not sure what you mean by moved considering that pkg-config pc generation was not part of the core/CMakeLists.txt before.

bretambrose: Just based on the diff, the configure_file was on line 378 and then it got moved after the "if(SIMPLE_INSTALL)" block to line 400.

Similarly, the "if(BUILD_SHARED_LIBS)" block within "if(SIMPLE_INSTALL)" got moved from line 388 to line 401.

Should all CMakeLists.txt have the same change applied to them? or was this just an harmless accidental change that doesn't actually change anything?

delhivery-automagic: Right, so this was moved because we'd want to retain the pc functionality even on non SIMPLE_INSTALL? If the changes are to be gated on SIMPLE_INSTALL then it should lie in the appropriate if expression block. I wasn't sure what you'd want to gate on SIMPLE INSTALL so I've left it out of the if blocks for now.

ambasta: Right, so this was moved because we'd want to retain the pc functionality even on non SIMPLE_INSTALL? If the changes are to be gated on SIMPLE_INSTALL then it should lie in the appropriate if expression block. I wasn't sure what you'd want to gate on SIMPLE INSTALL so I've left it out of the if blocks for now.

ambasta: Also, I've updated the generator to automatically replicate the changes for other CMakeFiles and stripped them from the generated files hopefully so you shouldn't have issues trying to merge it locally :)

ambasta:
Hey @bretambrose 

Any updates on this?

bretambrose:
I have it merged in and reworked a bit, but I haven't had time to test it
on anything but Windows.

For Unix-based builds, how should I test the package-related functionality?

delhivery-automagic:
You can test the package functionality by building + installing core separately. Building any other module after this should not trigger a re-build of core. The same would apply to any other module and those depending on it.

Finally, other projects should be automatically be able to pull the correct include/linking flags if they use any of aws-sdk modules as long as they use cmake.find_package  or pkg-config to check for dependencies

ambasta:
Can I offer any help to improve upon the patch :)

bretambrose:
Yeah, we're juggling a lot of things right now.  I will push the branch from the private repo to the public one today and you can take a look.  The main changes were the addition of some macros to reduce the amount of per-project boilerplate that was added.

bretambrose:
The FB-UnixPackagingPrep branch has the work-in-progress if you are able to take a look.

bretambrose:
So I finally had time to return to this.  I ended up making a couple of additional changes to support backwards compatibility with the way were previously doing exports (see https://aws.amazon.com/blogs/developer/using-cmake-exports-with-the-aws-sdk-for-c/).  In particular, I removed the AWS:: namespace prefix in front of the imported target names and I also resurrected the catch-all config (aws-sdk-cpp-config.cmake) that includes all the configs of the subprojects in a dependency-enforced order.

I have the pull request ready on our internal repository, but I will try and get it pushed to the branch I added to the public one so that you can review the changes.

bretambrose:
I made a new branch because I suck at git: FB-CMakeImportExport

bretambrose:
Updated and merged a while back.

PR #288, opened 2016-09-07
Remove (presumably) erroneous whitespace

pcolby:
The space in question looks to be accidental to me.

I'm not sure of the side-effects, if any, of this change.  At the very least, I _assume_ [ErrorCode::LunAlreadyAllocated_HASH](https://github.com/aws/aws-sdk-cpp/blob/d8a7539da21392d019c63cdabe3108a4923c51ad/aws-cpp-sdk-storagegateway/source/model/ErrorCode.cpp#L61) will change upon regeneration... is that a good or bad thing?

Cheers.

### DIFF
--- a/code-generation/api-descriptions/storagegateway-2013-06-30.normal.json
+++ b/code-generation/api-descriptions/storagegateway-2013-06-30.normal.json
@@ -1876,7 +1876,7 @@
         "InvalidParameters",
         "InvalidSchedule",
         "LocalStorageLimitExceeded",
-        "LunAlreadyAllocated ",
+        "LunAlreadyAllocated",
         "LunInvalid",
         "MaximumContentLengthExceeded",
         "MaximumTapeCartridgeCountExceeded",


bretambrose:
Those files are a reflection of an internal repository so we can't directly take a pull request on them but we will look into where that space is coming from and hopefully get it removed.

Thanks for noticing this.

JonathanHenson:
We are currently working on fixing this, it is actually a problem across all SDKs. It would hurt for us to trim those strings in the generator though :(

pcolby:
> Those files are a reflection of an internal repository so we can't directly take a pull request on them

Righto :smile:

Do you prefer that I close this PR or leave it open?  If you can't take PRs on that file, then I might as well close it, but if it helps in some way with how you guys track open issues then I'm happy to leave it open.  Either is fine with me :smile: 

Cheers.

JonathanHenson:
We've fixed this in the generator. Let me know if you have any more issues.

PR #312, opened 2016-10-17
AWSAuthSigner: PresignRequest also signs payload if m_signPayloads == true.

tvdijk:
IoT doesn't want "UNSIGNED-PAYLOAD" to be used in the canonical request string, like S3 does. Instead, you just hash the empty payload string (see http://docs.aws.amazon.com/iot/latest/developerguide/protocols.html). This change uses the same logic in PresignRequest() as that used in SignRequest() to conditionally sign the payload. The S3 client will have m_signPayloads == false so the behavior will not change there.


tvdijk:
Sure, see commit above, added new private method.

singku:
Hi, this problem has been solved in this commit: https://github.com/aws/aws-sdk-cpp/commit/b9ccf4d0ded1dccd1fe66fbba6b5c9c306fbb249.

PR #338, opened 2016-11-16
AWSAuthSigner: Add session token at end of query string.

tvdijk:
IoT specifically wants the session token to be at the end of the query string, see item 4 under "To create a Signature Version 4 request":

http://docs.aws.amazon.com/iot/latest/developerguide/protocols.html#mqtt-ws

S3 seems to express no preference:

http://docs.aws.amazon.com/AmazonS3/latest/dev/RESTAuthentication.html#UsingTemporarySecurityCredentials


tvdijk:
Quoting from the link in my comment above:

> If you have session credentials (from an STS server, AssumeRole, or Amazon Cognito), append the session token to the end of the URL string **after** signing:

This is the IoT documentation (and I can confirm this fixes the problem for connecting to the IoT message broker), not general AWS. I have not tested with other services. It would be problematic if other services have conflicting request signing rules.

PR #344, opened 2016-11-18
Allow using OpenSSL on windows (xp support) and better error checking on std::iostream

cedral:
CMAKE now can be told that bcrypt doesn't exist on windows by setting TARGET_ARCH_VERSION to 5

The following is all in WinSyncHttpClient:
switched from bad() to fail() to detect iostream failures (fail() detects both badbit and failbit)

added failure detection to writing to std::iostream (previously just assumed it worked)

switched from !good() to eof() to detect done reading iostream as !good() could indicate a failure rather than completion. The above mentioned switch to fail() makes this factually unneccessary because the other things that could make a !good() true are detected and success is set to false. But better to be symantically accurate. 

Don't call DoReadData if success is false and read data won't be used

stopped checking read > 0 inside while loop with read > 0 as one of its conditions (redundant)

for OpenSSL/CryptImpl.*
$ifndef _MSV_VER around all defaulted move constructors for OpenSSL/CryptoImpl.h (not supported in vs2013)
(std::streampos) cast to -1 in comparisons so it would compile on Visual Studio in Visual Studio 2013

all 196 core tests passing


cedral:
forgot to push first

PR #384, opened 2016-12-19
missing cmath include

WillBrennan:
cmath include missing, failed to compile under Clang++ 3.9 with libc++


bretambrose:
Please don't merge until we verify the android build against earlier NDKs.  Jsoncpp and cmath have been consistent problems across NDK versions.

WillBrennan:
added comment

bretambrose:
This should be fixed in the latest version of the SDK.  Thanks for your help.

PR #389, opened 2016-12-22
Respect ${INCLUDE_DIRECTORY} when installing the s3-encryption headers

edolstra:
`${INCLUDE_DIRECTORY}` is respected by all components of aws-sdk-cpp except `s3-encryption`. This causes problems in Nixpkgs which uses `${INCLUDE_DIRECTORY}` to install header files under a different prefix.


JonathanHenson:
thanks for your contribution. I've merged this internally, and the change will make its way through our CI and out to you shortly.

PR #425, opened 2017-01-25
Improved correctness of VERSION_STRING inside git repo

edenzik:
`VERSION_STRING` is now derived from a git repository if and only if that git repository is rooted at the source directory. This is to prevent behavior seen in aws#383 and elsewhere where the SDK repository is included as a tarball inside of another git repository but the `VERSION_STRING` is derived from the parent project instead of the SDK.


paresy:
This one hit us as well. Though --work-tree did not work as expected. We went with 

```
--git-dir=${CMAKE_CURRENT_SOURCE_DIR}/.git

marcomagdy:
@edenzik could you please update the PR to use `--git-dir=` as suggested by @paresy?
I'd like to merge it ASAP

PR #468, opened 2017-03-10
Add support for compiling Android using Clang up to API-9

akioCL:
Hi all,

I added support for compiling Android with Clang up to API-9. There was some NDK support headers missing that were causing compiling errors. Unfortunately the new headers use  "include_next" that generates a warning because of the "pedantic" mode. For NDK >= 12 there's a way to ignore the warning, but for lower versions I had to remove the "pedantic" mode.


bretambrose:
I've merged this in internally to staging.  I bumped the default api level back to 21 though (for backwards compatibility purposes).  We'll want to update our LY build script to use 19 though.

PR #477, opened 2017-03-21
Fix syntax in the generated pkgconfig .pc files

kromain:
The 'Libs:' line should only include the package's own libraries as linker flags,
while dependent packages should be listed on a separate 'Requires:' line.


kromain:
FYI I initially tried to import the aws-cpp-sdk-s3 module in my app via pkg-config, but it resulted in linker error because the linker line included the module names without the '-l' prefix.
With the patch from this PR, everything compiles and links fine. (I'm no pkg-config expert, but I checked against the syntax guide for .pc files in the pkg-config documentation and the generated files now look like what's specified in the docs)

PR #481, opened 2017-03-23
Add Auth v2 signing to S3 client

hkleynhans:


hkleynhans:
OK, will remove the S3 bits 👍

PR #482, opened 2017-03-23
Log error on failure to set HTTP headers

hkleynhans:


aws-cpp-sdk-core/source/http/windows/WinHttpSyncHttpClient.cpp
```diff
@@ -149,7 +149,8 @@ void WinHttpSyncHttpClient::DoAddHeaders(void* hHttpRequest, Aws::String& header
 
     Aws::WString wHeaderString = StringUtils::ToWString(headerStr.c_str());
 
-    WinHttpAddRequestHeaders(hHttpRequest, wHeaderString.c_str(), (DWORD)wHeaderString.length(), WINHTTP_ADDREQ_FLAG_REPLACE | WINHTTP_ADDREQ_FLAG_ADD);
+    if (!WinHttpAddRequestHeaders(hHttpRequest, wHeaderString.c_str(), (DWORD)wHeaderString.length(), WINHTTP_ADDREQ_FLAG_REPLACE | WINHTTP_ADDREQ_FLAG_ADD))
+        AWS_LOG_FATAL(GetLogTag(), "Failed to add HTTP request headers!");
```
JonathanHenson: FATAL is usually right before an exit(-1) or an assertion. ERROR is probably more appropriate. Also, let's use LOGSTREAM instead of LOG.

aws-cpp-sdk-core/source/http/windows/WinINetSyncHttpClient.cpp
```diff
@@ -120,7 +120,8 @@ void* WinINetSyncHttpClient::OpenRequest(const Aws::Http::HttpRequest& request,
 
 void WinINetSyncHttpClient::DoAddHeaders(void* hHttpRequest, Aws::String& headerStr) const
 {
-    HttpAddRequestHeadersA(hHttpRequest, headerStr.c_str(), (DWORD)headerStr.length(), HTTP_ADDREQ_FLAG_REPLACE | HTTP_ADDREQ_FLAG_ADD);
+    if (!HttpAddRequestHeadersA(hHttpRequest, headerStr.c_str(), (DWORD)headerStr.length(), HTTP_ADDREQ_FLAG_REPLACE | HTTP_ADDREQ_FLAG_ADD))
+        AWS_LOG_FATAL(GetLogTag(), "Failed to add HTTP request headers!");
```
JonathanHenson: same here.

PR #492, opened 2017-04-06
Fixing ClientConfiguration::verifySSL having no effect on Windows

imosk:
ClientConfiguration.verifySSL value used to have no effect with clients WinHttpSyncHttpClient and WinINetSyncHttpClient since you cannot set INTERNET_OPTION_SECURITY_FLAGS on session handles. This pull request makes INTERNET_OPTION_SECURITY_FLAGS to be set on request handles.

### DIFF
--- a/aws-cpp-sdk-core/source/http/windows/WinHttpSyncHttpClient.cpp
+++ b/aws-cpp-sdk-core/source/http/windows/WinHttpSyncHttpClient.cpp
@@ -40,7 +40,7 @@ using namespace Aws::Utils::Logging;
 static const uint32_t HTTP_REQUEST_WRITE_BUFFER_LENGTH = 8192;
 
 WinHttpSyncHttpClient::WinHttpSyncHttpClient(const ClientConfiguration& config) :
-    Base()
+    Base(), m_verifySSL(config.verifySSL)
 {
     AWS_LOGSTREAM_INFO(GetLogTag(), "Creating http client with user agent " << config.userAgent << " with max connections " << config.maxConnections 
         << " request timeout " << config.requestTimeoutMs << ",and connect timeout " << config.connectTimeoutMs);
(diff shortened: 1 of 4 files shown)


aws-cpp-sdk-core/source/http/windows/WinINetSyncHttpClient.cpp
```diff
@@ -115,6 +106,15 @@ void* WinINetSyncHttpClient::OpenRequest(const Aws::Http::HttpRequest& request,
         ss.str().c_str(), nullptr, nullptr, accept, requestFlags, 0);
     AWS_LOGSTREAM_DEBUG(GetLogTag(), "HttpOpenRequestA returned handle " << hHttpRequest);
 
+    if (!m_verifySSL)
+    {
+        AWS_LOG_WARN(GetLogTag(), "Turning ssl unknown ca verification off.");
+        DWORD flags = SECURITY_FLAG_IGNORE_UNKNOWN_CA | INTERNET_FLAG_IGNORE_CERT_CN_INVALID;
+
+        if (!InternetSetOptionA(hHttpRequest, INTERNET_OPTION_SECURITY_FLAGS, &flags, sizeof(flags)))
```
JonathanHenson: I can't imagine this is correct. The connection has already been made at this point. Am I missing something in the documentation?

aws-cpp-sdk-core/source/http/windows/WinHttpSyncHttpClient.cpp
```diff
@@ -133,6 +125,15 @@ void* WinHttpSyncHttpClient::OpenRequest(const Aws::Http::HttpRequest& request,
     HINTERNET hHttpRequest = WinHttpOpenRequest(connection, StringUtils::ToWString(HttpMethodMapper::GetNameForHttpMethod(request.GetMethod())).c_str(),
         wss.c_str(), nullptr, nullptr, accept, requestFlags);
 
+    if (!m_verifySSL)
```
JonathanHenson: After the tcp connection has already been made? We're at layer 7 already before setting these options.

rawmaze: To be honest, I am not an expert on the topic. But...
Take a look here https://msdn.microsoft.com/en-us/library/windows/desktop/aa384066(v=vs.85).aspx - the big table almost at the bottom of the page. According to it, WINHTTP_OPTION_SECURITY_FLAGS does not work on session handles. It was confirmed by the fact the old code didn't work. Every time I tested it, it reached this row 'AWS_LOG_FATAL(GetLogTag(), "Failed to turn ssl cert ca verification off.");' and the verifySSL flag had no effect at all.
On the other hand, the proposed code change works just fine and according to Microsoft's documentation it makes sense to be written this way.

crusader-mike:
I just spent a day trying to figure out why my code works fine on Win7, but fails on variety of newer versions. 

There is a code in SDK:
```
        if (!WinHttpSetOption(GetOpenHandle(), WINHTTP_OPTION_SECURITY_FLAGS, &flags, sizeof(flags)))
            AWS_LOG_FATAL(GetLogTag(), "Failed to turn ssl cert ca verification off.");
```

Many questions I have... Why this logging wasn't in earlier SDK versions? Why it logs error, but doesn't log GetLastError? Why it logs but continues without taking any meaningful action? Why I have to dig in log files to figure out reason for error that happens 5 miles away from it's cause?

And most importantly -- why WinHttpSetOption() fails in newer WinHTTP versions?

Jonathan, if you see this -- I'd really appreciate if you have a look at #542 and #537. I need my own sdk where exceptions aren't off.

rupal-bq:
We are using aws sdk for connecting to Opendistro for Elasticsearch. This issue is causing failure with default options. What is the plan for fixing this issue?

11destiny (approved):
good

PR #494, opened 2017-04-06
aws-sdk-cpp/core: Fix proxy when using schemes other than HTTPS

pcacjr:
There was no way to tell which scheme to use for the proxy settings, so
the user sets ClientConfiguration::scheme to HTTPS for making HTTPS
requests to S3, it also sets HTTPS for the proxy host -- which does not
make any sense. Given that, on some versions of Windows 10 (e.g. RS1),
by setting a proxy host with HTTPS when it *actually* works over HTTP,
the proxy fails.

This patch should fix issue #382

Signed-off-by: Paulo Alcantara <paulo.alc.cavalcanti@hp.com>

### DIFF
--- a/aws-cpp-sdk-core/source/http/curl/CurlHttpClient.cpp
+++ b/aws-cpp-sdk-core/source/http/curl/CurlHttpClient.cpp
@@ -292,7 +292,7 @@ CurlHttpClient::CurlHttpClient(const ClientConfiguration& clientConfig) :
     Base(),   
     m_curlHandleContainer(clientConfig.maxConnections, clientConfig.requestTimeoutMs, clientConfig.connectTimeoutMs),
     m_isUsingProxy(!clientConfig.proxyHost.empty()), m_proxyUserName(clientConfig.proxyUserName),
-    m_proxyPassword(clientConfig.proxyPassword), m_proxyHost(clientConfig.proxyHost),
+    m_proxyPassword(clientConfig.proxyPassword), m_proxyScheme(SchemeMapper::ToString(clientConfig.proxyScheme)), m_proxyHost(clientConfig.proxyHost),
     m_proxyPort(clientConfig.proxyPort), m_verifySSL(clientConfig.verifySSL), m_caPath(clientConfig.caPath),
     m_caFile(clientConfig.caFile), m_allowRedirects(clientConfig.followRedirects)
 {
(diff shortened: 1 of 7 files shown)


mmustafa-hp:
Can you please review? We need this change and would like to avoid creating our own build of the sdk.

pcacjr:
Hi Jonathan,

You're right. I was only worried about to fix our side -- which eventually uses WinHTTP API. Sorry for that. I'll go ahead and then change the other http implementations to start using the new proxyScheme field.

Thanks,
-Paulo

pcacjr:
Hi Jonathan,

I've added two commits on top of the original one that makes use of the new proxyScheme field for CurlHttpClient and WinInetSyncHttpClient implementations. IXmlHttpRequest2HttpClient does not support user defined proxy -- it relies on system default settings, so neither host or scheme could be used with it.

pcacjr:
@JonathanHenson ping? Are these changes what you expected?

Thanks,
-Paulo

aws-cpp-sdk-core/source/http/curl/CurlHttpClient.cpp
```diff
@@ -408,7 +408,7 @@ std::shared_ptr<HttpResponse> CurlHttpClient::MakeRequest(HttpRequest& request,
 
         if (m_isUsingProxy)
         {
-            curl_easy_setopt(connectionHandle, CURLOPT_PROXY, m_proxyHost.c_str());
+            curl_easy_setopt(connectionHandle, CURLOPT_PROXY, (m_proxyScheme + "://" + m_proxyHost).c_str());
```
JonathanHenson: This will likely not be portable across compilers. Aws::StringStream would probably be a better bet.

pcacjr: Yeah, probably. Thanks for catching that up. I will do the proposed change in a bit and update the pull request accordingly.

pcacjr:
Hi Jonathan,

I used Aws::StringStream class to build up the proxy URL in CurHttpClient as well as rebased the branch against master. Please, review it.

Thanks,
-Paulo

pcacjr:
Ping? :-)

marcomagdy:
Will merge this today.
Thank you.

PR #497, opened 2017-04-12
Allow non-custom memory management on Android.

chris-doclily:
On Android, do not set CUSTOM_MEMORY_MANAGEMENT to 1 if it is already set to 0.


bretambrose:
Looks fine.

PR #506, opened 2017-04-20
platform/linux: link against libatomic.so if needed

andred:
Some platforms, notably mips/mipsel, need to link against libatomic
in order to be able to use atomics.

Enhance the build to achieve just that.

We have to make sure to pass the same C++ standard as for the rest
of the compilation, and for old CMake versions we have to enable
support for C because of bugs in CMake.

Signed-off-by: André Draszik <git@andred.net>


andred:
Updated against current master

PR #362, opened 2016-12-05
Fix #361 adding conditions to differentiate OpenSSL v1.0.2 and v1.1.x in CryptoImpl

jliang1010:
OpenSSL v1.1 issue has been mentioned in both #361

PR #507, opened 2017-04-20
Openssl 1.1

andred:
I can't push to #362, so this is an updated version.

Compared to #362, this fixes a few issues detected so far:
- code involving EVP_CIPHER_CTX_copy() has been fixed
- EVP_CIPHER_CTX_init() / EVP_CIPHER_CTX_reset() / OpenSSLCipher::Reset() usages have been simplified in the OpenSSL 1.1.x case
- class Sha256HMACOpenSSLImpl allocates OpenSSL objects in its constructor now as requested
- this supports LibreSSL again

### DIFF
--- a/aws-cpp-sdk-core/include/aws/core/utils/crypto/openssl/CryptoImpl.h
+++ b/aws-cpp-sdk-core/include/aws/core/utils/crypto/openssl/CryptoImpl.h
@@ -191,7 +191,8 @@ namespace Aws
 
                 virtual size_t GetKeyLengthBits() const = 0;
 
-                EVP_CIPHER_CTX m_ctx;
+                EVP_CIPHER_CTX *m_ctx;
+                EVP_CIPHER_CTX _m_ctx;
 
                 void CheckInitEncryptor();
                 void CheckInitDecryptor();
--- a/aws-cpp-sdk-core/include/aws/core/utils/crypto/openssl/CryptoImpl.h
+++ b/aws-cpp-sdk-core/include/aws/core/utils/crypto/openssl/CryptoImpl.h
@@ -105,7 +105,9 @@ namespace Aws
 
             private:
                 HMAC_CTX *m_ctx;
+#if OPENSSL_VERSION_NUMBER < 0x10100003L
                 HMAC_CTX _m_ctx;
+#endif
             };
 
             /**
@@ -196,7 +198,9 @@ namespace Aws
                 virtual size_t GetKeyLengthBits() const = 0;
 
                 EVP_CIPHER_CTX *m_ctx;
+#if OPENSSL_VERSION_NUMBER < 0x10100003L
                 EVP_CIPHER_CTX _m_ctx;
+#endif
 
                 void CheckInitEncryptor();
                 void CheckInitDecryptor();
--- a/aws-cpp-sdk-core/include/aws/core/utils/crypto/openssl/CryptoImpl.h
+++ b/aws-cpp-sdk-core/include/aws/core/utils/crypto/openssl/CryptoImpl.h
@@ -21,6 +21,7 @@
 #include <aws/core/utils/GetTheLights.h>
 #include <openssl/ossl_typ.h>
 #include <openssl/evp.h>
+#include <openssl/hmac.h>
 #include <openssl/rand.h>
 #include <atomic>
 #include <mutex>
@@ -96,12 +97,15 @@ namespace Aws
             {
             public:
 
-                Sha256HMACOpenSSLImpl()
-                { }
+                Sha256HMACOpenSSLImpl();
 
-                virtual ~Sha256HMACOpenSSLImpl() = default;
+                virtual ~Sha256HMACOpenSSLImpl();
 
                 virtual HashResult Calculate(const ByteBuffer& toSign, const ByteBuffer& secret) override;
+
+            private:
+                HMAC_CTX *m_ctx;
+                HMAC_CTX _m_ctx;
             };
 
             /**
--- a/aws-cpp-sdk-core/include/aws/core/utils/crypto/openssl/CryptoImpl.h
+++ b/aws-cpp-sdk-core/include/aws/core/utils/crypto/openssl/CryptoImpl.h
@@ -26,6 +26,11 @@
 #include <atomic>
 #include <mutex>
 
+#if defined(LIBRESSL_VERSION_NUMBER) && (OPENSSL_VERSION_NUMBER == 0x20000000L)
+#undef OPENSSL_VERSION_NUMBER
+#define OPENSSL_VERSION_NUMBER 0x1000107fL
+#endif
+
 namespace Aws
 {
     namespace Utils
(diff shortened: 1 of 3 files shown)


aws-cpp-sdk-core/include/aws/core/utils/crypto/openssl/CryptoImpl.h
```diff
@@ -21,10 +21,16 @@
...
 #include <openssl/ossl_typ.h>
 #include <openssl/evp.h>
+#include <openssl/hmac.h>
 #include <openssl/rand.h>
 #include <atomic>
 #include <mutex>
 
+#if defined(LIBRESSL_VERSION_NUMBER) && (OPENSSL_VERSION_NUMBER == 0x20000000L)
+#undef OPENSSL_VERSION_NUMBER
+#define OPENSSL_VERSION_NUMBER 0x1000107fL
```
singku: How is this version number comes?

aws-cpp-sdk-core/include/aws/core/utils/crypto/openssl/CryptoImpl.h
```diff
@@ -191,7 +202,10 @@ namespace Aws
 
                 virtual size_t GetKeyLengthBits() const = 0;
 
-                EVP_CIPHER_CTX m_ctx;
+                EVP_CIPHER_CTX *m_ctx;
+#if OPENSSL_VERSION_NUMBER < 0x10100003L
```
singku: This is not good to appear in multiple locations.

aws-cpp-sdk-core/source/utils/crypto/openssl/CryptoImpl.cpp
```diff
@@ -229,19 +259,17 @@ namespace Aws
...
             OpenSSLCipher::OpenSSLCipher(OpenSSLCipher&& toMove) : SymmetricCipher(std::move(toMove)),
-                                                                   m_encDecInitialized(false)
+                                                                   m_ctx(nullptr), m_encDecInitialized(false)
             {
-                m_ctx = toMove.m_ctx;
-                toMove.m_ctx.cipher = nullptr;
-                toMove.m_ctx.cipher_data = nullptr;
-                toMove.m_ctx.engine = nullptr;
+                Init();
+                EVP_CIPHER_CTX_copy (m_ctx, toMove.m_ctx);
```
singku: In original impl, there are several fields set up to nullptr, and I checked the code of EVP_CIPHER_CTX_copy, it doesn't cleanup in arg, I guess we should call EVP_CIPHER_CTX_cleanup(toMove.m_ctx) here.

singku:
Why don't we just switch to pointer impl if OPENSSL made the type opaque?

aws-cpp-sdk-core/source/utils/crypto/openssl/CryptoImpl.cpp
```diff
@@ -411,14 +454,10 @@ namespace Aws
...
-                {
-                    EVP_CIPHER_CTX_cleanup(&m_ctx);
-                }
-
-                m_ctx.cipher = nullptr;
-                m_ctx.cipher_data = nullptr;
-                m_ctx.engine = nullptr;
+#if OPENSSL_VERSION_NUMBER < 0x10100003L
+                EVP_CIPHER_CTX_cleanup(m_ctx);
+                m_ctx = nullptr;
```
singku: After openssl 1.1 m_ctx will be acquired from heap, so here we should free it right?

aws-cpp-sdk-core/include/aws/core/utils/crypto/openssl/CryptoImpl.h
```diff
@@ -21,10 +21,16 @@
 #include <aws/core/utils/GetTheLights.h>
 #include <openssl/ossl_typ.h>
 #include <openssl/evp.h>
+#include <openssl/hmac.h>
 #include <openssl/rand.h>
 #include <atomic>
 #include <mutex>
 
+#if defined(LIBRESSL_VERSION_NUMBER) && (OPENSSL_VERSION_NUMBER == 0x20000000L)
```
andred: I don't think it makes sense to speculate what LibreSSL might or might not do in the future. If they change the API, code using LibreSSL will have to follow. This is no different behaviour than before this PR - with the current code-base, if LibreSSL decided to change the API tomorrow, things would break.

The approach suggested here is what I've seen several other open-source projects taking.
It avoids additional clutter, as it redefines the version number to correctly reflect the API that LibreSSL actually implements.

The alternative would be to have implement proper configuration checks using CMake, to detect features / APIs, rather than clutter the code with ifdefs. I don't really care about LibreSSL, so if you want I can drop this commit. It looked like a low-hanging fruit.

aws-cpp-sdk-core/include/aws/core/utils/crypto/openssl/CryptoImpl.h
andred: As per the commit message:

> ...
> While LibreSSL strives to be API compatible with OpenSSL,
> it does so against the version when it was forked, 1.0.1g
> ...

aws-cpp-sdk-core/source/utils/crypto/openssl/CryptoImpl.cpp
andred: Well spotted. EVP_CIPHER_CTX_cleanup() has been removed in OpenSSL 1.1 and instead EVP_CIPHER_CTX_reset() is to be used, so this adds another ifdef,

I'll do that.

aws-cpp-sdk-core/include/aws/core/utils/crypto/openssl/CryptoImpl.h
andred: I agree. Do you have a better suggestion? What about two subclasses, where each of them implements specifics for each OpenSSL version. I am not using C++ much day-to-day, so there might be better options?

singku:
I have pulled this PR to our local branch, after reviewing and making proper changes, we will push out to this repo, probably in early next week.

aws-cpp-sdk-core/source/utils/crypto/openssl/CryptoImpl.cpp
andred: I think this is OK. The code paths are slightly different (I think it's a bit ugly and hard to follow, actually). The idea was to be able to reuse the same EVP_CIPHER_CTX throughout the lifetime of a class OpenSSLCipher rather then freeing and re-allocating to achieve a more static memory allocation modelling, comparable to the OpenSSL 1.0 implementation.

Cleanup() is called from the destructor and from ::Reset(). For OpenSSL 1.1:
- Cleanup() itself doesn't do anything to the EVP_CIPHER_CTX
- the destructor does Cleanup(), followed by EVP_CIPHER_CTX_free()
- ::Reset() does Cleanup(), followed by Init(), which itself does EVP_CIPHER_CTX_reset()
hence we achieve re-use of the handle, rather than free() + malloc()

For OpenSSL 1.0:
- Cleanup() itself invalidates the static handle as per your comment
- the destructor does Cleanup()
- ::Reset() does Cleanup(), followed by Init(), which itself reinitialises the static handle

andred:
> Why don't we just switch to pointer impl if OPENSSL made the type opaque?

OpenSSL 1.0 doesn't have APIs for allocation: https://github.com/aws/aws-sdk-cpp/pull/362#discussion_r111515671

singku:
We have merged this pull request and did several changes. Check out the latest code in master branch.

Issue #509, opened 2017-04-23
Usage of CMAKE_SOURCE_DIR prevents building as part of another project


ComicSansMS:
The CMake scripts make use of `CMAKE_SOURCE_DIR` in a couple of places to determine paths. This variable always points to the top-most project of the current build. Therefore, when pulling the SDK into another build via `add_subdirectory`, these paths will break.

I propose to instead use a variable that is always guaranteed to point to the SDK root folder (even if that is not the same as the current CMake root folder). For instance, replacing all current occurrences of `CMAKE_SOURCE_DIR` with `AWS_NATIVE_SDK_ROOT` gets rid of the problem.

ComicSansMS:
I opened a pull request (#510) that fixes this for desktop builds.

There are a few places left in the android build that I did not touch, because I currently cannot test the android build on my machine.

singku:
Solved in https://github.com/awslabs/aws-sdk-cpp-staging/commit/230e5dfb51e929d47d0ccae6c9ca590d48ff625a

PR #510, opened 2017-04-23
Replaced usage of CMAKE_SOURCE_DIR with AWS_NATIVE_SDK_ROOT

ComicSansMS:
This will allow the SDK to be pulled into another project via
add_subdirectory. Note that there are still some occurrences of
CMAKE_SOURCE_DIR left in the android build.

This is a partial fix for issue #509.


singku:
Merged

PR #511, opened 2017-04-24
Fix race condition in TransferHandle::WaitUntilFinished() deadlocking

ferrouswheel:
I was debugging a rare deadlock condition we would observe on long running jobs processing data from s3. When I attached to the process it appears `TransferHandle::WaitUntilFinished()` would wait indefinitely.

From reading the code, it appears possible that `m_waitUntilFinishedSignal.notify_all();` may be called before `m_waitUntilFinishedSignal.wait(...);` if the `WaitUntilFinished` thread loses context after the conditional check on `m_status.load()`.

The fix (I think) is to broaden the scope of the lock to encompass setting the new status.

I'm not sure if the lock in `UpdateStatus` should actually be placed in a block to ensure `m_statusLock` is immediate available when `notify_all()` is called?

### DIFF
--- a/aws-cpp-sdk-transfer/source/transfer/TransferHandle.cpp
+++ b/aws-cpp-sdk-transfer/source/transfer/TransferHandle.cpp
@@ -229,6 +229,7 @@ namespace Aws
 
             if(IsTransitionAllowed(currentStatus, value))
             {
+                std::unique_lock<std::mutex> semaphoreLock(m_statusLock);
                 m_status.store(static_cast<long>(value));
 
                 if (IsFinishedStatus(value))
@@ -238,7 +239,6 @@ namespace Aws
                         CleanupDownloadStream();
                     }
 
-                    std::unique_lock<std::mutex> semaphoreLock(m_statusLock);
                     m_waitUntilFinishedSignal.notify_all();
                 }
             }
--- a/aws-cpp-sdk-transfer/source/transfer/TransferHandle.cpp
+++ b/aws-cpp-sdk-transfer/source/transfer/TransferHandle.cpp
@@ -246,9 +246,9 @@ namespace Aws
 
         void TransferHandle::WaitUntilFinished() const
         {
+            std::unique_lock<std::mutex> semaphoreLock(m_statusLock);
             if (!IsFinishedStatus(static_cast<TransferStatus>(m_status.load())) || HasPendingParts())
             {
-                std::unique_lock<std::mutex> semaphoreLock(m_statusLock);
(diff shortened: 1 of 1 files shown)


ferrouswheel:
Any thing I can do to help get this merged?

singku:
I will follow up on this with you

Eliyahu-Machluf:
Hi ferrouswheel,

Did the deadlock you encountered occur when transfer error occurred? I encountered a deadlock (from time to time), on transfer errors, and suggested a fix here: https://github.com/aws/aws-sdk-cpp/pull/520

If your case was on failure, I think my fix is better.

ferrouswheel:
@Eliyahu-Machluf afaikl the deadlock happened even on success. The file would be available on s3, but our processes would just hang in `WaitUntilFinished()`

Eliyahu-Machluf:
OK. so the fix of #520 is not relevant for the problem you encountered, as it fixes deadlock on transfer error.

singku:
@ferrouswheel After applying your changes, it's still possible incur deadlock. Thinking about this sequence:
Thread A: run line 249 (If condition succeed then ready to run line 251-252, lock & wait)
Thread B: run line 232 to line 242 (update status then lock and notify all)
Thread A: run line 251-252 (lock & wait) , now notification from Thread B has lost.

A better way is to apply you changes and move lock inside WaitUntilFinished before If condition

ferrouswheel:
@singku Yes - I see what you mean. I've made this change.

aws-cpp-sdk-transfer/source/transfer/TransferHandle.cpp
```diff
@@ -238,17 +239,16 @@ namespace Aws
...
 
         void TransferHandle::WaitUntilFinished() const
         {
+            std::unique_lock<std::mutex> semaphoreLock(m_statusLock);
             if (!IsFinishedStatus(static_cast<TransferStatus>(m_status.load())) || HasPendingParts())
             {
-                std::unique_lock<std::mutex> semaphoreLock(m_statusLock);
                 m_waitUntilFinishedSignal.wait(semaphoreLock, [this]()
                     { return IsFinishedStatus(static_cast<TransferStatus>(m_status.load())) && !HasPendingParts(); });
                 semaphoreLock.unlock();
```
singku: Is it necessary to unlock explicitly? Shouldn't the system will unlock it when lock get destructed?

singku: Is it necessary to unlock explicitly? Shouldn't the system unlock it when lock get destructed?

singku:
Merged

PR #515, opened 2017-05-04
fix deadlock case at core executor

Eliyahu-Machluf:
Hi,

I encountered a deadlock using the SDK for uploading a file to S3.
Looking at the code with debugger, i found out one thread called notify before the other thread called wait. this caused the notification to get lost.

the fix (I think) is to lock before notifying the thread, and to test again there are no tasks before entering sleep. I added another change here, to wait up to 5 seconds, as I think this is better checking the condition from time to time and not relaying only on the notification.

### DIFF
--- a/aws-cpp-sdk-core/source/utils/threading/Executor.cpp
+++ b/aws-cpp-sdk-core/source/utils/threading/Executor.cpp
@@ -39,13 +39,15 @@ PooledThreadExecutor::PooledThreadExecutor(size_t poolSize, OverflowPolicy overf
 
 PooledThreadExecutor::~PooledThreadExecutor()
 {
-    for(auto threadTask : m_threadTaskHandles)
-    {
-        threadTask->StopProcessingWork();       
-    }
+	{
+		std::unique_lock<std::mutex> locker(m_syncPointLock);
+		for(auto threadTask : m_threadTaskHandles)
+		{
+			threadTask->StopProcessingWork();       
+		}
 
-    m_syncPoint.notify_all();
-
+		m_syncPoint.notify_all();
+	}
     for (auto threadTask : m_threadTaskHandles)
     {
         Aws::Delete(threadTask);
@@ -67,6 +69,8 @@ PooledThreadExecutor::~PooledThreadExecutor()
 bool PooledThreadExecutor::SubmitToThread(std::function<void()>&& fn)
 {
     //avoid the need to do copies inside the lock. Instead lets do a pointer push.
+
+
     std::function<void()>* fnCpy = Aws::New<std::function<void()>>(POOLED_CLASS_TAG, std::forward<std::function<void()>>(fn));
 
     {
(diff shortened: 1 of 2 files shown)


marcomagdy:
There's a threading bug in the S3 TransferManager.
See https://github.com/aws/aws-sdk-cpp/pull/511

Eliyahu-Machluf:
Hi,

 #511 is another issue, I also encountered deadlock at WaitUntilFinished, and I think #520  is the right fix for that.

However regarding the fixes at this pull request I see I used namespace std::chrono_literals for '5s' for 5 seconds. however the chrono_literals namespace is valid only for C++14. So if this fix is merged, it shall be replaced with std::chrono::seconds(5)

Eliyahu-Machluf:
I closed this pull request by mistake, now reopened

aws-cpp-sdk-core/source/utils/threading/ThreadTask.cpp
```diff
@@ -44,9 +46,10 @@ void ThreadTask::MainTaskRunner()
         }
      
         std::unique_lock<std::mutex> locker(m_executor.m_syncPointLock);
-        if(m_continue)
+		
+        if(m_continue && !m_executor.HasTasks())
         {
-            m_executor.m_syncPoint.wait(locker);    
+            m_executor.m_syncPoint.wait_for(locker, 5s);
```
JonathanHenson: Isn't 5s c++14 only? would std::chrono::seconds(5) be better?

aws-cpp-sdk-core/source/utils/threading/Executor.cpp
```diff
@@ -80,6 +84,7 @@ bool PooledThreadExecutor::SubmitToThread(std::function<void()>&& fn)
         m_tasks.push(fnCpy);
     }
 
+	std::unique_lock<std::mutex> locker(m_syncPointLock);
```
singku: notify_one() here could cause false alarm (but harmless), since other thread may immediately fetch the task after line 84 executed. It's better to lock syncPointLock first before m_tasks.push then notify_one()

singku: The background is Executor create many threads to fetch and run task from taskQueue of Executor.
Also Executor is responsible for set up the threads' status: continue or not.
Check out ThreadTask::MainTaskRunner()

Without lock:
1: Thread A test m_continue and get true then get ready into waiting state but not yet
2: Executor run stopProcessingWork() and notify_all
3: Thread A go into waiting state

Now Thread A will never be waken up

marcomagdy: I see the problem now. I disagree with the solution though.
I think it would be much better to encapsulate that synchronization logic in a Semaphore/AutoResetEvent class. I'll have something to review by tomorrow.
Thanks

singku:
This has been merged

Issue #527, opened 2017-05-15
GNUInstallDirs.cmake included() too early


andred:
Because GNUInstallDirs is included before the call to project(), it doesn't actually work (except when using the CMake binaries from Ubuntu).

GNUInstallDirs uses CMAKE_SYSTEM_NAME, which is defined only after project().

It works in Ubuntu, because they add a local hack which ends up bypassing the relevant bits of GNUInstallDirs.cmake, but when using CMake from source, or when using any other distribution, the way GNUInstallDirs is used in the AWS C++ SDK doesn't work.

singku:
Could you please elaborate which purpose you want to achieve by using CMAKE_SYSTEM_NAME? If it is empty, cmake script just go into different branch after if condition

andred:
If you look at GNUInstallDirs.cmake (the original version, not Ubuntu's modified version), it uses CMAKE_SYSTEM_NAME to determine if it is compiling for Linux at all https://github.com/Kitware/CMake/blob/master/Modules/GNUInstallDirs.cmake#L220 . After that, it determines the OS, debian and derivates https://github.com/Kitware/CMake/blob/master/Modules/GNUInstallDirs.cmake#L222 or something else https://github.com/Kitware/CMake/blob/master/Modules/GNUInstallDirs.cmake#L232

Given CMAKE_SYSTEM_NAME is empty if GNUInstallDirs is included before project(), the AWS C++ SDK is installed into the wrong location.

PR #528, opened 2017-05-15
cmake: GNUInstallDirs must be included after project()

andred:
GNUInstallDirs uses CMAKE_SYSTEM_NAME, which is defined only
after project()

It works in Ubuntu, because they add a local hack which ends
up bypassing the relevant bits of GNUInstallDirs.cmake, but
when using CMake from source, or when using any other
distribution, the way GNUInstallDirs is used in the AWS C++
SDK doesn't work.

By moving the call to GNUInstallDirs into the
apply_post_project_platform_settings() macro, things seem to
work OK.

Tested on Debian Jessie/Stretch/Sid as well as
Ubuntu Trusty/Xenial.

See #527

Signed-off-by: André Draszik <git@andred.net>

### DIFF
--- a/cmake/platform/linux.cmake
+++ b/cmake/platform/linux.cmake
@@ -2,8 +2,6 @@
 message(STATUS "Generating linux build config")
 set(SDK_INSTALL_BINARY_PREFIX "linux")
 
-include(GNUInstallDirs)
-
 if(SIMPLE_INSTALL)
     include(CMakePackageConfigHelpers)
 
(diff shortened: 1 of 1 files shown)


singku:
On SIMPLE_INSTALL build, CMake will override several variables (CMAKE_INSTALL_[dir]) after these variable are available via call include(GNUInstallDirs). If moving this to post_project_platform_setting, these overrides will become invalid.

If adding include(CMakeDetermineSystem) at line 76 of CMakeLists.txt solve your problem, we can apply this change.

andred:
I have only tested using SIMPLE_INSTALL. It seems to have done correct thing.

I think my proposal works because of the following: While platform/linux.cmake sets ARCHIVE_DIRECTORY BINARY_DIRECTORY LIBRARY_DIRECTORY and INCLUDE_DIRECTORY on SIMPLE_INSTALL builds, all of these are unconditionally overridden anyway in the main CMakeLists.txt *after* platform/linux.cmake has executed

singku:
@andred You are right

singku:
@andred We have tested either on SIMPLE_INSTALL or normal INSTALL on different platforms. The SDK was installed in the correct directory on all platform. Could you please point out how it was wrong on your behalf?

singku:
merged

PR #548, opened 2017-05-30
Fixes NOMINMAX compilation error

supermassive:
Fixes compilation error when building with NOMINMAX global define(which disables min() from Windows.h). Also now #include <algorithm> actually makes sense.

### DIFF
--- a/aws-cpp-sdk-core/source/http/windows/WinConnectionPoolMgr.cpp
+++ b/aws-cpp-sdk-core/source/http/windows/WinConnectionPoolMgr.cpp
@@ -128,7 +128,7 @@ bool WinConnectionPoolMgr::CheckAndGrowPool(const Aws::String& host, HostConnect
     if (connectionContainer.currentPoolSize < m_maxConnectionsPerHost)
     {
         unsigned multiplier = connectionContainer.currentPoolSize > 0 ? connectionContainer.currentPoolSize : 1;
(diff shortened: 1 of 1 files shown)


marcomagdy:
Thanks. Will merge it soon.

aws-cpp-sdk-core/source/http/windows/WinConnectionPoolMgr.cpp
```diff
@@ -128,7 +128,7 @@ bool WinConnectionPoolMgr::CheckAndGrowPool(const Aws::String& host, HostConnect
     if (connectionContainer.currentPoolSize < m_maxConnectionsPerHost)
     {
         unsigned multiplier = connectionContainer.currentPoolSize > 0 ? connectionContainer.currentPoolSize : 1;
-        unsigned amountToAdd = min(multiplier * 2, m_maxConnectionsPerHost - connectionContainer.currentPoolSize);
+        unsigned amountToAdd = std::min(multiplier * 2, m_maxConnectionsPerHost - connectionContainer.currentPoolSize);
```
marcomagdy: This breaks the build for people who do not define NOMINMAX. So can you revise this to use the following syntax?
``(std::min)(...)``

singku:
@supermassive If you have confusion as I did why this parentheses works, check [here](https://stackoverflow.com/questions/5004858/stdmin-gives-error/5004893) and [this](https://stackoverflow.com/questions/13416418/define-nominmax-using-stdmin-max/13420838)

PR #549, opened 2017-05-30
Fixes UNICODE build on Windows

supermassive:
Explicitly call ANSI version of GetFileVersionInfoSize which is also consistent with other API calls in this function.


marcomagdy:
Will merge, thanks.

PR #570, opened 2017-06-14
Get rid of unused variable compiler warning

jt70471:
I don't know if this is the most elegant way to accomplish this, but this approach successfully prevents the unused variable compiler warning for m_giveSomeSize.

### DIFF
--- a/aws-cpp-sdk-core/include/aws/core/NoResult.h
+++ b/aws-cpp-sdk-core/include/aws/core/NoResult.h
@@ -22,9 +22,8 @@ namespace Aws
 
 class AWS_CORE_API NoResult
 {
-private:
-    //TODO: Remove this once we figure out how to make an empty class compile in a move on gcc.
-    int m_giveSomeSize;
+public:
+    virtual ~NoResult() {};
 };
 
 } // namespace Aws


singku:
What compiler are you using and what's the version and on which platform?

Bu11etmagnet:
Why not get rid of m_giveSomeSize? What does it try to accomplish (work around) ?

jt70471:
I'm using macOS with Xcode 8.3.2, which uses the clang 802.0.42 version.

jt70471:
@Bu11etmagnet, the code has the following comment, so must be related to GCC compiler if the class has no variables or methods (i.e. empty class):

    //TODO: Remove this once we figure out how to make an empty class compile in a move on gcc.
    int m_giveSomeSize;

singku:
@jt70471 I used the same version of clang on my mac, but there is no such compiler warning of "unused variable". Would you mind show your compilation options?

jt70471:
@singku, attached is the application compile listing where I'm receiving the unused variable warning. Perhaps my original post is/was not clear, I receive the working when I have an application that's using the library, not during the library cmake/make build itself.

[appmake.txt](https://github.com/aws/aws-sdk-cpp/files/1075515/appmake.txt)

singku:
Then I am wondering why building the SDK itself doesn't throw that warning but when building your own project.

singku:
I see, the SDK itself [turned off the warning](https://github.com/aws/aws-sdk-cpp/blob/master/cmake/compiler_settings.cmake#L74)

You can inherit the compiler flags generated by sdk's build to silence this warning.
Check those pkgconfig files such as:

```
cat /usr/local/lib64/pkgconfig/aws-cpp-sdk-core.pc
prefix=/usr/local
exec_prefix=${prefix}
includedir=${prefix}/include
libdir=${exec_prefix}/lib

Name: aws-cpp-sdk-core
Description: Core http and utility library for the AWS C++ SDK
Version: 1.0.110
Cflags: -I${includedir}   -Wall -Werror -pedantic -Wextra -Wno-unused-private-field -fno-exceptions -ggdb3 -std=c++11
Libs: -L${libdir} -laws-cpp-sdk-core
Libs.private: -lpthread -lcurl
Requires:
```

jt70471:
@singku, thanks for the information. I know that I can turn off that particular compiler warning, but I prefer to use -Wall and keep my code warning free. Perhaps overkill, but a personal preference that I'm sure is shared by many.

If it's not something that can be addressed in the Aws C++ SDK itself, I'll just keep the fork that I created and have patched already.

There may be other ways to prevent the warning, but I don't know what the original issue is with GCC that caused Aws to add the TODO and member variable in the first place.

JonathanHenson:
the original reason was because GCC wouldn't compile a class of size 0. The better solution is to drop a vtable in via a virtual destructor.

jt70471:
@JonathanHenson thanks for the suggestion, is that something planned for the SDK or would you prefer that I try to implement that and test it. I can only test macOSX/clang setup, but happy to do that and submit another pull request if that's what you'd like me to do.

jt70471:
Pardon my git trip down the rabbit hole to get this sorted out, but I have finally gotten it all sorted (I think). The virtual destructor approach mentioned by @JonathanHenson  seems to resolve the issue and the pull request has been reworked.

I have successfully tested this on macOSX/clang, however, I unfortunately can't test on Linux or Windows.

Thanks for considering this for inclusion into the SDK.

jt70471:
@JonathanHenson, I changed this to implement the virtual destructor as you suggested. Is that what you were looking for? Thanks

singku:
merged

PR #167, opened 2016-04-21
Retry S3 InternalError and RequestTimeout errors

edolstra:
This fixes "Unable to parse ExceptionName: InternalError Message: We encountered an internal error. Please try again." and "Unable to parse ExceptionName: RequestTimeout Message: Your socket connection to the server was not read from or written to within the timeout period. Idle connections will be closed." exceptions when doing an S3 request. Both errors now cause the request to be retried.

PR #576, opened 2017-06-20
Retry S3 InternalError and RequestTimeout errors

edolstra:
This is an updated version of https://github.com/aws/aws-sdk-cpp/pull/167, which added support for the S3 InternalError, RequestTimeout and SlowDown exceptions. @JonathanHenson's comment at https://github.com/aws/aws-sdk-cpp/pull/167#issuecomment-269376107 suggested that this had been merged in some way, but in fact, only support for the SlowDown exception was added (in 56385be06b).

I haven't seen RequestTimeout from S3 lately, but I see InternalError all the time, so the library should definitely recognize it and retry requests.

### DIFF
--- a/aws-cpp-sdk-core/source/client/CoreErrors.cpp
+++ b/aws-cpp-sdk-core/source/client/CoreErrors.cpp
@@ -29,6 +29,7 @@ static const int INVALID_SIGNATURE_EXCEPTION_HASH = HashingUtils::HashString("In
 static const int INVALID_SIGNATURE_HASH = HashingUtils::HashString("InvalidSignature");
 static const int INTERNAL_FAILURE_HASH = HashingUtils::HashString("InternalFailure");
 static const int INTERNAL_SERVER_ERROR_HASH = HashingUtils::HashString("InternalServerError");
+static const int INTERNAL_ERROR_HASH = HashingUtils::HashString("InternalError");
 static const int INVALID_ACTION_HASH = HashingUtils::HashString("InvalidAction");
 static const int INVALID_CLIENT_TOKEN_ID_HASH = HashingUtils::HashString("InvalidClientTokenId");
 static const int INVALID_CLIENT_TOKEN_ID_EXCEPTION_HASH = HashingUtils::HashString("InvalidClientTokenIdException");
@@ -80,7 +81,7 @@ AWSError<CoreErrors> CoreErrorsMapper::GetErrorForName(const char* errorName)
   {
       return AWSError<CoreErrors>(CoreErrors::INVALID_SIGNATURE, false);
   }
-  else if (errorHash == INTERNAL_FAILURE_HASH || errorHash == INTERNAL_SERVER_ERROR_HASH)
+  else if (errorHash == INTERNAL_FAILURE_HASH || errorHash == INTERNAL_SERVER_ERROR_HASH || errorHash == INTERNAL_ERROR_HASH)
   {
     return AWSError<CoreErrors>(CoreErrors::INTERNAL_FAILURE, true);
   }
--- a/aws-cpp-sdk-core/source/client/CoreErrors.cpp
+++ b/aws-cpp-sdk-core/source/client/CoreErrors.cpp
@@ -68,6 +68,7 @@ static const int INVALID_ACCESS_KEY_ID_HASH = HashingUtils::HashString("InvalidA
 static const int INVALID_ACCESS_KEY_ID_EXCEPTION_HASH = HashingUtils::HashString("InvalidAccessKeyIdException");
 static const int REQUEST_TIME_TOO_SKEWED_HASH = HashingUtils::HashString("RequestTimeTooSkewed");
 static const int REQUEST_TIME_TOO_SKEWED_EXCEPTION_HASH = HashingUtils::HashString("RequestTimeTooSkewedException");
+static const int REQUEST_TIMEOUT_HASH = HashingUtils::HashString("RequestTimeout");
 
 AWSError<CoreErrors> CoreErrorsMapper::GetErrorForName(const char* errorName)
 {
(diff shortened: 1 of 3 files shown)


aws-cpp-sdk-core/source/client/CoreErrors.cpp
```diff
@@ -164,6 +166,10 @@ AWSError<CoreErrors> CoreErrorsMapper::GetErrorForName(const char* errorName)
   {
     return AWSError<CoreErrors>(CoreErrors::REQUEST_TIME_TOO_SKEWED, true);
   }
+  else if (errorHash == REQUEST_TIMEOUT_HASH)
+  {
+    return AWSError<CoreErrors>(CoreErrors::REQUEST_TIMEOUT, true);
```
marcomagdy: Timeout errors are not retry-able. Change the last argument to `false`

marcomagdy: One thing is missing. Update `CoreErrorsMapper::GetErrorForHttpResponseCode` to map `HttpResponseCode::REQUEST_TIMEOUT` to `CoreErrors::REQUEST_TIMEOUT`

copumpkin:
The `InternalError` one is biting me repeatedly today. Can we get this merged?

aws-cpp-sdk-core/source/client/CoreErrors.cpp
copumpkin: I've seen a fair number of S3 calls time out (usually over unauthenticated HTTP) and then work fine straight afterwards. I currently have to overlay timeout retries one layer up from the SDK but it'd be nice for this to at least be optional. Or if not, can you elaborate on why they aren't retriable?

edolstra: Done in f37e1fff272baf4a9669ed6d593f58fb50a81ad7.

marcomagdy: HTTP 408 Request Timeout response status code means that the server would like to shut down this unused connection. It's basically an error on the client side. All 4xx errors are not retry-able by the SDK.

marcomagdy:
@copumpkin I'm more than happy to merge it. But the 408 error should not be retry-able.

copumpkin:
The PR is not my code, but 408 seems to be rather unique in the [4xx client-screwed-up category](https://en.wikipedia.org/wiki/List_of_HTTP_status_codes#4xx_Client_errors):

> 408 Request Timeout
> The server timed out waiting for the request. According to HTTP specifications: "The client did not produce a request within the time that the server was prepared to wait. The client MAY repeat the request without modifications at any later time."[40]

I think this is because it's one of the few where weirdness between the client server gets interpreted by the server as "client stopped talking to me; must be their fault", when actually, it's rarely actually a client-side bug leading to the issue, especially when there are often fallible load balancers and proxies sitting between the client and the server nowadays. 

Anyway, all I'm saying (and that @edolstra is saying I guess) is that we've seen this in the wild be a retriable condition. Either we have to retry by hand or the SDK can save us a bit of work. I won't keep pushing it, but I don't think it's ridiculous for the SDK to treat it as a retriable condition, and would actually save a few of us some work having to wrap an API that retries with more manual retries.

copumpkin:
Just as further precedent, it looks like aws-sdk-go (official) retries on `RequestTimeoutException` (which is 408) as well as 504 `ResponseTimeout`: https://github.com/aws/aws-sdk-go/blob/master/aws/request/retryer.go#L32

As far as other SDKs and timeouts, the story is inconsistent:

- [`aws-sdk-java` only seems to retry on 504](https://github.com/aws/aws-sdk-java/blob/master/aws-java-sdk-core/src/main/java/com/amazonaws/retry/RetryUtils.java#L53)
- I couldn't figure out what `botocore` does: it seems to handle a generic [`Timeout`](https://github.com/boto/botocore/blob/develop/botocore/retryhandler.py#L33) exception from 
`requests` which as far as I can tell encapsulates any sort of timeout error, but I'm not sure if it corresponds to particular HTTP error codes or not
- [`aws-sdk-php` only seems to retry on 504](https://github.com/aws/aws-sdk-php/blob/master/src/RetryMiddleware.php#L20)
- [`aws-sdk-net` seems to only retry on 408, oddly enough](https://github.com/aws/aws-sdk-net/blob/master/sdk/src/Core/Amazon.Runtime/Pipeline/RetryHandler/DefaultRetryPolicy.cs#L88)
- [`aws-sdk-ios` doesn't retry on either, unless I'm missing something](https://github.com/aws/aws-sdk-ios/blob/master/AWSCore/Serialization/AWSURLRequestRetryHandler.m#L124-L132)
- [`aws-sdk-ruby` seems to retry on both (the 5xx check is further down)](https://github.com/aws/aws-sdk-ruby/blob/master/gems/aws-sdk-core/lib/aws-sdk-core/plugins/retry_errors.rb#L47)

@marcomagdy I got sick of hunting down retry logic in the other SDKs, but I think it's clear that there isn't an obvious rule in favor or against retrying on 408. Given that we've both reported seeing retriable 408s in the wild and that (I think) I've provided a reasonable justification for retrying it above, can we leave the 408 retry in this PR and merge? Without it, I just have to wrap AWS's APIs in my own retry logic, and it can be painful to figure out that my AWS API failure was caused by a 408.

marcomagdy:
I also asked around the SDK teams and found discrepancies in behavior.
I'll merge this commit as it's obvious customer pain point, and we'll also try to standardize the behavior across the different SDKs.
Thank you.

copumpkin:
Thank you so much! This will simplify a lot of things on my side :)

copumpkin:
Will this make it into the next release?

singku:
This has been released. Thank you for your contribution.

copumpkin:
@singku hmm, I don't see the commit in master or in any released version. Where should I look?

copumpkin:
Oh I see, it's in https://github.com/aws/aws-sdk-cpp/commit/b05dfe47aa800a203b29a47c5efefba44d9ece7e. Any reason you strip the author on the commit?

singku:
https://github.com/aws/aws-sdk-cpp/commit/b05dfe47aa800a203b29a47c5efefba44d9ece7e

singku:
There are multiple commits of this PR and I had to merge to internal repo, then squash merge to this public repo. Unfortunately, squash merge erases the author's info. Next time, I will ask author to make it one commit.

copumpkin:
Awesome, thanks :smile:

PR #602, opened 2017-07-05
JsonValue : uninitialized member variable in copy process

yoshihiro-kato:
Hi, I found bug. Please check this.

[Detail]
When JsonValue created from copy constructor, WasParseSuccessful method return false. Because,  copy constructor does not copy m_wasParseSuccessful.


marcomagdy:
Merged your patch with some refactoring. Thanks.

yoshihiro-kato:
@marcomagdy 
Thanks!

PR #604, opened 2017-07-09
fix ExpiredObjectDeleteMarker xml

jt70471:
Hi, I was receiving an invalid XML document error when trying to use WithExpiredObjectDeleteMarker on a PutBucketLifecycleConfiguration call. It would appear that the XML validation is expecting true or false instead of 1 or 0.


singku:
All files under xx/model/ are generated based on api description under code-generation/. You might want to take a look at the code there. But I will fix it for you.

jt70471:
I was unable to determine what to change under code-generation/* so if you could make the appropriate change to ensure the generated XML passes the validation, I'd appreciate it.

I'm not opposed to doing it myself if you can give me a hint specifically what to change.

Thanks

singku:
This fix will be merged to master later today. Reopen if you still have the problem

PR #606, opened 2017-07-10
removed ref from m_cryptoConfig field

supermassive:
Without that fix S3EncryptionClient will eventually have garbage in m_cryptoConfig when it is constructed from temporary argument.


Bu11etmagnet:
Except `S3EncryptionClient` can't be constructed from a temporary argument. All its constructors take `const&` arguments, which means that the caller is responsible for ensuring the objects passed to `S3EncryptionClient` have a longer lifetime than the `S3EncryptionClient`.

marcomagdy:
`const&` will bind to a temporary. @supermassive is right. It's a bug. Will merge it shortly.

PR #607, opened 2017-07-10
fixes truncation of file size

supermassive:
Found that issue doing code review(trying to implement similar things for my needs, so was looking here as an example)
On vs2015x86 tellg() result holds 64bit long long which is truncated to 32bit size_t and then passed to TransferHandle which needs uint64_t. I've removed redundant cast.

Warning: while the fix looks correct, but I actually did not try to compile that.


aws-cpp-sdk-transfer/source/transfer/TransferManager.cpp
```diff
@@ -916,7 +916,7 @@ namespace Aws
                 const Aws::String& contentType, const Aws::Map<Aws::String, Aws::String>& metadata, const Aws::String& fileName)
         {
             fileStream->seekg(0, std::ios_base::end);
-            size_t length = static_cast<size_t>(fileStream->tellg());
+            auto length = fileStream->tellg();
```
marcomagdy: Keep the static case but use the explicit type `uint64_t`

PR #608, opened 2017-07-11
Possible fix for crashing EC2 requests

hallbjorn:
My colleague had two crashes in version 1.0.98 on Windows and sent me the minidumps.

I was unable to determine exactly what happened in EC2Client::MakeRequest but the request was reported as successful and included headers. The XmlDocument payload in the result did not have a root element and this caused the crashes when calling GetName(). 

Stack traces:

>  	Aws::External::tinyxml2::StrPair::GetStr() Line 201	
>  	Aws::Utils::Xml::XmlNode::GetName() Line 59
>  	Aws::EC2::Model::DescribeInstancesResponse::operator=(const Aws::AmazonWebServiceResult<Aws::Utils::Xml::XmlDocument> & result) Line 43
>  	Aws::EC2::EC2Client::DescribeInstances(const Aws::EC2::Model::DescribeInstancesRequest & request) Line 3860
>      ...

>  	Aws::External::tinyxml2::StrPair::GetStr() Line 201
>  	Aws::Utils::Xml::XmlNode::GetName() Line 59
>  	Aws::EC2::Model::DescribeRegionsResponse::operator=(const Aws::AmazonWebServiceResult<Aws::Utils::Xml::XmlDocument> & result) Line 43
>  	Aws::EC2::EC2Client::DescribeRegions(const Aws::EC2::Model::DescribeRegionsRequest & request) Line 4190
>      ...

I added an IsNull() check for the root element to avoid the crash. However, since returning an empty result is wrong for many operations I have also made a change to cause an empty payload to be treated as an error when the result is not ignored and not being streamed.

PR #612, opened 2017-07-14
Possible fix for crashing EC2 requests (fixed git identity)

hallbjorn:
Same as #608 with fixed git identity issue.


marcomagdy:
Unfortunately the patch failed our integration tests.
some operations for some services return empty payload when successful. e.g. [S3 PUT Bucket ](http://docs.aws.amazon.com/AmazonS3/latest/API/RESTBucketPUT.html#RESTBucketPUT-requests)
I'll merge your first commit though, because that's a bug.

marcomagdy:
merged

PR #681, opened 2017-09-28
corrected windows pcm driver bit width to be 16 instead of 16000

diablodale:
Fixed a typo where someone used KHZ_16 as a bit width.
We definitely don't have 16000 bit wide audio.  ;-)
one line change to BIT_WIDTH_16


wps132230:
Merged: https://github.com/aws/aws-sdk-cpp/commit/df5d84552a3e038f6941c0337b5d1981de0efbcb

PR #683, opened 2017-10-03
string trimming fix 

pozharskiyvadim:
The behavior of std::isspace is undefined if the argument's value is neither representable as unsigned char nor equal to EOF. To use this function safely with plain chars (or signed chars), the argument should first be converted to unsigned char


wps132230:
Hi, we've fixed this issue in another commit:
https://github.com/aws/aws-sdk-cpp/commit/c210e2f6486d8b4306cd06abf3d9eea662bee94a
Please reopen this PR or submit an issue if you have other questions.

PR #685, opened 2017-10-03
Added missing virtual destructors

drivehappy:
When building under Clang 6.0.0 (trunk 314768) the following warning (promoted to error) is caught when attempting to use derived classes from Directory and SecureRandomFactory:

`error: destructor called on non-final 'DefaultSecureRandFactory' that has virtual functions but non-virtual destructor [-Werror,-Wdelete-non-virtual-dtor]`


marcomagdy:
Those changes have already been added. Sorry for the super late reply.

Issue #679, opened 2017-09-28
TextToSpeech::SendTextCompletedHandler param SynthesizeSpeechOutcome unusable


diablodale:
The `SynthesizeSpeechOutcome` param of the `SendTextCompletedHandler` callback due to `TextToSpeechManager::SendTextToOutputDevice()` is an incomplete type. Or the VC++ 2015 compiler thinks it is incomplete. I was unable to isolate the cause nor a fix.

I saw in your test code, you ignore this param. Not naming or using that parameter allows it to compile in your tests and my code both. However, in my code, I want to check this outcome and get exception names and error messages when the error is network or services related.

https://github.com/aws/aws-sdk-cpp/blob/d2542152f560264094029e33f967cdb5ac19a8d7/aws-cpp-sdk-text-to-speech-tests/TextToSpeechManagerTests.cpp#L322

Ideas?

diablodale:
I found the problem. The Outcome header wasn't included, therefore the typedef was incompete.
Easy fix. Include the header. Commit and pull request already submitted.

PR #699, opened 2017-10-19
fix aws/aws-sdk-cpp#679

diablodale:
- included the Outcome header because the
  typedef SynthesizeSpeechOutcome needed it

### DIFF
--- a/aws-cpp-sdk-polly/include/aws/polly/PollyClient.h
+++ b/aws-cpp-sdk-polly/include/aws/polly/PollyClient.h
@@ -21,6 +21,7 @@
 #include <aws/core/client/AWSClient.h>
 #include <aws/core/utils/memory/stl/AWSString.h>
 #include <aws/core/utils/json/JsonSerializer.h>
+#include <aws/core/utils/Outcome.h>
 #include <aws/polly/model/DeleteLexiconResult.h>
 #include <aws/polly/model/DescribeVoicesResult.h>
 #include <aws/polly/model/GetLexiconResult.h>


wps132230:
Hi, we have a forward declaration in `PollyClient.h`.
https://github.com/aws/aws-sdk-cpp/blob/75edd57e762606a2724a5dff564e8359c66adfee/aws-cpp-sdk-polly/include/aws/polly/PollyClient.h#L46
And you can include this header in your application cpp file, for example in `text-to-speech` tests:
https://github.com/aws/aws-sdk-cpp/blob/75edd57e762606a2724a5dff564e8359c66adfee/aws-cpp-sdk-text-to-speech-tests/TextToSpeechManagerTests.cpp#L20
Or do you have any other concerns?

diablodale:
Those are workarounds for SDK errors. The SDK needs to be coherent and present an API that works clean and clear.

The fix that I provided solves this. It is the missing header file. Without that dependent header file, the API is nonfunctional. Please see the linked issue if you are unfamiliar with it.

It is not tenable for every one of your customers to digest your SDK, trace the bug, and apply the workaround. That is a tremendous waste of time and effort.

marcomagdy:
You make a good point. Even though, technically the header is not needed to be `#include`ed in the client's header, it would make the out-of-the-box SDK experience easier. That being said, we should fix this across the SDK, and not just the Polly client.

Thanks for feedback.

diablodale:
Closing PR, no action from Amazon.

Issue #702, opened 2017-10-20
buffer size unavail for PCMOutputDriver::WriteBufferToDevice; can't easily size buffers


diablodale:
The `WriteBufferToDevice()` operations on an audio output driver are bounded by `Aws::TextToSpeech::BUFF_SIZE`. Unfortunately, this value is a const variable in a cpp file and therefore not available to consumers of the SDK.

Awareness of this maximum buffer size is very useful when creating custom drivers and driver factories. For example, such knowledge can enable drivers to correctly size ring buffers.

### Fix

Put this BUFF_SIZE value as a constexpr in the header file for its current cpp. I already have a commit ready.

### Workaround

Read the source code on every updated SDK release to see if the value has changed. Then manually change the values in one's custom code.

PR #703, opened 2017-10-20
fix aws/aws-sdk-cpp#702 expose tts buffer size

diablodale:
I used a simple constexpr as that is supported in C++11 compilers *and* most importantly the value can be used as a template argument. Making it a `static const` in a header file would have restricted it to only runtime usage. I need compile time. `constexpr` provides it also at compile-time for C++11 template goodness.

### DIFF
--- a/aws-cpp-sdk-text-to-speech/include/aws/text-to-speech/TextToSpeechManager.h
+++ b/aws-cpp-sdk-text-to-speech/include/aws/text-to-speech/TextToSpeechManager.h
@@ -38,6 +38,12 @@ namespace Aws
 		typedef Aws::Vector<OutputDevicePair> OutputDeviceList;
 
         /**
+         * Maximum bytes size of audio to be sent to PCM drivers by TextToSpeechManager in one operation
+         * with PCMOutputDriver::WriteBufferToDevice()
+         */
+        constexpr size_t BUFF_SIZE = 8192;
+
+        /**
          * Manager for rendering text to the Polly service and then sending directly to an audio driver.
          * By default this uses our best guess at the correct drivers for you operating system.
          * On windows, this is the WaveOut API.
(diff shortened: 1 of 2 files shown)


marcomagdy:
Will merge it. Thanks

marcomagdy:
Can you do me a favor, rebase and fix the typo in the commit message.

diablodale:
There's only one line of commit message. I don't see a typo in it. Please assist.
```
fix aws/aws-sdk-cpp#702 expose tts buffer size
```

I don't see a typo. The text up to 'expose' is Github specific so that the commit, issue, and PR are linked together. "tts" is short for text-to-speech.

???🤔

marcomagdy:
haha tts was little confusing to me. Do you mind spelling it out in the commit message?

diablodale:
adjusted it

marcomagdy:
So, some bad news, when I merged this PR, it didn't build on one of the compilers we still support VS2013. They don't support `constexpr`. So, unfortunately, I have to revert it using `static const size_t`

diablodale:
Should be a fairly easy fix to enable. Your SDK codebase already uses the perfect macro to detect and specially handle compiler versions. For example...

https://github.com/aws/aws-sdk-cpp/blob/bacce5952061503537fca2ee96bb5cb4f2e41a74/aws-cpp-sdk-transfer/source/transfer/TransferManager.cpp#L112

That same define can be used to detect VS2013. `_MSC_VER == 1800` for VS2013.

I can adjust the code such that if that macro is present (msft compiler) and its value is less than 1900 (its value for VS2015), then it should use static const. Otherwise, it can use constexpr.

Would your team accept compiler handling like that?

marcomagdy:
We try to limit our use of `#ifdef` macros as much as we can.
In this case you, you can still use a `const size_t` as a template parameter. https://godbolt.org/g/BQsoQa

diablodale:
Mmm, undesired merge?
I have approach number 3 ready for a spin. Basically changing constexpr to static const while keeping it in the header file so it is accessible to others.

What should I do?

Thanks for the 411 on using static const's in templates. I would not have expected the compiler to recognize a runtime during compile time given the (remote) possibility of altering the constant at runtime.

marcomagdy:
I had to commit a fix on top of yours to unblock our release pipeline. You don’t have to update the PR.

PR #730, opened 2017-11-15
Updated to correctly find body size

### LINKED ISSUE #725: PutObjectRequest + Aws::FStream give inconsistent output
### What platform/OS are you using?
Linux (Ubuntu 16.04)

### What compiler are you using? what version?
gcc 5.4

### What's your CMake arguments?
`-DBUILD_ONLY="s3"`

### Issue
The `PutObject` request gives content length as `-1` for every odd index in the following code with the file being constant:
```c++
    auto inpData = Aws::MakeShared<Aws::FStream>("PutObjectInputStream",
            path.c_str(), std::ios_base::in | std::ios_base::binary);
    auto objReq = Aws::S3::Model::PutObjectRequest();
    objReq.WithBucket(bucketName).SetBody(inpData);
    for (size_t i = 0; i < loop; ++i)
    {
        String key_suffix = std::to_string(i).c_str();
        auto objRes = client->PutObject(objReq.WithKey(key + key_suffix));
    }
```

Some details about types and variables used:
```c++
    using String = Aws::String;
    String key = "some_key", path = "SomePath";
    size_t loop = 11;
    std::shared_ptr<Aws::S3::S3Client> client;
```

### Expected behavior
Consistent content length and content for every request (provided that the file doesn't change) without instantiating new instances for every loop. However, this is my personal expectation.

### Current Solution
```c++
    using String = Aws::String;
    String key = "some_key", path = "SomePath";
    size_t loop = 11;
    for (size_t i = 0; i < loop; ++i)
    {
        auto inpData = Aws::MakeShared<Aws::FStream>("PutObjectInputStream",
                path.c_str(), std::ios_base::in | std::ios_base::binary);
        auto objReq = Aws::S3::Model::PutObjectRequest();
        objReq.WithBucket(bucketName).SetBody(inpData);
        String key_suffix = std::to_string(i).c_str();
        auto objRes = client->PutObject(objReq.WithKey(key + key_suffix));
    }
```

kunaltyagi:
Ensures that tellg doesn't return -1 because of bad bits set in an earlier operation

Fixes #725

### DIFF
--- a/aws-cpp-sdk-core/source/client/AWSClient.cpp
+++ b/aws-cpp-sdk-core/source/client/AWSClient.cpp
@@ -244,11 +244,6 @@ HttpResponseOutcome AWSClient::AttemptExhaustively(const Aws::Http::URI& uri,
             if (!m_retryStrategy->ShouldRetry(outcome.GetError(), retries)) return outcome;
         
             AWS_LOGSTREAM_WARN(AWS_CLIENT_LOG_TAG, "Request failed, now waiting " << sleepMillis << " ms before attempting again.");
-            if(request.GetBody())
-            {
-                request.GetBody()->clear();
-                request.GetBody()->seekg(0);
-            }
 
             if (request.GetRequestRetryHandler())
             {
@@ -303,6 +298,12 @@ HttpResponseOutcome AWSClient::AttemptOneRequest(const Aws::Http::URI& uri,
     AWS_LOGSTREAM_DEBUG(AWS_CLIENT_LOG_TAG, "Request Successfully signed");
     std::shared_ptr<HttpResponse> httpResponse(
         m_httpClient->MakeRequest(*httpRequest, m_readRateLimiter.get(), m_writeRateLimiter.get()));
+
+    if(request.GetBody())
+    {
+        request.GetBody()->clear();
+        request.GetBody()->seekg(0);
+    }
 
     if (DoesResponseGenerateError(httpResponse))
     {


kunaltyagi:
@marcomagdy Moved the changes deeper into `AttemptOneRequest`. I feel this is the correct place for the code-block.

Otherwise we'd need quite a few blocks in each returning branch of `AttemptExhaustively`. Also could you please clarify

> I am saying the code should stay the same

As for test cases, it'll take some time.

marcomagdy:
This looks good. Just turn the same code snippet you had in the original issue #725 into a test, and then I'll merge it.

tyagikunal:
Hi

Any test using `stringstream` never fails to compute the content-length. Any other idea for testing?

marcomagdy:
It should set the EOF flag though

kunaltyagi:
Funny thing, it's not. Even with `request.SetComputeContentMd5(false)` which prevents the call to clear bad-bits, checking the stream after `client->MakeRequest(request)` gives EOF as false (unexpected), good as true (expected), fail as false (doesn't matter).

Manually reading the stream sets the flags correctly. I think I'm missing some data and control flow.

Here's my modification in AWSClientTest.cpp:
```C++
TEST_F(AWSClientTestSuite, TestMultipleHttpRequestWithBody)
{
    HeaderValueCollection headerValues, responseHeaders, testHeaders;
    headerValues["test1"] = "testValue1";

    QueueMockResponse(HttpResponseCode::OK, responseHeaders);
    QueueMockResponse(HttpResponseCode::OK, responseHeaders);

    AmazonWebServiceRequestMock request;
    request.SetHeaders(headerValues);
    request.SetComputeContentMd5(false);

    std::shared_ptr<Aws::StringStream> ss = Aws::MakeShared<Aws::StringStream>(ALLOCATION_TAG);
    std::string data;
    *ss << "test test\n";
    request.SetBody(ss);

    auto outcome = client->MakeRequest(request);

    std::cout << "****MANUAL OUTPUT****\n\";
    std::cout << ss->eof() << "\t" << ss->good() << "\t" << ss->fail() << "\n";
    while(*ss >> data) { std::cout << data << ";";} std::cout << "\n";
    std::cout << ss->eof() << "\t" << ss->good() << "\t" << ss->fail() << "\n";
    ASSERT_TRUE(outcome.IsSuccess());

    // check content length also?
    outcome = client->MakeRequest(request);
    std::cout << "****MANUAL OUTPUT****\n\";
    std::cout << ss->eof() << "\t" << ss->good() << "\t" << ss->fail() << "\n";
    while(*ss >> data) { std::cout << data << ";";} std::cout << "\n";
    std::cout << ss->eof() << "\t" << ss->good() << "\t" << ss->fail() << "\n";
    ASSERT_TRUE(outcome.IsSuccess());
}
```

aws-cpp-sdk-core/source/client/AWSClient.cpp
```diff
@@ -304,6 +299,12 @@ HttpResponseOutcome AWSClient::AttemptOneRequest(const Aws::Http::URI& uri,
     std::shared_ptr<HttpResponse> httpResponse(
         m_httpClient->MakeRequest(*httpRequest, m_readRateLimiter.get(), m_writeRateLimiter.get()));
 
+    if(request.GetBody())
+    {
+        request.GetBody()->clear();
+        request.GetBody()->seekg(0);
+    }
+
```
marcomagdy: Hey, this change needs to happen in the success case only.
So this needs to be added as an else block like:

```cpp
    if (DoesResponseGenerateError(httpResponse))
    {
        AWS_LOGSTREAM_DEBUG(AWS_CLIENT_LOG_TAG, "Request returned error. Attempting to generate appropriate error codes from response");
        return HttpResponseOutcome(BuildAWSError(httpResponse));
    }
    else if(request.GetBody())
    {
        request.GetBody()->clear();
        request.GetBody()->seekg(0);
    }
```

marcomagdy:
It's not setting the EOF because you added a unit test rather than an integration test. The unit test doesn't read the stream, so EOF is never set.

I suggested earlier you take the code snippet you had in when you filed the issue and create an S3 integration test.

kunaltyagi:
> create an S3 integration test.

I didn't realize it had to be an integration test. I thought a unit test would do. Anyways, running integration tests fails every single one of them at bucket creation. What's the correct way to go forwards?

aws-cpp-sdk-core/source/client/AWSClient.cpp
kunaltyagi: In case the retry parameter is set as true, then on failure, it'll result in more failures

marcomagdy: We should not clear the flags when the request fails. That's a dangerous assumption.

singku:
Per the discussion on issue #725 , close it for now, please reopen if requirements have been meet.

PR #740, opened 2017-11-23
Visual Studio 2017 (version 15.4.3) compilation.

svart-riddare:
Adding this type cast seems required for compilation with Visual Studio 2017 of 'core' component using openssl.


cedral:
I made the same change in #840

Bu11etmagnet:
Tsk, tsk, using C style casts.

cedral:
Tsk, tsk, worrying about c style casts of a constant value (-1).

PR #766, opened 2017-12-29
Parse INI-style file more like botocore (Python).

### LINKED ISSUE #765: Credentials parser doesn't accept values containing equals signs
Values that have equals signs will get corrupted by the reader, and may be dropped from the config. This is inconsistent with the AWS CLI.

This is primarily an issue with `aws_session_token`, which is base64-encoded.

Due to the root cause of this bug ([using `std::getline` to split a string](https://github.com/aws/aws-sdk-cpp/blob/0b16f4f465a65aca9f039af862f6ca92a83c5d9f/aws-cpp-sdk-core/source/utils/StringUtils.cpp#L92-L98)), this only shows up when the token is also double-quoted.

For example, reading a file with a session token like this:

```
aws_session_token="SSBhbSBhIGtleQo="
```

will return an empty value for the session token field.

This causes issues when using other tooling to edit / generate the credentials file.

jkinkead:
This splits on the first equals, regardless of the presence of additional equals.

Note that the basic parsing implementation in botocore is based on [RawConfigParser](https://docs.python.org/2/library/configparser.html#ConfigParser.RawConfigParser), which retains quotes in values.

Fixes #765 .


singku:
LGTM, will merge is soon.

jkinkead:
Thanks!

Issue #620, opened 2017-07-24
Mangled name compile error C++17


esemeniuc:
I'm running Arch Linux with g++ 7.1.1 and cmake 3.8.2. When running `make` I get the following error near the end
```shell
Scanning dependencies of target aws-cpp-sdk-dynamodb-integration-tests
[ 99%] Building CXX object aws-cpp-sdk-dynamodb-integration-tests/CMakeFiles/aws-cpp-sdk-dynamodb-integration-tests.dir/RunTests.cpp.o
[ 99%] Building CXX object aws-cpp-sdk-dynamodb-integration-tests/CMakeFiles/aws-cpp-sdk-dynamodb-integration-tests.dir/TableOperationTest.cpp.o
In file included from /home/user/dev/cpps3/deps/aws-sdk-cpp/aws-cpp-sdk-core/include/aws/core/utils/ratelimiter/DefaultRateLimiter.h:21:0,
                 from /home/user/dev/cpps3/deps/aws-sdk-cpp/aws-cpp-sdk-dynamodb-integration-tests/TableOperationTest.cpp:29:
/home/user/dev/cpps3/deps/aws-sdk-cpp/aws-cpp-sdk-core/include/aws/core/utils/memory/stl/AWSFunction.h:57:3: error: mangled name for ‘F Aws::BuildFunction(F) [with F = std::chrono::time_point<std::chrono::_V2::system_clock, std::chrono::duration<long int, std::ratio<1, 1000000000> > > (*)() noexcept]’ will change in C++17 because the exception specification is part of a function type [-Werror=noexcept-type]
 F BuildFunction(F f)
   ^~~~~~~~~~~~~
cc1plus: all warnings being treated as errors
make[2]: *** [aws-cpp-sdk-dynamodb-integration-tests/CMakeFiles/aws-cpp-sdk-dynamodb-integration-tests.dir/build.make:89: aws-cpp-sdk-dynamodb-integration-tests/CMakeFiles/aws-cpp-sdk-dynamodb-integration-tests.dir/TableOperationTest.cpp.o] Error 1
make[1]: *** [CMakeFiles/Makefile2:6073: aws-cpp-sdk-dynamodb-integration-tests/CMakeFiles/aws-cpp-sdk-dynamodb-integration-tests.dir/all] Error 2
make: *** [Makefile:130: all] Error 2```

JonathanHenson:
[...]

PR #775, opened 2018-01-17
Silent warning about mangled name changing in C++17

vincentbernat:
When compiling with a recent gcc (for example, 7.2.0), the compiler
warns about mangled name to change in C++17 because the exception
specification is part of the function type:

```
/home/…/aws-sdk-cpp/aws-cpp-sdk-core/include/aws/core/utils/memory/stl/AWSFunction.h:57:3: error: mangled name for ‘F Aws::BuildFunction(F) [with F = std::chrono::time_point<std::chrono::_V2::system_clock, std::chrono::duration<long int, std::ratio<1, 1000000000> > > (*)() noexcept]’ will change in C++17 because the exception specification is part of a function type [-Werror=noexcept-type]
 F BuildFunction(F f)
   ^~~~~~~~~~~~~
```

As per #620, this warning is considered as harmless as people are
unlikely to build the library with one compiler and use it from a
different one. However, as warnings trigger errors by default, we just
silent this warning.

### DIFF
--- a/cmake/compiler_settings.cmake
+++ b/cmake/compiler_settings.cmake
@@ -49,7 +49,7 @@ macro(set_gcc_flags)
 endmacro()
 
 macro(set_gcc_warnings)
-    list(APPEND AWS_COMPILER_WARNINGS "-Wall" "-Werror" "-pedantic" "-Wextra")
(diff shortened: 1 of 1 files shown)


Bu11etmagnet:
This warning is useless because it can't be fixed, short of removing `noexcept()` everywhere.

singku:
See #620

PR #780, opened 2018-01-22
Prevent the SDK uninstall target to clash with a another uninstall target

mdashti:
This SDK is usually used as a sub-project and if this change is not applied, the user gets an error if she wants to add a custom "uninstall" target.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -174,4 +174,6 @@ add_sdks()
 include(setup_cmake_find_module)
 
 # for generating make unstaill target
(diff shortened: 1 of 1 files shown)


marcomagdy:
Hi, can you instead of conditionally adding the target, rename it to something like `uninstall-awssdk`.
Thanks for the patch.

mdashti:
@marcomagdy Thanks for having a look at this. I applied your suggestion.

marcomagdy:
@wps132230 I'm good with this PR. Review it and merge it when you get a chance please.

singku:
@mdashti This has been merged, thanks again for your contribution

Issue #781, opened 2018-01-22
Segmentation fault in Aws::S3::S3Client::GetObject


oliora:
We build SDK of version https://github.com/aws/aws-sdk-cpp/commit/47030e94869586f4b9d29bba8fd0f3315f1c2f58 with GCC 4.9.4 and run it on Linux @ x86_64, although the issue should occur with any platform/OS/compiler.

Segfault Call stack:

```
Thu Jan 18 12:38:16 2018: [0x3501701C47A11A44] [FATAL  ] Caught SIGSEGV (11), traceback:
Thu Jan 18 12:38:16 2018: [0x3501701C47A11A44] [ERROR  ] 00000000005932c1 <our-server-executable>(sigsegv(int)+0xd1) [0x5932c1]
Thu Jan 18 12:38:16 2018: [0x3501701C47A11A44] [ERROR  ] 00007ff0652ee7e0 /lib64/libpthread.so.0(+0xf7e0) [0x7ff0652ee7e0]
Thu Jan 18 12:38:16 2018: [0x3501701C47A11A44] [ERROR  ] 00007ff06407a624 libaws-cpp-sdk-core.so(Aws::External::tinyxml2::StrPair::GetStr()+0x14) [0x7ff06407a624]
Thu Jan 18 12:38:16 2018: [0x3501701C47A11A44] [ERROR  ] 00007ff0640bbbb4 libaws-cpp-sdk-core.so(Aws::Utils::Xml::XmlNode::GetName() const+0x14) [0x7ff0640bbbb4]
Thu Jan 18 12:38:16 2018: [0x3501701C47A11A44] [ERROR  ] 00007ff0640a0f5e libaws-cpp-sdk-core.so(Aws::Client::XmlErrorMarshaller::ContainsError(Aws::Http::HttpResponse const&)+0x8e) [0x7ff0640a0f5e]
Thu Jan 18 12:38:16 2018: [0x3501701C47A11A44] [ERROR  ] 00007ff0640969d9 libaws-cpp-sdk-core.so(Aws::Client::AWSClient::AttemptOneRequest(Aws::Http::URI const&, Aws::AmazonWebServiceRequest const&, Aws::Http::HttpMethod, char const*) const+0x369) [0x7ff0640969d9]
[...]

Issue #658, opened 2017-09-12
S3Client::CompleteMultipartUpload returns success for failed requests


dyfrgi:
Complete Multipart Upload is a bit weird in the REST API. Rather than return a failed HTTP code for a failure, it returns 200 OK immediately and then may later, at its leisure, send back an error. This is described in the API docs here:
http://docs.aws.amazon.com/AmazonS3/latest/API/mpUploadComplete.html

However, if an error is received for a Complete Multipart Upload along with a 200 OK, as described in the example given in the doc there, the CompleteMultipartUploadOutcome returned by S3Client::CompleteMultipartUpload will have have IsSuccess()==true.

We are specifically receiving 200 OK SlowDown errors, but I expect other errors may occur as there's one given as an example in the API doc.

JonathanHenson:
Thanks for reporting will get it queued up for fix.

marcomagdy:
This has been fixed in v1.3.30

marcomagdy:
We'll have to revert this change f5c0f797e8267 See https://github.com/aws/aws-sdk-cpp/issues/781 for more details.

andriy-bulynko:
Hi! We're hitting this bug as well. Do you have any idea when you might post a fix? Thanks!

pitrou:
Why isn't this still fixed? This sounds like a critical bug. We just got word that our users can be hit by it: https://issues.apache.org/jira/browse/ARROW-14523

pitrou:
@wps132230 @sdavtaker @KaibaLopez

pitrou:
I'll also note that the docstring, while stating the problem, is unhelpful as to how to solve it:

> Processing of a Complete Multipart Upload request could take several minutes to complete. After Amazon S3 begins processing the request, it sends an HTTP response header that specifies a 200 OK response. While processing is in progress, Amazon S3 periodically sends white space characters to keep the connection from timing out. Because a request could fail after the initial 200 OK response has been sent, it is important that you check the response body to determine whether the request succeeded.

pitrou:
Our own [workaround](https://github.com/apache/arrow/pull/11594/files) is to intercept XML error response using a DataReceivedHandler. Unfortunately, we can't really test it (those errors seem quite sporadic), but it should theoretically work.

sdavtaker:
It seems like the Bug in TinyXML2 was resolved and we recently and we should be able to merge this fix again.

pitrou:
[...]

PR #783, opened 2018-01-22
Handle S3 API failures that return 200 OK

amsa:
The SDK relies on HTTP status codes to determine whether or not the operation succeeded. However, in certain APIs such as CompleteMultipartUpload, the API may return 200 OK with an error in the response body. In those cases, the response body needs to be inspected for errors. This fix only checks response bodies of non-GET requests that return XML as the response type. #658 #781


marcomagdy:
I don't have a better solution.
@singku and @wps132230 please chime in if you have any.
@amsa can you rebase this PR on master please?

PR #789, opened 2018-01-25
Queue: Do not wait to complete poll period if there are messages waiting

m-seker:
After trying to receive an element from the queue the loop sleeps until poll period gets expired even though the queue is not empty.
In other words, if there are more than one element in the queue, it should continue receiving messages without waiting the poll period.


marcomagdy:
Can you add a reason for closing this PR, please?

marcomagdy:
nvm i see you have another one open for the same change.

PR #790, opened 2018-01-25
Queue: Do not wait to complete poll period if there are messages waiting

m-seker:
After trying to receive an element from the queue the loop sleeps until poll period gets expired even though the queue is not empty.
In other words, if there are more than one element in the queue, it should continue receiving messages without waiting the poll period.

### DIFF
--- a/aws-cpp-sdk-queues/source/sqs/SQSQueue.cpp
+++ b/aws-cpp-sdk-queues/source/sqs/SQSQueue.cpp
@@ -54,7 +54,7 @@ SQSQueue::SQSQueue(const std::shared_ptr<SQSClient>& client, const char* queueNa
 {
 }
 
-Message SQSQueue::Top() const
+bool SQSQueue::Top(Message& message) const
 {
     if(IsInitialized())
     {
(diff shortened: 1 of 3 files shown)


aws-cpp-sdk-queues/source/sqs/SQSQueue.cpp
```diff
@@ -71,7 +71,8 @@ Message SQSQueue::Top() const
             if (receiveMessageOutcome.IsSuccess() && receiveMessageOutcome.GetResult().GetMessages().size() > 0)
             {
                 AWS_LOGSTREAM_DEBUG(CLASS_TAG, "Message found, returning");
-                return receiveMessageOutcome.GetResult().GetMessages()[0];
+                message = receiveMessageOutcome.GetResult().GetMessages()[0];
```
marcomagdy: I'd change `SetMaxNumberOfMessages(1)` to be configurable (add a ctor parameter with default = 1. And change the return type of `Top` to return a vector of messages. And in the loop check if the vector is empty to decide whether to sleep or not.

marcomagdy:
bump

m-seker:
Hey bot(and developers), unfortunately I don't work with AWS anymore. Feel free to modify or close it.

PR #793, opened 2018-01-29
Fix compilation on GCC 4.8 because of lack of std::max_align_t

oliora:
We need to build (parts of) SDK under GCC 4.8 and we can except for `std::max_align_t`. This PR fixes it.


marcomagdy:
LGTM, @wps132230 review and merge it when you get a chance.

PR #813, opened 2018-02-23
Fix for Mangled name compile error C++17 #620

aliakseis:
Issue https://github.com/aws/aws-sdk-cpp/issues/620


singku:
Closing this because we have remove AwsFunction.h

PR #822, opened 2018-03-09
Authentication V2 for Amazon S3

svart-riddare:
This pull request implements (deprecated) authentication V2 for Amazon S3 Storage REST API.

I understand that authentication V2 is deprecated and that Amazon deliberately omitted it from the C++ SDK. However, third parties (eg HGST - Western Digital) have implemented storage solutions compatible with the S3 API and older versions of their solutions support only V2 authentication. These solutions are in production and it is not possible to have them upgraded, hence the need for V2 authentication in the SDK.

The pull request is made of four distinct commits :
1. Addition of _Sha1HMAC_ in the crypto algorithms. Required for V2 authentication.
2. Addition of _AwsAuthV2Signer_, a signature V2 implementation.
3. Addition of a _signer_ parameter to the 'ClientConfiguration' structure, which defaults to "SignatureV4".
4. Enabling of the authentication V2 for _S3Client_.

There are probably better ways of implementing 3 and 4.


svart-riddare:
Note that I pulled some code from an old version of aws-cpp-sdk (from https://github.com/aws/aws-sdk-cpp/commit/31a09d2ef2bd644791f38cf5905a4d54adaab3e1#diff-b85e8661a69528c45f96e352b3745537), code which seems to conform to the AWS documentation (https://docs.aws.amazon.com/general/latest/gr/signature-version-2.html) but did not succeeded in making it work.

I instead implemented the process found in Python utility _s3cmd_ (https://github.com/s3tools/s3cmd/blob/master/S3/Crypto.py).

singku:
Yes, Auth V2 has been deprecated and C++ SDK omitted it in the first place and won't integrate it back. I suggest you to fork the repo and integrate it back to your own source.

PR #829, opened 2018-03-27
Update DateTime.h

Bousk:
* fix const-correctness


Bousk:
Sure, when working with GameLift, I wanted to use the FleetAttributes::GetCreationTime::WasParseSuccessful as it sounds like a way to check if my fleet is allocated (I simply move its result into my object), but couldn't compile due to WasParseSuccessful being non-const. Given the getter nature of this method, that appeared to be strange to me.
Though after more testing, turns out this pretty much always returns true so i'll use now give a try to FleetAttributes::GetStatus.

singku:
Thank you, will merge is ASAP

PR #841, opened 2018-04-11
crash attempting finalize unitialized ciphers

cedral:
CryptoCipherStream supports resetting m_cipher when seek goes back to the beginning of the file. If you do that and then don't read or write to re initialize the underlying openssl cipher it crashes when finalizing in the destructor. I altered the live cipher test in aws-cpp-sdk-core-tests to reproduce the crash and then fix it. This was discovered running the s3-encryption-integration tests when building against openssl on windows. Finalize is noop with bcrypt so it doesn't have this problem.


singku:
Thank you for your contribution. We have an internal PR in code review which will fix this problem and more potential problems. We will get you posted when that PR is merged.

marcomagdy:
This has been fixed in v1.4.38
see https://github.com/aws/aws-sdk-cpp/commit/ded84836cd7bf15aa2375a6c1f7143f34d985df1 for more details

PR #846, opened 2018-04-18
Reduce the refresh of AWS credentials threshold to 5 minutes.

AmineChikhaoui:
According to [1] new credentials are available in the EC2 metadata service
at least 5 minutes before the expiration of the current ones, with the
current 15 minutes threshold for refresh we could end up with expired
credentials in case the refresh is done between 5 to 15 minutes from the
expiration and end up with errors such as signature mismatch.

[1]: https://docs.aws.amazon.com/AWSEC2/latest/UserGuide/iam-roles-for-amazon-ec2.html#instance-metadata-security-credentials

### DIFF
--- a/aws-cpp-sdk-core/include/aws/core/auth/AWSCredentialsProvider.h
+++ b/aws-cpp-sdk-core/include/aws/core/auth/AWSCredentialsProvider.h
@@ -35,7 +35,7 @@ namespace Aws
 
     namespace Auth
     {
-        static int REFRESH_THRESHOLD = 1000 * 60 * 15;
+        static int REFRESH_THRESHOLD = 1000 * 60 * 5;
 
         /**
          * Simple data object around aws credentials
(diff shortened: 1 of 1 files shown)


AmineChikhaoui:
I see commits such as https://github.com/aws/aws-sdk-cpp/commits/a38488cd52c, can that be a possible fix to the issues I'm having such as getting `The request signature we calculated does not match the signature you provided. Check your key and signing method.` sporadically using the ec2 metadata service ?

AmineChikhaoui:
ping @marcomagdy :)

marcomagdy:
Hey sorry for the late reply.
>I see commits such as https://github.com/aws/aws-sdk-cpp/commits/a38488cd52c, can that be a possible fix to the issues I'm having such as getting The request signature we calculated does not match the signature you provided. Check your key and signing method. sporadically using the ec2 metadata service ?

Glad to hear it.

I'm good with the PR, but please rebase on the latest AND update the comments above the constructors that state 15 minutes.

AmineChikhaoui:
@marcomagdy done

marcomagdy:
Thanks, will merge it now internally to go through our CI.

PR #852, opened 2018-04-25
Fix code example for setting HttpClientFactory

AndrewGuenther:
The previous example would fail to compile because arg1 is undefined.

### DIFF
--- a/README.md
+++ b/README.md
@@ -259,7 +259,7 @@ Install custom memory manager:
 Override default http client factory:
 ```
     Aws::SDKOptions options;
-    options.httpOptions.httpClientFactory_create_fn = [](){ return Aws::MakeShared<MyCustomHttpClientFactory>("ALLOC_TAG", arg1); };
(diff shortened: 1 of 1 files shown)


marcomagdy:
this was an example based on a hypothetical http client library named `MyCustomHttpClientFactory`. It's there just for illustration purposes.

AndrewGuenther:
Isn't it a bit ridiculous to not fix example code on the back that it is just an example? It's a single line change, why not just take it?

marcomagdy:
I'd happily take the PR if it was fixing something broken.

AndrewGuenther:
The example is broken.

singku:
Would you mind point us where does `MyCustomHttpClientFactory` exist except in ReadMe and comments in Aws.h? Because I can't find it's existence in a real example, so I don't understand why it would break an example.

singku:
Correct, you can see my updated comments and that's what @marcomagdy said, it's for illustration purpose.

PR #876, opened 2018-05-21
Make the code compile with c++17.

yasushi-saito:
std::ptr_fun has been removed in c++17.

Tested on Ubuntu16.04 + clang 6.0 + libc++, with -std=c++17.


marcomagdy:
Looks good, will merge soon. Thank you

PR #879, opened 2018-05-23
Fix race condition when initializing Enum overflow container

marcomagdy:


marcomagdy:
I'll need to think of another way to solve this problem

marcomagdy:
@singku Yes, that sounds like a better approach.

PR #880, opened 2018-05-23
Including functional header to use std::function

amartya00:
Fixing build error function in namespace std does not name a template type


marcomagdy:
ugh, we didn't catch it because we do unity builds. Thank you.

marcomagdy:
@amartya00 which compiler did you see this error with?

amartya00:
Hello!

My compiler version is g++ 7.3.0
My system is ubuntu 18.04.

Thanks!
-Amartya

Issue #924, opened 2018-07-31
Segmentation Fault on DynamoDB AttributeValue when it is empty


namiazad:
Default values in [AttributeValueValue.h](https://github.com/aws/aws-sdk-cpp/blob/master/aws-cpp-sdk-dynamodb/include/aws/dynamodb/model/AttributeValueValue.h#L79) is changed from _static_ to _non-static_ as part of commit 45305f4e5c4886654639ab7ecdbe486d7ec0f852
This will cause segmentation fault on [AttributeValue](https://github.com/aws/aws-sdk-cpp/blob/master/aws-cpp-sdk-dynamodb/source/model/AttributeValue.cpp#L33) when m_value is nullptr.

marcomagdy:
Thanks for the PR.
Couple of problems with this though:
* We have to do the fix in the codegenerator.
* We can't use statics. All allocations/deallocations must go through the custom allocator.

marcomagdy:
reverted the broken commit.
Fixing this will require changing the API of `AttributeValue` class to return an optional value rather than relying on static variables.

PR #925, opened 2018-07-31
Fix segmentation fault on AttributeValue when it is empty - issue #924

namiazad:


namiazad:
As the fix should be done in code generator, closing the pull request

Issue #923, opened 2018-07-30
Better throttling handling in EC2MetadataClient & co


oliora:
Currently, `AWSHttpResourceClient` and derived classes `EC2MetadataClient` and `ECSCredentialsClient` uses default configuration to call metadata service and there is no way to configure it. Neither default timeout (3s) nor default retries configuration (no retries) is well suitable for the metadata service when it throttles. I propose to allow to pass `ClientConfiguration` object to `AWSHttpResourceClient` (and derived classes) so user can control retries and timeouts used to call the metadata service.

marcomagdy:
That's a fair ask.
Can you submit a PR please?

oliora:
It’s merged and fixed

PR #927, opened 2018-08-02
Add timeouts and retries configuration to metadata service clients (#923)

oliora:
Fix for #923

The commit also makes the default configuration to be useful. Now it has 5s request timeout and does 4 retries with 1s exponential back-off.

### DIFF
--- a/aws-cpp-sdk-core-tests/aws/auth/AWSHttpResourceClientTest.cpp
+++ b/aws-cpp-sdk-core-tests/aws/auth/AWSHttpResourceClientTest.cpp
@@ -56,7 +56,7 @@ namespace
         }
     };
 
-    TEST_F(AWSHttpResourceClientTest, TestAWSHttpResourceClientWithNullResponse)
+    TEST_F(AWSHttpResourceClientTest, TestAWSHttpResourceClientWithPermanentNullResponse)
     {
         auto awsHttpResourceClient = Aws::MakeShared<Aws::Internal::AWSHttpResourceClient>(ALLOCATION_TAG, ALLOCATION_TAG);
         Aws::String result = awsHttpResourceClient->GetResource("http://www.uri.com", "/path/to/res", ""/*authToken*/);
@@ -68,6 +68,8 @@ namespace
         ASSERT_EQ("/path/to/res", mockRequest.GetUri().GetPath());
         ASSERT_EQ(Aws::Http::HttpMethod::HTTP_GET, mockRequest.GetMethod());
         ASSERT_EQ("", result);
+        // Retries should be done when request can't be made at all
+        ASSERT_EQ(5u, mockHttpClient->GetAllRequestsMade().size());
     }
 
     TEST_F(AWSHttpResourceClientTest, TestAWSHttpResourceClientWithEmptyResponse)
@@ -90,6 +92,8 @@ namespace
         ASSERT_EQ("/path/to/res", mockRequest.GetUri().GetPath());
         ASSERT_EQ(Aws::Http::HttpMethod::HTTP_GET, mockRequest.GetMethod());
         ASSERT_EQ("", result);
+        // No retries should be made
+        ASSERT_EQ(1u, mockHttpClient->GetAllRequestsMade().size());
     }
 
     TEST_F(AWSHttpResourceClientTest, TestAWSHttpResourceClientWithHttpBadRequest)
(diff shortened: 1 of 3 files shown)


aws-cpp-sdk-core/include/aws/core/internal/AWSHttpResourceClient.h
```diff
@@ -29,6 +30,15 @@ namespace Aws
...
     {
+        /**
+         * This mutable structure is used to configure AWSHttpResourceClient.
+         * Default values can only be overwritten prior to passing to the client constructors.
+         */
+        struct HttpResourceClientConfiguration: public Client::ClientConfiguration
+        {
+            HttpResourceClientConfiguration();
+        };
+
```
marcomagdy: This is the only wrinkle I see in this PR.

I'd rather not introduce another type just for convenience.
I prefer having two constructors one that takes `ClientConfiguration` parameter and another that does not. Using the constructor without the client configuration would set the defaults you have in this PR.

Thoughts?

oliora: Sounds reasonable. I’ll update my PR with it.

oliora:
I've updated my commit in-place (and seems it was wrong and I should add a fixing commit instead but now I can't undo it)

aws-cpp-sdk-core-tests/aws/auth/AWSHttpResourceClientTest.cpp
```diff
@@ -68,6 +68,8 @@ namespace
         ASSERT_EQ("/path/to/res", mockRequest.GetUri().GetPath());
         ASSERT_EQ(Aws::Http::HttpMethod::HTTP_GET, mockRequest.GetMethod());
         ASSERT_EQ("", result);
+        // Retries should be done when request can't be made at all
+        ASSERT_EQ(5u, mockHttpClient->GetAllRequestsMade().size());
```
singku: Could you please add an inline comment here explaining how `5` come up?, E.g. `5/* 1 initial request + 4 retries*/`

aws-cpp-sdk-core-tests/aws/auth/AWSHttpResourceClientTest.cpp
```diff
@@ -134,6 +164,63 @@ namespace
         ASSERT_EQ("/path/to/res", mockRequest.GetUri().GetPath());
         ASSERT_EQ(Aws::Http::HttpMethod::HTTP_GET, mockRequest.GetMethod());
         ASSERT_EQ("{ \"Resource\": \"TestResource\" }", result);
+        // No retries should be made
+        ASSERT_EQ(1u, mockHttpClient->GetAllRequestsMade().size());
+    }
+
+    TEST_F(AWSHttpResourceClientTest, TestAWSHttpResourceClientWithTemporaryNullResponse)
```
singku: What about `WithSuccessResponseAfterOneNullResponse`? This can help explain why it's `2` in the last ASSERT

aws-cpp-sdk-core-tests/aws/auth/AWSHttpResourceClientTest.cpp
```diff
@@ -134,6 +164,63 @@ namespace
...
+        ASSERT_EQ(Aws::Http::Scheme::HTTP, mockRequest.GetUri().GetScheme());
+        ASSERT_EQ("www.uri.com", mockRequest.GetUri().GetAuthority());
+        ASSERT_EQ("/path/to/res", mockRequest.GetUri().GetPath());
+        ASSERT_EQ(Aws::Http::HttpMethod::HTTP_GET, mockRequest.GetMethod());
+        ASSERT_EQ("{ \"Resource\": \"TestResource\" }", result);
+        // There should be one retry
+        ASSERT_EQ(2u, mockHttpClient->GetAllRequestsMade().size());
+    }
+
+    TEST_F(AWSHttpResourceClientTest, TestAWSHttpResourceClientWithTemporaryInternalServerError)
```
singku: Same thought here.

aws-cpp-sdk-core/source/internal/AWSHttpResourceClient.cpp
```diff
@@ -33,35 +35,62 @@ using namespace Aws::Internal;
...
 
+static const char* RESOURCE_CLIENT_CONFIGURATION_ALLOCATION_TAG = "AWSHttpResourceClient";
 static const char* EC2_METADATA_CLIENT_LOG_TAG = "EC2MetadataClient";
 static const char* ECS_CREDENTIALS_CLIENT_LOG_TAG = "ECSCredentialsClient";
 
-AWSHttpResourceClient::AWSHttpResourceClient(const char* logtag)
-    : m_logtag(logtag), m_httpClient(nullptr)
+
+namespace {
+    ClientConfiguration makeDefaultHttpResourceClientConfiguration(const char *logtag)
```
singku: make -> Make

aws-cpp-sdk-core/source/internal/AWSHttpResourceClient.cpp
```diff
@@ -72,40 +101,67 @@ Aws::String AWSHttpResourceClient::GetResource(const char* endpoint, const char*
...
+        else
+        {
+            // Client error, retrying the same request will not help
+            AWS_LOGSTREAM_ERROR(m_logtag.c_str(), "Http request to retrieve credentials failed with non-retriable error code "
+                                << static_cast<int>(response->GetResponseCode()));
+            return {};
+        }
+
+        // Internal error that is retryable
+        Aws::Client::AWSError<Aws::Client::CoreErrors> error(Aws::Client::CoreErrors::INTERNAL_FAILURE, true);
```
singku: I think it's better to refer `isRetryableHttpResponseCode` in AWSClient.cpp. That's a static function though. But we can update that.

oliora: I don't get it. I thought that the second parameter to `AWSError` specifies whether it is a retryable error or not.

oliora: TBH, I don't get the whole design of the error/retries handling (for me it seems more complicated that it should be). It's unclear who is responsible for making decisions about retrying. From one perspective, the `AWSError` has `isRetryable` field that tells whether the error is retryable or not and it seems pretty valid because it is set by a particular service who perfectly knows what errors are retryable. From other perspective there is `RetryStrategy::ShouldRetry` which is responsible for final decision about retrying but none of the standard strategies has any logic beyond checking `AWSError::isRetryable` and this part seems not really needed.

oliora: Seems I'm getting it. Do you actually mean to replace condition `if (response->GetResponseCode() >= HttpResponseCode::INTERNAL_SERVER_ERROR)` with `if (isRetryableHttpResponseCode(responseCode))`?

oliora: Btw, there is also `CoreErrorsMapper::GetErrorForHttpResponseCode` functions. One seems to have more correct logic than `isRetryableHttpResponseCode` and it is not static. Should not I use it instead?

singku: You bring up a good point. These two functions are actually not consistent with each other. We need to resolve this.

oliora:
@marcomagdy @singku I've updated the PR with using of `CoreErrorsMapper::GetErrorForHttpResponseCode` and added more tests.

singku:
Already merged.

PR #941, opened 2018-08-16
Fixes compile issue with "Client': is not a class or namespace name"

Vawx:
Resolved scope with Aws::Client


wps132230:
Thank you for your contribution! We've merged your commit into our code.

PR #943, opened 2018-08-22
Add missing #include <cassert>

edolstra:
This fixes

```
/build/source/aws-cpp-sdk-core/source/http/standard/StandardHttpRequest.cpp: In member function 'virtual const String& Aws::Http::Standard::Stand>
/build/source/aws-cpp-sdk-core/source/http/standard/StandardHttpRequest.cpp:72:5: error: 'assert' was not declared in this scope
     assert (iter != headerMap.end());
     ^~~~~~
```


singku:
Thanks for submitting this PR, but a fix is already on the way, we will update this repo ASAP, before that, you can specify -DENABLE_UNITY_BUILD to cmake to work this around.

PR #964, opened 2018-09-13
Enable CURLOPT_TCP_NODELAY to avoid unnecessary latencies

cygnus9:


marcomagdy:
This is enabled by default in curl!

PR #981, opened 2018-09-30
Mistyping fix - makes build failed for Android target

atlascoder:


marcomagdy:
`errno` usage is correct in this context. And Android builds pass just fine in our CI.
maybe update this PR with more specifics about the error you're getting?

PR #1026, opened 2018-11-29
Fixed ConfigFileProfileFSM to flush last profile correctly

LantosDavid:
The code as-is works, but it's not technically correct.  When the end of file is reached, the last profile should be flushed by FlushProfileAndReset, and that's it.  Instead, FlushProfileAndReset will still try to extract another profile name (which doesn't exist) because of the following line:

if(!line.empty() && openPos != std::string::npos && closePos != std::string::npos)


singku:
Merged. Thanks again.

PR #1031, opened 2018-12-03
only set proxy username and password options when specified for curl

alanbirtles:
If the proxy username and password are blank then sending them to curl causes curl to generate a
```Proxy-Authorization: Basic Og==``` header. This causes problems with proxies that don't require authentication (e.g. CNTLM) as they reject this incorrect username and password.


marcomagdy:
LGTM Thank you. Will merge it soon.

marcomagdy:
Taking a second look at this, it might be better to set the username only if the username is set and set the password only if the password is set. Basically, split the if condition.

alanbirtles:
either way results in the same behaviour

marcomagdy:
we'll take it as it is and change it later if needed.
Thanks

singku:
This is released.

PR #1034, opened 2018-12-14
Adding missing request parameter interpolation to Aws::Pinpoint::PinpointClient methods

julio-gorge:
Ex-Amazon SDE here :) - All request making methods in Aws::Pinpoint::PinpointClient were missing interpolation of request parameters, effectively making the Pinpoint package unusable.

I have added the missing interpolation of request parameters to all method clients and succesfully tested some of the most basic requests against a live Pinpoint projects on my personal AWS account.


singku:
Thank you @julio-gorge for bring it up. This is actually a bug in our code-generator. We are working on fixing it. Thanks again.

PR #1038, opened 2018-12-20
Move to tagged versions of C dependency libraries

david-koenig:
Needed because we were previously pinning to old version
of aws-c-common before some breaking changes.


david-koenig:
Hold off on merging this. It looks like we're going to have to update to the *.*.1 tag of each of these libraries, but I need to do one more check in the morning that those ones will work properly.

david-koenig:
The ``*.*.1`` versions are not working properly in the build here and we will need to pin new versions again. This is a work in progress. Hold off on merging.

david-koenig:
This has been updated to use the versions we need.

wps132230:
For Jonathan's concern, we've already passed `CMAKE_INSTALL_PREFIX` and `CMAKE_PREFIX_PATH`, and for `CMAKE_C_FLAGS`, I think customers can build their own version of dependencies, and make SDK link to these libraries.

JonathanHenson:
If you don’t pass CMAKE_C_FLAGS your 32 bit builds and cross compile builds will fail.

PR #1064, opened 2019-02-09
Fix potential out of bound read in CalculateBase64DecodedLength

Voultapher:
If the length of base64 input is only one character.
len - 2 would wrap to size_t::max - 1, causing an out of bound read.

Albeit being an invalid base64 string, this should not cause UB.


Voultapher:
I see this has been approved, is there something else that needs to happend for this to get merged?

marcomagdy:
We run it through our internal CI, and if it all looks good we merge it.
This PR is super simple, so I anticipate no issues. We'll have it merged ASAP. 
Thank you

marcomagdy:
I merged the PR to our internal repo. It will be show up on the public repo with the next release.

singku:
Merged to master

PR #1081, opened 2019-03-05
Added check in AWSHttpResourceClient if authToken is actually set.

zokir:
*If you set Authorization header with custom HttpClientFactory, AWSHttpResourceClient will overwrite it again because authToken check always passes. We should check for the actual content of the authToken passed.*

### DIFF
--- a/aws-cpp-sdk-core/source/internal/AWSHttpResourceClient.cpp
+++ b/aws-cpp-sdk-core/source/internal/AWSHttpResourceClient.cpp
@@ -172,7 +172,7 @@ EC2MetadataClient::~EC2MetadataClient()
 
 Aws::String EC2MetadataClient::GetResource(const char* resourcePath) const
 {
-    return GetResource(m_endpoint.c_str(), resourcePath, ""/*authToken*/);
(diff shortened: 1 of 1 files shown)


marcomagdy:
>because authToken check always passes

Why would that check always pass? it shouldn't if you pass nullptr.

zokir:
So does the intention of this call is to clear header then?

https://github.com/aws/aws-sdk-cpp/blob/65394d753c27adc1734d73475304a609ee693d4b/aws-cpp-sdk-core/source/internal/AWSHttpResourceClient.cpp#L175

Is API written with intention where "" means one thing and std::nullptr other? If that's the case, i might be doing something wrong as a client. 

Please confirm, and I will close the this pr. Sorry for wasting time.

marcomagdy:
Oh I see. You're referring to EC2MetadataClient. That makes sense.
Can you change this PR to do the following instead?
```cpp
return GetResource(m_endpoint.c_str(), resourcePath, nullptr/*authToken*/); 
```

singku:
This PR has been merged.

Issue #1086, opened 2019-03-07
Can't build x86 on an x86-64 machine because of third-party


kmaragon:
### What platform/OS are you using?
Linux x86-64 machine with -m32 build

### What compiler are you using? what version?
gcc / clang are both affected

### What's your CMake arguments?
The important ones are CMAKE_CXX_FLAGS="-m32", CMAKE_SHARED_LINKER_FLAGS="-m32", and CMAKE_CXX_FLAGS="-m32" and CMAKE_C_FLAGS="-m32"

### Can you provide a TRACE level log? (sanitize any sensitive information)
not really but there's this: https://travis-ci.org/kmaragon/conan-aws-sdk-cpp/jobs/503283625

Ultimately the issue is that when calling ExternalProject_Add at https://github.com/aws/aws-sdk-cpp/blob/master/third-party/cmake/BuildAwsEventStream.cmake#L45
https://github.com/aws/aws-sdk-cpp/blob/master/third-party/cmake/BuildAwsChecksums.cmake#L38
and https://github.com/aws/aws-sdk-cpp/blob/master/third-party/cmake/BuildAwsCCommon.cmake#L39

The build doesn't forward the aformentioned flags. it seems like there was a commit:
https://github.com/aws/aws-sdk-cpp/commit/312db5e6940e67e81bda38edd09eeee7e0316899
that likely largely resolved the issue for OSX. It only forwards C_FLAGS but I assume these 3 libraries are pure C from their names so that's sufficient. But the fix wasn't rolled out to any case *except* OSX.

The result is that when building aws-sdk-cpp with `-m32` on Linux, with either clang or gcc, you get a nicely compiled set of 32 bit aws cpp libraries that can't be loaded because they ultimately depend on these 3 c libraries which have all been built for 64 bit only.

singku:
Looking at the Travis job, AWS SDK for CPP has been built and installed correctly. It failed when building the example project because it can't locate the 3rd-party libraries.
CMAKE_C_FLAGS is actually passed to 3rd-party libraries and they are installed in `lib` instead of `lib64` directory. So I don't think this is the cause.

kmaragon:
It found the library. But it couldn't link: 
https://travis-ci.org/kmaragon/conan-aws-sdk-cpp/jobs/503283625#L2103

singku:
What I meant is based on that build output, AWS SDK CPP itself can link against third parties. Means both of them must be successfully built to 32bit libraries. 

[...]

PR #1087, opened 2019-03-08
Carry C_FLAGS / CXX_FLAGS into the aws-c-* libs

kmaragon:
*Issue #, if available:*
#1086

*Description of changes:*
Moving the project declaration up was for good form and it enables tools like conan to know what language they are running in so they can inject library search paths and compilation flags *before* running the execute_process subprocess builds for the aws-c-* libs.

The rest was to ensure that at least C_FLAGS and CXX_FLAGS are carried into the separate child cmake processes that are executed first with execute_process and then later via ExternalProject. Without this, building an x86 build on an x86-64 machine would succeed but would produce an unusable set of libs where the aws-c-* libs were all 64 bit and the main aws-sdk-cpp libs were all 32 bit as requested.

This doesn't fix an outstanding issue though. When using `/MT` on Windows the same kind of problem as above occurs. The aws-c-* libs all link against the shared multi-threaded runtime - /MD. But the aws-sdk-cpp libs do as requested and use the static runtime. And again, these build correctly but they are unusable because of the conflicting runtimes.

It's not enough to carry the CMAKE_C_FLAGS that contain /MT. There seems to be some mechanism (maybe in the way the child cmake runs or in the aws-c-* libs themselves) that clobbers the /MT and converts it to /MD. So it is not possible to build a usable aws-sdk-cpp statically on Windows using the static runtime. And this PR won't fix that. But it will at least enable building x86 libraries on an x86-64 machine for Linux and Windows.


singku:
https://github.com/awslabs/aws-c-common/commit/8e36c092bd103272a7162026dafea136d151617f

I think this is the place that switches between /MT and /MD

CMakeLists.txt
```diff
@@ -27,6 +27,9 @@ if(POLICY CMP0056)
     cmake_policy(SET CMP0056 NEW)
 endif()
 
+# build the sdk targets
+project("aws-cpp-sdk-all" VERSION "${PROJECT_VERSION}" LANGUAGES CXX)
```
marcomagdy: project() defines the INSTALL_PREFIX which we test for later.
The call is intentionally placed *after* the cmake commands to build the third-party libs.
We should have documented that better.

kmaragon:
> [awslabs/aws-c-common@8e36c09](https://github.com/awslabs/aws-c-common/commit/8e36c092bd103272a7162026dafea136d151617f)
> 
> I think this is the place that switches between /MT and /MD

This code isn't in the commit that is being checked out with the ExternalProject. So that's probably not butchering it. But for what it's worth, BUILD_SHARED_LIBS is OFF in this case - at least at the top level.

CMakeLists.txt
marcomagdy: ping!
any chance of updating this PR?

ihnorton:
One problem I've encountered with static builds on Windows is that CMake `ExternalProject_Add` *does not* pass the `CMAKE_GENERATOR_PLATFORM` explicitly down to sub-builds unless the generator was explicitly set in the top level CMake invocation. So even though cmake is running in the correct x64 vcvars environment and chooses the correct generator at the top level with `-A x64`, EP sub-builds don't get the `-A x64` and fall back to x86. This was a headache because even though I was passing `-A x64` in my superbuild `ExternalProject_Add` for AWS SDK, the EP_Add calls inside of AWS SDK itself (for c-common etc.) did not thread the `-A x64` down, causing linkage failures.

I've ended up using `-GVisual Studio 15 2017 Win64`, although I believe `-GVisual Studio 15 2017 -A x64` would also work.

PR #1182, opened 2019-06-27
Bug fix for "Avoid disabling speed checks when timeout is less than 1 second"

derpferd:
*Description of changes:*
This makes the hack to avoid disabling speed checks take into account configurations that disable speed checks with a configured time of 0.
Examples:
1000 ms timeout -> 1 second
40 ms timeout -> 1 second
timeout disabled (0 ms) -> 0 second

Original commit: https://github.com/aws/aws-sdk-cpp/commit/9260512d52cdb686f707128d4696f384678abc74


derpferd:
Was merge in commit b5440ef

PR #1183, opened 2019-07-01
Added config to allow setting Curl's request timeout as well as Curl's low speed time.

derpferd:
*Description of changes:*
Added config to allow setting Curl's request timeout as well as Curl's low speed time.
Also added more information to the configuration docs


singku:
Merged similar PR

Issue #713, opened 2017-11-01
Potential hang in `CurlHandleContainer` and `WinConnectionPoolMgr` on getting new connection


oliora:
I suspect that both `CurlHandleContainer::AcquireCurlHandle()` and `WinConnectionPoolMgr::AquireConnectionForHost()` can hang when they not supposed to.

I'll describe the scenario for `CurlHandleContainer`. `WinConnectionPoolMgr` has absolutely the same behavior.

```
CURL* CurlHandleContainer::AcquireCurlHandle()
{
    AWS_LOGSTREAM_DEBUG(CURL_HANDLE_CONTAINER_TAG, "Attempting to acquire curl connection.");

    if(!m_handleContainer.HasResourcesAvailable())  // [1]
    {
        AWS_LOGSTREAM_DEBUG(CURL_HANDLE_CONTAINER_TAG, "No current connections available in pool. Attempting to create new connections.");
        CheckAndGrowPool();  // [2]
    }

    CURL* handle = m_handleContainer.Acquire();  // [3]
    AWS_LOGSTREAM_INFO(CURL_HANDLE_CONTAINER_TAG, "Connection has been released. Continuing.");
    AWS_LOGSTREAM_DEBUG(CURL_HANDLE_CONTAINER_TAG, "Returning connection handle " << handle);
    return handle;
}
```

Scenario:
- There is a `CurlHandleContainer` object with empty`m_handleContainer`.
- Two threads call this object's `CurlHandleContainer::AcquireCurlHandle()`
- *Thread A* calls `m_handleContainer.HasResourcesAvailable()` [1] and get `false`
- *Thread A* calls `CurlHandleContainer::CheckAndGrowPool()` [2], which creates a new CURL handle and puts it into `m_handleContainer`
- *Thread B* calls `m_handleContainer.HasResourcesAvailable()` [1] and get `true`
- *Thread B* calls `m_handleContainer.Acquire()` [3] and get handle created by *Thread A*. `m_handleContainer` is empty again.
- *Thread A* calls `m_handleContainer.Acquire()` [3] assuming there is a CURL handle available for him (which is not true anymore) and hangs there until someone has released a CURL handle.

The issue exist in mainline.

oliora:
[...]

PR #1184, opened 2019-07-03
Adds lock in CurlHandleContainer::AcquireCurlHandle

ziliangpeng:
*Issue #, if available:*

#1185 

*Description of changes:*

There is a race condition in checking and acquiring connection.

## The bug

When a program need to acquire a connection, 2 things are done:

1) check if there's available connection (`CurlHandleContainer::AcquireCurlHandle`)
2) if not, double the number of total connection (`CurlHandleContainer::CheckAndGrowPool`)

While 2) is guarded by a mutex, 1) is not. This results in a race condition that if 2 threads tries to acquire a connection at the same time when no connection is available, both threads will double the number (sequentially, because mutex).

The issue is significantly worsen when a large number of threads acquire connections at the very beginning, with each thread trying to double the number, the number of connections increases exponentially, much more than requested.

## The Fix

add a mutex for 1).

@JonathanHenson

### DIFF
--- a/aws-cpp-sdk-core/source/http/curl/CurlHandleContainer.cpp
+++ b/aws-cpp-sdk-core/source/http/curl/CurlHandleContainer.cpp
@@ -44,6 +44,8 @@ CurlHandleContainer::~CurlHandleContainer()
 
 CURL* CurlHandleContainer::AcquireCurlHandle()
 {
+    static std::mutex m_acquireHandleLock;
+    std::lock_guard<std::mutex> locker(m_acquireHandleLock);
     AWS_LOGSTREAM_DEBUG(CURL_HANDLE_CONTAINER_TAG, "Attempting to acquire curl connection.");
 
     if(!m_handleContainer.HasResourcesAvailable())


singku:
Thanks for your inputs, I believe this problem has been discussed in this thread.
https://github.com/aws/aws-sdk-cpp/issues/713

ziliangpeng:
@singku thanks for the quick response.

I believe the issue described in #713 is related but not exactly the same. Actually it's the exact opposite of the issue here.

In #713 the issue described was not enough connections being created, thus blocking threads.

In this issue, the race condition results in threads creating much more connections (exponentially), which exhausts resource.

ziliangpeng:
To further illustrate, consider the following code (stolen from #713) and scenario:

```
CURL* CurlHandleContainer::AcquireCurlHandle()
{
    AWS_LOGSTREAM_DEBUG(CURL_HANDLE_CONTAINER_TAG, "Attempting to acquire curl connection.");

    if(!m_handleContainer.HasResourcesAvailable())  // [1]
    {
        AWS_LOGSTREAM_DEBUG(CURL_HANDLE_CONTAINER_TAG, "No current connections available in pool. Attempting to create new connections.");
        CheckAndGrowPool();  // [2]
    }

    CURL* handle = m_handleContainer.Acquire();  // [3]
    AWS_LOGSTREAM_INFO(CURL_HANDLE_CONTAINER_TAG, "Connection has been released. Continuing.");
    AWS_LOGSTREAM_DEBUG(CURL_HANDLE_CONTAINER_TAG, "Returning connection handle " << handle);
    return handle;
}
```
We have A, B, C, D threads.

- A calls [1], gets false
- B calls [1], gets false
- C calls [1], gets false
- D calls [1], gets false
- A calls [2], creates 1 connection (1 in total)
- B calls [2], creates 2 connections (3 in total)
- C calls [2], creates 6 connections (12 in total)
- D calls [2], creates 24 connections (36 in total)

A very small number of threads very quickly max out the connection pool size.

singku:
Following @marcomagdy's comments there:

>Depending on the order of thread execution you could potentially over-optimize by growing the pool too much (calling Grow more than once)

ziliangpeng:
@singku yeah, I think that describes the behavior.

What I'd like to discuss is whether this makes sense (that we allow over-optimize and prematurely exhaust resource) and if a fix is wanted?

marcomagdy:
it seems like that piece of code is uncomfortable to many people. So we might as well fix it.

I have a few comments. Some general and some regarding the patch.
1. Do you have more concrete data on what exhaust resources in this context means? The cost of a curl handle is a memory allocation and initialization values of some structs.
2.  Adding a `static` mutex means every client in the process has to content with that lock. There's no reason to make it static.
3. Adding the lock at the top of the function makes all the other locks in the sub-functions useless(`CheckAndGrowPool` and `HasResourcesAvailable` both have locks), which means we should either get rid of those other locks, or put this new lock in a better spot. For example, after checking `HasResourcesAvailable` acquire the lock and check again to avoid the double grow.

marcomagdy:
Can you please update the PR with the changes above?

PR #1187, opened 2019-07-06
Pass CMAKE_GENERATOR_PLATFORM to third-party build if set

ihnorton:
*Issue #, if available:*

Specifically resolves an issue I encountered while building the SDK within a larger project.

May also help with https://github.com/aws/aws-sdk-cpp/pull/1087 (in combination with adding `-A x64` to the CMake call creating the AWSSDK project in conan).

*Description of changes:*

Pass the `CMAKE_GENERATOR_PLATFORM` to the `execute_process` calls used to create cmake projects for the `third-party` builds, if that parameter has been set. This fixes an issue I observed when setting the top-level generator to `Visual Studio 15 2017` (rather than `Visual Studio 15 2017 Win64`), where the third-party modules were all built as 32-bit x86 objects, causing LNK2019 errors.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -133,6 +133,9 @@ if (BUILD_DEPS)
             set(AWS_DEPS_INSTALL_DIR ${CMAKE_CURRENT_BINARY_DIR}/.deps/install CACHE STRING "A string describes the path where 3rd-party dependencies will be or have been installed")
         endif()
     endif()
+    if (DEFINED CMAKE_GENERATOR_PLATFORM AND "${CMAKE_GENERATOR}" MATCHES "Visual Studio.*")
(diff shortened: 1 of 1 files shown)


ihnorton:
Note that the underlying issue here is, per the CMake documentation:

> The default target platform name (architecture) is Win32.

https://cmake.org/cmake/help/latest/generator/Visual%20Studio%2015%202017.html

So currently the only way I can see to specify the target platform for the third-party build, at least when using the Visual Studio 2017 generator, is to specify `Visual Studio 15 2017 Win64`. However, that generator specification is mutually exclusive with using the "modern cmake" with the `-A x64` argument.

ihnorton:
Bump

singku:
This has been merged.

Issue #1196, opened 2019-07-17
Disabled URL Encoding of Reserved Characters Caused Signature Mismatch on Some S3-Compatible Storage Servers


shanebarnes:
Are there any plans to enable URL encoding of reserved characters (e.g., $, @, &, ...) in the SDK core?

In https://github.com/aws/aws-sdk-cpp/blob/master/aws-cpp-sdk-core/source/http/URI.cpp#L124

```
            switch(c)
            {
                // §2.3 unreserved characters
                case '-': case '_': case '.': case '~':
                // The path section of the URL allow reserved characters to appear unescaped
                // RFC 3986 §2.2 Reserved characters
                // NOTE: this implementation does not accurately implement the RFC on purpose to accommodate for
                // discrepancies in the implementations of URL encoding between AWS services for legacy reasons.
                case '$': case '&': case ',':
                case ':': case '=': case '@':
                    ss << c;
                    break;
                default:
                    ss << '%' << std::setfill('0') << std::setw(2) << (int)((unsigned char)c) << std::setw(0);
            }
```

Files with reserved characters, for example, `my@file.bin`, are uploaded to S3 successfully but the request path is not URL encoded (e.g., `PUT /my-bucket/my@file.bin HTTP/1.1`). However, some third party S3-compatible storage services (SwiftStack) fail with a signature mismatch because the SDK does not URL encode the path.

It looks like the Python botocore package does enable URL encoding for reserved characters so I am just curious if the C++ SDK will enable URL encoding for reserved characters as well.

### What platform/OS are you using?

I am using Linux and macOS platforms:

* CentOS 7.5, and
* macOS 10.14

### Which version of the SDK?

1.7.138 (https://github.com/aws/aws-sdk-cpp/tree/1.7.138)

### What compiler are you using? what version?

On Linux, I am using g++ 4.9.2. On macOS, I am using Xcode 10.

### Can you provide a TRACE level log? (sanitize any sensitive information)

KaibaLopez:
Hi @shanebarnes, thanks for pointing this out.
It might be a bug, I will take a look at it and as always, we're open for contributions, if you have a solution for this in mind.

PR #1202, opened 2019-07-19
core-http-uri: percent encode reserved characters in URL paths as sug…

shanebarnes:
…gested by S3 guidelines

*Issue #, if available:*

This PR addresses issue #1196 when using the S3 library with S3-compatible storage servers (e.g., SwiftStack) that require reserved characters in object key names (e.g., my@file.txt) to be percent encoded in order to prevent a signing mismatch.

It is worth noting that the AWS CLI (e.g., botocore) does percent encode S3 object key names that contain reserved characters.

*Description of changes:*

The percent encoding of URL paths algorithm was modified to be compliant with S3 guidelines which indicates that reserved characters, as defined in RFC 3986, probably require encoding. The algorithm currently prevents encoding of six reserved characters in URL paths (`$`, `&`, `,`, `:`, `=`, and `@`) for legacy reasons related to AWS services.   

### DIFF
--- a/aws-cpp-sdk-core-tests/http/URITest.cpp
+++ b/aws-cpp-sdk-core-tests/http/URITest.cpp
@@ -188,7 +188,7 @@ TEST(URITest, TestParseWithColon)
     EXPECT_STREQ("test.com", uri.GetAuthority().c_str());
     EXPECT_EQ(443, uri.GetPort());
     EXPECT_STREQ("/path/1234:_Some_Path", uri.GetPath().c_str());
-    EXPECT_STREQ(strUri, uri.GetURIString().c_str());
+    EXPECT_STREQ("https://test.com/path/1234%3A_Some_Path", uri.GetURIString().c_str());
 
     const char* strUriWithPort = "https://test.com:8080/path/1234:_Some_Path";
     URI uriWithPort(strUriWithPort);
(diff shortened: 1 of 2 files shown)


shanebarnes:
@singku Good question. Based on the comment in the existing code that I have changed, it is difficult to tell what legacy services could potentially break. My assumption, which could be a bad one, is that there is unit test coverage for services that require unencoded URL paths for some reserved characters.

@marcomagdy Can you comment on services that still require URL paths with unencoded reserved characters?

marcomagdy:
We can't take this patch as is because it would break services like Greengrass that happens to have a '$' as part of its API URI. Encoding it would result in signature mismatch errors. 
See https://docs.aws.amazon.com/greengrass/latest/apireference/resetdeployments-post.html and https://docs.aws.amazon.com/greengrass/latest/apireference/stopbulkdeployment-put.html

The fix would be in the S3 specific code. And since S3's code is auto-generated, then you/we would need to patch the code generator template files https://github.com/aws/aws-sdk-cpp/blob/master/code-generation/generator/src/main/resources/com/amazonaws/util/awsclientgenerator/velocity/cpp/xml/rest/

To be completely honest, this will not be a priority of ours since it's not affecting S3. However, we would happily take a patch, although I understand that adds a burden of figuring out the Java  code generator logic, etc.

shanebarnes:
Thanks for the explanation, @marcomagdy. I will look into creating a patch specifically for the S3 code.

Issue #1199, opened 2019-07-18
SDK fails to compile on Debian 10 armv7l


skyhisi:
### What platform/OS are you using?
Debian 10 (Buster)
Raspberry Pi 4B (armv7l)
Linux rpi.lan 4.19.57-v7l+ #1244 SMP Thu Jul 4 18:48:07 BST 2019 armv7l GNU/Linux

### Which version of the SDK?
Git revision: 234716a

### What compiler are you using? what version?
g++ (Raspbian 8.3.0-6+rpi1) 8.3.0
cmake version 3.13.4
ninja version 1.8.2

### What are your CMake arguments?
-G Ninja
-D CMAKE_BUILD_TYPE=Release
-DBUILD_ONLY=email;lambda;s3;sns;sqs
-DENABLE_TESTING=OFF

### Can you provide a TRACE level log? (sanitize any sensitive information)
End of console log
```plain
-- Performing Test HAVE_ATOMICS_WITHOUT_LIBATOMIC
-- Performing Test HAVE_ATOMICS_WITHOUT_LIBATOMIC - Failed
-- Looking for __atomic_load_8 in atomic
-- Looking for __atomic_load_8 in atomic - not found
CMake Error at cmake/platform/linux.cmake:46 (message):
  Could not determine support for atomic operations.
Call Stack (most recent call first):
  CMakeLists.txt:206 (apply_post_project_platform_settings)
```

CMakeError.log
```plain
Performing C++ SOURCE FILE Test HAVE_ATOMICS_WITHOUT_LIBATOMIC failed with the following output:
Change Dir: /var/lib/jenkins/jobs/aws-sdk-cpp-raspi/workspace/build/CMakeFiles/CMakeTmp

Run Build Command:"/usr/bin/ninja" "cmTC_144a6"
[1/2] Building CXX object CMakeFiles/cmTC_144a6.dir/src.cxx.o
[2/2] Linking CXX executable cmTC_144a6
FAILED: cmTC_144a6 
: && /usr/lib/ccache/c++  -DHAVE_ATOMICS_WITHOUT_LIBATOMIC -std=c++11  -rdynamic CMakeFiles/cmTC_144a6.dir/src.cxx.o  -o cmTC_144a6   && :
/usr/bin/ld: CMakeFiles/cmTC_144a6.dir/src.cxx.o: in function `std::__atomic_base<unsigned long long>::operator unsigned long long() const':
[...]

PR #1243, opened 2019-09-22
cmake: fix libatomic check with GCC 8

lopsided98:
*Issue #, if available:*
#1199 

*Description of changes:*

Fixes detection of libatomic with GCC 8. This solution is based on https://bugs.debian.org/cgi-bin/bugreport.cgi?bug=907277#25, which describes why the error occurs (in that case with autotools).

Some of the possible solutions are:
1. Unconditionally call `enable_language(C)` so that the check runs in C mode (untested)
2. Run a compile test that calls `__atomic_load_8` with the right number of arguments
3. Drop the explicit `libatomic` test altogether and just check whether C++ atomics work when linked to `libatomic`

I chose option 2 because it maintained the existing semantics and seemed cleaner than using the C compiler to work around this issue.

*Check all that applies:*

- [x] Linux
- [ ] Windows
- [ ] Android
- [ ] MacOS
- [ ] IOS
- [ ] Other Platforms

### DIFF
--- a/cmake/platform/linux.cmake
+++ b/cmake/platform/linux.cmake
@@ -15,6 +15,12 @@ int main() {
     return x + y;
 }")
 
+set(LIBATOMIC_TEST_CXX_SOURCE "
+int main() {
+    __atomic_load_8(nullptr, 0);
+    return 0;
+}")
+
 macro(apply_post_project_platform_settings)
     include(GNUInstallDirs)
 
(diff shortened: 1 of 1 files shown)


emersonknapp:
Hi! Will this be merged soon? This problem is affecting us on the latest Raspbian

KaibaLopez:
Sorry for hte lack of update here but the PR was merged, closing it.

Issue #1307, opened 2020-01-16
Update the "Using CMake exports with AWS sdk for C++"


Lectem:
The documentation points to https://aws.amazon.com/fr/blogs/developer/using-cmake-exports-with-the-aws-sdk-for-c/ which suggests to use `find_package(aws-sdk-cpp)` while from what I gathered we are now supposed to use `find_package(AWSSDK)` ?

It would also be nice if the exported target were namespaced with `AWS::`. This is already the case for a few targets such as `AWS::aws-c-common` or the lambda runtime `AWS::aws-lambda-runtime`. 

Currently we have to use either the non-namespaced targets (which has known issues in cmake, as if the target doesnt exist the configuration will not fail) or `${AWSSDK_LINK_LIBRARIES}` which does not give any flexibility.

I tried modifying the `do_packaging` macro to use namespace (add `NAMESPACE AWS::` here https://github.com/aws/aws-sdk-cpp/blob/master/cmake/utilities.cmake#L70 and here https://github.com/aws/aws-sdk-cpp/blob/master/cmake/utilities.cmake#L88 ).

And it seems to work fine so far.

KaibaLopez:
First, you're right, we should update that, thanks for pointing it out to us.
The rest would take a bit more time and discussion but does make sense, could you make a pull request with the changes you made?

dblock:
I PRed https://github.com/awsdocs/aws-cpp-developer-guide/pull/37 and https://github.com/awsdocs/aws-doc-sdk-examples/pull/1022 that should cover the documentation changes described here.

PR #1308, opened 2020-01-20
Use AWS:: namespace for cmake targets

Lectem:
Issue #1307  

*Description of changes:*

The namespace AWS was added to the installed target. This let the consumers use the targets prefixed with `AWS::` like so:

`target_link_libraries(my_lib PRIVATE AWS::aws-cpp-sdk-s3)`

However users who were using the version without namespace will not get the transitive properties of the target(it will however link the library):

`target_link_libraries(my_lib PRIVATE aws-cpp-sdk-s3)`

It seems that AWSSDK_LINK_LIBRARIES is not populated correctly either since it uses the non-namespaced version of the targets. I did not find where the variables such as `AWSSDK_PLATFORM_DEPS_LIBS` `AWSSDK_CRYPTO_LIBS` etc are populated, and hence was unable to fix this.

This was only tested with `find_package(AWSSDK REQUIRED COMPONENTS s3)` but should work correctly when multiple components are required, or none specified.

*Check all that applies:*
- [X] Checked if this PR is a breaking (APIs have been changed) change. **Yes, using non-namespaced target names might cause consumer builds to fail. Fix is to add `AWS::`**
- [X] Checked if this PR would require a ReadMe/Wiki update. **This will be needed.**

- [X] Linux
- [X] Windows
- [ ] Android
- [ ] MacOS
- [ ] IOS
- [ ] Other Platforms

### DIFF
--- a/cmake/utilities.cmake
+++ b/cmake/utilities.cmake
@@ -68,6 +68,7 @@ macro(do_packaging)
 
         export(EXPORT "${PROJECT_NAME}-targets"
             FILE "${CMAKE_CURRENT_BINARY_DIR}/${PROJECT_NAME}-targets.cmake"
+            NAMESPACE AWS::
         )
 
     if(${PROJECT_NAME} STREQUAL "aws-cpp-sdk-core")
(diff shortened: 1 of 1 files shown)


Lectem:
Updated the original PR description to mention that this is not ready for merge yet.
`AWSSDK_LINK_LIBRARIES` is not populated with the targets prefixed by the namespace, which causes missing transitive properties of the CMake targets.

PR #1333, opened 2020-03-03
Add support for building with MinGW

trilader:
*Issue #, if available:*

*Description of changes:*
This PR adds support for building aws-sdk-cpp with the MinGW toolchain on Windows.
Needs https://github.com/awslabs/aws-c-common/pull/604 in order to work.

The main changes are adding checks for the __MINGW32__ define which is preset on both 32 and 64 bit MinGW installations, silencing some bogus warnings and addressing some `0` vs `NULL` cases which gcc complains about.

I don't this this needs explicit tests but might need additional CI work to ensure MinGW continues to work in the future.
Regarding ReadMe/Wiki: It might need mentioning that building with MinGW is supported.

*Check all that applies:*

- [x] Linux
- [x] Windows
- [ ] Android
- [ ] MacOS
- [ ] IOS
- [ ] Other Platforms


jeroen:
FYI, we have adopted a modified version of this patch in msys2 [mingw-w64-aws-sdk-cpp](https://github.com/msys2/MINGW-packages/tree/master/mingw-w64-aws-sdk-cpp). To install binaries from msys2 use 


```sh
pacman -Syu
pacman -S mingw-w64-{i686,x86_64}-aws-sdk-cpp
```

jmklix:
Thanks for taking the time to make this PR. We are still interested in adding MinGW support for this sdk, but the implementation will need to be changed slightly and this isn't a high priority currently. Also we will have to add CI so this isn't broken in with any future releases. Because of this I will be converting this PR to a draft.

PR #1342, opened 2020-03-11
Update opendjk 8 link for Debian

alanyee:
*Description of changes:*
Remove old link in the Dockerfile. The old link returns a 404 response, so the current Dockerfile cannot successfully build. With the updated link, however, it successfully builds.

*Check all that applies:*

- [ ] Linux
- [ ] Windows
- [ ] Android
- [X] MacOS
- [ ] IOS
- [X] Other Platforms

### DIFF
--- a/CI/docker-file/Debain/Dockerfile
+++ b/CI/docker-file/Debain/Dockerfile
@@ -10,14 +10,14 @@ RUN apt-get update && \
 RUN pip install awscli --upgrade
 
 # We can install openjdk by "apt install openjdk-8-jdk", but it has some issues during building code-generation, we have to install it manually.
-RUN wget --no-check-certificate -c --header "Cookie: oraclelicense=accept-securebackup-cookie" https://download.oracle.com/otn-pub/java/jdk/8u191-b12/2787e4a523244c269598db4e85c51e0c/jdk-8u191-linux-x64.tar.gz && \
-    tar zxvf jdk-8u191-linux-x64.tar.gz && \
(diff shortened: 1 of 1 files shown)


alanyee:
Yes, this is still an issue in the latest release

wps132230:
Hi @alanyee 
I am testing JDK installation on Debian, and I found that manual download is not required any more, we could install JDK by `apt-get install default-jdk`. But it's OpenJDK 11, rather than 8 (both work with C++ SDK though).
Is that fine to use `apt-get` to install OpenJDK 11?
If the solution is fine to you, we will update the dockerfile and no action need to take from your side.

alanyee:
@wps132230 `apt-get` is fine to me

KaibaLopez:
This PR has been accepted and will be merged on the next release tomorrow, thanks and sorry again for the long time it took.

alanyee:
@KaibaLopez will the PR be accepted as is, or will there be changes to my PR as according to @wps132230? If the former, the current link in my PR is returning `404` now, but I can update them to existing, current versions. Otherwise, if the latter, I have done a quick check via `apt-get install -y default-jdk` per wps132230's suggestion, and that seems to work.

wps132230:
Hi @alanyee, we will use `apt-get install default-jdk`.
The reason we download jdk manually is the default-jdk at that moment is not compatible with C++ SDK.

alanyee:
Closing via https://github.com/aws/aws-sdk-cpp/commit/6af954a8dbcea742c87bf7cb96ce9b160626344b. Thank you everyone!

Issue #1188, opened 2019-07-06
cmake: cmake can't find libraries on linux


dmitryash:
Hello,

I get cmake error when run cmake for project which uses AWSSDK:

```
CMake Error at /usr/lib64/cmake/AWSSDK/AWSSDKConfig.cmake:122 (message):
  AWS SDK for C++ headers found, but we were unable to locate the binaries.
  Have you deleted or moved it?

              Please make sure header files and binaries are located in INSTALL_ROOT_DIR/INCLUDE_DIR/ and INSTALL_ROOT_DIR/LIB_DIR/[PLATFORM_PREFIX]/[Debug|Config|OtherConfigs]
Call Stack (most recent call first):
  CMakeLists.txt:5 (find_package)


-- Configuring incomplete, errors occurred!
```

This error happens because header files are installed to `/usr/include` and libraries are installed to `/usr/lib64` directory. These directories are set by rpm macro `%cmake` provided by Linux OS maintainers.

I see that AWSSDKConfig.cmake finds header files because it enabled system directories. However then it never looks at system directories. As you can see below, system directoryes with libraries are just skipped.

https://github.com/aws/aws-sdk-cpp/blob/ec618a42ae702c86e5466ae637ff903028af11f6/cmake/AWSSDKConfig.cmake#L110-L119

dmitryash:
I found that platformDeps.cmake is generated with:
```
set(AWSSDK_INSTALL_LIBDIR /usr/lib64)
```

This was most likely because `%cmake` macro defines CMAKE_INSTALL_LIBDIR to /usr/lib64

singku:
Hi, we don't ship rpm install package for AWS SDK for CPP. I suggest you to compile and install sdk from source to make it work properly.

dmitryash:
Hi, it is OK that you don't ship rpm package. Basically distro maintainers knowns details of their distro and pack rpm in a proper way. I'm trying to pack sdk into rpm and let users install rpm packages instead. Installation from sources is not a solution.
What really matters is that SDK should use cmake correctly. In this particular case cmake documentation says that `CMAKE_INSTALL_LIBDIR` is supposed to be put to `imstall()` command. At the same time this `install()` command allows specifying both absolute and relative paths. As a result it is allowed that `CMAKE_INSTALL_LIBDIR` is absolute path.

singku:
Oh I see. I thought you are integrating SDK with a 3rd party rpm package. 
[...]

PR #1365, opened 2020-04-15
Make generated CMake files valid when install paths are absolute

thequux:
*Issue #, if available:*
NixOS/nixpkgs#70075, #1188 

Similar alternative patch: #1315 

The difference between this patch and #1315 is that the change to the logic in this patch is much simpler, and it also fixes the pkgconfig files.

*Description of changes:*

Problem
-------

If the install directories are given as absolute paths, the generated
CMake files break completely; in particular, on NixOS, they will try
to read headers from

/nix/store/somelonghash-aws-version-dev//nix/store/somelonghash-aws-sdk-cpp-version-dev/include

and find the library binaries in

/nix/store/somelonghash-aws-version-dev//nix/store/somelonghash-aws-sdk-cpp-version/lib

The part before the `//` is what is discovered as AWSSDK_ROOT_DIR.

Fix
---

If CMAKE_INSTALL_LIBDIR and friends are absolute, then they should not
have AWSSDK_ROOT_DIR prepended to them.

Unfortunately, this is tricky to do in a small patch, but on the
assumption that if one of these variables is absolute, then so are the
rest, I have picked AWSSDK_INSTALL_LIBDIR as the representative sample.

If this variable is unset, then AWSSDK_ROOT_DIR is forced to be "" so
that in can still be prepended to the install paths and get a valid
path as a result.

*Check all that applies:*

Tests are not applicable because this is purely a change to the build system.

I strongly suspect that this will not introduce cross-platform inconsistent behaviour, as I don't know enough about cmake's path handling on Windows to be sure. However, I am completely sure that this will behave the same across all the various Unix-like platforms.

- [x] Linux
- [ ] Windows
- [ ] Android
- [ ] MacOS
- [ ] IOS
- [ ] Other Platforms


rehno-lindeque:
Hi, another nix user here - just wanted to bump this PR since it's been a while since it was raised and I encountered the same problem.

jmklix:
https://github.com/aws/aws-sdk-cpp/issues/1888

jmklix:
Closing in favor of updated PR

PR #1395, opened 2020-06-12
Made S3Client::ComputeEndpointString() public for ease of endpoint de…

dburken:
Would like to make S3Client::ComputeEndpointString() public for ease of client configuration / endpoint debugging on disconnected systems such as snowballs. This is a const method so it's a safe change.

*Issue #, if available:* N/A

*Description of changes:* Moved method from private to public in class declaration.

*Check all that applies:*
- [x ] Did a review by yourself.

- [x ] Linux
- [x ] Windows
- [ ] Android
- [ ] MacOS
- [ ] IOS
- [ ] Other Platforms

### DIFF
--- a/aws-cpp-sdk-s3/include/aws/s3/S3Client.h
+++ b/aws-cpp-sdk-s3/include/aws/s3/S3Client.h
@@ -9416,11 +9416,13 @@ namespace Aws
         virtual bool MultipartUploadSupported() const;
 
         void OverrideEndpoint(const Aws::String& endpoint);
+
+        ComputeEndpointOutcome ComputeEndpointString() const;
+        
     private:
         void init(const Client::ClientConfiguration& clientConfiguration);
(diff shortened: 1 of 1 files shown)


KaibaLopez:
Hi @dburken ,
Sorry for the late response and thanks for the PR but, I'm not convinced this is a good change to implement, could you maybe give more context of a situation where this change would be necessary? Also, this change would have to be done through the code generator so that also complicates the pr a bit...

Having said that, nothing really stopping you from doing it for yourself or creating a custom s3client where you expose this, as you say it is const so it should be safe to do so.

dburken:
@KaibaLopez If it is too painful, you can close/kill it.  I did not realize that was generated code. I was working with snowballs on a disconnected system and just wanted to see what the client configuration resolved to endpoint wise, so I made that public and wrapped it around trace code. Anyway thanks...

PR #1429, opened 2020-08-05
Copy only the newly read bytes fromn the buffer

cosu:
*Issue #, if available:*

*Description of changes:*

 With the current behaviour the last iteration appends both newly read bytes and bytes from the previous read. This results in a string which contains the json file and some fragments. Because the generator uses a very old version of gson in the pom.xml it will happily work with broken json but newer versions will refuse to load that string and throw an invalid json exception. This change ensures that only the new bytes are appended to the string. 

*Check all that applies:*

- [x] Linux
- [ ] Windows
- [ ] Android
- [ ] MacOS
- [ ] IOS
- [ ] Other Platforms

### DIFF
--- a/code-generation/generator/src/main/java/com/amazonaws/util/awsclientgenerator/main.java
+++ b/code-generation/generator/src/main/java/com/amazonaws/util/awsclientgenerator/main.java
@@ -67,8 +67,9 @@ public class main {
                 StringBuilder stringBuilder = new StringBuilder();
 
                 byte[] buffer = new byte[1024];
-                while (stream.read(buffer) > 0) {
-                    stringBuilder.append(new String(buffer, StandardCharsets.UTF_8));
+                int bytes;
+                while ((bytes = stream.read(buffer)) > 0) {
+                    stringBuilder.append(new String(buffer, 0, bytes, StandardCharsets.UTF_8));
                 }
 
                 arbitraryJson = stringBuilder.toString();


code-generation/generator/src/main/java/com/amazonaws/util/awsclientgenerator/main.java
```diff
@@ -67,8 +68,9 @@ public static void main(String[] args) throws IOException {
                 StringBuilder stringBuilder = new StringBuilder();
 
                 byte[] buffer = new byte[1024];
-                while (stream.read(buffer) > 0) {
-                    stringBuilder.append(new String(buffer, StandardCharsets.UTF_8));
+                int bytes;
+                while ((bytes = stream.read(buffer)) > 0) {
+                    stringBuilder.append(new String(Arrays.copyOf(buffer, bytes), StandardCharsets.UTF_8));
```
webbnh, line 73: Why interpose an `Arrays` object?  Why not just use a better `String` constructor:
```
String(byte[] bytes, int offset, int length, Charset charset)
```

On a related note, does this code actually need to instantiate a new `String`, or can we get away with just passing a subset of `buffer` directly?

cosu: That's a good callout. I've updated the code to skip the extra buffer copy. 

the bytes need to be converted to string/char somewhere and StringBuffer doesn't expose that functionality.

KaibaLopez:
this pr has been merged so I'll be closing it, thanks for the contribution!

PR #1441, opened 2020-08-21
Update CMakeLists.txt replace find_package with aws_use_package for aws-c-event-stream

mattparks:
*Issue #, if available:*
When building alongside aws-c-event-stream find_package pulls in the system library instead of the in source target. 

*Description of changes:*
This change brings over the concept of `aws_use_package` from aws-c-common to aws-cpp-sdk-core. I did not add any testing because with `IN_SOURCE_BUILD=OFF` the same fine_package behaviour is maintained.

*Check all that applies:*

- [x] Linux
- [x] Windows
- [x] Android
- [ ] MacOS
- [ ] IOS
- [ ] Other Platforms

### DIFF
--- a/cmake/AwsFindPackage.cmake
+++ b/cmake/AwsFindPackage.cmake
@@ -1,0 +1,22 @@
+# Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.
+# SPDX-License-Identifier: Apache-2.0.
+
+option(IN_SOURCE_BUILD "If the CRT libs are being built from your source tree (add_subdirectory), set this to ON" OFF)
+
+# This function handles dependency list building based on if traditional CMAKE modules via. find_package should be
+# used, vs if this is an in source build via. something like git submodules and add_subdirectory.
+# This is largely because CMake was not well planned out, and as a result, in-source and modules don't play well
+# together. Only use this on CRT libraries (including S2N), libcrypto will stay as an assumed external dependency.
+#
+# package_name: is the name of the package to find
+# DEP_CRT_LIBS: output variable will be appended after each call to this function. You don't have to use it,
+#    but it can be passed directly target_link_libraries and it will be the properly qualified library
+#    name and namespace based on configuration.
+function(aws_use_package package_name)
+    if (IN_SOURCE_BUILD)
+        set(DEP_AWS_LIBS ${DEP_AWS_LIBS} ${package_name} PARENT_SCOPE)
+    else()
+        find_package(${package_name} REQUIRED)
+        set(DEP_AWS_LIBS ${DEP_AWS_LIBS} AWS::${package_name} PARENT_SCOPE)
+    endif()
+endfunction()
--- a/cmake/AwsFindPackage.cmake
+++ b/cmake/AwsFindPackage.cmake
@@ -9,7 +9,7 @@ option(IN_SOURCE_BUILD "If the CRT libs are being built from your source tree (a
 # together. Only use this on CRT libraries (including S2N), libcrypto will stay as an assumed external dependency.
 #
 # package_name: is the name of the package to find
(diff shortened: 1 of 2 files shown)


cmake/AwsFindPackage.cmake
```diff
@@ -0,0 +1,22 @@
...
+
+option(IN_SOURCE_BUILD "If the CRT libs are being built from your source tree (add_subdirectory), set this to ON" OFF)
+
+# This function handles dependency list building based on if traditional CMAKE modules via. find_package should be
+# used, vs if this is an in source build via. something like git submodules and add_subdirectory.
+# This is largely because CMake was not well planned out, and as a result, in-source and modules don't play well
+# together. Only use this on CRT libraries (including S2N), libcrypto will stay as an assumed external dependency.
+#
+# package_name: is the name of the package to find
+# DEP_CRT_LIBS: output variable will be appended after each call to this function. You don't have to use it,
```
webbnh, line 12: Presumably you meant `DEP_AWS_LIBS` instead of `DEP_CRT_LIBS`, here.

cmake/AwsFindPackage.cmake
```diff
@@ -0,0 +1,22 @@
+# Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.
+# SPDX-License-Identifier: Apache-2.0.
+
+option(IN_SOURCE_BUILD "If the CRT libs are being built from your source tree (add_subdirectory), set this to ON" OFF)
+
+# This function handles dependency list building based on if traditional CMAKE modules via. find_package should be
+# used, vs if this is an in source build via. something like git submodules and add_subdirectory.
+# This is largely because CMake was not well planned out, and as a result, in-source and modules don't play well
+# together. Only use this on CRT libraries (including S2N), libcrypto will stay as an assumed external dependency.
```
webbnh, line 9: This is rather tortured English.  And, presumably there is no need to comment on whether CMake, etc. was "well planned" or not.  And, it's unclear (at least to me) what you mean by "CRT libraries" (i.e., presumably you don't mean the C Language Run-Time Library or any of the analogous C++ Language ones).

How about something like:
> This function handles building a list of external dependencies, taking into account whether the dependency is built as a submodule or found elsewhere on the build platform.  It should be used as a replacement for `find_package()` in cases where the dependency might satisfied either by using an installed library or by building it as part of the build.  The `IN_SOURCE_BUILD` option determines how the dependency will be satisfied.  For dependencies which are never satisfied as part of the build (such as `libcrypto`), `find_package()` should be used instead of this function.

mattparks: This file is from https://github.com/awslabs/aws-c-common/blob/master/cmake/AwsFindPackage.cmake
If you want me to update the comment I can add to this pr.

webbnh: Sorry -- Github's presentation made it look like this was a new file that you were adding, so I figured it would be best if it had better verbiage before it was merged in.

In any case, I was just making a suggestion -- I'm not an `aws-sdk-cpp` maintainer, so I'll defer to them on what they want.

wps132230:
Hi there, thank you for your pull request.
We are working with maintainers of aws-c-* to have a better approach to consume their libraries.
And one improvement is to use `aws_use_package`.

wps132230:
Hi @mattparks We've updated a lot in `aws-cpp-sdk-core/CMakeLists.txt` since 1.9.x.
And it's using `aws_use_package(aws-crt-cpp)` instead of `aws-c-event-stream` now.
Does it work in your scenario? And do you still need changes in this PR?

jmklix:
https://github.com/aws/aws-sdk-cpp/issues/1888

jmklix:
Closing this PR. Thanks for taking the time to make this PR, but it is now outdated thanks to some changes in the build process of this sdk.

PR #1443, opened 2020-08-24
Add SOCKS5 and SOCKS5H schemes

ahmedyarub:
*Description of changes:*
Added schemes required for proxying connections to socks5 and socks5h (with host resolution) with CURL library

*Check all that applies:*
Not applicable: it doesn't add anything functional. It just enables prepending socks5 and socks5h to the host name

- [X] Linux
- [X] Windows
- [X] Android
- [X] MacOS
- [X] IOS
- [X] Other Platforms
WebAssembly


KaibaLopez:
Hi @ahmedyarub ,
Could you give some description to your use case and also, could you use the endpoint override function in the clientconfiguration to use these or is there a limitation there?

ahmedyarub:
proxifying your connections to AWS is a very common requirement when using AWS SDK. unfortunately, since the default client only accepts http and https schemas, we cannot use socks5 and socks5h proxies that are already supported by curl. socks5h is especially useful in my case since I'm running my WebAssembly application on Google Chrome, and the only way to connect to AWS servers with WebAssembly is by proxifying the requests to a SOCKS proxy AND doing the DNS resolution on the proxy itself.
I could create a new client and use the overriding function but I found it too much of a hassle to create a new client and implement the overriding method and then setting as a default client configuration just so that I can add the schema.

anubes19 (approved):

ahmedalnuaimi:
Any chance this will be merged?

PR #1453, opened 2020-09-10
Prevent deadlock in CurlHandleContainer::AcquireCurlHandle()

amitojsinghbhatia:
*Issue #, if available:*

### Brief
When a heavily multi-threaded application creates a lot of S3Clients, a deadlock is observed in CurlHandleContainer::AcquireCurlHandle()

### Issue Description:
Consider this snippet from AcquireCurlHandle:

    if(!m_handleContainer.HasResourcesAvailable())   
    {  
        CheckAndGrowPool();  
    }  
    CURL* handle = m_handleContainer.Acquire();

If the pool has already reached max limit, it will not grow further and caller will block at `m_handleContainer.Acquire();`
However, if the thread that has acquired the handle, encounters an error, it calls DestroyCurlHandle() instead of ReleaseCurlHandle(), which does not release the handle back into the pool. If none of the threads are released back, then the waiting threads will never get the handle.
Added a unit test to reproduce the scenario.

### Fix details
I am assuming there was a good reason to call DestroyCurlHandle() instead of just resetting and release it back. As a proposed fix, I have created a new handle to replace the one being destroyed so that pool size remains the same.

*Check all that applies:*

- [x] Linux
- [ ] Windows
- [ ] Android
- [ ] MacOS
- [ ] IOS
- [ ] Other Platforms

### DIFF
--- a/aws-cpp-sdk-core-tests/http/HttpClientTest.cpp
+++ b/aws-cpp-sdk-core-tests/http/HttpClientTest.cpp
@@ -11,17 +11,18 @@
 #include <aws/core/http/standard/StandardHttpRequest.h>
 #include <aws/core/client/ClientConfiguration.h>
 #include <aws/core/utils/logging/LogMacros.h>
+#include <future>
+#include <chrono>
 
 using namespace Aws::Http;
 using namespace Aws::Utils;
 using namespace Aws::Client;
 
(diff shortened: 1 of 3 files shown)


KaibaLopez:
This PR has been merged, thanks for the contribution.

PR #1480, opened 2020-09-30
Don't require git to download external projects

nealrichardson:
*Issue #, if available:* N/A

*Description of changes:*

Building `aws-sdk-cpp` with the default `BUILD_DEPS=ON` fails if `git` is not found on the system:

```
-- Configuring incomplete, errors occurred!

-- stderr output is:
CMake Error at /tmp/RtmpqRJAdR/file1835205980df/cmake-3.18.1-Linux-x86_64/share/cmake-3.18/Modules/ExternalProject.cmake:2350 (message):
  error: could not find git for clone of AwsCCommon
Call Stack (most recent call first):
  /tmp/RtmpqRJAdR/file1835205980df/cmake-3.18.1-Linux-x86_64/share/cmake-3.18/Modules/ExternalProject.cmake:3228 (_ep_add_download_command)
  cmake/BuildAwsCCommon.cmake:39 (ExternalProject_Add)
  CMakeLists.txt:19 (include)


CMake Error at CMakeLists.txt:218 (message):
  Failed to configure third-party libraries.
```

These `third-party` cmake files specify a git repository and tag for the source for each library, but we can download the source without `git` by constructing a URL.

On https://github.com/apache/arrow/pull/8304 I've tested that this works as expected by applying it in a `PATCH_COMMAND`.

*Check all that applies:*
- [x] Added proper tests to cover this PR. (If tests are not applicable, explain.) *Any CI you have that tests these cmake files will cover the changes; there are no behavior changes*

- [x] Linux
- [ ] Windows
- [ ] Android
- [ ] MacOS
- [ ] IOS
- [ ] Other Platforms

### DIFF
--- a/third-party/CMakeLists.txt
+++ b/third-party/CMakeLists.txt
@@ -14,15 +14,15 @@ set(AWS_DEPS_INSTALL_DIR "${CMAKE_INSTALL_PREFIX}" CACHE PATH "Dependencies inst
 set(AWS_DEPS_BUILD_DIR "${CMAKE_BINARY_DIR}/build" CACHE PATH "Dependencies build directory.")
 set(AWS_DEPS_DOWNLOAD_DIR "${AWS_DEPS_BUILD_DIR}/downloads" CACHE PATH "Dependencies download directory.")
 
-set(AWS_C_COMMON_URL "https://github.com/awslabs/aws-c-common.git")
+set(AWS_C_COMMON_URL "https://github.com/awslabs/aws-c-common")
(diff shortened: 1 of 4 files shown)


wps132230:
Hi there, we are using git submodule instead of CMake external project now.
If you don't have git on your system, you could run this script to download all submodules: https://github.com/aws/aws-sdk-cpp/blob/main/prefetch_crt_dependency.sh
Does that meet your requests?

nealrichardson:
I think we worked around this by building the dependencies ourselves rather than relying on BUILD_DEPS=ON

PR #1541, opened 2020-12-19
Do not force to run a test when crosscompiling

diegonieto:
*Issue #, if available:*

*Description of changes:*
I'm using Amazon TranscribeStreaming in both PC64 and ARM64 machines.

When crosscompiling the SDK there is a check_c_source_runs CMake sentence that forces to run a CURL code in order to enable its functionality. Since the crosscompiled code is not runnable in the host machine, the checker fails when CURL is actually capable to work with both HTTP2 and TLS Proxy connections. This is affecting to the TranscribeStreaming service.
My proposal is that for SDK cross-compilations CMake should check weather CURL is compilable instead of runnable. Of course, both HTTP2 and TLS Proxy may not be enable, can be checked in the target machine, but at least the SDK does not discard this feature just because it is not runnable in the target machine at compile time.

*Check all that applies:*

- [x] Linux
- [ ] Windows
- [ ] Android
- [ ] MacOS
- [ ] IOS
- [ ] Other Platforms

### DIFF
--- a/aws-cpp-sdk-core/CMakeLists.txt
+++ b/aws-cpp-sdk-core/CMakeLists.txt
@@ -86,23 +86,30 @@ file(GLOB UTILS_CRYPTO_FACTORY_SOURCE "${CMAKE_CURRENT_SOURCE_DIR}/source/utils/
 include(CheckCXXSourceCompiles)
 include(CheckCSourceRuns)
 include(CheckCXXSourceRuns)
+include(CheckCSourceCompiles)
 
 # http client implementations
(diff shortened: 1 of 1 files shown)


diegonieto:
Hi guys, do you have any news about this?

KaibaLopez:
Hi @diegonieto ,
Sorry for the long wait here, but yea we have merged this PR, and should be available in the newest version of the SDK once the new release happens (at some point later today).
Thanks and sorry again for the wait.

PR #1668, opened 2021-06-04
fix: check for minimal supported macos version

gjasny:
Before this change the only the version of the macOS SDK
was checked, to find out if the modern function declarations
are available at all.

But the code does not take the minimal supported macos version
into account. Even on macos 11.0 you could set
`-mmacosx-version-min=10.8`. Before ths change this led to the
following error:

```
error: 'CCCryptorGCMSetIV' is only available on macOS 10.13 or newer
```

The code is now using the compiler built-in runtime check to
dynamically switch between versions. This also plays along nicely
with the Apple availability macros.

If a newer macOS SDK than 10.13 SDK is detected it could also be newer
than 10.14.4 at the same time.

*Check all that applies:*

- [ ] Linux
- [ ] Windows
- [ ] Android
- [x] MacOS
- [ ] IOS
- [ ] Other Platforms

### DIFF
--- a/aws-cpp-sdk-core/source/utils/crypto/commoncrypto/CryptoImpl.cpp
+++ b/aws-cpp-sdk-core/source/utils/crypto/commoncrypto/CryptoImpl.cpp
@@ -539,10 +539,13 @@ AWS_SUPPRESS_DEPRECATION(
                                                                  nullptr, m_key.GetUnderlyingData(), m_key.GetLength(),
                                                                  nullptr, 0, 0, kCCModeOptionCTR_BE, &m_encryptorHandle);
 #ifdef MAC_13_AVAILABLE
-                status |= CCCryptorGCMSetIV(m_encryptorHandle, m_initializationVector.GetUnderlyingData(), m_initializationVector.GetLength());
-#else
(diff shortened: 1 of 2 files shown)


gjasny:
CC: @singku

Terrybrar97 (approved):
Hmm

aws-cpp-sdk-core/source/utils/crypto/commoncrypto/CryptoImpl.cpp
```diff
@@ -602,16 +613,20 @@ AWS_SUPPRESS_DEPRECATION(
...
                 size_t tagLength = TagLengthBytes;
 
                 /* Note that CCCryptorGCMFinal is deprecated in Mac 10.13. It also doesn't compare the tag with expected tag
                  * https://opensource.apple.com/source/CommonCrypto/CommonCrypto-60118.1.1/include/CommonCryptorSPI.h.auto.html
                  */
 #ifdef MAC_13_AVAILABLE
-                status = CCCryptorGCMFinalize(m_decryptorHandle, m_tag.GetUnderlyingData(), tagLength);
+                if (__builtin_available(macos 10.13, ios 11.0, *)) {
+                  status = CCCryptorGCMFinalize(m_decryptorHandle, m_tag.GetUnderlyingData(), tagLength);
+                } else
```
singku, line 625: This is a bit hard to read. Can you please make it more clear?

Issue #1745, opened 2021-08-19
Unable to compile debug version of AWS sdk on Windows using CMake generated solution files.


jorgen:
Confirm by changing [ ] to [x] below to ensure that it's a bug:
- [x ] I've gone though [Developer Guide](https://docs.aws.amazon.com/sdk-for-cpp/v1/developer-guide/welcome.html) and [API reference](http://sdk.amazonaws.com/cpp/api/LATEST/index.html)
- [x ] I've searched for [previous similar issues](https://github.com/aws/aws-sdk-cpp/issues) and didn't find any solution
  
If I generate a Visual Studio Solution and try and build it will stop with the error:
6>Building Custom Rule C:/Projects/aws-cpp-sdk-1.9.83/crt/aws-crt-cpp/crt/aws-checksums/CMakeLists.txt
5>cl : command line error D8016: '/O2' and '/RTC1' command-line options are incompatible

When inspecting some of the sub-projects in aws-crt-cpp it seems like they end up having O2 set for the Debug build. This is incompatible with some of the other debug flags.

This does not happen if I clone the relevant sub-projects and build them directly.

**SDK version number**
1.9.45 -> 1.9.83

**Platform/OS/Hardware/Device**
Windows 10

**To Reproduce (observed behavior)**
Generate a Visual Studio solution and build debug.

KaibaLopez:
Hi @jorgen ,
I'm a bit lost trying to reproduce this, you say this happens on Debug build, does that mean that for you it worked on Release? Can you walk me through how you build the SDK (git command & cmake command)? 
And kind of a shot in the dark here but, did you include the submodules in your clone command?

jorgen:
git pull --recursive. I have checked out the 5aa7d65f7f75225d7b7f884626d2754a912d0b8b commit since it is tagged with 1.9.83.
I use cmake-gui to generate a Visual Studio 19 solution (Win64 is the default architecture) . If you open the solution and build a debug build you will see that the build fails because of incompatible compiler flags. The release config builds fine.

KaibaLopez:
Can you check on the cmake-gui select the build binaries folder you've built it on, and see what the attribute `DCMAKE_BUILD_TYPE` is? 
By default it would be Release and that would explain this errors can you verify?

jorgen:
Ah! I understand whats going on! I have created a merge-request: #1748

KaibaLopez:
[...]

PR #1748, opened 2021-08-20
Fix multiconfig generators not working by forcing single build type

jorgen:
*Issue #1745

By setting the cache variable CMAKE_BUILD_TYPE multi config generators such as the Visual Studio Solution generator are unable to produce more than one build configuration, leading to unexpected build failures. Setting this CMAKE_ variable is a regression from the 1.8.x series.

Our build scripts just did a cmake and then used cmake --build x --config y to build the two configurations. Its not a huge problem to fix on our end, but it forces developers that uses Visual Studio generated solutions to have two solutions.


jorgen:
A comment on this would be nice

KaibaLopez:
Hi @jorgen, 
This change has been merged, thank you for opening the PR, and sorry for the lack of response, it's something we need to improve in our repo.

Issue #1750, opened 2021-08-25
Compile Error Version 1.9.88 gcc 11.1.0


theShmoo:
Compile error because of warning:

Output:
```
[ 89%] Building CXX object aws-cpp-sdk-core/CMakeFiles/aws-cpp-sdk-core.dir/source/utils/event/EventStreamDecoder.cpp.o
In file included from aws-sdk-cpp/aws-cpp-sdk-core/include/aws/core/utils/event/EventHeader.h:13,
                 from /aws-sdk-cpp/aws-cpp-sdk-core/source/utils/event/EventStreamDecoder.cpp:7:
/aws-sdk-cpp/aws-cpp-sdk-core/source/utils/event/EventStreamDecoder.cpp: In function 'static void Aws::Utils::Event::EventStreamDecoder::onPayloadSegment(aws_event_stream_streaming_decoder*, aws_byte_buf*, int8_t, void*)':
/aws-sdk-cpp/aws-cpp-sdk-core/source/utils/event/EventStreamDecoder.cpp:77:77: error: 'this' pointer is null [-Werror=nonnull]
   77 |                         "ErrorMessage: " << handler->GetEventPayloadAsString());
      |                                             ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~^~
/aws-sdk-cpp/aws-cpp-sdk-core/include/aws/core/utils/logging/LogMacros.h:143:30: note: in definition of macro 'AWS_LOGSTREAM_ERROR'
  143 |                 logStream << streamExpression; \
      |                              ^~~~~~~~~~~~~~~~
/aws-sdk-cpp/aws-cpp-sdk-core/source/utils/event/EventStreamDecoder.cpp: In static member function 'static void Aws::Utils::Event::EventStreamDecoder::onHeaderReceived(aws_event_stream_streaming_decoder*, aws_event_stream_message_prelude*, aws_event_stream_header_value_pair*, void*)':
/aws-sdk-cpp/aws-cpp-sdk-core/source/utils/event/EventStreamDecoder.cpp:134:77: error: 'this' pointer is null [-Werror=nonnull]
  134 |                         "ErrorMessage: " << handler->GetEventPayloadAsString());
      |                                             ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~^~
[...]

PR #1754, opened 2021-08-26
Fixed dereferencing a nullptr (#1750)

theShmoo:
#1750

Fixed dereferencing a nullptr in aws-cpp-sdk-core/source/utils/event/EventStreamDecoder.cpp


sdavtaker:
Merged, added a few lines of other places where we have similar bug.

theShmoo:
Thanks!

PR #1755, opened 2021-08-29
Add memory ordering to default executor for better performance on aarch64, also change a cmpexchng in a loop to weak

jweinst1:
On weakly ordered platforms, such as aarch64, it is desirable to have the memory ordering of atomic operations as relaxed as possible for the best performance. By default, with no explicit annotation, memory ordering uses either sequentially consistent ordering or acquire - release ordering, when in most cases, it only need either acquire or releases ordering, or just even relaxed ordering.

In `DefaultExecutor::SubmitToThread(std::function<void()>&&  fx)` , the `compare_exchange_strong()`  on success, is not cache coherent and does not require synchronization, just atomicity, so it's fine to relax it, but the `store()` should be release ordering.

In either `SubmitToThread` or `Detach`, we want the failed compare exchange to be in synchronization with the `store()` that happens before the return on a successful  compare exchange. This way, at the end of the `do .. while()` loop, the loaded state is properly checked and not our of order.

Additionally, the shutdown compare - exchange loop in `DefaultExecutor::~DefaultExecutor()` can be `compare_exchange_weak` instead because it is run inside the clause of a `while` loop. allowing it to tolerate spurious failures.

*Check all that applies:*

- [x] Linux
- [x] Windows
- [ ] Android
- [x] MacOS
- [ ] IOS
- [x] Other Platforms

### DIFF
--- a/aws-cpp-sdk-core-tests/utils/threading/DefaultExecutorTest.cpp
+++ b/aws-cpp-sdk-core-tests/utils/threading/DefaultExecutorTest.cpp
@@ -11,6 +11,20 @@
 
 using namespace Aws::Utils::Threading;
 
+TEST(DefaultExecutor, InOrderMemoryTest)
+{
+    std::atomic<unsigned> count(10);
+    {
+        DefaultExecutor exec;
+        auto first = [&] {
+            count.fetch_add(2, std::memory_order_relaxed);
+            count.fetch_sub(3, std::memory_order_relaxed);
+        };
+        exec.Submit(first);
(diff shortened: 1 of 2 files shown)


jweinst1:
```
[----------] Global test environment tear-down
[==========] 398 tests from 70 test cases ran. (7268 ms total)
[  PASSED  ] 398 tests.
[100%] Built target aws-cpp-sdk-core-tests
```

On an AWS graviton ARM machine

PR #1810, opened 2021-11-11
Fix build break in cmake 3.22.0-rc2

Anstow:
The build fails in cmake 3.22.0-rc2 when LIB_SEARCH_PATH is empty. Adding quotes ensures there is always a string to compare to. 

Questions about process:
* This is not a code change just a build fix so I'm not sure tests are relevant (please let me know if that's isn't the case)
* I've made sure it fixes the issue I was seeing on windows I don't think this will have any effect on building the library so I haven't built the SDK on any platform (I've included it in another project in windows though which is probably sufficient windows testing)
* I'm not sure what doing a review by myself involves for this PR

- [ ] Linux
- [ ] Windows
- [ ] Android
- [ ] MacOS
- [ ] IOS
- [ ] Other Platforms

### DIFF
--- a/cmake/AWSSDKConfig.cmake
+++ b/cmake/AWSSDKConfig.cmake
@@ -129,7 +129,7 @@ endif()
 get_filename_component(TEMP_PATH "${AWSSDK_CORE_LIB_FILE}" PATH)
 get_filename_component(TEMP_NAME "${TEMP_PATH}" NAME)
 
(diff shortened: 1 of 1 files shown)


elfprince13:
+1

jeroen:
+1, cmake 3.22 is now released and this has broken the build for many distros.

ahmedyarub (approved):
LGTM tested on multiple OSes

KaibaLopez:
This Pr has been merged, thanks for your contribution and sorry for the delay.

PR #1839, opened 2022-01-09
Fix pkgconfig files on windows

nirbheek:
```
build: Fix libdir in pkgconfig files for Windows

Windows import libraries are installed inside ARCHIVE_DIRECTORY, not
LIBRARY_DIRECTORY. So we need to use that for libdir on Windows.
```

```
build: Set -DUSE_IMPORT_EXPORT in pkgconfig files on Windows

MSVC consumers of the shared library need to define this macro when
linking to aws-cpp-sdk-* libs, or they will get undefined reference
errors. Also doesn't hurt to define this with MinGW, so add it
unconditionally on Windows.
```

This is necessary for fixing https://github.com/amzn/amazon-s3-gst-plugin/issues/13

*Check all that applies:*
    - Build system fix, doesn't require a test

- [x] Linux
- [x] Windows
- [ ] Android
- [ ] MacOS
- [ ] IOS
- [ ] Other Platforms


nacho:
LGTM

nirbheek:
Could someone please review this? It is necessary for an *Amazon* product to work correctly :)

nirbheek:
Poke?

jmklix:
#1888

jawilson:
Poke?

PR #1845, opened 2022-01-14
Convert 2 log messages WARN->INFO

VirtualTim:
As far as I can tell `AWS_ROLE_ARN` and `AWS_WEB_IDENTITY_TOKEN_FILE` are optional, so I don't think it's appropriate to log those as WARN, as that implies something has gone wrong.

My motivation at creating this PR is that my logs are getting filled with warning messages like:
`WARNING: Token file must be specified to use STS AssumeRole web identity creds provider.`

*Check all that applies:*

- [ ] Linux
- [ ] Windows
- [ ] Android
- [ ] MacOS
- [ ] IOS
- [ ] Other Platforms

### DIFF
--- a/aws-cpp-sdk-core/source/auth/STSCredentialsProvider.cpp
+++ b/aws-cpp-sdk-core/source/auth/STSCredentialsProvider.cpp
@@ -59,7 +59,7 @@ STSAssumeRoleWebIdentityCredentialsProvider::STSAssumeRoleWebIdentityCredentials
 
     if (m_tokenFile.empty())
     {
-        AWS_LOGSTREAM_WARN(STS_ASSUME_ROLE_WEB_IDENTITY_LOG_TAG, "Token file must be specified to use STS AssumeRole web identity creds provider.");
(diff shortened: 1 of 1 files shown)


nanish94 (approved):
comment

serzhick:
Hello,
Sorry to say this, but these log lines must be present in order to notify about the error use-case because both token file and role ARN must be present for STSAssumeRoleWebIdentityCredentialsProvider to work.
I would suggest checking why you see these warnings in your logs, i.e. why you need to use STSAssumeRoleWebIdentityCredentialsProvider when it fails to properly initialize.
Best regards,
Sergey

VirtualTim:
Thanks for the feedback @serzhick. I don't intentionally/directly use `STSAssumeRoleWebIdentityCredentialsProvider`, however I do use `DefaultAWSCredentialsProviderChain`, which creates a `STSAssumeRoleWebIdentityCredentialsProvider`. How would you suggest I get around this? Should I be creating my own class derived from `AWSCredentialsProviderChain`?

PR #1856, opened 2022-01-29
Fix crashes in S3 CRT GetObjectAsync / PutObjectAsync

### LINKED ISSUE #1655: S3CrtClient::GetObjectAsync Doesn't copy handler
Confirm by changing [ ] to [x] below to ensure that it's a bug:
- [x] I've gone though [Developer Guide](https://docs.aws.amazon.com/sdk-for-cpp/v1/developer-guide/welcome.html) and [API reference](http://sdk.amazonaws.com/cpp/api/LATEST/index.html)
- [x] I've searched for [previous similar issues](https://github.com/aws/aws-sdk-cpp/issues) and didn't find any solution
  
**Describe the bug**
The API for the new `S3CrtClient::GetObjectAsync` function looks like the old API `S3Client::GetObjectAsync`, but unlike the old API the new one does not copy the handler. This should either be changed or documented because code cannot safely switch from old to new API without potentially hidden bugs.

Example code that is buggy:
```
void MyFunction() {
  std::string local_thing;
  auto callback = [local_thing](...) {
     ... use local_thing
  };
  s3_crt_client.GetObjectAsync(..., callback, ...);
```

The captured variable `local_think` could be accessed after its destructor is called because GetObjectAsync takes a pointer rather than a copy of the handler:

` userData->userCallback = static_cast<const void*>(&handler);`

What's the point of taking `std::function` as input if the caller still has to manage the lifetime of the handler?

Thanks for reading, let me know if it would be helpful for me to provide additional information.

**SDK version number**
1.9.22

**Platform/OS/Hardware/Device**
What are you running the sdk on?
Linux

**To Reproduce (observed behavior)**
Steps to reproduce the behavior (please share code)

Something like the example code.

**Expected behavior**
Either the difference between new and old API should be documented, or the new API should manage ownership of the handler, or the new API should accept a function pointer instead of `std::function`

**Additional context**
The same problem exists with PutObjectAsync, but not with other methods that still like the thread executor (like `CompleteMultipartUploadAsync`)
Thanks for reading!

grrtrr:
_This fixes undefined behaviour in `Get/PutObjecAsync` which results in crashes and program termination._

*Issue #, if available:*
Resolves #1655.

*Description of changes:*

The `Get/PutObjectAsync` methods use const-references to `std::function` as callbacks:
```c++
typedef std::function<void(const S3CrtClient*, const Model::GetObjectRequest&, Model::GetObjectOutcome, const std::shared_ptr<const Aws::Client::AsyncCallerContext>&) > GetObjectResponseReceivedHandler;
typedef std::function<void(const S3CrtClient*, const Model::PutObjectRequest&, const Model::PutObjectOutcome&, const std::shared_ptr<const Aws::Client::AsyncCallerContext>&) > PutObjectResponseReceivedHandler;

void S3CrtClient::GetObjectAsync(const GetObjectRequest& request, const GetObjectResponseReceivedHandler& handler, const std::shared_ptr<const Aws::Client::AsyncCallerContext>& context) const;
void S3CrtClient::PutObjectAsync(const PutObjectRequest& request, const PutObjectResponseReceivedHandler& handler, const std::shared_ptr<const Aws::Client::AsyncCallerContext>& context) const;
```
The _address_ of these const references is taken and later dereferenced (the `PutObjectAsync` path is similar):
```c++
// S3Client.h:
        struct CrtRequestCallbackUserData {
          const S3CrtClient *s3CrtClient;
          const void *userCallback;  // <== HERE
          // ...
   };

// S3CrtClient.cpp:
  CrtRequestCallbackUserData *userData = Aws::New<CrtRequestCallbackUserData>(ALLOCATION_TAG);
  // ...
  userData->userCallback = static_cast<const void*>(&handler);
```
By the time the `userCallback` is executed within `GetObjectRequestShutdownCallback`, the `handler` object is out of scope, since the `GetObjectAsync` function returns quickly. The _address_ now points to a destructed `std::function` object.

### Fix
Rather than taking the address of a (temporary) object and risk its destruction, copy-construct the handlers within `CrtRequestCallbackUserData` to ensure that the required `std::function` objects are still alive when the shutdown callbacks get called.

*Check all that applies:*
  Tested behaviour manually, using the code in #1655 and a similar local application. Please see below for more details.

- [X] Linux
- [ ] Windows
- [ ] Android
- [ ] MacOS
- [ ] IOS
- [ ] Other Platforms

### DIFF
--- a/aws-cpp-sdk-s3-crt/include/aws/s3-crt/S3CrtClient.h
+++ b/aws-cpp-sdk-s3-crt/include/aws/s3-crt/S3CrtClient.h
@@ -11322,8 +11322,9 @@ namespace Aws
 
         struct CrtRequestCallbackUserData {
           const S3CrtClient *s3CrtClient;
-          const void *userCallback;
-          std::shared_ptr<const Aws::Client::AsyncCallerContext> userCallbackContext;
+          GetObjectResponseReceivedHandler getResponseHandler;
+          PutObjectResponseReceivedHandler putResponseHandler;
+          std::shared_ptr<const Aws::Client::AsyncCallerContext> asyncCallerContext;
           const Aws::AmazonWebServiceRequest *originalRequest;
           std::shared_ptr<Aws::Http::HttpRequest> request;
           std::shared_ptr<Aws::Http::HttpResponse> response;
--- a/aws-cpp-sdk-s3-crt/source/S3CrtClient.cpp
+++ b/aws-cpp-sdk-s3-crt/source/S3CrtClient.cpp
@@ -458,8 +458,7 @@ static void GetObjectRequestShutdownCallback(void *user_data)
   auto *userData = static_cast<S3CrtClient::CrtRequestCallbackUserData*>(user_data);
   // call user callback and release user_data
   S3Crt::Model::GetObjectOutcome outcome(userData->s3CrtClient->GenerateStreamOutcome(userData->response));
-  auto handler = static_cast<const GetObjectResponseReceivedHandler*>(userData->userCallback);
-  (*handler)(userData->s3CrtClient, *(reinterpret_cast<const GetObjectRequest*>(userData->originalRequest)), std::move(outcome), userData->userCallbackContext);
+  userData->getResponseHandler(userData->s3CrtClient, *(reinterpret_cast<const GetObjectRequest*>(userData->originalRequest)), std::move(outcome), userData->asyncCallerContext);
 
   Aws::Delete(userData);
 }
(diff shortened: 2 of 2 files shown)


wmorgan6796:
What is the status of this PR? I've been facing this issue as well and feel that this fixes a critical bug in the S3Crt lib that is hard to debug.

jmklix:
Merged with code-gen changes [here](https://github.com/aws/aws-sdk-cpp/pull/2191)

Issue #1768, opened 2021-09-09
S3 crt client doesn't pick up the default region


dhananjays:
Confirm by changing [ ] to [x] below to ensure that it's a bug:
- [ x] I've gone though [Developer Guide](https://docs.aws.amazon.com/sdk-for-cpp/v1/developer-guide/welcome.html) and [API reference](http://sdk.amazonaws.com/cpp/api/LATEST/index.html)
- [ x] I've searched for [previous similar issues](https://github.com/aws/aws-sdk-cpp/issues) and didn't find any solution
  
**Describe the bug**
S3-crt client doesn't automatically get the region unlike the S3 client which automatically fetches region (using the chain of AWS_DEFAULT_REGION -> profile -> instance metadata). The user is forced to provide the correct region when creating the s3-crt client

**SDK version number**
1.9

**Platform/OS/Hardware/Device**
EC2 Linux instance

**To Reproduce (observed behavior)**
Steps to reproduce the behavior (please share code)
Create S3 client without setting region in the client config. It automatically picks up the default region from the profile/instance metadata (This is as expected from version 1.8 as mentioned here - https://aws.amazon.com/blogs/developer/aws-sdk-for-c-version-1-8-developer-preview/)

Create S3-crt client without setting region in the client config. It does not pick the default region as expected.

**Expected behavior**
S3-crt client should automatically pick up the default region as done by the S3 client

KaibaLopez:
Hi @dhananjays ,
So, it seems like you've figured out the workaround , but for the record this behavior is not expected of the s3-crt default constructor. 

The s3-crt default provider chain is reduced to make it faster to go through, and it will only check for the region in the environment variable, the profile in the config file or the one hard coded.
Having said that, this could be a feature request, but I can't make any promises on when we could work on it.

grrtrr:
Here is how we back-ported that in our code, since we ran into the same issue: the CRT does not pickup `AWS_{,DEFAULT_}REGION`:
```c++
        Aws::S3Crt::ClientConfiguration config;
        // ...
        config.region = []() -> Aws::String {
            // Return the configured AWS region, using us-east-1 as fallback. This is normally part
            // of Aws::Client::ClientConfiguration, but is not implemented for the S3CrtClient.
            Aws::String region = os::getenv("AWS_DEFAULT_REGION");
            if (!region.empty()) {
                return region;
            }

            region = os::getenv("AWS_REGION");
            if (!region.empty()) {
                return region;
            }

[...]

Issue #1654, opened 2021-05-21
Segmentation fault in Aws::S3Crt::S3CrtClient::UploadPartCallable


Johnparke888:
Confirm by changing [ ] to [x] below to ensure that it's a bug:
- [ x] I've gone though [Developer Guide](https://docs.aws.amazon.com/sdk-for-cpp/v1/developer-guide/welcome.html) and [API reference](http://sdk.amazonaws.com/cpp/api/LATEST/index.html)
- [ x] I've searched for [previous similar issues](https://github.com/aws/aws-sdk-cpp/issues) and didn't find any solution
  
**Describe the bug**
Attempting to do a multipart upload using the new S3Crt client. I started with a working version of code using the S3 client. When the code executes Aws::S3Crt::S3CrtClient::UploadPartCallable, I get a segmentation fault.
Address sanitizer indicates an attempt to read from memory address which points to the zero page:
==11503==ERROR: AddressSanitizer: SEGV on unknown address 0x000000000000 (pc 0x0000005888f3 bp 0x60300023b7b0 sp 0x7fff12af8210 T0)
==11503==The signal is caused by a READ memory access.
==11503==Hint: address points to the zero page.
    #0 0x5888f3 in Aws::S3Crt::S3CrtClient::UploadPartCallable(Aws::S3Crt::Model::UploadPartRequest const&) const (.../Linux/test2+0x5888f3)
    #1 0x4ddf5b in main /.../aws-multipart-upload-new/main.cpp:176
    #2 0x7feaa9037544 in __libc_start_main (/lib64/libc.so.6+0x22544)
    #3 0x410241  (.../Linux/test2+0x410241)

I would be very grateful for any help.

John
**SDK version number**
1.9.23
**Platform/OS/Hardware/Device**
What are you running the sdk on?
Red Hat Enterprise Linux Server release 7.5 (Maipo)
gcc version 11.1.0 (GCC)

**To Reproduce (observed behavior)**
Steps to reproduce the behavior (please share code)

int main ()
{

   int i = 0;
   int j = 0;
   int position1 = 0;
   int position2 = 0;

   int partNumber = 0;
   int outputBufferLength = 1024 * 1024 * 20;
   unsigned long int contentLength = 0;

   const unsigned long int fixedPartSize = 1024 * 1024 * 8;
   int bytesRead = 0;
   const double throughput_target_gbps = 5;

   Aws::Utils::Logging::InitializeAWSLogging(
[...]

PR #1884, opened 2022-03-19
S3 Crt ClientConfiguration: expose Aws::Client::ClientConfiguration

### LINKED ISSUE #1882: S3CrtClient: regression due to missing AwsClient fields in  Aws::S3Crt::ClientConfiguration
### Describe the bug

The `S3CrtClient` still uses the parent class `MakeRequest` for the majority of its operations.

The S3Client allows to configure the behaviour of the `Aws::Client::AWSXMLClient` via `ClientConfiguration` options.

These are not exposed via the `S3CrtClient` constructor, hence defaults are used.

This means that people wishing to upgrade to the new `S3CrtClient` suffer a regression: for the majority of requests, the S3CrtClient will use default behaviour, with the user unable to influence this.

Please see list of minimally required configuration parameters. The defaults are not sufficient for production use.

### Expected behavior

Minimally the following `Aws::Client::ClientConfiguration` parameters should be included in the `Aws::S3Crt::ClientConfiguration`:
- `maxConnections`,
- `httpRequestTimeoutMs` (Linux),
- `requestTimeoutMs`,
- `connectTimeoutMs`,
- `enableTcpKeepAlive`,
- `tcpKeepAliveIntervalMs`,
- `lowSpeedLimit`,
- `retryStrategy`
- `caPath` / `caFile`.

### Current behavior

The above parameters are not configurable, and so the S3CrtClient falls back to defaults via `operator Aws::Client::ClientConfiguration()`.

### Steps to Reproduce

Reproducible with any S3CrtClient as of today's master.

### Possible Solution

Above.

### AWS CPP SDK version used

1.9

### Compiler and Version used

gcc / clang

Linux, ubuntu 18.04

grrtrr:
Since the S3CrtClient is a sub-class of the AwsClient, expose the
configuration needed by the parent class.

Many API calls use the parent class, so preserving user configurations is important. 

This now uses slicing instead of a conversion operator, so that the
parent struct only sees the value relevant to it.

Since it initializes the base struct first, this also addresses #1768.

*Issue #, if available:*
Fixes #1882, #1768 and #1654.

*Description of changes:*

*Check all that applies:*
       Tested with large application in our repo

- [X] Linux
- [ ] Windows
- [ ] Android
- [ ] MacOS
- [ ] IOS
- [ ] Other Platforms

### DIFF
--- a/aws-cpp-sdk-s3-crt/include/aws/s3-crt/ClientConfiguration.h
+++ b/aws-cpp-sdk-s3-crt/include/aws/s3-crt/ClientConfiguration.h
@@ -19,16 +19,9 @@ namespace Aws
     {
         struct AWS_S3CRT_API ClientConfiguration : Aws::Client::ClientConfiguration
         {
-            ClientConfiguration() :
+            ClientConfiguration() : Aws::Client::ClientConfiguration(),
                 partSize(5 * 1024 * 1024),
-                throughputTargetGbps(2.0)
-            {
-                /* Parent configuration settings: */
-                region = Aws::Region::US_EAST_1;
-                useDualStack = false;
-                enableHostPrefixInjection = true;
-                scheme = Aws::Http::Scheme::HTTPS;
-            }
+                throughputTargetGbps(2.0) {}
 
             /** Client bootstrap used for common staples such as event loop group, host resolver, etc..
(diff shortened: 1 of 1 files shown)


grrtrr:
Replaced by https://github.com/aws/aws-sdk-cpp/pull/2098

PR #1891, opened 2022-03-29
Created LEGACY_BUILD behaviour

sdavtaker:
*Issue #, if available:* 1888

*Description of changes:*
All cmake modules in cmake were moved to cmake_legacy.
Most of the main CMakeLists.txt was moved to cmake_legacy/legacy_main.cmake
Option for disabling LEGACY_BUILD was created. But, no implementation yet.

*Check all that applies:*
Tested manually by building with and without the LEGACY_BUILD=OFF flag
No an API breaking changing, but forces upgrade of CMAKE for those having older than 3.12 in their toolchain. We were warning in every build about it since a year ago, and we are going to hold this from reaching main branch until version 1.10 is released.
- [x] Checked if this PR will _not_ introduce cross-platform inconsistent behaviour.

- [x] Linux
- [ ] Windows
- [ ] Android
- [x] MacOS
- [ ] IOS
- [ ] Other Platforms


CMakeLists.txt
```diff
@@ -2,284 +2,35 @@
...
+            DESCRIPTION ${DESCRIPTION_STRING}
+            HOMEPAGE_URL "https://docs.aws.amazon.com/sdk-for-cpp"
+            )
+    set(CMAKE_CXX_STANDARD 11)
+    set(CMAKE_CXX_EXTENSIONS OFF)
+    set(CMAKE_CXX_STANDARD_REQUIRED ON)
+
+    message(STATUS "Building ${PROJECT_NAME} ${PROJECT_VERSION}")
+
+    message(WARNING "This builds nothing yet.")
```
sbiscigl, line 35: should this be `FATAL_ERROR` ([docs](https://cmake.org/cmake/help/latest/command/message.html)) since it will do nothing yet?

CMakeLists.txt
```diff
@@ -2,284 +2,35 @@
...
+    message(STATUS "Building with 1.10 CMake scripts.")
+    string(CONCAT DESCRIPTION_STRING "The AWS SDK for C++ provides a modern C++ (standard version C++11 or later) "
+            "interface for Amazon Web Services (AWS).")
+    project("aws-cpp-sdk"
+            LANGUAGES CXX
+            VERSION 1.10.0
+            DESCRIPTION ${DESCRIPTION_STRING}
+            HOMEPAGE_URL "https://docs.aws.amazon.com/sdk-for-cpp"
+            )
+    set(CMAKE_CXX_STANDARD 11)
```
sbiscigl, line 29: since we're requiring a higher cmake version should we do something crazy and update to 14 or even higher?

sbiscigl (approved):

CMakeLists.txt
sdavtaker: Not in the plans for now.

CMakeLists.txt
sdavtaker: Could be, its just a place holder for next PR code. Going to be removed soon.

serzhick (approved):

PR #1893, opened 2022-03-30
Moving generated libraries to generated directory

sdavtaker:
*Issue #, if available:* 1888

*Description of changes:*
- Moved generated code to generated directory
- modified legacy codebase to generate, build, and install using the new location for generated code

*Check all that applies:*
Manually tested by running configuration, build, unit test, and install from clean environment with and without BUILD_ONLY parameter.
No an API breaking changing, but forces upgrade of CMAKE for those having older than 3.12 in their toolchain. We were warning in every build about it since a year ago, and we are going to hold this from reaching main branch until version 1.10 is released.

- [x] Linux
- [ ] Windows
- [ ] Android
- [x] MacOS
- [ ] IOS
- [ ] Other Platforms

### DIFF
--- a/cmake_legacy/dependencies.cmake
+++ b/cmake_legacy/dependencies.cmake
@@ -50,7 +50,7 @@ endfunction()
 #
 # Uses selection sort (stable).
 function(sort_links targets_list)
-    # Special case of empty input list. Futher code assumes list to be non-empty.
+    # Special case of empty input list. Further code assumes list to be non-empty.
     if(NOT ${targets_list})
         return()
     endif()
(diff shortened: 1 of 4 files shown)


serzhick:
Why don't you split the PR into 2 commits
* update config and generator
* manual re-generation run?

sdavtaker:
> Why don't you split the PR into 2 commits
> 
> * update config and generator
> * manual re-generation run?

The code in legacy cmake was used for finding where to generate, where to remove and  where to include all intertwined, by separating in 2 first commit would effectively be broken without extra work.

sdavtaker:
Isolated real change from generation for easier review. This commit does not run by itself, needs a regeneration to be buildable and runable.

serzhick (approved):

sbiscigl (approved):

Issue #1888, opened 2022-03-23
Improvement to build, install, and distribution mechanisms


sdavtaker:
### Describe the issue

Feedback from our customers show that AWS SDK for C++ complexity to build is high. Our customers come from very diverse disciplines and backgrounds, that requires AWS SDK for C++ to build under a vast set of constraints. These constraints usually come from using different operating systems, toolchains, memory management processes, and domain specific practices to mention a few. 

Since AWS SDK for C++ was first released, we extended our cmake scripts to fulfill needs from new platforms, new uses cases we identified, and new tooling mostly as a collection of small individual changes with no significant impact on the build as a whole. However, as the collection of them grew, the build as a whole became a very complex piece of code to work with. This building complexity many times translates in time spent for developers who want to install the AWS SDK for C++ to work in their projects, for contributors and reviewers wanting to implement changes to fix issues with it, for package managers to release distributions, for release managers to distribute it as a dependency of their projects. 

We are working in a redesign of our cmake scripts to be released at v1.10. We expect to provide a simpler and idiomatic modern cmake experience based in standard industry patterns. 
Here is a list of some of the changes we will introduce:

- [ ] Implement [superbuild](https://www.kitware.com/cmake-superbuilds-git-submodules/) pattern.
- [ ] Remove the need to pull dependencies via git submodules.
- [x] Restructure project directories to easily distinct between generated-src and src.
- [ ] Provide CPack targets.
- [ ] Declaring dependencies closer to their targets so they are not required when not used. For example, no test tools will be required when tests are not expected to be run.
- [ ] Make configuration parameters names reflect better cmake community practices.
- [ ] A legacy mode that will mantain backward compatibility with current build scripts for allowing an easy transition to customers.
- [ ] Simplify toolchain customizations.
- [x] Set cmake_minimum_required to 3.12 version.

We will develop these changes in branch build-dev-1.10 until ready for release. We are going to publish PRs in each step of the way and link them to this Issue. We encourage everyone interested to participate in this Issue discussion thread, and the PRs. 

### AWS CPP SDK version used

1.9

### compiler and version used

any

### Operating System and version

any

sdavtaker:
First preparation PR: https://github.com/aws/aws-sdk-cpp/pull/1886

from81:
[...]

PR #1935, opened 2022-05-26
build dev 1.10 pr8

sdavtaker:
- Preparing toolchain and dependencies

*Issue #1888 , if available:*

*Description of changes:*
Added logic to detect dependencies and setup compiler basic options. This is an early draft, still needs more dependencies to deal with other platforms. 

*Check all that applies:*

- [x] Linux
- [ ] Windows
- [ ] Android
- [x] MacOS
- [ ] IOS
- [ ] Other Platforms

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -40,9 +40,14 @@ else ()
     set(CMAKE_CXX_EXTENSIONS OFF)
     set(CMAKE_CXX_STANDARD_REQUIRED ON)
 
-    message(STATUS "Building ${PROJECT_NAME} ${CMAKE_PROJECT_VERSION}")
+    # Setting build to hide symbols in targets by default
+    set(CMAKE_CXX_VISIBILITY_PRESET hidden)
+    set(CMAKE_VISIBILITY_INLINES_HIDDEN YES)
 
-    # Setting version header for library
+    # Preventing writes to package registry by default
+    set(CMAKE_EXPORT_NO_PACKAGE_REGISTRY YES)
+
+    message(STATUS "Generating source file for introspection of version to ${CMAKE_PROJECT_VERSION} ")
     configure_file(
             VersionConfig.h.in
             src/aws-cpp-sdk-core/include/aws/core/VersionConfig.h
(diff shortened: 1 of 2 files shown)


CMakeLists.txt
```diff
@@ -62,10 +67,31 @@ else ()
...
+        enable_testing()
+        # Testing dependency
+        find_package(GTest 1.11 REQUIRED )
+        include(GoogleTest) # for gtest_discover_tests()
+    endif ()
+
+    # Project dependencies http library (curl or OS), tinyxml2, cJSON, ssl (open or other), CRT
+    find_package(CURL REQUIRED)
+    find_package(ZLIB REQUIRED)
+    find_package(OpenSSL REQUIRED COMPONENTS Crypto)
```
SergeyRyabinin, line 86: What about aws-lc and s2n?
There are so much done in CRT to use aws-lc.

CMakeLists.txt
```diff
@@ -62,10 +67,31 @@ else ()
             message(FATAL_ERROR "Unknown build type: ${CMAKE_BUILD_TYPE}")
         endif ()
     endif ()
+
+    # Options definition
     option(BUILD_TESTS "If enabled, the SDK will include tests in the build" OFF)
 
+    message(STATUS "Setting toolchain and dependencies")
+    if (BUILD_TESTS)
+        message(STATUS "Building tests")
```
graebm, line 76: unrelated: isn't [BUILD_TESTING](https://cmake.org/cmake/help/latest/command/enable_testing.html) the official cmake option?

sdavtaker: Good catch, updating.

CMakeLists.txt
```diff
@@ -40,9 +40,14 @@ else ()
     set(CMAKE_CXX_EXTENSIONS OFF)
     set(CMAKE_CXX_STANDARD_REQUIRED ON)
 
-    message(STATUS "Building ${PROJECT_NAME} ${CMAKE_PROJECT_VERSION}")
+    # Setting build to hide symbols in targets by default
+    set(CMAKE_CXX_VISIBILITY_PRESET hidden)
```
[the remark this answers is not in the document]
sdavtaker: It seems like this only affected C++ until version 3.12 of CMAKE, not applies also to C. 
https://cmake.org/cmake/help/latest/module/GenerateExportHeader.html?highlight=cmake_cxx_visibility_preset
I will leave it like this for now, we shouldn't expose anything C at this SDK right now, but worth to dig into this further for future use cases.

PR #1938, opened 2022-05-28
Add an option to fix non-compliant RFC3986 encoding.

### LINKED ISSUE #1196: Disabled URL Encoding of Reserved Characters Caused Signature Mismatch on Some S3-Compatible Storage Servers

jeking3:
*Issue #, if available:*

This fixes #1196
This closes #1202

*Description of changes:*

Based on legacy design (see description in #1202 for one case) the SDK core URI encoder does not treat the characters `$&,:=@` as encodable.  This causes problems when using the SDK code against third party implementations of some of the AWS APIs, most notably the S3 API.

The Python SDK does not contain this legacy encoding strategy and is compatible with third party implementations.  Additionally, the Python SDK is compatible with AWS services as well, so this change is compatible all around.  To solve this in a manner which did not affect backwards compatibility, I added an opt-in HttpOption called `compliantRfc3986Encoding`.  If this option is set, or if `Aws::Http::SetCompliantRfc3986Encoding(true);` is called, the legacy behavior will be defeated.

I first added the option and refactored some duplicated code paths in URI.cpp, and ensured all the tests were still passing and no backwards compatibility issues were introduced.  Then I added test code explicitly for when this option is enabled.

I also augmented the S3 integration test to optionally set this flag (by default it does not, maintaining baseline behavior) and ran it without errors when the flag is set and without the flag set.

*Note:*

If accepted, I will need and provide a backport to the 1.8.x branch.

*Check all that applies:*

- [ ] Linux
- [ ] Windows
- [ ] Android
- [X] MacOS
- [ ] IOS
- [ ] Other Platforms

### DIFF
--- a/aws-cpp-sdk-core-tests/http/URITest.cpp
+++ b/aws-cpp-sdk-core-tests/http/URITest.cpp
@@ -281,7 +281,20 @@ TEST(URITest, TestParseWithColon)
     EXPECT_EQ(80, complexUri.GetPort());
     EXPECT_STREQ("/awsnativesdkputobjectstestbucket20150702T200059Z/TestObject:1234/awsnativesdkputobjectstestbucket20150702T200059Z/TestObject:Key", complexUri.GetPath().c_str());
     EXPECT_STREQ(strComplexUri, complexUri.GetURIString().c_str());
+}
 
+TEST(URITest, TestParseWithColonCompliant)
+{
+    Aws::Http::SetCompliantRfc3986Encoding(true);
+    const char* strComplexUri = "http://s3.us-east-1.amazonaws.com/awsnativesdkputobjectstestbucket20150702T200059Z/TestObject:1234/awsnativesdkputobjectstestbucket20150702T200059Z/TestObject:Key";
+    URI complexUri(strComplexUri);
+    const char* compliantStrComplexUri = "http://s3.us-east-1.amazonaws.com/awsnativesdkputobjectstestbucket20150702T200059Z/TestObject%3A1234/awsnativesdkputobjectstestbucket20150702T200059Z/TestObject%3AKey";
+    EXPECT_EQ(Scheme::HTTP, complexUri.GetScheme());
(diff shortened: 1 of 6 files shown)


jmklix:
Already merged

PR #1977, opened 2022-06-28
Fix EventStream code to work with latest version of aws-c-event-stream

graebm:
*Issue:*
The latest aws-c-event-stream changed a struct's members, which breaks code here that was accessing these members.

*Description of changes:*
Make code compatible with older and newer versions of aws-c-event-stream. Also revamp logic to be a bit safer, specifically:
- Don't access struct members directly. Instead, use public functions.
- Don't touch a struct that failed initialization. Report failed initialization rather than trying to detect it later by inspecting the struct's members.

*Check all that applies:*
- [ X ] Did a review by yourself.
- [ NOPE ] Added proper tests to cover this PR. (If tests are not applicable, explain.)
  - Not sure how to make the failure branches occur
- [ X ] Checked if this PR is a breaking (APIs have been changed) change.
- [ X ] Checked if this PR will _not_ introduce cross-platform inconsistent behavior.
- [ X ] Checked if this PR would require a ReadMe/Wiki update.

*Check which platforms you have built SDK on to verify the correctness of this PR:*
I've only compiled, not run any tests. I compiled against an old version of aws-c-event-stream and a new version of aws-c-event-stream to confirm it works with both.

### DIFF
--- a/aws-cpp-sdk-core/source/utils/event/EventStreamEncoder.cpp
+++ b/aws-cpp-sdk-core/source/utils/event/EventStreamEncoder.cpp
@@ -83,10 +83,10 @@ namespace Aws
                 Aws::Vector<unsigned char> outputBits;
 
                 aws_event_stream_message encoded;
-                if (Encode(&encoded, msg))
+                if (InitEncodedStruct(msg, &encoded))
                 {
                     aws_event_stream_message signedMessage;
-                    if (Sign(&signedMessage, &encoded))
+                    if (InitSignedStruct(&encoded, &signedMessage))
                     {
                         // success!
                         const auto signedMessageBuffer = aws_event_stream_message_buffer(&signedMessage);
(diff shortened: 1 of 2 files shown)


aws-cpp-sdk-core/source/utils/event/EventStreamEncoder.cpp
```diff
@@ -80,19 +80,32 @@ namespace Aws
 
             Aws::Vector<unsigned char> EventStreamEncoder::EncodeAndSign(const Aws::Utils::Event::Message& msg)
             {
-                aws_event_stream_message encoded = Encode(msg);
-                aws_event_stream_message signedMessage = Sign(&encoded);
+                Aws::Vector<unsigned char> outputBits;
 
-                const auto signedMessageLength = signedMessage.message_buffer ? aws_event_stream_message_total_length(&signedMessage) : 0;
+                aws_event_stream_message encoded;
```
SergeyRyabinin, line 85: don't we need to initialize/cleanup the struct?

aws-cpp-sdk-core/source/utils/event/EventStreamEncoder.cpp
```diff
@@ -103,57 +116,56 @@ namespace Aws
...
-                    // return {};
-                    aws_event_stream_message empty{nullptr, nullptr, 0};
-                    return empty;
-                }
+                    aws_array_list headers;
+                    EncodeHeaders(signedMessage, &headers);
 
-                aws_array_list headers;
-                EncodeHeaders(signedMessage, &headers);
+                    aws_byte_buf payload;
```
SergeyRyabinin, line 147: worth initing to = {0} in case new members are added to the struct.

aws-cpp-sdk-core/source/utils/event/EventStreamEncoder.cpp
graebm: If the `Encode()` function succeeds, then it initializes the struct. I'll rename the function to be more clear

SergeyRyabinin: I've asked because we have an example in history when CRT was updated to actually access data in the dst buffer:
https://github.com/aws/aws-sdk-cpp/commit/0227ef625c2c53689c61cb9d09eeeea2a78611fb

SergeyRyabinin (approved):

aws-cpp-sdk-core/source/utils/event/EventStreamEncoder.cpp
graebm: I changed this to use a function for aws_byte_buf construction, instead of setting the fields manually

PR #2020, opened 2022-08-04
Delegate CRT credentials

sbiscigl:
*Description of changes:*

The S3CrtClient right now uses a different credential chain than the rest of the SDK. Although functionally the same, they are both called at the same time which could lead to issues. This delegates the CRT credential chain call to use the SDK credential chain.

*Check all that applies:*

- [x] Linux
- [x] Windows
- [ ] Android
- [x] MacOS
- [ ] IOS
- [ ] Other Platforms

### DIFF
--- a/aws-cpp-sdk-s3-crt/include/aws/s3-crt/S3CrtClient.h
+++ b/aws-cpp-sdk-s3-crt/include/aws/s3-crt/S3CrtClient.h
@@ -5178,7 +5178,7 @@ namespace Aws
         Aws::Client::StreamOutcome GenerateStreamOutcome(const std::shared_ptr<Http::HttpResponse>& response) const;
 
     private:
-        void init(const S3Crt::ClientConfiguration& clientConfiguration, const Aws::Auth::AWSCredentials* credentials = nullptr);
+        void init(const S3Crt::ClientConfiguration& clientConfiguration, const std::shared_ptr<Aws::Auth::AWSCredentialsProvider> credentialsProvider);
 
         struct CrtClientShutdownCallbackDataWrapper {
           void *data;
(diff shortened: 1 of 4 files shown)


aws-cpp-sdk-core/source/auth/AWSCrtCredentialsProviderAdapter.cpp
```diff
@@ -0,0 +1,33 @@
...
+    std::function<AWSCredentials()> credFunc =
+        std::bind([](const std::shared_ptr<AWSCredentialsProvider>& provider){
+            return provider->GetAWSCredentials();
+        }, this->credentialsProvider);
+
+    std::function<std::shared_ptr<Aws::Crt::Auth::Credentials>()> handler =
+        std::bind([](const std::function<AWSCredentials()>& credFunc) {
+            AWSCredentials credentials = credFunc();
+            return Aws::MakeShared<Aws::Crt::Auth::Credentials>(ALLOCATION_TAG,
+                Aws::Crt::ByteCursorFromCString(credentials.GetAWSAccessKeyId().c_str()),
```
SergeyRyabinin, line 26: here ByteCursor will contain a pointer to a local object std::string c_str buffer.

SergeyRyabinin (changes requested):

aws-cpp-sdk-core/source/auth/AWSCrtCredentialsProviderAdapter.cpp
SergeyRyabinin: but then in c-tor of Aws::Crt::Auth::Credentials, in the function aws_credentials_new,
byte buffers will be copied before local credentials object is destroyed.

SergeyRyabinin (approved):

aws-cpp-sdk-core/source/auth/AWSCrtCredentialsProviderAdapter.cpp
sbiscigl: as discussed offline, moving all of this too a anon function in the client class itself to address ownership issues, will push up once i have updated the templates

sdavtaker (approved):

SergeyRyabinin (approved):

PR #2028, opened 2022-08-10
Build dev 1.10 pr11

sdavtaker:
*Issue #1888 , if available:*

*Description of changes:*
Added support for detecting and linking external dependencies to the build. Core and Core-Test build, but they do not work yet as expected. Sharing PR now since the diff is already too big. LEGACY_BUILD got broken because of issues finding CRT. 

Next steps:
- Core Tests can be run and pass
- Legacy Build is back in a working state
- Build Generated Clients

*Check all that applies:*

- [x] Linux
- [ ] Windows
- [ ] Android
- [x] MacOS
- [ ] IOS
- [ ] Other Platforms


CMakeLists.txt
```diff
@@ -86,13 +96,25 @@ else ()
 
     # -- Dependencies --
     add_subdirectory(dependencies)
+    if (CMAKE_SOURCE_DIR STREQUAL CMAKE_CURRENT_SOURCE_DIR)
+        # Testing Dependencies
+        if (BUILD_TESTING)
+            message(STATUS "Building tests")
+            # Testing dependency
+            find_package(GTest 1.11 REQUIRED)
```
sbiscigl, line 104: why not follow HEAD? from [their page](https://github.com/google/googletest)

```
GoogleTest now follows the Abseil Live at Head philosophy. We recommend [updating to the latest commit in the main branch as often as possible.
```

sbiscigl (approved):

CMakeLists.txt
sdavtaker: Because we know what version works with our build at this particular version. If head is not backward compatible at some point tests of old versions of SDK will not work anymore.

PR #2052, opened 2022-08-23
Draft DynamoDB changes for build scripts

sdavtaker:
*Issue #, if available:* 1888

*Description of changes:*
Build and export header changes to use as reference
for the required code-generation changes in following
commits.

*Check all that applies:*

- [ ] Linux
- [ ] Windows
- [ ] Android
- [x] MacOS
- [ ] IOS
- [ ] Other Platforms


generated/aws-cpp-sdk-dynamodb/source/CMakeLists.txt
```diff
@@ -0,0 +1,12 @@
+file(GLOB AWS_DYNAMODB_MODEL_SOURCE
```
sbiscigl, line 1: the creators of cmake advise against [globbing](https://cmake.org/cmake/help/v3.15/command/file.html?highlight=glob#glob).

Since this is autogenerated is there a reason we should not do this? if the reasons is "readability" which is a valid concern when there is 100s of model files, could we mitigate that by introducing a `files.cmake` generated file with the intent of that file only listing files.

sbiscigl (approved):

generated/aws-cpp-sdk-dynamodb/source/CMakeLists.txt
sdavtaker: I was back and forward on this one, I decided not to use GLOB in any other place because of the reasons CMAKE recommends to avoid it, but at the end I was wondering if I wasn't just pushing the GLOB to the generator and not really fixing that. Anyway, it makes sense to push it to the generator, it can be extracted from the model list generating it and that is something we already obtaining from the model itself. I'm going to fix and push again.

sbiscigl (approved):

PR #2060, opened 2022-08-26
Generation support for 1.11 - build dev 1.10 pr16

sdavtaker:
* Issue #1888 *

*Description of changes:*
- Updating Generator Java code
- Adding message about REGENERATE_CLIENT removal
- Refactor generate_sdks.py
- Updating Generator templates for 1.10

This change doesn't include regeneration of all clients, only DDB, neither the inclussion of all of them at the generated directory level. Holding that for follow up PR since will create a lot of noise here.

*Check all that applies:*

- [ ] Linux
- [ ] Windows
- [ ] Android
- [ ] MacOS
- [ ] IOS
- [ ] Other Platforms


tools/code-generation/generator/src/main/java/com/amazonaws/util/awsclientgenerator/domainmodels/codegeneration/ServiceModel.java
```diff
@@ -44,4 +44,25 @@ public boolean hasStreamingRequestShapes() {
     public Collection<Error> getNonCoreServiceErrors() {
         return serviceErrors.stream().filter(e -> !e.isCoreError()).collect(Collectors.toSet());
     }
+
+    public boolean hasSigV4Auth() {
+        if(metadata.getSignatureVersion().equals("v4") || metadata.getSignatureVersion().equals("s3v4")) {
+            return true;
+        }
+        return operations.values().parallelStream().anyMatch(operation -> operation.getSignerName().equals("Aws::Auth::SIGV4_SIGNER"));
```
sbiscigl, line 52: parallel steams are generally advised against in the java world because they use the common fork-join pool, meaning they share an executor pool with the rest of the world. I would either use sequential streams which is often faster. or alternatively it can have its own executor pool.

tools/code-generation/generator/src/main/java/com/amazonaws/util/awsclientgenerator/generators/cpp/CppClientGenerator.java
```diff
@@ -261,10 +264,10 @@ protected SdkFileEntry generateEventStreamHandlerSourceFile(ServiceModel service
     protected SdkFileEntry generateErrorSourceFile(final ServiceModel serviceModel) throws Exception {
 
         Set<String> retryableErrors = getRetryableErrors();
-        for(Error error : serviceModel.getServiceErrors()) {
-           if(retryableErrors.contains(error.getName())) {
-               error.setRetryable(true);
-           }
+        for (Error error : serviceModel.getServiceErrors()) {
```
sbiscigl, line 267: could be simplified to
```
error.setRetryable(serviceModel.getServiceErrors()
  .findAny(error -> retryableErrors.contains(error.getName())))
```

tools/code-generation/generator/src/main/java/com/amazonaws/util/awsclientgenerator/generators/cpp/CppClientGenerator.java
```diff
@@ -447,6 +447,46 @@ private SdkFileEntry generateCmakeFile(final ServiceModel serviceModel) throws E
...
+    }
+
+    private SdkFileEntry generateCmakeFileSourceModel(final ServiceModel serviceModel, final List<SdkFileEntry> fileList) throws Exception {
+
+        Template template = velocityEngine.getTemplate("/com/amazonaws/util/awsclientgenerator/velocity/cpp/CMakeFileSourceModel.vm", StandardCharsets.UTF_8.name());
+
+        VelocityContext context = createContext(serviceModel);
+
+        List<String> files = new ArrayList<>();
+        for (int i = 0; i < fileList.size(); i++) {
```
sbiscigl, line 475: could be simplified using streams to
```
fileList.stream()
  .filter(Objects::nonNull)
  .map(File:: getPathRelativeToRoot)
  .filter(filePath -> filePath.endsWith("cpp") && filePath.contains("/model/")
  .map(filePath -> Paths.get(filePath).getFileName().toString())
  .collect(Collectors.toList())
```

tools/code-generation/generator/src/main/java/com/amazonaws/util/awsclientgenerator/generators/cpp/s3/S3RestXmlCppClientGenerator.java
```diff
@@ -123,6 +125,11 @@ public SdkFileEntry[] generateSourceFiles(ServiceModel serviceModel) throws Exce
             replicationStatus.getEnumValues().set(indexOfComplete, "COMPLETED");
         }
 
+        // Some S3 operations have embedded errors, and we need to search for errors in the response.
```
sbiscigl, line 128: i recognize this code 🙂

sbiscigl (approved):

tools/scripts/generate_sdks.py
```diff
@@ -1,152 +1,190 @@
-﻿# Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.
```
SergeyRyabinin, line 1: this is a wrong script to modify.
You could just delete it.

sdavtaker: I'm going to do a clean up of all no longer scripts in this directory and merge this back to the right place. 
Moving to draft in the mean time.

Issue #1843, opened 2022-01-13
Expire token error code?


lsong0905:
### Describe the issue

Hello,

When I issue a S3 request(PUT, GET, etc) with an expire credentials (generated by STS service), the S3 SDK seems returning  UNKNOWN = 100 error code from S3Errors.h. The error message "ExpiredToken:Unable to parse ExceptionName: ExpiredToken Message: The provided token has expired." , on the other hand, seems containing some hint that indicates why the request failed. 

Other than parsing the error message, is there any better way to verify the request failed due to "ExpiredToken"? I expected to receive an error code that maps to "ExpiredToken" in S3Errors.h, but it is returning UNKNOWN = 100 error code. I am not certain if this is intentional. 

Thank you in advance!

### AWS CPP SDK version used

1.7.29

### compiler and version used

na

### Operating System and version

na

KaibaLopez:
Hi @lsong0905 ,
Short answer is no.
Error handling is always somewhat problematic for the SDK as it is heavily limited to what the service APIs models provide. We have to keep parity with them and so there is a limit to what we can change there, and the api sends error 100 (which can mean anything really) and the message with the description. Admittedly, it could be "catched" by reading the error but making those kinds of checks is a rabbit hole we are just not looking to get into.

Let me know if that answers your question.

PR #2062, opened 2022-08-29
Set BAD_REQUEST error code for 400 error instead of UNKNOWN

pcmoritz:
*Issue #, if available:*

This fixes https://github.com/aws/aws-sdk-cpp/issues/1843

More crucially, it removes the current inconsistency between boto3 and the AWS CLI on the one hand and the C++ SDK on the other hand when it comes to 400 return codes.

*Description of changes:*

Convert a 400 http status code into a `BAD_REQUEST` error instead of `UNKNOWN`.

This PR might need to be re-done from the Amazon side since I'm not familiar with the code generation tools used. I just fixed the occurrences manually.

*Check all that applies:*

- [ ] Linux
- [ ] Windows
- [ ] Android
- [ ] MacOS
- [ ] IOS
- [ ] Other Platforms


aws-cpp-sdk-core/source/client/AWSClient.cpp
```diff
@@ -73,6 +73,8 @@ static CoreErrors GuessBodylessErrorType(Aws::Http::HttpResponseCode responseCod
         return CoreErrors::ACCESS_DENIED;
     case HttpResponseCode::NOT_FOUND:
         return CoreErrors::RESOURCE_NOT_FOUND;
+    case HttpResponseCode::BAD_REQUEST:
```
pcmoritz, line 76: This is the main change of this PR, everything else is plumbing to make the BAD_REQUEST error code available in every error class.

pcmoritz:
This PR is part of trying to improve the error messages that are raised from Apache Arrow when interacting with S3: https://issues.apache.org/jira/browse/ARROW-17079

PR #2074, opened 2022-09-01
Fix potential namespace collisions in codegen.

sbiscigl:
*Description of changes:*

Fixes a potential namespace collision when an Operation and Shape have the same name, specifically a Getter method.

*Check all that applies:*

- [x] Linux
- [x] Windows
- [ ] Android
- [x] MacOS
- [ ] IOS
- [ ] Other Platforms

### DIFF
--- a/code-generation/generator/src/main/resources/com/amazonaws/util/awsclientgenerator/velocity/cpp/ModelClassMembersAndInlines.vm
+++ b/code-generation/generator/src/main/resources/com/amazonaws/util/awsclientgenerator/velocity/cpp/ModelClassMembersAndInlines.vm
@@ -63,16 +63,20 @@
 #set($setterWithCopy = " { ${required}${memberVariableName} = value; }")
 #set($setterWithMove = " { ${required}${memberVariableName} = std::move(value); }")
 #end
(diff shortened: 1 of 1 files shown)


code-generation/generator/src/main/resources/com/amazonaws/util/awsclientgenerator/velocity/cpp/ModelClassMembersAndInlines.vm
```diff
@@ -63,16 +63,20 @@
 #set($setterWithCopy = " { ${required}${memberVariableName} = value; }")
 #set($setterWithMove = " { ${required}${memberVariableName} = std::move(value); }")
 #end
+#set($getterFunctionName = "Get${memberKeyWithFirstLetterCapitalized}")
+#if($getterFunctionName == ${otherMember.value.shape.name})
```
SergeyRyabinin, line 67: This place is executed in a loop
in case of very huge models, it may affect the performance/increase the code gen time.
Please try to run ec2 code generation with and without change to see the diff.

SergeyRyabinin (approved):

PR #2077, opened 2022-09-02
[aws-cpp-sdk-s3-crt]: handle s3 meta_request errors in Get/PutObjectAsync

### LINKED ISSUE #2076: [aws-cpp-sdk-s3-crt]: Get/PutObjectAsync hang when s3 meta_request creation fails
### Describe the bug

We are observing that `PutObjectAsync` calls hang if the `aws_s3_meta_request` fails and returns a `NULL` pointer.

The _last log lines_ are in this case:
```
9/2/2022, 1:26:16 PM UTC	stderr	[ERROR] 2022-09-02 13:26:16.826 S3MetaRequest [140368700321792] Could not create auto-ranged-put meta request; required part size for put request is 1844674407370955, but current maximum part size is 33554432
9/2/2022, 1:26:16 PM UTC	stderr	[ERROR] 2022-09-02 13:26:16.826 S3Client [140368700321792] id=0x7fa849e7f000: Could not create new meta request.
```

The `rawRequest` is `NULL` in this case, and hence the `PutObjectResponseReceivedHandler` is never invoked, causing the program to hang indefinitely.

### Expected Behavior

Program/job should not hang indefinitely on failure.

### Current Behavior

Explanation of failure:
```c++
// source/S3CrtClient.cpp
void S3CrtClient::PutObjectAsync(const PutObjectRequest& request, const PutObjectResponseReceivedHandler& handler, const std::shared_ptr<const Aws::Client::AsyncCallerContext>& context) const
{
  // ...
  std::lock_guard<std::mutex> lockRequest{userData->underlyingS3RequestMutex};
  aws_s3_meta_request *rawRequest = aws_s3_client_make_meta_request(m_s3CrtClient, &options);
  userData->underlyingS3Request = rawRequest; // <== not checked for NULL here
}}
```
There is no check for `nullptr` of `rawRequest`, hence the `handler` is never invoked, hence the program hangs.
The identical problem exists in `GetObjectAsync` (which also affects `GetObject`).

### Reproduction Steps

Could write a (mock) unit test where the returned `aws_s3_meta_request` is `NULL`.

Otherwise, anything that can cause `aws_s3_client_make_meta_request` to fail.

### Possible Solution

Check if `rawRequest` is `NULL`.

_No response_

### AWS CPP SDK version used

### Compiler and Version used

gcc / clang

Linux, ubuntu 18.04

grrtrr:
When `aws_s3_client_make_meta_request` returns `NULL`, `{Get,Put}ObjectAsync` hang,
since the response handlers never get invoked.

Handle the case by returning an error instead.

Resolves #2076.

*Check all that applies:*
   Tested function via test program that handled the case of a NULL s3 meta_request.

- [X] Linux
- [ ] Windows
- [ ] Android
- [ ] MacOS
- [ ] IOS
- [ ] Other Platforms

### DIFF
--- a/aws-cpp-sdk-s3-crt/source/S3CrtClient.cpp
+++ b/aws-cpp-sdk-s3-crt/source/S3CrtClient.cpp
@@ -500,8 +500,12 @@ void S3CrtClient::GetObjectAsync(const GetObjectRequest& request, const GetObjec
   options.message= crtHttpRequest->GetUnderlyingMessage();
   userData->crtHttpRequest = crtHttpRequest;
 
-  std::lock_guard<std::mutex> lockRequest{userData->underlyingS3RequestMutex};
   aws_s3_meta_request *rawRequest = aws_s3_client_make_meta_request(m_s3CrtClient, &options);
+  if (rawRequest == nullptr)
+  {
+    return handler(this, request, GetObjectOutcome(Aws::Client::AWSError<S3CrtErrors>(S3CrtErrors::INTERNAL_FAILURE, "INTERNAL_FAILURE", "Unable to create s3 meta request", false)), context);
+  }
+  std::lock_guard<std::mutex> lockRequest{userData->underlyingS3RequestMutex};
   userData->underlyingS3Request = rawRequest;
 }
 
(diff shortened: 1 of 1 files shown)


grrtrr:
Changes have been absorbed into #2112, please see there.

PR #2083, opened 2022-09-07
Remove AsyncHelper functions by manually inlining

SergeyRyabinin:
*Issue #, if available:*
Client header size and read-ability by humans.
In general, having these functions private on a service clients are not mandatory.
Even having these functions as a separate functions are not needed.

*Description of changes:*
So inline them in the lambdas in a service client translation unit.

This is another take on https://github.com/aws/aws-sdk-cpp/pull/2082

*Check all that applies:*

- [x] Linux
- [x] Windows
- [ ] Android
- [ ] MacOS
- [ ] IOS
- [ ] Other Platforms


sdavtaker (approved):
Great work simplifying the generated headers. Way more readable and simpler to follow code.

PR #2088, opened 2022-09-09
Fix USE_OPENSSL is ignored

temideoye:
Currently, `-DUSE_OPENSSL` has no effect. In its stead, an undocumented variable – `ENABLE_OPENSSL_ENCRYPTION` –  is what gets evaluated to verify if OpenSSL was requested. `ENABLE_OPENSSL_ENCRYPTION` defaults to undefined/OFF, which invariably means OpenSSL is actually never used, regardless of the value of `USE_OPENSSL`. This push proposes a fix by linking `USE_OPENSSL` to `ENABLE_OPENSSL_ENCRYPTION`.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -94,6 +94,9 @@ if(NOT USE_OPENSSL)
     message(WARNING "Turning off USE_OPENSSL will install AWS-LC as replacement of OpenSSL in the system default directory. This is an experimental feature. Do not use if you have an OpenSSL installation in your system already.")
 endif()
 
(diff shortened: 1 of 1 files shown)


jmklix:
Thanks for taking the time to make this PR. We recently changed this sdk to depend on CRT's crypto implementation. aws-crt-cpp correctly respects the usage of `USE_OPENSSL`. You can look into that implementation in this [file](https://github.com/awslabs/aws-crt-cpp/blob/main/CMakeLists.txt#L9). Closing this PR as it is no longer needed, but please let us know if you have any other problems with this sdk.

PR #2094, opened 2022-09-13
[aws-cpp-sdk-s3-crt]: handle bad input stream passed to PutObjectAsync

### LINKED ISSUE #2093: [aws-cpp-sdk-s3-crt]: PutObject fails badly if input stream is in a bad state
### Describe the bug

If the input stream to `PutObject` (provided via `SetBody` of the parent `AmazonStreamingWebServiceRequest` method) is in a bad state ([`fail()`](https://en.cppreference.com/w/cpp/io/basic_ios/fail) returns `true`), the `PutObject{,Async}` call fails with an arcane error:
```
[ERROR] 2022-09-13 13:23:17.055 S3MetaRequest [139790909245696] Could not create auto-ranged-put meta request; required part size for put request is 1844674407370955, but current maximum part size is 5000000000000
[ERROR] 2022-09-13 13:23:17.055 S3Client [139790909245696] id=0x7f23992eb200: Could not create new meta request.
```

Without #2076/#2077, the request in addition hangs.

The huge value of `1844674407370955` is the decoded `Content-Length` header of the `aws-c-s3` request, divided by `g_s3_max_num_upload_parts = 1000. The cause of the failure becomes clearer when printing out the Content-Length header string, and the parsed `content_length` value in `aws-c-s3`:
```c
// aws-c-s3/source/s3_client.c
    if (!aws_http_headers_get(initial_message_headers, g_content_length_header_name, &content_length_cursor)) {
        struct aws_string *content_length_str = aws_string_new_from_cursor(client->allocator, &content_length_cursor);
        char *content_length_str_end = NULL;

        content_length = strtoull((const char *)content_length_str->bytes, &content_length_str_end, 10);
       AWS_LOGF_ERROR(  // <== ADDED THIS STATEMENT
                AWS_LS_S3_META_REQUEST,
                "Content-Length header: %s, parsed: %" PRIu64 ", max: %" PRIu64,
                (const char *)content_length_str->bytes,
                content_length, ULLONG_MAX);
        aws_string_destroy(content_length_str);

        content_length_str = NULL;
        content_length_header_found = true;
    }
```
The combined log output now is:
```
[ERROR] 2022-09-13 13:23:17.055 S3MetaRequest [139790909245696] Content-Length header: -1, parsed: 18446744073709551615, max: 18446744073709551615
[ERROR] 2022-09-13 13:23:17.055 S3MetaRequest [139790909245696] Could not create auto-ranged-put meta request; required part size for put request is 1844674407370955, but current maximum part size is 5000000000000
[ERROR] 2022-09-13 13:23:17.055 S3Client [139790909245696] id=0x7f23992eb200: Could not create new meta request.
```
This shows that the `strtoull(3)` conversion fails, indicated by returning `ULLONG_MAX`. This failure was caused by the `Content-Length` string to be set to `"-1"`.

The _cause_ for that value is due to the use of `tellg` in `aws-crt-cpp` `InputStream`:
```c++
// aws-crt-cpp/source/io/Stream.cpp
            int64_t StdIOStreamInputStream::GetLengthImpl() const noexcept
            {   
                auto currentPosition = m_stream->tellg();

                m_stream->seekg(0, std::ios_base::end);
                int64_t retVal = -1;

                if (*m_stream)
                {   
                    retVal = static_cast<int64_t>(m_stream->tellg()); // <=== HERE, RETURNS -1
                }

                m_stream->seekg(currentPosition);

                return retVal;
            }
```
The `InputStream` provides the above get-length method for `aws-c-s3` to fill in the value of the `Content-Length` header.
Since the input file stream is in a bad state (`failbit` or `badbit` set), [`std::istream::tellg` returns -1](https://en.cppreference.com/w/cpp/io/basic_istream/tellg).

This then causes the `Content-Length` string to appear as `"-1"`, which then produces the above described failure of `strtoull` and the unrelated "part size too big" error.

### Expected Behavior

Fail early, with a clear error message informing the caller that the provided input stream (which is outside of the control of the SDK code) is in a bad/failed state.

### Current Behavior

Arcane error due to late failure.

### Reproduction Steps

Supply a `std::istream` to `PutObject` via `SetBody` which has the badbit or failbit set.

### Possible Solution

Check for a bad input stream in `aws-cpp-sdk-s3-crt/source/S3CrtClient.cpp` and return an error if the stream is in a bad state.

_No response_

### AWS CPP SDK version used

### Compiler and Version used

gcc / clang

Linux, ubuntu 18.04

grrtrr:
If `PutObjectAsync` is passed an input stream which is in a bad state (failbit or badbit set), the `aws-c-s3` meta request fails with arcane errors, since the Content-Length is computed as the string `"-1"`  (see #2093).

Avoid the late failure (with confusing error) by returning early with error if the user supplied an unusable input stream.

 Resolves #2093.

*Check all that applies:*

- [X] Linux
- [ ] Windows
- [ ] Android
- [ ] MacOS
- [ ] IOS
- [ ] Other Platforms


jmklix:
Merged changes with additional required code-gen changes. closing this PR

PR #2101, opened 2022-09-21
[aws-cpp-sdk-s3-crt]: Configure TCP connection-monitoring options

### LINKED ISSUE #2107: [aws-cpp-sdk-s3-crt]: need default settings for TCP connection-monitoring options
### Describe the bug

We have repeatedly experienced problems with `S3CrtClient` programs that were hanging due to stuck/low-progress TCP connections. 
These can be avoided by TCP keep-alive probes and TCP connection-speed monitoring, both are [enabled by default](https://github.com/aws/aws-sdk-cpp/blob/main/aws-cpp-sdk-core/include/aws/core/client/ClientConfiguration.h#L107-L122).

The `S3CrtClient` and `aws_s3_client` do [_not_ enable either option by default](https://github.com/awslabs/aws-c-s3/blob/main/source/s3_client.c#L252):
 - [TCP keep-alive](https://github.com/awslabs/aws-c-s3/blob/main/source/s3_client.c#L327-L330) and 
 - [TCP connection speed monitoring](https://github.com/awslabs/aws-c-s3/blob/main/source/s3_client.c#L332-L336).
 
 have to be explicitly configured.

### Expected Behavior

Programs should recover from bad TCP connections. 

We documented problems caused by the absence of TCP keep-alive probes in https://github.com/awslabs/aws-c-s3/issues/210.

Keep-alive probes do not cover all cases - a connection may respond to keep-alive probes, but lack progress/speed.
One possible cause is [CPU starvation](https://github.com/amzn/amzn-drivers/blob/master/kernel/linux/ena/ENA_Linux_Best_Practices.rst#cpu-starvation) - we have observed stuck programs which do no more than reading a small file from `s3`, and write out a few files to `s3` when the CPU load was near 100%.

There is likely reason why the main SDK has been enabling both TCP keep-alives and TCP connection-speed monitoring [by default](https://github.com/aws/aws-sdk-cpp/blob/main/aws-cpp-sdk-core/include/aws/core/client/ClientConfiguration.h#L107-L122).

We should be able to expect the same defaults (and recovery from bad TCP connections) in the `S3CrtClient`.

### Current Behavior

There is no recovery from bad TCP connections currently.

### Reproduction Steps

Please see https://github.com/awslabs/aws-c-s3/issues/210 for details on TCP keep-alive issues.

These problems will occur with bad TCP connections. Tools like `tc` and/or `iptables` can be used to emulate these.

### Possible Solution

- https://github.com/awslabs/aws-c-s3/pull/204 exposed TCP keep-alive and connection-monitoring options for the `s3_client`;
- https://github.com/aws/aws-sdk-cpp/issues/1882 extended the options for the `S3CrtClient`,
- https://github.com/aws/aws-sdk-cpp/pull/2101 enables TCP keep-alive and connection-monitoring by default (i.e. fixes the problem reported above).

_No response_

### AWS CPP SDK version used

### Compiler and Version used

gcc / clang

Linux, ubuntu 18.04

grrtrr:
The main SDK enables TCP keep-alive and connection-speed monitoring by default, these options are also important for the S3CrtClient (e.g. https://github.com/awslabs/aws-c-s3/issues/210).

Hence, based on https://github.com/awslabs/aws-c-s3/pull/204, expose the following `ClientConfiguration` options to CRT:
- `connect_timeout_ms`: S3 socket connection timeout;
- `enableTcpKeepAlive`: enable TCP keep-alive probes;
- `tcpKeepAliveIntervalMs`: TCP keep-alive wait/probe interval;
- `lowSpeedLimit`: `aws-c-http` TCP connection speed monitoring;
- `requestTimeoutMs`: to provide upper bound on the monitoring interval for `aws-c-http` TCP connection speed monitoring.

*Issue #, if available:* Resolves #2107.

*Description of changes:*

*Check all that applies:*

- [X] Linux
- [ ] Windows
- [ ] Android
- [ ] MacOS
- [ ] IOS
- [ ] Other Platforms


aws-cpp-sdk-s3-crt/source/S3CrtClient.cpp
```diff
@@ -201,6 +201,32 @@ void S3CrtClient::init(const S3Crt::ClientConfiguration& config, const std::shar
...
   AWS_ZERO_STRUCT(s3CrtConfig);
   s3CrtConfig.region = Aws::Crt::ByteCursorFromCString(config.region.c_str());
+  s3CrtConfig.connect_timeout_ms = config.connectTimeoutMs;
+
+  aws_s3_tcp_keep_alive_options *tcp_keep_alive_options = nullptr;
+  if (config.enableTcpKeepAlive)
+  {
+    const uint16_t keep_intvl = std::max<uint16_t>(15, config.tcpKeepAliveIntervalMs/1000);
+
+    tcp_keep_alive_options = static_cast<aws_s3_tcp_keep_alive_options *>(aws_mem_calloc(Aws::get_aws_allocator(),
```
fboranek, line 211: Do you need to allocate it when at the end of scope deallocate it? What about keep it on the stack?

This is the very same code
```
   aws_s3_tcp_keep_alive_options tcp_keep_alive_options;
   if (config.enableTcpKeepAlive)
   {
     const uint16_t keep_intvl = std::max<uint16_t>(15, config.tcpKeepAliveIntervalMs/1000);
     tcp_keep_alive_options.keep_alive_interval_sec = keep_intvl;
     tcp_keep_alive_options.keep_alive_timeout_sec = keep_intvl;

    s3CrtConfig.tcp_keep_alive_options = &tcp_keep_alive_options;
   }
```

grrtrr: I used this in an earlier variant before submitting the PR. This part of the code was flagged by the address sanitizer, which flagged using the address of a local variable as an error. Hence switched to allocating it on the heap instead.

grrtrr: Just re-ran the tests. It was a `stack-use-after-scope` error, and it was my fault, since I had declared the variable inside the `if`-block, like this:
```c++
   aws_s3_tcp_keep_alive_options tcp_keep_alive_options;
   if (config.enableTcpKeepAlive)
   {
     const uint16_t keep_intvl = std::max<uint16_t>(15, config.tcpKeepAliveIntervalMs/1000);
     tcp_keep_alive_options.keep_alive_interval_sec = keep_intvl;
     tcp_keep_alive_options.keep_alive_timeout_sec = keep_intvl;

    s3CrtConfig.tcp_keep_alive_options = &tcp_keep_alive_options;
   }
```
Will update the PR to use stack-allocated variables. Thank you for pointing this out.

aws-cpp-sdk-s3-crt/source/S3CrtClient.cpp
```diff
@@ -201,6 +201,32 @@ void S3CrtClient::init(const S3Crt::ClientConfiguration& config, const std::shar
...
+    s3CrtConfig.tcp_keep_alive_options = &tcp_keep_alive_options;
+  }
+
+  aws_http_connection_monitoring_options tcp_monitoring_options;
+  if (config.lowSpeedLimit) {
+    // Use the same monitor interval default as used by the curl client, but allow override via requestTimeoutMs:
+    const uint32_t monitor_intvl = std::max<uint32_t>(3, config.requestTimeoutMs/1000);
+
+    AWS_ZERO_STRUCT(tcp_monitoring_options);
+    tcp_monitoring_options.minimum_throughput_bytes_per_second = config.lowSpeedLimit;
```
grrtrr, line 224: Will use a `static_cast` here, `lowSpeedLimit` is `long`, while the LHS is `uint64_t`.

aws-cpp-sdk-s3-crt/source/S3CrtClient.cpp
```diff
@@ -201,6 +201,32 @@ void S3CrtClient::init(const S3Crt::ClientConfiguration& config, const std::shar
   aws_s3_client_config s3CrtConfig;
   AWS_ZERO_STRUCT(s3CrtConfig);
   s3CrtConfig.region = Aws::Crt::ByteCursorFromCString(config.region.c_str());
+  s3CrtConfig.connect_timeout_ms = config.connectTimeoutMs;
+
+  aws_s3_tcp_keep_alive_options tcp_keep_alive_options;
+  if (config.enableTcpKeepAlive)
+  {
+    const uint16_t keep_intvl = std::max<uint16_t>(15, config.tcpKeepAliveIntervalMs/1000);
```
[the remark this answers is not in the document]
jmklix: The warning was `warning C4244: '=': conversion from 'const int' to 'uint16_t', possible loss of data `
and to fix that I changed it to this
```cpp
    uint16_t configKeepAliveS = static_cast<uint16_t>(std::min(static_cast<unsigned long>(std::numeric_limits<uint16_t>::max()), config.tcpKeepAliveIntervalMs / 1000ul));
    static const uint16_t MAX_CRT_KEEP_ALIVE = 15; // seconds
    const uint16_t keep_intvl = std::max(MAX_CRT_KEEP_ALIVE, configKeepAliveS);
```

jmklix:
Changes have been merged [here](https://github.com/aws/aws-sdk-cpp/pull/2166/files). Closing this PR

Issue #1850, opened 2022-01-21
aws-core: segfault if CleanupConfigAndCredentialsCacheManager() is called too late


grrtrr:
### Describe the bug

`CleanupConfigAndCredentialsCacheManager()` is not robust against being called late, i.e. after the `static` variable `s_configManager` has been destructed already once (as part of the `exit` sequence, calling the destructors of static variables).

In this case, the `ConfigAndCredentialsCacheManager` destructor gets called _twice_:
```
AWSConfigFileProfileConfigLoader() 0x7fac9d0d9408
AWSConfigFileProfileConfigLoader() 0x7fac9d0d9580
ConfigAndCredentialsCacheManager() 0x7fac9d0d9300
InitConfigAndCredentialsCacheManager() 0x7fac9d0d9300
AWSConfigFileProfileConfigLoader() 0x7fac9d032508
... test runs ...
ConfigAndCredentialsCacheManager() 0x7fac9d0d9300
~AWSConfigFileProfileConfigLoader(/home/grenker/.aws/config) 0x7fac9d0d9580
~AWSProfileConfigLoader() 0 0x7fac9d0d9580
~AWSConfigFileProfileConfigLoader(/home/grenker/.aws/credentials) 0x7fac9d0d9408
~AWSProfileConfigLoader() 10 0x7fac9d0d9408
... shutdown ...
CleanupConfigAndCredentialsCacheManager() 0x7fac9d0d9300
~AWSConfigFileProfileConfigLoader(/home/grenker/.aws/config) 0x7fac9d0d9580
~AWSProfileConfigLoader() 0 0x7fac9d0d9580
~AWSConfigFileProfileConfigLoader(/home/grenker/.aws/credentials) 0x7fac9d0d9408
~AWSProfileConfigLoader() 10 0x7fac9d0d9408
```
This prints out the object addresses and the sequence of constructor/destructor calls. The `0` and `10` numbers print the length of the `m_profiles` element at the time of call. 
The call to `CleanupConfigAndCredentialsCacheManager` happens after "shutdown", during the application exit sequence.

Here is the _detailed `gdb` trace:
```
(gdb) bt
#0  0x00007ffff6b67fbf in free ()
   from _solib_k8/libthird_Uparty_Sjemalloc_Sv4.5.0_Slibjemalloc.so
[...]

PR #2105, opened 2022-09-22
#1850 fix multiple destructor invocations of a static UniquePtr

SergeyRyabinin:
*Issue #, if available:*
[#1850 ](aws-core: segfault if CleanupConfigAndCredentialsCacheManager() is called too late #1850)
*Description of changes:*
Try to avoid compiler optimizations that skips unique_ptr underlying pointer nullifying.
*Check all that applies:*

- [x] Linux
- [x] Windows
- [ ] Android
- [ ] MacOS
- [ ] IOS
- [ ] Other Platforms

### DIFF
--- a/aws-cpp-sdk-core/source/monitoring/MonitoringManager.cpp
+++ b/aws-cpp-sdk-core/source/monitoring/MonitoringManager.cpp
@@ -25,7 +25,8 @@ namespace Aws
         /**
          * Global factory to create global metrics instance.
          */
-        static Aws::UniquePtr<Monitors> s_monitors;
+        using MonitorsDeleter = Aws::StaticUniquePtrDeleter<Monitors>;
+        static Aws::UniquePtr<Monitors, MonitorsDeleter> s_monitors(nullptr, MonitorsDeleter(nullptr));
 
         Aws::Vector<void*> OnRequestStarted(const Aws::String& serviceName, const Aws::String& requestName, const std::shared_ptr<const Aws::Http::HttpRequest>& request)
         {
@@ -93,7 +94,7 @@ namespace Aws
             {
                 return;
             }
-            s_monitors = Aws::MakeUnique<Monitors>(MonitoringTag);
+            s_monitors = Aws::MakeUniqueWithDeleter<Monitors>(MonitoringTag, MonitorsDeleter(&s_monitors));
             for (const auto& function: monitoringFactoryCreateFunctions)
             {
                 auto factory = function();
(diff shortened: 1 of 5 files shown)


aws-cpp-sdk-core/source/monitoring/MonitoringManager.cpp
```diff
@@ -25,7 +25,8 @@ namespace Aws
         /**
          * Global factory to create global metrics instance.
          */
-        static Aws::UniquePtr<Monitors> s_monitors;
+        using MonitorsDeleter = Aws::StaticUniquePtrDeleter<Monitors>;
+        static Aws::UniquePtr<Monitors, MonitorsDeleter> s_monitors(nullptr, MonitorsDeleter(nullptr));
```
graebm, line 29: Could we reduce boilerplate by defining an  `StaticUniquePtr<T>`? and then this code would be like:

`static Aws::StaticUniquePtr<Monitors> s_monitors(nullptr);`

aws-cpp-sdk-core/source/monitoring/MonitoringManager.cpp
```diff
@@ -93,7 +94,7 @@ namespace Aws
             {
                 return;
             }
-            s_monitors = Aws::MakeUnique<Monitors>(MonitoringTag);
+            s_monitors = Aws::MakeUniqueWithDeleter<Monitors>(MonitoringTag, MonitorsDeleter(&s_monitors));
```
graebm, line 97: and maybe reduce boilerplate here so it looks like:

`Aws::MakeStaticUnique(&s_monitors, MonitoringTag);`


and that helper function's definition is something like:
```
template<class T, typename ... Args>
StaticUniquePtr<T> MakeStaticUnique(T* pointer, Args&&... args)
```

it's been a while since I've ~done~ fought templates ... so no idea if that's all accurate

aws-cpp-sdk-core/source/monitoring/MonitoringManager.cpp
SergeyRyabinin: No, we can't for all cases, only for the zero-initialization.
The reason is that the 2nd c-tor argument (the deleter itself) should know the actual pointer value of the object being initialized on this line.
(or nullptr, it is just a workaround for not initialized)

SergeyRyabinin: well, we could implement the wrapper on top of unique_ptr or implement our own unique_ptr.
However, I haven't lost my mind yet (I hope), so won't do it.

aws-cpp-sdk-core/source/monitoring/MonitoringManager.cpp
SergeyRyabinin: The problem with make_static is that C++ standard semantic/logic enforces that unique_ptrs allocated via std::make_unique to not have a custom deleter.... so "WithDeleter" is a very important addition to this function mocking the std::make_unique and violating the initial std::unique_ptr design.

However, an overload with a default deleter provided is feasible here.

SergeyRyabinin:
Another approach has been merged:
https://github.com/aws/aws-sdk-cpp/pull/2108

PR #2155, opened 2022-10-21
Endpoints 2.0 core interface

SergeyRyabinin:
*Issue #, if available:*
Endpoints 2.0
*Description of changes:*
Changes to the core of the SDK to support new endpoints 2.0 API
*Check all that applies:*

- [ ] Linux
- [ ] Windows
- [ ] Android
- [ ] MacOS
- [ ] IOS
- [ ] Other Platforms


aws-cpp-sdk-core/include/aws/core/endpoint/EndpointProviderBase.h
```diff
@@ -26,18 +27,18 @@ namespace Aws
           * EndpointProviderBase is an interface definition that resolves the provided
           *   EndpointParameters to either an Endpoint or an error.
           */
-        template<typename EndpointParametersT>
         class AWS_CORE_API EndpointProviderBase
         {
         public:
+            using EndpointParameters = Aws::Vector<EndpointParameter>;
```
SergeyRyabinin, line 33: because template does not make much sense here.

aws-cpp-sdk-core/source/endpoint/AWSPartitions.cpp
```diff
@@ -0,0 +1,14 @@
...
+ * SPDX-License-Identifier: Apache-2.0.
+ */
+
+#include <aws/core/endpoint/AWSPartitions.h>
+
+namespace Aws
+{
+namespace Endpoint
+{
+constexpr AWSPartitions::PartitionsBlobT AWSPartitions::PartitionsBlob;
```
SergeyRyabinin, line 12: this is required for constexpr in C++11 (that we have)

aws-cpp-sdk-core/include/aws/core/client/AWSClient.h
```diff
@@ -153,6 +154,16 @@ namespace Aws
              */
             Aws::String GeneratePresignedUrl(Aws::Http::URI& uri, Aws::Http::HttpMethod method, const char* region, const char* serviceName, const char* signerName, const Aws::Http::HeaderValueCollection& customizedHeaders, long long expirationInSeconds = 0);
 
+#if AWS_SDK_VERSION_MAJOR > 1 || AWS_SDK_VERSION_MINOR >= 10
```
sbiscigl, line 157: dont we want this to be `&&`? technically [`0.14.9`](https://github.com/aws/aws-sdk-cpp/blob/0.14.9/aws-cpp-sdk-core/include/aws/core/VersionConfig.h#L16) would pass this check

aws-cpp-sdk-core/include/aws/core/endpoint/AWSEndpoint.h
```diff
@@ -19,14 +21,10 @@ namespace Aws
...
-            {
-            public:
-                virtual ~AWSEndpointAttribute();
-            };
+            using EndpointAttributes = Internal::Endpoint::EndpointAttributes;
 
-
-            virtual ~AWSEndpoint();
+            virtual ~AWSEndpoint()
+            {};
```
sbiscigl, line 27: nit: no need for new lline

aws-cpp-sdk-core/include/aws/core/endpoint/BuiltInParameters.h
```diff
@@ -0,0 +1,35 @@
...
+        class AWS_CORE_API BuiltInParameters
+        {
+        public:
+            using EndpointParameter = Aws::Endpoint::EndpointParameter;
+
+            virtual ~BuiltInParameters(){};
+
+            const EndpointParameter& GetParameter(const Aws::String& name) const;
+            void SetParameter(EndpointParameter param);
+            void SetStringParameter(Aws::String name, Aws::String value);
```
sbiscigl, line 26: why not have name and value be const references?

aws-cpp-sdk-core/include/aws/core/endpoint/internal/AWSEndpointAttribute.h
```diff
@@ -0,0 +1,89 @@
...
+                    m_disableDoubleEncoding = disableDoubleEncoding;
+                }
+
+            private:
+                Aws::String m_name;
+
+                Crt::Optional<Aws::String> m_signingName;
+                Crt::Optional<Aws::String> m_signingRegion;
+                Crt::Optional<Aws::String> m_signingRegionSet;
+                Crt::Optional<bool> m_disableDoubleEncoding;
```
sbiscigl, line 74: does this need to be optional since a boolean would just be initialized to false?

aws-cpp-sdk-core/source/client/AWSClient.cpp
```diff
@@ -608,6 +608,50 @@ StreamOutcome AWSClient::MakeRequestWithUnparsedResponse(const Aws::Http::URI& u
...
+#if AWS_SDK_VERSION_MAJOR > 1 || AWS_SDK_VERSION_MINOR >= 10
+StreamOutcome AWSClient::MakeRequestWithUnparsedResponse(const Aws::AmazonWebServiceRequest& request,
+                                                         const Aws::Endpoint::AWSEndpoint& endpoint,
+                                                         Http::HttpMethod method,
+                                                         const char* signerName,
+                                                         const char* signerRegionOverride,
+                                                         const char* signerServiceNameOverride) const
+{
+    Aws::Http::URI uri = endpoint.GetURL();
+    if (endpoint.GetAttributes()) {
```
sbiscigl, line 620: if the endpoint has no attributes can we even make a request? should we log an error and short circuit the request if it is not present?

aws-cpp-sdk-core/source/endpoint/internal/AWSEndpointAttribute.cpp
```diff
@@ -0,0 +1,65 @@
+/**
+ * Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.
+ * SPDX-License-Identifier: Apache-2.0.
+ */
+
+#include <aws/core/endpoint/internal/AWSEndpointAttribute.h>
+#include "aws/core/utils/logging/LogMacros.h"
```
sbiscigl, line 7: use `<` instead of `"`

aws-cpp-sdk-core/source/endpoint/internal/AWSEndpointAttribute.cpp
```diff
@@ -0,0 +1,65 @@
...
+                        } else if (mapItemProperty.first == "signingRegionSet") {
+                            authScheme.SetSigningRegionSet(mapItemProperty.second.AsString());
+                        } else if (mapItemProperty.first == "disableDoubleEncoding") {
+                            authScheme.SetDisableDoubleEncoding(mapItemProperty.second.AsBool());
+                        } else {
+                            AWS_LOG_WARN(ENDPOINT_AUTH_SCHEME_TAG, Aws::String("Unknown Endpoint authSchemes attribute property: " + mapItemProperty.first).c_str());
+                        }
+                    }
+                }
+                /*
```
sbiscigl, line 46: nit: remove commented code?

aws-cpp-sdk-core/source/endpoint/internal/AWSEndpointAttribute.cpp
```diff
@@ -0,0 +1,65 @@
...
+    {
+        Aws::Map<Aws::String, Utils::Json::JsonView> jsonMap = jsonObject.View().GetAllObjects();
+        for (const auto& mapItemAttribute : jsonMap)
+        {
+            if (mapItemAttribute.first == "authSchemes" && mapItemAttribute.second.IsListType()) {
+                Aws::Utils::Array<Utils::Json::JsonView> jsonAuthSchemeArray = mapItemAttribute.second.AsArray();
+
+                for (size_t arrayIdx = 0; arrayIdx < jsonAuthSchemeArray.GetLength(); ++arrayIdx)
+                {
+                    const Utils::Json::JsonView& property = jsonAuthSchemeArray.GetItem(arrayIdx);
```
sbiscigl, line 28: any reason not to do

```
for(const auto &property: jsonAuthSchemeArray) {
  ...
}
```

aws-cpp-sdk-core/include/aws/core/endpoint/BuiltInParameters.h
SergeyRyabinin: pass-by-value and move
also, it avoids potential issue of bad_alloc at the callee, because now caller is going to allocate memory for these strings.

aws-cpp-sdk-core/include/aws/core/endpoint/internal/AWSEndpointAttribute.h
SergeyRyabinin: I think there is difference between value to be set and actual value itself.
This API tells us if endpoints rules specify the behavior, and if so, what is specified.

aws-cpp-sdk-core/source/endpoint/internal/AWSEndpointAttribute.cpp
SergeyRyabinin: Aws::Utils::Array does not provide std:: range-for required functions.

aws-cpp-sdk-core/source/endpoint/internal/AWSEndpointAttribute.cpp
SergeyRyabinin: thanks

aws-cpp-sdk-core/source/endpoint/internal/AWSEndpointAttribute.cpp
SergeyRyabinin: thanks

aws-cpp-sdk-core/source/client/AWSClient.cpp
SergeyRyabinin: Frankly speaking I'm not 100% sure
It is mostly for the case when endpoint rules do not care and we fallback to the c2j-model defaults.

aws-cpp-sdk-core/include/aws/core/client/AWSClient.h
SergeyRyabinin: thanks

sdavtaker: I would just replace for exact match on 1.10, we only want to have it to work around merging this code before the last 1.9 release. After first release of 1.10 is done done, we should just remove the conditional behavior on version number.

PR #2175, opened 2022-11-03
switch sprintf to snprintf

jmklix:
*Issue #, if available:*
Ran into this warning when building on mac:
```
"This function is provided for compatibility reasons only.  Due to security concerns inherent in the design of sprintf(3), it is highly recommended that you use snprintf(3) instead."
```

*Description of changes:*
switch sprintf to snprintf
```cpp
//from
sprintf(buf, "%lld", num);
//to
snprintf(buf, sizeof(buf), "%lld", num);
```
*Check all that applies:*

- [X] Linux
- [X] Windows
- [ ] Android
- [X] MacOS
- [ ] IOS
- [ ] Other Platforms

### DIFF
--- a/aws-cpp-sdk-core/source/external/cjson/cJSON.cpp
+++ b/aws-cpp-sdk-core/source/external/cjson/cJSON.cpp
@@ -1018,7 +1018,7 @@ static cJSON_AS4CPP_bool print_string_ptr(const unsigned char * const input, pri
                     break;
                 default:
                     /* escape and print as unicode codepoint */
-                    snprintf((char*)output_pointer, sizeof(output_pointer), "u%04x", *input_pointer);
+                    snprintf((char*)output_pointer, output_buffer->length - (output_pointer - output_buffer->buffer), "u%04x", *input_pointer);
                     output_pointer += 4;
                     break;
             }
--- a/aws-cpp-sdk-core/source/external/cjson/cJSON.cpp
+++ b/aws-cpp-sdk-core/source/external/cjson/cJSON.cpp
@@ -120,7 +120,7 @@ CJSON_AS4CPP_PUBLIC(double) cJSON_AS4CPP_GetNumberValue(const cJSON * const item
 CJSON_AS4CPP_PUBLIC(const char*) cJSON_AS4CPP_Version(void)
 {
     static char version[15];
-    sprintf(version, "%i.%i.%i", CJSON_AS4CPP_VERSION_MAJOR, CJSON_AS4CPP_VERSION_MINOR, CJSON_AS4CPP_VERSION_PATCH);
+    snprintf(version, sizeof(version), "%i.%i.%i", CJSON_AS4CPP_VERSION_MAJOR, CJSON_AS4CPP_VERSION_MINOR, CJSON_AS4CPP_VERSION_PATCH);
 
     return version;
 }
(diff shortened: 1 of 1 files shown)


aws-cpp-sdk-core/source/external/cjson/cJSON.cpp
```diff
@@ -1018,7 +1018,7 @@ static cJSON_AS4CPP_bool print_string_ptr(const unsigned char * const input, pri
                     break;
                 default:
                     /* escape and print as unicode codepoint */
-                    sprintf((char*)output_pointer, "u%04x", *input_pointer);
+                    snprintf((char*)output_pointer, sizeof(output_pointer), "u%04x", *input_pointer);
```
graebm, line 1021: this is not correct

`output_pointer` is not an array on the stack like `char output_pointer[8]` that you can simply do `sizeof()` on

I don't know what it is without a bunch of investigation...

aws-cpp-sdk-core/source/external/cjson/cJSON.cpp
```diff
@@ -120,7 +120,7 @@ CJSON_AS4CPP_PUBLIC(double) cJSON_AS4CPP_GetNumberValue(const cJSON * const item
 CJSON_AS4CPP_PUBLIC(const char*) cJSON_AS4CPP_Version(void)
 {
     static char version[15];
-    sprintf(version, "%i.%i.%i", CJSON_AS4CPP_VERSION_MAJOR, CJSON_AS4CPP_VERSION_MINOR, CJSON_AS4CPP_VERSION_PATCH);
+    snprintf(version, sizeof(version), "%i.%i.%i", CJSON_AS4CPP_VERSION_MAJOR, CJSON_AS4CPP_VERSION_MINOR, CJSON_AS4CPP_VERSION_PATCH);
```
graebm, line 123: are we sure we want to be laying changes on top of this copy/pasted CJSON implementation?

if we take an update, won't these changes just get wiped away?

graebm: since this is external code, it might be simpler to just disable the warnings from the top of the file, or in the CMakeLists.txt somehow ... something isolated with a high chance of being preserved if someone copy/pastes an updated cJSON.cpp into the repo

... discuss with the CPP team how they want to handle it

aws-cpp-sdk-core/source/external/cjson/cJSON.cpp
jmklix: Is the fix as simple as `(*output_pointer)`?

jmklix: Fixed with `output_buffer->length - (output_pointer - output_buffer->buffer)`
https://github.com/DaveGamble/cJSON/commit/ae49da2b61a378b9c0e07572ca759cd305c8b9c8

aws-cpp-sdk-core/source/external/cjson/cJSON.cpp
SergeyRyabinin: Since we are already C99, since we have already modified version of cJSON, we will unblock existing mac builds (i.e. I'm for merging this change).
And deal with external dependencies later.

SergeyRyabinin (approved):

PR #2220, opened 2022-11-25
Generate each client in docs as an external project.

sbiscigl:
*Description of changes:*

Our current [API Docs](https://sdk.amazonaws.com/cpp/api/LATEST/index.html) take a considerable amount of time to build and are monolithic and hard to navigate. This treats each service as a [external project](https://www.doxygen.nl/manual/external.html) with a home page that will be able to navigate all sub-projects. This way a customer will be able to view more relevant documentation quicker as well as aiding our process to build documentation quicker

*Check all that applies:*

- [x] Linux
- [ ] Windows
- [ ] Android
- [x] MacOS
- [ ] IOS
- [ ] Other Platforms


SergeyRyabinin (approved):

doxygen/modules/template/module-template.config
```diff
@@ -0,0 +1,322 @@
+# Doxyfile 1.8.11
+
+#---------------------------------------------------------------------------
+# Project related configuration options
+#---------------------------------------------------------------------------
+DOXYFILE_ENCODING      = UTF-8
+PROJECT_NAME           = "AWS SDK for C++"
+PROJECT_NUMBER         = "1.10.12"
```
sdavtaker, line 8: Shouldn't this one be a placeholder?

sbiscigl: Its replaced with a `sed` statement, and was previously i.e.

```
'sed -i "s/PROJECT_NUMBER .*/PROJECT_NUMBER         = $VERSION_NUM/" ./doxygen/modules/static/root.config',
```


 at runtime, so it doesnt matter what is there. previously, actually currently until we do this, [it is also hard coded](https://github.com/aws/aws-sdk-cpp/blob/main/doxygen/doxygen.config#L8) and replaced by sed at runtime. its actually set to `1.3.52` previously and replaced. in reality anything can be there.

PR #2244, opened 2022-12-16
[aws-cpp-sdk-core]: replace MonitoringManager assertions with guard statements

### LINKED ISSUE #2243: [aws-cpp-sdk-core]: assertions triggered in MonitoringManager.cpp at program exit
_This is a follow-up issue to #2135 and happens after #2136 had been added. Problem occurs in 1.10.18 and later revisions._

### Problem description

Programs using the AWS C++ SDK run to _completion_, but then fail at program exit due to a _failed assertion_:
```console
aws-cpp-sdk-core/source/monitoring/MonitoringManager.cpp:57: 
void Aws::Monitoring::OnRequestFailed(const String&, const String&, const std::shared_ptr<const Aws::Http::HttpRequest>&, 
                                      const HttpResponseOutcome&, const Aws::Monitoring::CoreMetricsCollection&, 
                                      Aws::Vector<void*>&): 
Assertion `s_monitors' failed.
```

As in #2135, this has to do with the shutdown of the global `s_monitors` variable:
```c++
// aws-cpp-sdk-core/source/monitoring/MonitoringManager.cpp
        void OnRequestFailed(const Aws::String& serviceName, const Aws::String& requestName, const std::shared_ptr<const Aws::Http::HttpRequest>& request,
                const Aws::Client::HttpResponseOutcome& outcome, const CoreMetricsCollection& metricsFromCore, const Aws::Vector<void*>& contexts)
        {   
            assert(s_monitors);  // <== HERE
            assert(contexts.size() == s_monitors->size());
            size_t index = 0;
            for (const auto& interface: *s_monitors)
            {   
                interface->OnRequestFailed(serviceName, requestName, request, outcome, metricsFromCore, contexts[index++]);
            }
        }
```

For more context, I added `fprintf` statements for the constructor calls, and to print out _information about the failed request_:
```console
InitMonitoring(): s_monitors = 0x7ff218489b20 
...
~ApiHandle()
after Aws::CleanupCrt()
~UniquePtrSafeDeleted() 0x7ff218489b20
~UniquePtrSafeDeleted() (nil)

FAILED REQUEST: S3 GetObject https://our.bucket/our.object?versionId=CnW9k4hIg0RaI7cZTdcRbKCHb5AbPXqF 

ERROR MESSAGE:
HTTP response code: -1
Resolved remote host IP address: 52.216.58.170
Request ID: 
~UniquePtrSafeDeleted() (nil)  
Exception name: 
Error message: Request cancelled by user's continuation handler
13 response headers:
accept-ranges : bytes
content-length : 5242880
content-range : bytes 0-5242879/33037704
content-type : binary/octet-stream
...
aws-cpp-sdk-core/source/monitoring/MonitoringManager.cpp:61: void Aws::Monitoring::OnRequestFailed(const String&, const String&, const std::shared_ptr<const Aws::Http::HttpRequest>&, const HttpResponseOutcome&, const Aws::Monitoring::CoreMetricsCollection&, Aws::Vector<void*>&): Assertion `s_monitors' failed.
```

### Analysis

At program start, `InitMonitoring()` sets  `s_monitors` to `0x7ff218489b20`. The program then successfully runs to the end.
After `main()` returns, first the `ApiHandle()` destructor is called. 

The statement `"after Aws::CleanupCrt()"` is printed right after `Aws::CleanupCrt()` returns in `Aws::ShutdownAPI()`.

We then see the `s_monitor` `UniquePtrSafeDeleted()` destructor called for `s_monitors` identified by ` 0x7ff218489b20` (there are multiple subsequent calls with a pointer argument of `nil`, but these are not relevant here).

The _cause_ of calling `OnRequestFailed` was that the request had been canceled (`"Request cancelled by user's continuation handler"`). Canceling outstanding request is done by our program at the end, before `main()` returns.

We are using the `TransferManager`, which uses the `Aws::Utils::Threading::DefaultExecutor`. Hence it is possible that the _detached threads_ spawned by the `DefaultExecutor` continue to be alive after `main()` returns, concurrent with the shutdown sequence.

_That means the fix in #2136 was not sufficient_, since
- detached threads may run to completion even after `main()` returns,
- they will want to call the `Aws::Monitoring` functions as part of their execution sequence,
- the destruction order of `static` global variables across translation units is undefined,
- hence we can not time the moment `s_monitors` is destructed.

We added #2136 to 1.9.170 and only observed the recurrence of the `SIGABRT` failed assertions after upgrading to 1.10.18.
The upgrade is not to blame, I don't think that this is the cause.

_Rather, the fix in #2136 was insufficient - monitoring functions must be robust against being called even after `ShutdownAPI()` has been called_.

### What to do

Improve on #2136 by adding `nullptr` checks for `s_monitors`, making each of the functions a no-op when the check holds.

To avoid a _race condition_  between testing the `if`-condition whether `s_monitors` is NULL, #2136 should also be reverted.
This will cause the `Aws::Monitoring` instance to be shut down first, at the begin of the shutdown sequence.

### Additional information

This was observed on Ubuntu 18.04, using clang-11. It will very likely happen on other platforms also.

grrtrr:
Due to using _detached threads_, `Aws::Monitoring` functions may be called at any time even after `main()` has exited.

This triggers asserting `s_monitors` to not be `NULL`, causing successfully completed programs to fail at exit.

Fix the problem by using guard statements in place of assertions. Also revert the changes in #2136, to avoid
a race condition of testing `s_monitors` to be NULL during shutdown.

Resolves #2243.

*Check all that applies:*

- [X] Linux
- [ ] Windows
- [ ] Android
- [ ] MacOS
- [ ] IOS
- [ ] Other Platforms

### DIFF
--- a/aws-cpp-sdk-core/source/monitoring/MonitoringManager.cpp
+++ b/aws-cpp-sdk-core/source/monitoring/MonitoringManager.cpp
@@ -29,12 +29,14 @@ namespace Aws
 
         Aws::Vector<void*> OnRequestStarted(const Aws::String& serviceName, const Aws::String& requestName, const std::shared_ptr<const Aws::Http::HttpRequest>& request)
         {
-            assert(s_monitors);
             Aws::Vector<void*> contexts;
-            contexts.reserve(s_monitors->size());
-            for (const auto& interface: *s_monitors)
+            if (s_monitors)
             {
-                contexts.emplace_back(interface->OnRequestStarted(serviceName, requestName, request));
+                contexts.reserve(s_monitors->size());
+                for (const auto& interface: *s_monitors)
+                {
+                    contexts.emplace_back(interface->OnRequestStarted(serviceName, requestName, request));
+                }
             }
             return contexts;
         }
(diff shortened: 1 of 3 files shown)


sbiscigl:
Hey @grrtrr thanks for coming up with this, running CI on it right now and will merge after it passes, just one small nit pick, do you think you could squash your commit history?

grrtrr:
> do you think you could squash your commit history?

Will do shortly (force-push).

sbiscigl:
Thanks for squashing! it passes CI, but i think some other contributors want to look at it a little more carefully, so I think we can hopefully merge monday after a quick review.

aws-cpp-sdk-core/source/monitoring/MonitoringManager.cpp
```diff
@@ -29,61 +29,71 @@ namespace Aws
...
         {
-            assert(s_monitors);
             Aws::Vector<void*> contexts;
-            contexts.reserve(s_monitors->size());
-            for (const auto& interface: *s_monitors)
+            if (s_monitors)
             {
-                contexts.emplace_back(interface->OnRequestStarted(serviceName, requestName, request));
+                contexts.reserve(s_monitors->size());
+                for (const auto& interface: *s_monitors)
```
SergeyRyabinin, line 36: Thanks for this change,
this is definitely a great improvement.
The only thing I can note is still a possibility for race condition:
if (s_monitors) can be true, but next line a separate thread can still destroy s_monitors.

We should merge this PR, but also follow-up on the SDK monitoring design.

SergeyRyabinin (approved):

aws-cpp-sdk-core/source/monitoring/MonitoringManager.cpp
grrtrr: Yes that is true. I was about to add a comment for @jmklix on the issue - this PR is not a fix, as there is still a chance of the race condition to occur.

It would be great if users could _disable the MonitoringManager_,
- e.g. at compile time via a `#define`, or
- via a configuration option.

grrtrr: @SergeyRyabinin - we ran into a similar situation in https://github.com/aws/aws-sdk-cpp/issues/2274. Could you please take a look to see if this is correct? To my understanding, the asynchronous threads of the `TransferManager` (used by the `S3Client`, not CRT) are the culprit, since these may complete after `main()` has returned.

I will look into a follow-up PR for this problem, since the presence of the race condition will mean that eventually a segmentation fault will occur.

PR #2252, opened 2022-12-21
Adding sigpipe ignore to intengration tests in linux

sdavtaker:
*Issue #, if available:*

*Description of changes:*

*Check all that applies:*

- [x] Linux
- [x] Windows
- [ ] Android
- [x] MacOS
- [ ] IOS
- [ ] Other Platforms

### DIFF
--- a/aws-cpp-sdk-cloudfront-integration-tests/RunTests.cpp
+++ b/aws-cpp-sdk-cloudfront-integration-tests/RunTests.cpp
@@ -6,20 +6,24 @@
 #include <gtest/gtest.h>
 #include <aws/core/Aws.h>
 #include <aws/testing/platform/PlatformTesting.h>
+#include <aws/testing/TestingEnvironment.h>
 #include <aws/testing/MemoryTesting.h>
 
 int main(int argc, char** argv)
 {
+    Aws::Testing::SetDefaultSigPipeHandler();
     Aws::SDKOptions options;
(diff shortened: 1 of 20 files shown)


sbiscigl (approved):

aws-cpp-sdk-cloudfront-integration-tests/RunTests.cpp
```diff
@@ -6,20 +6,26 @@
 #include <gtest/gtest.h>
 #include <aws/core/Aws.h>
 #include <aws/testing/platform/PlatformTesting.h>
+#include <aws/testing/TestingEnvironment.h>
 #include <aws/testing/MemoryTesting.h>
 
 int main(int argc, char** argv)
 {
+#ifdef __linux__
```
SergeyRyabinin, line 14: you already include PlatformTesting.h / MemoryTesting.h
You can reduce code duplication by extracting logic required into one of this common headers and just declare there and call a method here called smth like "InstallSigPipeHandler()".

We should also consider making the handler more smart that tries to detect if it is a known SIGPIPE from curl from OpenSSL or something different that should not be ignored.

sdavtaker: New revision including this change.

SergeyRyabinin: Thank you, ship it.

SergeyRyabinin (approved):

PR #2272, opened 2023-01-05
Fix poor implementation of string hash

jeking3:
*Issue #, if available:*

None.

*Description of changes:*

The original hashing algorithm used in the SDK produced 56,897 collisions on an input sample of 65,536 two-character strings. Changing to use `Aws::Crt::HashString` has made the hash more reliable.

*Check all that applies:*

- [ ] Linux
- [ ] Windows
- [ ] Android
- [X] MacOS
- [ ] IOS
- [ ] Other Platforms

### DIFF
--- a/aws-cpp-sdk-core-tests/utils/HashingUtilsTest.cpp
+++ b/aws-cpp-sdk-core-tests/utils/HashingUtilsTest.cpp
@@ -6,6 +6,7 @@
 #include <gtest/gtest.h>
 
 #include <aws/core/utils/HashingUtils.h>
+#include <aws/core/utils/memory/stl/AWSMap.h>
 #include <aws/core/utils/memory/stl/AWSStringStream.h>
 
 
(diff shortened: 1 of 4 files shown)


jeking3:
For building on MacOS, I also had to define sprintf as snprintf in cJSON to avoid deprecation warnings as errors.  That is not included in the PR.

aws-cpp-sdk-core/include/aws/core/utils/HashingUtils.h
```diff
@@ -107,7 +109,17 @@ namespace Aws
...
             static ByteBuffer CalculateCRC32C(Aws::IOStream& stream);
 
-            static int HashString(const char* strToHash);
+            /**
+             * Hash a string to a 32-bit number.  Returning the native size_t
+             * would be better, but there seems to be a lot of code that expects
+             * this to return a signed int, so keeping it that way for now.
+             */
+            static inline int HashString(const char* strToHash) {
+                // take 0x00000FFFFFFFF000 from the 64-bit value
```
sdavtaker, line 118: Nice experiment

aws-cpp-sdk-core-tests/utils/HashingUtilsTest.cpp
```diff
@@ -484,4 +485,40 @@ TEST(HashingUtilsTest, TestCRC32CFromStream)
...
+            check[2] = 0x00;
+            int hashValue = Aws::Utils::HashingUtils::HashString(check);
+            auto insertResult = hashResults.insert(
+                Aws::Map<size_t, std::pair<int, int>>::value_type(hashValue, std::pair<int, int>(c1, c2)));
+            if (!insertResult.second && c1 && c2) {
+                if (++failureCount < 10) {
+                    printf("HashString result %d identical for inputs %02x%02x and %02x%02x\n",
+                        hashValue, insertResult.first->second.first, insertResult.first->second.second, c1, c2);
+                }
+            }
```
sdavtaker, line 519: This can be more readable using a find and insert if the find reaches end, in place of relaying on a side effect of insert that may refer many to look into the standard to figure out what is the intention of the test.

sdavtaker (commented):
Thanks for the contribution, added small style comments, but seems to work fine as it is.

aws-cpp-sdk-core/include/aws/core/utils/HashingUtils.h
jeking3: I'd love to have fixed the rest of the SDK to use size_t but there are static const int declarations all over, and I'm not sure how all those files get generated or if I needed to power my way through it.

PR #2283, opened 2023-01-10
Update checking IMDS response

### LINKED ISSUE #2282: IMDSv2 required breaks EC2MetadataClient/AWSHttpResourceClient getting credentials
### Describe the bug

When the IMDS service fails or throttles for whatever reason, the behaviour in EC2MetadataClient will be broken forever.
The boolean [`m_tokenRequired` is set to false](https://github.com/aws/aws-sdk-cpp/blob/main/aws-cpp-sdk-core/source/internal/AWSHttpResourceClient.cpp#L262) and IMDSv1 is tried without a token. This fails because our instances require the token to be passed. IMDS returns a 401 **without** any body contents. This error is [translated to a network connection error](https://github.com/aws/aws-sdk-cpp/blob/main/aws-cpp-sdk-core/source/internal/AWSHttpResourceClient.cpp#L143-L146). When this happens the `m_tokenRequired` is never set to true anymore [because the check is on UNAUTHORIZED](https://github.com/aws/aws-sdk-cpp/blob/main/aws-cpp-sdk-core/source/internal/AWSHttpResourceClient.cpp#L209-L212)

Our long lived program now sticks with the old credentials until they expire: we never get new credentials.

### Expected Behavior

Credentials should not expire when IMDS returns an occasional error.

### Current Behavior

Credentials cannot be updated anymore

### Reproduction Steps

This is tough because IMDS needs to produce an error while credentials are fetched. This seems to happen about once a week in our fleet.

### Possible Solution

* Never fallback to IMDSv1. Not sure if all old EC2 hosts also have the latest IMDS behaviours.
* When 401 is returned, the code should not look at the body or its size at all and properly propagate the error

_No response_

### AWS CPP SDK version used

1.9

### Compiler and Version used

7.3.1

AmazonLinux2

jkamp-aws:
Do not check the body presence so we keep the 401 Unauthorized error returned when the token is missing in an empty response from IMDS. This should flip back the m_tokenRequired boolean and a retry will get the credentials through the secure method again.

Keep the body check when using a Marshall for the error response as used in the STS client from a subclass.

Fixes #2282

*Issue #2282 , if available:*

*Description of changes:*

*Check all that applies:*
- [o] Added proper tests to cover this PR. 
- Updated existing tests, but a test checking the downgrade/upgrade mechanism of IMDSv1 / IMDSv2 is missing.
- [-] Checked if this PR is a breaking (APIs have been changed) change.
- [-] Checked if this PR will _not_ introduce cross-platform inconsistent behavior.
- [o] Checked if this PR would require a ReadMe/Wiki update.

- [x] Linux
- [ ] Windows
- [ ] Android
- [ ] MacOS
- [ ] IOS
- [ ] Other Platforms

### DIFF
--- a/aws-cpp-sdk-core-tests/aws/auth/AWSHttpResourceClientTest.cpp
+++ b/aws-cpp-sdk-core-tests/aws/auth/AWSHttpResourceClientTest.cpp
@@ -635,7 +635,7 @@ namespace
                                                                        HttpMethod::HTTP_GET, Aws::Utils::Stream::DefaultResponseStreamFactoryMethod);
         std::shared_ptr<StandardHttpResponse> profileResponse = Aws::MakeShared<StandardHttpResponse>(ALLOCATION_TAG, profileRequest);
         profileResponse->SetResponseCode(HttpResponseCode::UNAUTHORIZED);
-        profileResponse->GetResponseBody() << "token required";
+        // Do not send a body as IMDS also doesn't send one. Just the HTTP error code.
         mockHttpClient->AddResponseToReturn(profileResponse);
 
         auto cred = ec2MetadataClient->GetDefaultCredentialsSecurely();
--- a/aws-cpp-sdk-core/source/internal/AWSHttpResourceClient.cpp
+++ b/aws-cpp-sdk-core/source/internal/AWSHttpResourceClient.cpp
@@ -140,12 +140,12 @@ namespace Aws
                 }
 
                 const Aws::Client::AWSError<Aws::Client::CoreErrors> error = [this, &response]() {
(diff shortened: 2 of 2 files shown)


SergeyRyabinin (approved):

SergeyRyabinin:
Hello,
Thank you a lot for your PR, we would like to merge it.
Could you please rebase it, if possible, otherwise I will merge one I rebased in a different PR: https://github.com/aws/aws-sdk-cpp/pull/2340

Best regards,
Sergey

SergeyRyabinin:
Merged in https://github.com/aws/aws-sdk-cpp/pull/2340

PR #2314, opened 2023-01-28
[aws-cpp-sdk-core]: add note that InitAPI and ShutdownAPIshould be called from the same thread

grrtrr:
Running `InitAPI()` and `ShutdownAPI()` from different threads can cause subtle problems, see e.g.
- https://github.com/aws/s2n-tls/issues/3525
- https://github.com/awslabs/aws-c-common/pull/891

Add a note for users to use the same thread.

*Check all that applies:*
- [X] Added proper tests to cover this PR. (If tests are not applicable, explain.) - Not applicable.

- [X] Linux
- [ ] Windows
- [ ] Android
- [ ] MacOS
- [ ] IOS
- [ ] Other Platforms

### DIFF
--- a/src/aws-cpp-sdk-core/include/aws/core/Aws.h
+++ b/src/aws-cpp-sdk-core/include/aws/core/Aws.h
@@ -286,7 +286,10 @@ namespace Aws
 
     /**
      * Shutdown SDK wide state for the SDK. This method must be called when you are finished using the SDK.
-     * Do not call any other SDK methods after calling ShutdownAPI.
+     * Notes:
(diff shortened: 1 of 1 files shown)


SergeyRyabinin (approved):

sdavtaker (approved):

grrtrr:
Here is a _third case_ (in `aws-c-io`) where calling `InitAPI/ShutdownAPI` from different threads causes problems/race conditions: https://github.com/awslabs/aws-c-io/pull/478.

PR #2346, opened 2023-02-10
Adding optional support for compressing requests

sdavtaker:
*Issue #, if available:*

*Description of changes:*

*Check all that applies:*

- [x] Linux
- [x] Windows
- [ ] Android
- [x] MacOS
- [ ] IOS
- [ ] Other Platforms


CMakeLists.txt
```diff
@@ -110,6 +112,13 @@ if (LEGACY_BUILD)
     find_package(PythonInterp)
     set(PYTHON3_CMD ${PYTHON_EXECUTABLE})
 
+    # Find zlib (gzip) and use it if available
+    find_package(ZLIB QUIET)
+    if ( NOT ZLIB_FOUND)
```
SergeyRyabinin, line 117: Is there a way in cmake to fail if the variable was set externally but not by default?
I.e. 
explicitly requested on AND !zlib => fail.
explicitly requested off AND zlib => still of

sdavtaker: We can add another parameter FORCE_... that would make it fail if set and cannot find the Zlib, or define it as a non-option (using set) parameter and parse it internally when set.
First option is more solid, since it will always have the right values. Second option is more flaky but does all in a single parameter.

SergeyRyabinin (approved):

PR #2347, opened 2023-02-10
Fix generating shapes with member but empty json response body

lucastetreault:
*Issue #, if available:*
n/a 

*Description of changes:*

Given an exception in a C2J model like this: 

```
<structure name="MyRedirectException">
    <member name="location" target="Location" />
</structure>
<exception target="MyRedirectException" />
<httperror target="MyRedirectException">
    <httpresponsecode value="302"/>
</httperror>
<httpheader target="MyRedirectException$location">
    <header value="Location" />
</httpheader>
```

When we generate the SDK it fails with the following message since the response body is empty: 
```
error : unused parameter 'jsonValue'
```

It looks like this was meant to be handled with `AWS_UNREFERENCED_PARAM(jsonValue);` but the if statement just needs to be tweaked for situations where the shape does have a member but it's bound to the header like `MyRedirectException` so the response body is empty and that value ends up unused. 

*Check all that applies:*
I didn't see any unit tests for the templates. I generated the full sdk and there was no difference in the generated code before/after my changes. For my custom C2J model this change allows me to build without errors. 

- [X] Linux
- [ ] Windows
- [ ] Android
- [ ] MacOS
- [ ] IOS
- [ ] Other Platforms


SergeyRyabinin (approved):
waiting for a dry run

Issue #2319, opened 2023-01-30
[aws-cpp-sdk-core] undefined behaviour when flushing ResponseStream before deletion


grrtrr:
### Problem description

We encountered _segmentation faults_ after switching from `fstream` to in-memory downloads (`stringstream`, boost `vectorstream`). It seems the original condition in #58 was not fixed. A related issue (looking at a different aspect) is #1732.

In our downloads we set the [`rdbuf()`](https://en.cppreference.com/w/cpp/io/basic_ios/rdbuf) member of the `Aws::IOStream` (`std::iostream`) returned by the `ResponseStreamFactory` of the `GetRequest`, in order to reduce copying.

This would not be a problem since the `std::iostream` destructor [does not destroy the associated `rdbuf()`](https://en.cppreference.com/w/cpp/io/basic_iostream/~basic_iostream). 

However, the `flush()` operation in `~ResponseStream`, which occurs after the download completed, uses the associated `rdbuf()` (which by now has gone out of scope), as shown in the following core dump. 

### Core dump

```console
(gdb) bt
#0  0x00007f0e7b41bc4b in std::ostream::flush() () from /usr/lib/x86_64-linux-gnu/libstdc++.so.6
#1  0x00007f0e78e5888d in Aws::Utils::Stream::ResponseStream::ReleaseStream (this=0x7f0e477fc2b0) at external/com_github_aws-sdk-cpp/aws-cpp-sdk-core/source/utils/stream/ResponseStream.cpp:61
#2  0x00007f0e78e5885e in Aws::Utils::Stream::ResponseStream::~ResponseStream (this=0x7f0e477fc2b0, __in_chrg=<optimized out>) at external/com_github_aws-sdk-cpp/aws-cpp-sdk-core/source/utils/stream/ResponseStream.cpp:54
#3  0x00007f0e786714a8 in Aws::S3Crt::Model::GetObjectResult::~GetObjectResult (this=0x7f0e477fc2b0, __in_chrg=<optimized out>) at external/com_github_aws-sdk-cpp/aws-cpp-sdk-s3-crt/include/aws/s3-crt/model/GetObjectResult.h:30
[...]

Issue #1732, opened 2021-08-10
TransferManager::DoSinglePartDownload() should flush the stream before marking the transfer completed


igorcanadi:
`TransferManager::DoSinglePartDownload()` marks the transfer as completed here: https://github.com/rockset/rocksdb-cloud/blob/8c0ff476d77c0df0af92b1466269404bdfb8e89e/cloud/aws/aws_s3.cc#L196

This happens before the underlying response stream is flushed. The response stream's ownership is tied to local variable `getObjectOutcome` and the stream is flushed only in `getObjectOutcome`'s destructor when it out of scope (here: https://github.com/aws/aws-sdk-cpp/blob/5543842bdd1bd0a318295a6f2247909883054b0e/aws-cpp-sdk-core/source/utils/stream/ResponseStream.cpp#L61).

This means that a code that uses `std::fstream` to download S3 object to a local file and does:
```
handle->WaitUntilFinished();
// do something with the file
```
is incorrect, because the file might not be flushed when `WaitUntilFinished()` returns.

This is similar to this 6 year old bug: https://github.com/aws/aws-sdk-cpp/issues/58

KaibaLopez:
Hi @igorcanadi ,
Yea thanks for bringing attention to this,and you're right, this is an ongoing issue with the SDK, I'll try to make it a priority for the team.

igorcanadi:
Hi @KaibaLopez any news on this issue? Thanks!

SergeyRyabinin:
Hi,
I think it should be fixed now by https://github.com/aws/aws-sdk-cpp/pull/2351
At least a flush() is called before a task completion notification.
Closing the issue, please submit a new one or re-open should you see more issues.

Best regards,
Sergey

Issue #58, opened 2015-12-08
FileStream should flush on network operation complete


cjhanks:
In gist I have code like this..

```
std::string filePath = "some/local/path";
std::shared_ptr<Aws::Http::HttpRequest>  request;
request->SetResponseStreamFactory(
        [&filePath, &offset]() -> Aws::IOStream* {
            constexpr auto OpenFlag = std::ios_base::out
                                    | std::ios_base::binary;
            Aws::IOStream* ofs =
                Aws::New<Aws::FStream>(ALLOCATION_TAG, filePath, OpenFlag);
            CHECK(ofs->good());
            CHECK(ofs->seekp(offset, ofs->beg));
            return ofs;
        });

std::shared_ptr<Aws::Http::HttpResponse> response(std::move(m_client->MakeRequest(*request)));

std::ifstream ifs(filePath); // ! DATA NOT YET FLUSHED TO FILE HERE
```

However if I artificially scope everything as follows...
```
std::string filePath = "some/local/path";
{ // BEGIN BOGUS SCOPE
std::shared_ptr<Aws::Http::HttpRequest>  request;
request->SetResponseStreamFactory(
        [&filePath, &offset]() -> Aws::IOStream* {
            constexpr auto OpenFlag = std::ios_base::out
                                    | std::ios_base::binary;
            Aws::IOStream* ofs =
                Aws::New<Aws::FStream>(ALLOCATION_TAG, filePath, OpenFlag);
            CHECK(ofs->good());
            CHECK(ofs->seekp(offset, ofs->beg));
            return ofs;
        });

std::shared_ptr<Aws::Http::HttpResponse> response(std::move(m_client->MakeRequest(*request)));
} // END BOGUS SCOPE 

std::ifstream ifs(filePath); // DATA IS NOW FLUSHED
```

This indicates to me a few things:
-  You're not flushing the file stream.
-  You're lazily deleting the Aws::IOstream object

This behavior is surprising and cost me ~45 minute of debug time.

This may /seem obvious/ in this contrived example.  But the entire AWS SDK is designed using *shared pointers practically everywhere*, and as such... it can be come less clear in complicated code what the precise lifetime of objects will be.

JonathanHenson:
Christopher,

The output stream is allocated and given to the response object. In this
case the response object is move only and the stream has the lifetime of
its containing response object. It will be cleaned up when the response is
deleted. Good catch on the missing flush, we'll get an update out to you as
soon as possible.

[...]

PR #2351, opened 2023-02-16
ResponseStream underlying buffer improvements

SergeyRyabinin:
*Issue #, if available:*

https://github.com/aws/aws-sdk-cpp/issues/58, https://github.com/aws/aws-sdk-cpp/issues/1732, https://github.com/aws/aws-sdk-cpp/issues/2319

The issue: 
* ResponseStream design can attach to an underlying buffer it does not own, a double destruction may occur;
* TransferManager marks a download as complete before the actual destination buffer is flushed.

*Description of changes:*
* Use [std::ios_base::register_callback](https://en.cppreference.com/w/cpp/io/ios_base/register_callback) to get notified if underlying stream is destructed;
* Do not flush in ResponseStream d-tor: it does not make sense, on the next line, proper underlying stream implementation should do it anyway in it's destructor (and should not if flush is not required in their design).
* Flush outcome's bodies in TransferManager before sending a complete notification;
* Update Result's Get<Aws::IOStream object> to be const (for the change above);
*Check all that applies:*

- [x] Linux
- [x] Windows
- [ ] Android
- [ ] MacOS
- [ ] IOS
- [ ] Other Platforms

### DIFF
--- a/src/aws-cpp-sdk-core/source/utils/stream/ResponseStream.cpp
+++ b/src/aws-cpp-sdk-core/source/utils/stream/ResponseStream.cpp
@@ -5,6 +5,7 @@
 
 #include <aws/core/utils/stream/ResponseStream.h>
 #include <aws/core/utils/memory/stl/AWSStringStream.h>
+#include <aws/core/utils/logging/LogMacros.h>
 
 #if defined(_GLIBCXX_FULLY_DYNAMIC_STRING) && _GLIBCXX_FULLY_DYNAMIC_STRING == 0 && defined(__ANDROID__)
 #include <aws/core/utils/stream/SimpleStreamBuf.h>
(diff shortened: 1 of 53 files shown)


src/aws-cpp-sdk-core/source/utils/stream/ResponseStream.cpp
```diff
@@ -43,12 +50,26 @@ ResponseStream& ResponseStream::operator=(ResponseStream&& toMove)
...
+    RegisterStream();
 
     return *this;
 }
 
+Aws::IOStream& ResponseStream::GetUnderlyingStream() const
+{
+    if (!m_underlyingStream)
+    {
+        assert(m_underlyingStream);
```
sdavtaker, line 65: Should we log first and fail assert after during debug?

sdavtaker (approved):

src/aws-cpp-sdk-core/source/utils/stream/ResponseStream.cpp
SergeyRyabinin: They are equal in messaging to the developer.
Except FATAL tracing is just a last resort if library is built in Release mode.
It is a really just "not crash", the next line will highly likely produce a memory leak reported by custom allocator set in the API Handle.

PR #2354, opened 2023-02-18
request compression

sdavtaker:
- refactor breakdown ClientTest in multiple files
- Reading Env and Config for UseRequestCompression setting

*Issue #, if available:*

*Description of changes:*

*Check all that applies:*

- [x] Linux
- [x] Windows
- [ ] Android
- [x] MacOS
- [ ] IOS
- [ ] Other Platforms


src/aws-cpp-sdk-core/include/aws/core/client/ClientConfiguration.h
```diff
@@ -265,7 +274,14 @@ namespace Aws
             Aws::String profileName;
 
             /**
-             * A helper function to read config value from env variable of aws profile config
+             * Use request compression when available.
+             * To use this feature, the service needs to provide the support, and the compression
+             * algorithms needs to be available at SDK build time.
+             */
+            Aws::Client::UseRequestCompression useRequestCompression;
```
sbiscigl, line 281: Should we have this at the client level or the request level? if at the request level users could choose on a per request basis whether or not to compress, and the client level once created there would be no way to opt out at request level i would think, unless we want it that way.

sbiscigl (approved):

PR #2361, opened 2023-02-23
Implementation of RequestCompression core for ZLIB

sdavtaker:
*Issue #, if available:*

*Description of changes:*
Adding core support for compressing requests using ZLIB for API methods that support it.

*Check all that applies:*

- [x] Linux
- [x] Windows
- [ ] Android
- [x] MacOS
- [ ] IOS
- [ ] Other Platforms

### DIFF
--- a/src/aws-cpp-sdk-core/include/aws/core/AmazonWebServiceRequest.h
+++ b/src/aws-cpp-sdk-core/include/aws/core/AmazonWebServiceRequest.h
@@ -7,15 +7,16 @@
 
 #include <aws/core/Core_EXPORTS.h>
 
-#include <aws/core/utils/memory/stl/AWSString.h>
-#include <aws/core/utils/UnreferencedParam.h>
-#include <aws/core/http/HttpTypes.h>
-#include <aws/core/http/HttpRequest.h>
-#include <aws/core/utils/memory/stl/AWSStreamFwd.h>
-#include <aws/core/utils/stream/ResponseStream.h>
+#include <aws/core/client/RequestCompression.h>
 #include <aws/core/auth/AWSAuthSigner.h>
 #include <aws/core/client/CoreErrors.h>
 #include <aws/core/endpoint/EndpointParameter.h>
+#include <aws/core/http/HttpRequest.h>
+#include <aws/core/http/HttpTypes.h>
+#include <aws/core/utils/UnreferencedParam.h>
+#include <aws/core/utils/memory/stl/AWSStreamFwd.h>
+#include <aws/core/utils/memory/stl/AWSString.h>
+#include <aws/core/utils/stream/ResponseStream.h>
 
 namespace Aws
 {
@@ -180,6 +181,9 @@ namespace Aws
 
         using EndpointParameters = Aws::Vector<Aws::Endpoint::EndpointParameter>;
         virtual EndpointParameters GetEndpointContextParams() const;
+
+        virtual Aws::Client::CompressionAlgorithm
+        GetSelectedCompressionAlgorithm(Aws::Client::RequestCompressionConfig) const { return Aws::Client::CompressionAlgorithm::NONE; }
 
     protected:
         /**
--- a/src/aws-cpp-sdk-core/include/aws/core/client/RequestCompression.h
+++ b/src/aws-cpp-sdk-core/include/aws/core/client/RequestCompression.h
@@ -1,0 +1,66 @@
+/**
+ * Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.
+ * SPDX-License-Identifier: Apache-2.0.
+ */
+
+#pragma once
+
+#include <aws/core/Core_EXPORTS.h>
+#include <aws/core/client/ClientConfiguration.h>
+#include <aws/core/utils/Outcome.h>
+#include <aws/core/utils/memory/stl/AWSStreamFwd.h>
+#include <aws/core/utils/memory/stl/AWSString.h>
+#include <aws/core/utils/memory/stl/AWSVector.h>
+
+using iostream_outcome =
+    Aws::Utils::Outcome<std::shared_ptr<Aws::IOStream>, bool>;
+
+namespace Aws {
+namespace Client {
+enum class CompressionAlgorithm { NONE, GZIP };
+
+/**
+ * Converts a compression Algorithms enum to String to be used as content-type
+ * header value when compressing a request.
+ * @param algorithm
+ * @return string with HTTP content type algorithm id
+ */
+Aws::String AWS_CORE_API
+GetCompressionAlgorithmId(const CompressionAlgorithm &algorithm);
+
+/**
+ * Request compression API
+ */
+class AWS_CORE_API RequestCompression final {
+public:
+  /**
+   * Select the best matching algorithm based in proposed ones, config, length
+   * of content and the available algorithms.
+   * @param proposedAlgorithms
+   * @param config
+   * @param payloadLength
+   * @return selected compression algorithm
+   */
+  CompressionAlgorithm
+  selectAlgorithm(const Aws::Vector<CompressionAlgorithm> &proposedAlgorithms,
+                  const Aws::Client::RequestCompressionConfig &config,
+                  const size_t payloadLength);
+  /**
+   * Compress a IOStream input using the requested algorithm.
+   * @param input
+   * @param algorithm
+   * @return IOStream compressed
+   */
+  iostream_outcome compress(std::shared_ptr<Aws::IOStream> input,
+                            const CompressionAlgorithm &algorithm) const;
+  /**
+   * Uncompress a IOStream input using the requested algorithm.
+   * @param input
+   * @param algorithm
+   * @return
+   */
+  iostream_outcome uncompress(std::shared_ptr<Aws::IOStream> input,
+                              const CompressionAlgorithm &algorithm) const;
+};
+} // namespace Client
+} // namespace Aws
--- a/src/aws-cpp-sdk-core/source/client/AWSClient.cpp
+++ b/src/aws-cpp-sdk-core/source/client/AWSClient.cpp
@@ -13,6 +13,7 @@
 #include <aws/core/client/ClientConfiguration.h>
 #include <aws/core/client/CoreErrors.h>
 #include <aws/core/client/RetryStrategy.h>
+#include <aws/core/client/RequestCompression.h>
 #include <aws/core/http/HttpClient.h>
 #include <aws/core/http/HttpClientFactory.h>
 #include <aws/core/http/HttpResponse.h>
@@ -124,7 +125,8 @@ AWSClient::AWSClient(const Aws::Client::ClientConfiguration& configuration,
     m_customizedUserAgent(!m_userAgent.empty()),
     m_hash(Aws::Utils::Crypto::CreateMD5Implementation()),
     m_requestTimeoutMs(configuration.requestTimeoutMs),
-    m_enableClockSkewAdjustment(configuration.enableClockSkewAdjustment)
+    m_enableClockSkewAdjustment(configuration.enableClockSkewAdjustment),
+    m_requestCompressionConfig(configuration.requestCompressionConfig)
 {
     AWSClient::SetServiceClientName("AWSBaseClient");
 }
@@ -143,7 +145,8 @@ AWSClient::AWSClient(const Aws::Client::ClientConfiguration& configuration,
     m_customizedUserAgent(!m_userAgent.empty()),
     m_hash(Aws::Utils::Crypto::CreateMD5Implementation()),
     m_requestTimeoutMs(configuration.requestTimeoutMs),
-    m_enableClockSkewAdjustment(configuration.enableClockSkewAdjustment)
+    m_enableClockSkewAdjustment(configuration.enableClockSkewAdjustment),
+    m_requestCompressionConfig(configuration.requestCompressionConfig)
 {
     AWSClient::SetServiceClientName("AWSBaseClient");
 }
@@ -696,6 +699,20 @@ void AWSClient::AddHeadersToRequest(const std::shared_ptr<Aws::Http::HttpRequest
     AddCommonHeaders(*httpRequest);
 }
 
+void AWSClient::AppendHeaderValueToRequest(const std::shared_ptr<HttpRequest> &httpRequest, const String header, const String value) const
+{
+    if (!httpRequest->HasHeader(header.c_str()))
+    {
+        httpRequest->SetHeaderValue(header, value);
+    }
+    else
+    {
+        Aws::String contentEncoding = httpRequest->GetHeaderValue(header.c_str());
+        contentEncoding.append(",").append(value);
+        httpRequest->SetHeaderValue(header, contentEncoding);
+    }
+}
+
 void AWSClient::AddChecksumToRequest(const std::shared_ptr<Aws::Http::HttpRequest>& httpRequest,
     const Aws::AmazonWebServiceRequest& request) const
 {
@@ -704,8 +721,8 @@ void AWSClient::AddChecksumToRequest(const std::shared_ptr<Aws::Http::HttpReques
     // Request checksums
     if (!checksumAlgorithmName.empty())
     {
-        // For non streaming payload, the resolved checksum location is always header.
-        // For streaming payload, the resolved checksum location depends on whether it's unsigned payload, we let AwsAuthSigner decide it.
+        // For non-streaming payload, the resolved checksum location is always header.
+        // For streaming payload, the resolved checksum location depends on whether it is an unsigned payload, we let AwsAuthSigner decide it.
         if (checksumAlgorithmName == "crc32")
         {
             if (request.IsStreaming())
@@ -874,14 +891,11 @@ Aws::String Aws::Client::GetAuthorizationHeader(const Aws::Http::HttpRequest& ht
     return authHeader.substr(signaturePosition + strlen(Aws::Auth::SIGNATURE) + 1);
 }
 
-void AWSClient::BuildHttpRequest(const Aws::AmazonWebServiceRequest& request,
-    const std::shared_ptr<HttpRequest>& httpRequest) const
+void AWSClient::BuildHttpRequest(const Aws::AmazonWebServiceRequest& request, const std::shared_ptr<HttpRequest>& httpRequest) const
 {
-    //do headers first since the request likely will set content-length as it's own header.
+    //do headers first since the request likely will set content-length as its own header.
     AddHeadersToRequest(httpRequest, request.GetHeaders());
     AddHeadersToRequest(httpRequest, request.GetAdditionalCustomHeaders());
-
-    AddChecksumToRequest(httpRequest, request);
 
     if (request.IsEventStreamRequest())
     {
@@ -889,9 +903,31 @@ void AWSClient::BuildHttpRequest(const Aws::AmazonWebServiceRequest& request,
     }
     else
     {
-        AddContentBodyToRequest(httpRequest, request.GetBody(), request.ShouldComputeContentMd5(), request.IsStreaming() && request.IsChunked() && m_httpClient->SupportsChunkedTransferEncoding());
+        //Check if compression is required
+        CompressionAlgorithm selectedCompressionAlgorithm =
+            request.GetSelectedCompressionAlgorithm(m_requestCompressionConfig);
+        if (Aws::Client::CompressionAlgorithm::NONE != selectedCompressionAlgorithm) {
+            Aws::Client::RequestCompression rc;
+            auto compressOutcome = rc.compress(request.GetBody(), selectedCompressionAlgorithm);
+
+            if (compressOutcome.IsSuccess()) {
+                Aws::String compressionAlgorithmId = Aws::Client::GetCompressionAlgorithmId(selectedCompressionAlgorithm);
+                AppendHeaderValueToRequest(httpRequest, CONTENT_ENCODING_HEADER, compressionAlgorithmId);
+                AddContentBodyToRequest(
+                    httpRequest, compressOutcome.GetResult(),
+                    request.ShouldComputeContentMd5(),
+                    request.IsStreaming() && request.IsChunked() &&
+                        m_httpClient->SupportsChunkedTransferEncoding());
+            } else {
+                AWS_LOGSTREAM_ERROR(AWS_CLIENT_LOG_TAG, "Failed to compress request, submitting uncompressed");
+                AddContentBodyToRequest(httpRequest, request.GetBody(), request.ShouldComputeContentMd5(), request.IsStreaming() && request.IsChunked() && m_httpClient->SupportsChunkedTransferEncoding());
+            }
+        } else {
+            AddContentBodyToRequest(httpRequest, request.GetBody(), request.ShouldComputeContentMd5(), request.IsStreaming() && request.IsChunked() && m_httpClient->SupportsChunkedTransferEncoding());
+        }
     }
 
+    AddChecksumToRequest(httpRequest, request);
     // Pass along handlers for processing data sent/received in bytes
     httpRequest->SetDataReceivedEventHandler(request.GetDataReceivedEventHandler());
     httpRequest->SetDataSentEventHandler(request.GetDataSentEventHandler());
--- a/src/aws-cpp-sdk-core/source/client/ClientConfiguration.cpp
+++ b/src/aws-cpp-sdk-core/source/client/ClientConfiguration.cpp
@@ -30,6 +30,8 @@ namespace Client
 static const char* CLIENT_CONFIG_TAG = "ClientConfiguration";
 static const char* USE_REQUEST_COMPRESSION_ENV_VAR = "USE_REQUEST_COMPRESSION";
 static const char* USE_REQUEST_COMPRESSION_CONFIG_VAR = "use_request_compression";
+static const char* REQUEST_MIN_COMPRESSION_SIZE_BYTES_ENV_VAR = "REQUEST_MIN_COMPRESSION_SIZE_BYTES";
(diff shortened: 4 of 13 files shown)


src/aws-cpp-sdk-core/source/client/RequestCompression.cpp
```diff
@@ -0,0 +1,76 @@
...
+Aws::Client::RequestCompression::compress(const Aws::String &input) const {
+#ifdef ENABLED_ZLIB_REQUEST_COMPRESSION
+  // Preparing output
+  std::shared_ptr<Aws::IOStream> output =
+      Aws::MakeShared<Aws::StringStream>("REQUEST_COMPRESSION");
+
+  // Prepare ZLIB to compress
+  int ret;
+  z_stream strm;
+  unsigned char in[ZLIB_CHUNK];
```
SergeyRyabinin, line 33: Be careful when allocating buffers on stack. Stack overflow may happen :-).

src/aws-cpp-sdk-core/source/client/RequestCompression.cpp
```diff
@@ -0,0 +1,76 @@
...
+#define SET_BINARY_MODE(file) setmode(fileno(file), O_BINARY)
+#else
+#define SET_BINARY_MODE(file)
+#endif
+// Defining zlib chunks to be 256k
+#define ZLIB_CHUNK 263144
+#endif
+
+Aws::Utils::Outcome<std::shared_ptr<Aws::IOStream>, bool>
+Aws::Client::RequestCompression::compress(const Aws::String &input) const {
```
SergeyRyabinin, line 24: why not Vector?

src/aws-cpp-sdk-core/source/client/RequestCompression.cpp
sdavtaker: It also is ignoring CUSTOM_MEMORY_ALLOCATORs here, I'm going to work on the memory model in a follow up PR. This is a first very early iteration of what would be at least 4 PRs. Just want to keep it functional.

src/aws-cpp-sdk-core/source/client/ClientConfiguration.cpp
```diff
@@ -81,12 +83,28 @@ void setLegacyClientConfigurationParameters(ClientConfiguration& clientConfig)
...
+      clientConfig.requestCompressionConfig.useRequestCompression = Aws::Client::UseRequestCompression::ENABLE;
+      AWS_LOGSTREAM_DEBUG(CLIENT_CONFIG_TAG, "Request Compression enabled");
     }
 
+    // Getting min request compression length
+    Aws::String minRequestCompressionString = Aws::Environment::GetEnv(REQUEST_MIN_COMPRESSION_SIZE_BYTES_ENV_VAR);
+    if (minRequestCompressionString.empty())
+    {
+      minRequestCompressionString = Aws::Config::GetCachedConfigValue(REQUEST_MIN_COMPRESSION_SIZE_BYTES_CONFIG_VAR);
+    }
```
SergeyRyabinin, line 99: there is a static method
ClientConfiguration::LoadConfigFromEnvOrProfile that fetches the value from our all config sources

src/aws-cpp-sdk-core/source/client/RequestCompression.cpp
```diff
@@ -0,0 +1,99 @@
...
+    return false;
+  // Compress
+  //  do {
+  // Fill the buffer
+  if (input.length() > ZLIB_CHUNK) {
+    // Skipping multi chunk for now
+    return false;
+  } else {
+    int flush = Z_FINISH;
+    memcpy(in, input.c_str(), input.length());
```
SergeyRyabinin, line 72: I would suggest 
```
memcpy(in, input.c_str(), std::min(input.length(), ZLIB_CHUNK);
```

src/aws-cpp-sdk-core/source/client/RequestCompression.cpp
sdavtaker: No longer in the stack, using unique pointers to new memory.

src/aws-cpp-sdk-core/source/client/RequestCompression.cpp
sdavtaker: Using IOStream everywhere, thats the only thing a request will pass.

src/aws-cpp-sdk-core/include/aws/core/client/RequestCompression.h
```diff
@@ -0,0 +1,35 @@
...
+#include <aws/core/utils/memory/stl/AWSStreamFwd.h>
+#include <aws/core/utils/memory/stl/AWSString.h>
+#include <aws/core/utils/memory/stl/AWSVector.h>
+
+namespace Aws {
+namespace Client {
+
+enum class CompressionAlgorithm { NONE, GZIP };
+
+Aws::String GetCompressionAlgorithmId(const CompressionAlgorithm &algorithm);
```
SergeyRyabinin, line 20: I think this needs AWS_CORE_API

src/aws-cpp-sdk-core/source/client/RequestCompression.cpp
sdavtaker: Good catch, adding.

src/aws-cpp-sdk-core/include/aws/core/client/RequestCompression.h
sdavtaker: Yes!

src/aws-cpp-sdk-core/include/aws/core/client/RequestCompression.h
```diff
@@ -0,0 +1,35 @@
...
+
+class AWS_CORE_API RequestCompression {
```
SergeyRyabinin, line 22: please add doxygen comments and some spacing between public APIs

src/aws-cpp-sdk-core/include/aws/core/client/RequestCompression.h
```diff
@@ -0,0 +1,35 @@
...
+public:
+  CompressionAlgorithm
+  selectAlgorithm(const Aws::Vector<CompressionAlgorithm> &proposedAlgorithms,
+                  const Aws::Client::RequestCompressionConfig &config,
+                  const size_t payloadLength);
+  Aws::Utils::Outcome<std::shared_ptr<Aws::IOStream>, bool>
```
SergeyRyabinin, line 28: might worth aliasing to some CompressionOutcome type alias.

src/aws-cpp-sdk-core/source/client/AWSClient.cpp
```diff
@@ -874,24 +877,53 @@ Aws::String Aws::Client::GetAuthorizationHeader(const Aws::Http::HttpRequest& ht
...
     if (request.IsEventStreamRequest())
     {
         httpRequest->AddContentBody(request.GetBody());
     }
     else
     {
-        AddContentBodyToRequest(httpRequest, request.GetBody(), request.ShouldComputeContentMd5(), request.IsStreaming() && request.IsChunked() && m_httpClient->SupportsChunkedTransferEncoding());
+        //Check if compression is required
+        if (Aws::Client::CompressionAlgorithm::NONE !=
+            request.GetSelectedCompressionAlgorithm(m_requestCompressionConfig)) {
```
SergeyRyabinin, line 894: store a result of
request.GetSelectedCompressionAlgorithm in a variable to avoid calling it 3 times

src/aws-cpp-sdk-core/include/aws/core/client/RequestCompression.h
SergeyRyabinin, line 22: it does not define destructor as virtual, thus mark this class as final

src/aws-cpp-sdk-core/source/client/AWSClient.cpp
```diff
@@ -874,24 +877,53 @@ Aws::String Aws::Client::GetAuthorizationHeader(const Aws::Http::HttpRequest& ht
...
+                      request.GetSelectedCompressionAlgorithm(m_requestCompressionConfig));
+
+                // Update/Add Header for compressionAlgorithmId
+                if (!httpRequest->HasHeader(CONTENT_ENCODING_HEADER)) {
+                    httpRequest->SetHeaderValue(CONTENT_ENCODING_HEADER, compressionAlgorithmId);
+                } else {
+                    Aws::String contentEncoding = httpRequest->GetHeaderValue(CONTENT_ENCODING_HEADER);
+                    contentEncoding.append(",").append(compressionAlgorithmId);
+                    httpRequest->SetHeaderValue(CONTENT_ENCODING_HEADER, contentEncoding);
+                }
```
SergeyRyabinin, line 910: It worth extracting this block into a (possibly inline) method on HttpRequest with a name such as "AppendHeaderValue".

src/aws-cpp-sdk-core/source/client/AWSClient.cpp
```diff
@@ -874,24 +877,53 @@ Aws::String Aws::Client::GetAuthorizationHeader(const Aws::Http::HttpRequest& ht
...
+
+                AddContentBodyToRequest(
+                    httpRequest, compressOutcome.GetResult(),
+                    request.ShouldComputeContentMd5(),
+                    request.IsStreaming() && request.IsChunked() &&
+                        m_httpClient->SupportsChunkedTransferEncoding());
+            } else {
+                AWS_LOGSTREAM_ERROR(AWS_CLIENT_LOG_TAG, "Failed to compress request, submitting uncompress");
```
SergeyRyabinin, line 918: uncompressed

src/aws-cpp-sdk-core/source/client/RequestCompression.cpp
```diff
@@ -0,0 +1,229 @@
...
+#if defined(MSDOS) || defined(OS2) || defined(WIN32) || defined(__CYGWIN__)
+#include <fcntl.h>
+#include <io.h>
+#define SET_BINARY_MODE(file) setmode(fileno(file), O_BINARY)
+#else
+#define SET_BINARY_MODE(file)
+#endif
+// Defining zlib chunks to be 256k
+#define ZLIB_CHUNK 263144
+#endif
```
SergeyRyabinin, line 21: please add helper closing comments i.e. 
```
#endif // defined(MSDOS) || defined(OS2) || defined(WIN32) || defined(__CYGWIN__)
```

src/aws-cpp-sdk-core/source/client/RequestCompression.cpp
```diff
@@ -0,0 +1,229 @@
...
+    AWS_LOGSTREAM_ERROR(
+        AWS_REQUEST_COMPRESSION_LOG_TAG,
+        "Compressed request requested in runtime without support for it");
+    return false;
+  }
+#else
+  // If there is no support to compress, always fail calls to this method.
+  AWS_LOGSTREAM_ERROR(
+      AWS_REQUEST_COMPRESSION_LOG_TAG,
+      "Compressed request requested in runtime without support for it");
```
SergeyRyabinin, line 115: worth tracing the algorithm that was requested

src/aws-cpp-sdk-core/source/client/RequestCompression.cpp
```diff
@@ -0,0 +1,229 @@
...
+                        "Compressing request of " << streamSize << " bytes.");
+
+    // Preparing output
+    std::shared_ptr<Aws::IOStream> output = Aws::MakeShared<Aws::StringStream>(
+        AWS_REQUEST_COMPRESSION_ALLOCATION_TAG);
+
+    // Prepare ZLIB to compress
+    int ret = Z_NULL;
+    z_stream strm;
+    std::unique_ptr<unsigned char[]> in(new unsigned char[ZLIB_CHUNK]);
```
SergeyRyabinin, line 61: use Aws::UniquePtr and Aws::MakeUnique

src/aws-cpp-sdk-core/source/client/RequestCompression.cpp
```diff
@@ -0,0 +1,229 @@
...
+    std::unique_ptr<unsigned char[]> out(new unsigned char[ZLIB_CHUNK]);
+    // Cleaning Zlib stream state flags
+    strm.zalloc = Z_NULL;
+    strm.zfree = Z_NULL;
+    strm.opaque = Z_NULL;
+
+    const int max_window_gzip = 31;
```
SergeyRyabinin, line 68: nit: magic constants are typically uppercase

src/aws-cpp-sdk-core/source/client/RequestCompression.cpp
```diff
@@ -0,0 +1,229 @@
...
+    return "gzip";
+  default:
+    return "";
+  }
+}
+
+Aws::Utils::Outcome<std::shared_ptr<Aws::IOStream>, bool>
+Aws::Client::RequestCompression::compress(
+    std::shared_ptr<Aws::IOStream> input,
+    const CompressionAlgorithm &algorithm) const {
```
SergeyRyabinin, line 42: mixing blank space in this file.
SDK typically uses curly braces on a new line and not that strict line width limit

src/aws-cpp-sdk-core/source/client/RequestCompression.cpp
```diff
@@ -0,0 +1,229 @@
...
+      return false;
+    // Compress
+    //  do {
+    // Fill the buffer
+    if (streamSize > ZLIB_CHUNK) {
+      // Skipping multi chunk for now
+      return false;
+    } else {
+      int flush = Z_FINISH;
+      input->read(reinterpret_cast<char *>(in.get()), streamSize);
```
SergeyRyabinin, line 83: Because of disabled exceptions, we are not able to verify that read succeeds 
Please check for badbit or [fail()](https://en.cppreference.com/w/cpp/io/basic_ios/fail) after read/get is called

src/aws-cpp-sdk-core/source/client/RequestCompression.cpp
```diff
@@ -0,0 +1,229 @@
...
+    AWS_LOGSTREAM_TRACE(AWS_REQUEST_COMPRESSION_LOG_TAG,
+                        "Uncompressing request of " << streamSize << " bytes.");
+
+    // Preparing output
+    std::shared_ptr<Aws::IOStream> output = Aws::MakeShared<Aws::StringStream>(
+        AWS_REQUEST_COMPRESSION_ALLOCATION_TAG);
+
+    // Prepare ZLIB to uncompress
+    int ret = Z_NULL;
+    z_stream strm;
```
SergeyRyabinin, line 142: it is a C POD type, I would use memset to zero-out it's initial state.
Especially to the fact that it has `reserved` field that you keep uninitialized that may bring troubles with future zlib versions.

src/aws-cpp-sdk-core/source/client/RequestCompression.cpp
```diff
@@ -0,0 +1,229 @@
...
+    AWS_LOGSTREAM_TRACE(AWS_REQUEST_COMPRESSION_LOG_TAG,
+                        "Compressing request of " << streamSize << " bytes.");
+
+    // Preparing output
+    std::shared_ptr<Aws::IOStream> output = Aws::MakeShared<Aws::StringStream>(
+        AWS_REQUEST_COMPRESSION_ALLOCATION_TAG);
+
+    // Prepare ZLIB to compress
+    int ret = Z_NULL;
+    z_stream strm;
```
SergeyRyabinin, line 60: it is a C POD type, I would use memset to zero-out it's initial state.
the same comment on the occurrence below.

src/aws-cpp-sdk-core/source/client/RequestCompression.cpp
SergeyRyabinin: there is also a dedicated macro in C CRT library to initialize a C POD struct.

src/aws-cpp-sdk-core/source/client/RequestCompression.cpp
```diff
@@ -0,0 +1,229 @@
...
+      strm.next_in = in.get();
+      do {
+        // Run deflate on buffers
+        strm.avail_out = ZLIB_CHUNK;
+        strm.next_out = out.get();
+
+        ret = deflate(&strm, flush);
+
+        // writing the output
+        unsigned output_size = ZLIB_CHUNK - strm.avail_out;
```
SergeyRyabinin, line 94: although I understand that this will never happen, but please add 
```
assert(ZLIB_CHUNK >= strm.avail_out)
```

src/aws-cpp-sdk-core/source/client/RequestCompression.cpp
```diff
@@ -0,0 +1,229 @@
...
+    // calculating stream size
+    input->seekg(0, input->end);
+    auto streamSize = input->tellg();
+    input->seekg(0, input->beg);
+
+    AWS_LOGSTREAM_TRACE(AWS_REQUEST_COMPRESSION_LOG_TAG,
+                        "Uncompressing request of " << streamSize << " bytes.");
+
+    // Preparing output
+    std::shared_ptr<Aws::IOStream> output = Aws::MakeShared<Aws::StringStream>(
```
SergeyRyabinin, line 137: this is only a nit now but something I would prefer to enforce
Because of disabled exceptions, we need to check the actual allocated pointer value to handle failed memory allocation cases.

src/aws-cpp-sdk-core/source/client/RequestCompression.cpp
```diff
@@ -0,0 +1,229 @@
...
+        AWS_REQUEST_COMPRESSION_ALLOCATION_TAG);
+
+    // Prepare ZLIB to uncompress
+    int ret = Z_NULL;
+    z_stream strm;
+    std::unique_ptr<unsigned char[]> in(new unsigned char[ZLIB_CHUNK]);
```
SergeyRyabinin, line 143: same as above,
please use Aws::UniquePtr and Aws::MakeUnique to be compatible with existing memory allocation model of the SDK.

src/aws-cpp-sdk-core/source/client/RequestCompression.cpp
```diff
@@ -0,0 +1,229 @@
...
+              AWS_REQUEST_COMPRESSION_LOG_TAG,
+              "Compressed request failed to inflate with code: Z_DATA_ERROR");
+          return false;
+        case Z_MEM_ERROR:
+          (void)inflateEnd(&strm);
+          AWS_LOGSTREAM_ERROR(
+              AWS_REQUEST_COMPRESSION_LOG_TAG,
+              "Compressed request failed to inflate with code: Z_MEM_ERROR");
+          return false;
+        }
```
SergeyRyabinin, line 193: what if ret does not fall into any of the switch statement?

src/aws-cpp-sdk-core/source/client/ClientConfiguration.cpp
sdavtaker: I'm using it for the other config option, but for minRequestCompression which is an integer value it wouldn't pass the validator.

src/aws-cpp-sdk-core/source/client/RequestCompression.cpp
sdavtaker: Actually, no longer using memcpy here

src/aws-cpp-sdk-core/include/aws/core/client/RequestCompression.h
sdavtaker: done

src/aws-cpp-sdk-core/include/aws/core/client/RequestCompression.h
sdavtaker: done

src/aws-cpp-sdk-core/source/client/AWSClient.cpp
sdavtaker: done

src/aws-cpp-sdk-core/include/aws/core/client/RequestCompression.h
sdavtaker: done with final

src/aws-cpp-sdk-core/source/client/RequestCompression.cpp
sdavtaker: Then it is not an error.

src/aws-cpp-sdk-core/source/client/RequestCompression.cpp
sdavtaker: Adding

src/aws-cpp-sdk-core/source/client/RequestCompression.cpp
sdavtaker: Isn't better to `= {}` in this particular case?

src/aws-cpp-sdk-core/source/client/RequestCompression.cpp
sdavtaker: fixed

src/aws-cpp-sdk-core/source/client/RequestCompression.cpp
sdavtaker: Added

src/aws-cpp-sdk-core/source/client/AWSClient.cpp
sdavtaker: extracted into AwsClient to keep it consistent with AddHeaders and AddBody

src/aws-cpp-sdk-core/include/aws/core/AmazonWebServiceRequest.h
```diff
@@ -7,15 +7,16 @@
 
 #include <aws/core/Core_EXPORTS.h>
 
-#include <aws/core/utils/memory/stl/AWSString.h>
-#include <aws/core/utils/UnreferencedParam.h>
-#include <aws/core/http/HttpTypes.h>
-#include <aws/core/http/HttpRequest.h>
-#include <aws/core/utils/memory/stl/AWSStreamFwd.h>
-#include <aws/core/utils/stream/ResponseStream.h>
+#include "aws/core/client/RequestCompression.h"
```
sbiscigl, line 10: nitpick: use `<` instead of `"`

src/aws-cpp-sdk-core/source/client/RequestCompression.cpp
```diff
@@ -0,0 +1,284 @@
...
+
+        auto out = Aws::MakeUniqueArray<unsigned char>(ZLIB_CHUNK, AWS_REQUEST_COMPRESSION_ALLOCATION_TAG);
+        if(!out)
+        {
+           AWS_LOGSTREAM_ERROR(AWS_REQUEST_COMPRESSION_LOG_TAG, "Failed to allocate out buffer while compressing")
+           return false;
+        }
+
+        // Cleaning Zlib stream state flags
+        strm.zalloc = Z_NULL;
```
sbiscigl, line 79: could make a RAII container for this like

```
class ZStreamRAIIContainer {
public:
  ZStreamRAIIContainer() {
     strm = {};
  }
  
  ~ZStreamRAIIContainer() {
    strm.zalloc = Z_NULL;
    strm.zfree = Z_NULL;
    strm.opaque = Z_NULL;
  }

private:
 z_stream strm;
}
```

that way we wouldnt need to worry about forgetting to clean up if we moved this.

sbiscigl (approved):

src/aws-cpp-sdk-core/source/client/RequestCompression.cpp
```diff
@@ -0,0 +1,284 @@
...
+        {
+           return false;
+        }
+
+        //Adding one to the stream size counter to account for the EOF marker.
+        streamSize++;
+        size_t toRead;
+        // Compress
+        do {
+            toRead = (streamSize < ZLIB_CHUNK)?streamSize:ZLIB_CHUNK;
```
SergeyRyabinin, line 96: std::min ?

src/aws-cpp-sdk-core/source/client/RequestCompression.cpp
```diff
@@ -0,0 +1,284 @@
...
+            strm.avail_in = toRead;
+            strm.next_in = in.get();
+            do
+            {
+                // Run inflate on buffers
+                strm.avail_out = ZLIB_CHUNK;
+                strm.next_out = out.get();
+
+                ret = inflate(&strm, Z_NO_FLUSH);
+                // Catch errors
```
SergeyRyabinin, line 237: >>what if ret does not fall into any of the switch statement?

>Then it is not an error.

maybe worth catching all other unhanded errors

```
if (ret < 0)
{
    switch (ret)
    {
    case Z_NEED_DICT:
        AWS_LOGSTREAM_ERROR(AWS_REQUEST_COMPRESSION_LOG_TAG, "Compressed request failed to inflate with code: Z_NEED_DICT");
        return false;
    case Z_DATA_ERROR:
        AWS_LOGSTREAM_ERROR(AWS_REQUEST_COMPRESSION_LOG_TAG, "Compressed request failed to inflate with code: Z_DATA_ERROR");
        return false;
    case Z_MEM_ERROR:
        (void)inflateEnd(&strm);
        AWS_LOGSTREAM_ERROR(AWS_REQUEST_COMPRESSION_LOG_TAG, "Compressed request failed to inflate with code: Z_MEM_ERROR");
        return false;
    case default:
        (void)inflateEnd(&strm);
        AWS_LOGSTREAM_ERROR(AWS_REQUEST_COMPRESSION_LOG_TAG, "Compressed request failed to inflate with unknown error code: " << ret);
        return false;
    }
}
```

SergeyRyabinin:
That looks great now, thank you for updating all comments.

SergeyRyabinin (approved):

src/aws-cpp-sdk-core/source/client/RequestCompression.cpp
sdavtaker: Tested it, it will catch also some internal non-error status if only checking it is not Z_OK and break on good cases.

PR #1996, opened 2022-07-12
[aws-cpp-sdk-core] shut down CRT logging after shutting down CRT

grrtrr:
*Issue #, if available:*
Resolves #1995.

*Description of changes:*
Shut down CRT logging only after the CRT subsystem has been shut down, since logging does happen during shutdown.
Currently CRT logging is shut down first, which causes undefined behaviour.

*Check all that applies:*
- [X] Did a review by yourself.
- [X] Added proper tests to cover this PR. (If tests are not applicable, explain.)
   Ran before/after tests with `clang` TSAN. Confirmed that this fixes the race condition documented in #1995.
- [X] Checked if this PR is a breaking (APIs have been changed) change.
- [X] Checked if this PR will _not_ introduce cross-platform inconsistent behavior.
- [X] Checked if this PR would require a ReadMe/Wiki update.

Check which platforms you have built SDK on to verify the correctness of this PR.
- [X] Linux
- [ ] Windows
- [ ] Android
- [ ] MacOS
- [ ] IOS
- [ ] Other Platforms

By submitting this pull request, I confirm that you can use, modify, copy, and redistribute this contribution, under the terms of your choice.

Issue #1995, opened 2022-07-12
[aws-cpp-sdk-core] shut down CRT logging after shutting down CRT


grrtrr:
### Describe the bug

The `ShutdownAPI` call in `source/Aws.cpp` first calls `Aws::Utils::Logging::ShutdownCRTLogging()` (if logging was enabled), and then `Aws::CleanupCrt()`.

This causes issues on shutdown of the CRT subsystem - threads that are in the process of being shut down may still emit logging messages to the (already shut down) `CRTLogSystemInterface`, leading to undefined behaviour and a race condition (please see below).

### Expected Behavior

`ShutdownAPI` cleanly shuts down all involved components.

### Current Behavior

When running test using the clang TSAN analyzer, one can observe the above described race conditions:
```
WARNING: ThreadSanitizer: data race (pid=9800)
  Write of size 8 at 0x7b1000000e20 by main thread:
    #0 free ??:? (evolved_log_reader_integration_test+0x5c518)
    #1 std::_Sp_counted_ptr_inplace<Aws::Utils::Logging::DefaultCRTLogSystem, std::allocator<Aws::Utils::Logging::DefaultCRTLogSystem>, (__gnu_cxx::_Lock_policy)2>::_M_destroy() ??:? (libcloud_Saws_Slibaws_Uinit
.so+0x896e)
    #2 Aws::Utils::Logging::ShutdownCRTLogging() ??:? (libexternal_Scom_Ugithub_Uaws-sdk-cpp_Slibaws-core.so+0x1675b2)
    #3 Aws::ShutdownAPI(Aws::SDKOptions const&) ??:? (libexternal_Scom_Ugithub_Uaws-sdk-cpp_Slibaws-core.so+0x7dbe8)
    #4 av::cloud::aws::AWSInit::~AWSInit() ??:? (libcloud_Saws_Slibaws_Uinit.so+0x6f62)
    #5 cxa_at_exit_wrapper(void*) tsan_interceptors_posix.cpp:? (our_integration_test+0x9a46f)

  Previous read of size 8 at 0x7b1000000e20 by thread T2:
    #0 s_main_loop epoll_event_loop.c:? (libexternal_Saws-c-io_Slibaws-c-io.so+0x1d84a)
    #1 thread_fn thread.c:? (libexternal_Saws-c-common_Slibaws-c-common.so+0x2b680)

  Thread T2 (tid=9812, running) created by main thread at:
[...]

PR #2410, opened 2023-03-29
issue-2409: CRT logging thread runs beyond usable lifetime of CRT and dependents

### LINKED ISSUE #2409: Loggers run beyond the usable lifetime of CRT
### Describe the bug

#1996 a change was made to shutdown the CRT before shutting down both loggers.  In general, the shutdown sequence should always be the exact reverse of the init sequence to avoid lifetime issues, otherwise there is an architectural issue to solve.  If the CRT logs during shutdown, it needs to be smart enough not to fall over if there is no logger.  I believe the fix for the issue observed was performed at the wrong code layer.

If one is using CUSTOM_MEMORY_MANAGEMENT, when the CRT is shutdown, the custom memory allocator is invalidated.  At that time, the logging thread is still active and may call for an allocation.

The following assertion from aws-c-common demonstrates this behavior.  Placed log message wrappers around the shutdown calls.  There is no message indicating logging was shut down.  "Shutting Down" is logged immediately before we call ShutdownAPI.  The fix is to restore the original ordering.

```
[INFO] Libas3Test.test[51845]: main(SdkMgr.cpp:89):  AwsSdkCpp Shutting Down
Fatal error condition occurred in /opt/pkg-cache/packages/Aws-c-common/Aws-c-common-0.2.x.141369.0/generic-flavor/src/source/allocator.c:121: allocator != ((void *)0)
Exiting Application
################################################################################
Stack trace:
################################################################################
lib/libaws-c-common.so.1(aws_backtrace_print+0x4f) [0x7f5d3a33169f]
lib/libaws-c-common.so.1(aws_fatal_assert+0x44) [0x7f5d3a31d1d4]
lib/libaws-c-common.so.1(aws_mem_acquire+0x37) [0x7f5d3a31b5b7]
Libas3Test.test(_ZN3Aws3Crt12StlAllocatorIcE8allocateEmPKv+0x33) [0x65bc4d]
Libas3Test.test(_ZNSt16allocator_traitsIN3Aws3Crt12StlAllocatorIcEEE8allocateERS3_m+0x31) [0x65796b]
Libas3Test.test(_ZNSt7__cxx1112basic_stringIcSt11char_traitsIcEN3Aws3Crt12StlAllocatorIcEEE9_M_createERmm+0x110) [0x65bd68]
Libas3Test.test(_ZNSt7__cxx1112basic_stringIcSt11char_traitsIcEN3Aws3Crt12StlAllocatorIcEEE9_M_mutateEmmPKcm+0x95) [0x65c011]
Libas3Test.test(_ZNSt7__cxx1112basic_stringIcSt11char_traitsIcEN3Aws3Crt12StlAllocatorIcEEE10_M_replaceEmmPKcm+0x351) [0x657d67]
lib/libaws-cpp-sdk-core.so(+0xfd4f7) [0x7f5d3a4474f7]
lib/libaws-cpp-sdk-core.so(+0xfdabf) [0x7f5d3a447abf]
lib/libaws-cpp-sdk-core.so(_ZNSt6thread11_State_implINS_8_InvokerISt5tupleIJPFvPN3Aws5Utils7Logging16DefaultLogSystem22LogSynchronizationDataERKSt10shared_ptrISoERKNSt7__cxx1112basic_stringIcSt11char_traitsIcENS3_3Crt12StlAllocatorIcEEEEbES8liib/libstdc++.so.6(+0xe71a0) [0x7f5d388bb1a0] 
```

### Expected Behavior

The CRT logger's lifetime cannot exceed the CRT's lifetime.

No assertion.

### Current Behavior

The logger thread is running beyond the CRT lifetime and causing an error.

### Reproduction Steps

Difficult to reproduce, timing related.

### Possible Solution

Revert the change.

Should provide a unit test that does init/shutdown many times in a row, in the same executable, to prove the behavior is correct.

### AWS CPP SDK version used

1.11.47

### Compiler and Version used

9.5.0

Linux (AL2)

jeking3:
The CRT logging thread was outliving the CRT and dependents, including aws-c-common.  If one is using CUSTOM_MEMORY_MANAGEMENT then the aws-c-common allocator will be NULL, but the logging thread is still running, and may try to allocate, causing an assert.

In general, shutdown should be the reverse sequence of init to avoid situations like this.  #1996 may have fixed something, but the fix is in the wrong place.

*Issue #, if available:*

This fixes #2409 

*Description of changes:*

1. Restore the correct shutdown sequence, an inverse of the init sequence.
2. Clean up cJSON for good measure, properly.
3. Add a test that proves it is okay to Init and Shutdown many times in the same executable run.

*Check all that applies:*

- [X] Linux
- [ ] Windows
- [ ] Android
- [X] MacOS
- [ ] IOS
- [ ] Other Platforms

### DIFF
--- a/src/aws-cpp-sdk-core/source/Aws.cpp
+++ b/src/aws-cpp-sdk-core/source/Aws.cpp
@@ -157,19 +157,21 @@ namespace Aws
         Aws::Monitoring::CleanupMonitoring();
         Aws::Internal::CleanupEC2MetadataClient();
         Aws::Net::CleanupNetwork();
+        cJSON_AS4CPP_InitHooks(nullptr);
         Aws::CleanupEnumOverflowContainer();
         Aws::Http::CleanupHttp();
         Aws::Utils::Crypto::CleanupCrypto();
 
         Aws::Config::CleanupConfigAndCredentialsCacheManager();
 
-        Aws::Client::CoreErrorsMapper::CleanupCoreErrorsMapper();
-        Aws::CleanupCrt();
         if (options.loggingOptions.logLevel != Aws::Utils::Logging::LogLevel::Off)
         {
             Aws::Utils::Logging::ShutdownCRTLogging();
             Aws::Utils::Logging::ShutdownAWSLogging();
         }
+
+        Aws::Client::CoreErrorsMapper::CleanupCoreErrorsMapper();
+        Aws::CleanupCrt();
 #ifdef USE_AWS_MEMORY_MANAGEMENT
         if(options.memoryManagementOptions.memoryManager)
         {
--- a/tests/aws-cpp-sdk-core-tests/RunTests.cpp
+++ b/tests/aws-cpp-sdk-core-tests/RunTests.cpp
@@ -13,6 +13,8 @@
 #include <sys/stat.h>
 #endif
 
+Aws::SDKOptions options;
+
 int main(int argc, char** argv)
 {
 #if defined(HAS_UMASK)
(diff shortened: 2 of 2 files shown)


jeking3:
This is not a valid fix for the issue; it re-introduces the instability described in #1995 and fixed in #1996.  We need another way to fix this for CUSTOM_MEMORY_MANAGEMENT.

PR #2432, opened 2023-04-11
S3Client::GeneratePresignedUrl return invalid url when key contains spaces

wcw84:
*Issue #, if available:*
The issue is here: https://github.com/aws/aws-sdk-cpp/issues/2396

*Description of changes:*
Change urlencode twice to urlencode once.

*Check all that applies:*
- [ *] Did a review by yourself.
- [ *] Added proper tests to cover this PR. (If tests are not applicable, explain.)
- [ *] Checked if this PR is a breaking (APIs have been changed) change.
- [ *] Checked if this PR will _not_ introduce cross-platform inconsistent behavior.
- [ *] Checked if this PR would require a ReadMe/Wiki update.

- [ *] Linux
- [ ] Windows
- [ ] Android
- [ ] MacOS
- [ ] IOS
- [ ] Other Platforms

### DIFF
--- a/generated/src/aws-cpp-sdk-s3/source/S3Client.cpp
+++ b/generated/src/aws-cpp-sdk-s3/source/S3Client.cpp
@@ -2027,9 +2027,7 @@ Aws::String S3Client::GeneratePresignedUrl(const Aws::String& bucket,
         return {};
     }
     Aws::Endpoint::AWSEndpoint& endpoint = computeEndpointOutcome.GetResult();
-    URI uri(endpoint.GetURL());
-    uri.SetPath(uri.GetPath() + "/" + key);
-    endpoint.SetURL(uri.GetURIString());
+    endpoint.AddPathSegments(key);
     return AWSClient::GeneratePresignedUrl(endpoint, method, customizedHeaders, expirationInSeconds);
 }
 
--- a/tools/code-generation/generator/src/main/resources/com/amazonaws/util/awsclientgenerator/velocity/cpp/s3/S3ClientSource.vm
+++ b/tools/code-generation/generator/src/main/resources/com/amazonaws/util/awsclientgenerator/velocity/cpp/s3/S3ClientSource.vm
@@ -121,9 +121,7 @@ Aws::String ${className}::GeneratePresignedUrl(const Aws::String& bucket,
         return {};
     }
     Aws::Endpoint::AWSEndpoint& endpoint = computeEndpointOutcome.GetResult();
-    URI uri(endpoint.GetURL());
-    uri.SetPath(uri.GetPath() + "/" + key);
-    endpoint.SetURL(uri.GetURIString());
+    endpoint.AddPathSegments(key);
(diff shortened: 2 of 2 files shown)


generated/src/aws-cpp-sdk-s3/source/S3Client.cpp
```diff
@@ -2027,9 +2027,7 @@ Aws::String S3Client::GeneratePresignedUrl(const Aws::String& bucket,
         return {};
     }
     Aws::Endpoint::AWSEndpoint& endpoint = computeEndpointOutcome.GetResult();
-    URI uri(endpoint.GetURL());
-    uri.SetPath(uri.GetPath() + "/" + key);
-    endpoint.SetURL(uri.GetURIString());
+    endpoint.SetURL(endpoint.GetURL() + "/" + key);
```
sbiscigl, line 2030: Hey Thanks for taking the time to contribute! Was able to replicate and take a look at this solution. just a few things about this.

* thinking we should go with `endpoint.AddPathSegments(key);` instead of this as to avoid creating the temporary URI object all together. Also this is how we handle this in [`GetObject` correctly](https://github.com/aws/aws-sdk-cpp/blob/main/generated/src/aws-cpp-sdk-s3/source/S3Client.cpp#L1028) so we should follow the same pattern. The reason the problem arose intially is because these diverged, and this results in a accidental double encoding when calling `uri.GetURIString()`.

i.e.

```
Aws::Endpoint::AWSEndpoint& endpoint = computeEndpointOutcome.GetResult();
endpoint.AddPathSegments(key);
return AWSClient::GeneratePresignedUrl(endpoint, method, customizedHeaders, expirationInSeconds);
```

* you will also need to update the [code-gen file that has this line in it](https://github.com/aws/aws-sdk-cpp/blob/main/tools/code-generation/generator/src/main/resources/com/amazonaws/util/awsclientgenerator/velocity/cpp/s3/S3ClientSource.vm#L124-L126). We run a [codegenerator](https://github.com/aws/aws-sdk-cpp/blob/main/docs/CMake_Parameters.md#regenerate_clients) to generate aws clients based on modeled json. Since S3 is so specialized though, we have singular location for it.

sbiscigl (changes requested):

generated/src/aws-cpp-sdk-s3/source/S3Client.cpp
wcw84: Hey, already made the new commit following your Suggestion.

sbiscigl: nice let me run some tests and make sure its all good, then we should be able to merge

sbiscigl (changes requested):
Hey just checking in, please squash your commits and we're good to merge

wcw84:
> Hey just checking in, please squash your commits and we're good to merge

Hey, just finished, please check it

sbiscigl:
counds good, running one last test, then will give approval

sbiscigl (approved):

SergeyRyabinin (approved):

Issue #2056, opened 2022-08-25
Built aws-sdk-cpp with multi-threaded (\MT) runtime library (including aws-crt-cpp)


raphasampaio:
### Describe the bug

I'm trying to build aws-sdk-cpp on windows with the multi-threaded (\MT) runtime library. 

### Expected Behavior

I want all the projects to be built with the runtime library \MT configuration.

### Current Behavior

The aws-cpp projects are correctly built with the \MT flag; however, the aws-c projects ([https://github.com/awslabs/aws-crt-cpp](https://github.com/awslabs/aws-crt-cpp)) are built with multi-threaded DLL (\MD). 

### Reproduction Steps

I'm running the following cmake command:

```bash
cmake ..\aws -DBUILD_SHARED_LIBS=OFF -DFORCE_SHARED_CRT=OFF -DSTATIC_CRT=ON -DBUILD_ONLY="s3;athena;transfer"
```

### Possible Solution

Inherit those flags to the aws-crt-cpp project

### Additional Information/Context

Possible related PR https://github.com/aws/aws-sdk-cpp/pull/1087

### AWS CPP SDK version used

v1.9.251

### Compiler and Version used

Visual Studio

### Operating System and version

Windows 10

jmklix:
We are looking into why this flag isn't being forwarded correctly.

PR #2434, opened 2023-04-12
Set static/shared msvc runtime flags using a CMake property instead of overwriting the flags

TinyTinni:
**Issue #, if available:** #2056

**Description of changes:**
This PR changes how the flags for the MSVC C++ runtime linkage are set. Before this PR, the CMAKE_CXX_FLAGS were changed. WIth CMake 3.15, there exists the property [CMAKE_MSVC_RUNTIME_LIBRARY](https://cmake.org/cmake/help/latest/variable/CMAKE_MSVC_RUNTIME_LIBRARY.html). Based on the options provided by this SDK, this property is set instead of changing the flags.

Defining the property so early has 2 advantages:
- it is close to the option handing
- from the CMake doc `This variable is used to initialize the MSVC_RUNTIME_LIBRARY property on all targets as they are created` which results in that all targets have the right runtime. This solves the problem described in #2056

As you can read in the CMake documentation, any non msvc builds are not affected by defining the property. 
> The value is ignored on compilers not targeting the MSVC ABI, but an unsupported value will be rejected as an error when using a compiler targeting the MSVC ABI.

Please ask and/or edit, if I missed something.

**Check all that applies:**
_Changes in the buildsystem. Test would be, if the project compiled or not. I cannot see any CI builds in the .guthub/workflows. If I should add a build target, please let me know._ 
_Requires CMake 3.15 instead of 3.12. CMake 3.15 is from 2019, so I don't see it as a big problem_
_Only Windows is affected._
_No Readme changes required. Options are still the same with the same semantic._

- [ ] Linux
- [x] Windows 
_Build and unit-tested locally with dynamic and static linkage_
- [ ] Android
- [ ] MacOS
- [ ] IOS
- [ ] Other Platforms

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -2,7 +2,7 @@
 # SPDX-License-Identifier: Apache-2.0.
 #
 
-cmake_minimum_required(VERSION 3.13 FATAL_ERROR)
+cmake_minimum_required(VERSION 3.15 FATAL_ERROR)
 
 option(LEGACY_BUILD "If enabled, the SDK will use 1.11.0 version of CMake files to build" ON)
 if (LEGACY_BUILD)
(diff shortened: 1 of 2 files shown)


TinyTinni:
Hi,
any updates on this? Do you need any more input from me?

PR #2497, opened 2023-05-19
Embed SVGs into html

SergeyRyabinin:
*Issue #, if available:*

*Description of changes:*

*Check all that applies:*

- [x] Linux
- [ ] Windows
- [ ] Android
- [ ] MacOS
- [ ] IOS
- [ ] Other Platforms

### DIFF
--- a/docs/doxygen/doxygen_wrapper.py
+++ b/docs/doxygen/doxygen_wrapper.py
@@ -2,6 +2,8 @@
 
 # Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.
 # SPDX-License-Identifier: Apache-2.0.
+
+import base64
 import fileinput
 import os
 import re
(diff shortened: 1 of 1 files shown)


docs/doxygen/doxygen_wrapper.py
```diff
@@ -18,6 +20,11 @@
 
 
 class DoxygenWrapper(object):
+    SVG_IMAGE_PATTERN = re.compile(
+        """ src="(?P<svg_filename>[a-z0-9_]+\\.svg)" """
```
DmitriyMusatkin, line 24: are all filenames guaranteed to be lower case? no other weird chars besides _?

docs/doxygen/doxygen_wrapper.py
```diff
@@ -167,6 +174,35 @@ def _cleanup_dir_by_extension(child_dir: Path, extension: str):
...
+        svg_content = "data:image/svg+xml;base64,"
+        with open(svg_filename) as svg_file:
+            file_content = svg_file.read()
+            svg_content += base64.b64encode(file_content.encode()).decode()
+        return svg_content
+
+    def _embed_svg_images(self, html_dir):
+        p = Path(html_dir)
+        html_files = p.glob(f"**/*.html")
+        files_to_remove = set()
```
DmitriyMusatkin, line 188: is it common for doc to share same file across many html pages? i.e. are we increasing size by a lot by instead embedding image into every page?

DmitriyMusatkin (approved):

docs/doxygen/doxygen_wrapper.py
SergeyRyabinin: Yes, at least at this point from doxygen.
The key is to stay consistent and don't break pages (i.e. if file is not embedded - it won't get deleted).

docs/doxygen/doxygen_wrapper.py
SergeyRyabinin: So far we get a single SVG per single Class doc page.
I haven't seen an example of generated shared images (non-shared non-generated are in the SVG_TO_KEEP list).

PR #2516, opened 2023-05-31
add smithy tracing interfaces

sbiscigl:
*Description of changes:*

Creates Interfaces in a new smithy namespace to closer align with the future of tracing in aws sdks.

*Check all that applies:*

- [x] Linux
- [x] Windows
- [] Android
- [x] MacOS
- [] IOS
- [] Other Platforms


SergeyRyabinin (commented):
Please add docstrings/comments to the new classes.

src/aws-cpp-sdk-core/include/smithy/tracing/RootTraceProbe.h
```diff
@@ -0,0 +1,33 @@
...
+             * A specific Implementation of TraceProbe that keeps track
+             * of a list of TraceProbes, that will dispatch a list of events
+             * to each probe. The intention is that this class will be used
+             * as the root entry point to emitting traces to a list of
+             * arbitrary backends.
+             */
+            class SMITHY_API RootTraceProbe : public TraceProbe {
+            public:
+                void dispatchEvents(const Aws::Vector<TraceEvent> &events) override;
+
```
DmitriyMusatkin, line 25: whats the ownership model? does it make sense for this to be unique pointer thats moved to the class

src/aws-cpp-sdk-core/include/smithy/Smithy_EXPORTS.h
```diff
@@ -0,0 +1,28 @@
...
+ * SPDX-License-Identifier: Apache-2.0.
+ */
+#pragma once
+
+#ifdef _MSC_VER
+//disable windows complaining about max template size.
+    #pragma warning (disable : 4503)
+#endif // _MSC_VER
+
+#if defined (USE_WINDOWS_DLL_SEMANTICS) || defined (_WIN32)
```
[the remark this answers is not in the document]
sbiscigl: We decided to make smithy a different set of exports in the event that it would be split off into its own library, and we need the win annotations for creating the lib. It [largely copies the export model that we already have for core](https://github.com/aws/aws-sdk-cpp/blob/main/src/aws-cpp-sdk-core/include/aws/core/Core_EXPORTS.h#L10-L32).

src/aws-cpp-sdk-core/include/smithy/tracing/RootTraceProbe.h
```diff
@@ -0,0 +1,33 @@
...
+ * SPDX-License-Identifier: Apache-2.0.
+ */
+#pragma once
+
+#include <smithy/Smithy_EXPORTS.h>
+#include <smithy/tracing/TraceEvent.h>
+#include <smithy/tracing/TraceProbe.h>
+#include <aws/core/utils/memory/stl/AWSVector.h>
+
+namespace smithy {
```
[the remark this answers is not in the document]
sbiscigl: this is the scheme we are going with for smithy components.

src/aws-cpp-sdk-core/include/smithy/Smithy_EXPORTS.h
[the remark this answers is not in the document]
sbiscigl: talked and since these are new exports we will align with how [CRT does visibility on linux](https://github.com/awslabs/aws-c-common/blob/9f36d07189435db27f746c10e39ba58511fadd0d/include/aws/common/exports.h#L20-L24)

src/aws-cpp-sdk-core/include/smithy/tracing/ScopedTracer.h
```diff
@@ -0,0 +1,96 @@
...
+            public:
+                ScopedTracer(std::shared_ptr<TraceProbe> probe) :
+                    m_currentSpanScope(std::make_shared<TraceSpan>(std::move(probe))) {}
+
+                ScopedTracer(ScopedTracer &parent) :
+                    m_currentSpanScope(parent.getCurrentSpan()->newChildSpan()) {}
+
+                ~ScopedTracer() = default;
+
+                void CountEvent(Aws::String &&eventName,
```
DmitriyMusatkin, line 29: nit: api names read a bit awkward since they are nouns. maybe OnCountEvent(), etc...

src/aws-cpp-sdk-core/include/smithy/tracing/TraceMessageEvent.h
```diff
@@ -0,0 +1,36 @@
...
+        namespace tracing {
+            /**
+             * A Trace event that is a message event.
+             */
+            class SMITHY_API TraceMessageEvent : public TraceEventDataInterface {
+            private:
+                Aws::String message;
+            public:
+                explicit TraceMessageEvent(Aws::String message) :
+                    message(std::move(message)) {}
```
DmitriyMusatkin, line 24: nit: string is not passed by ref, what is move trying to do?

sbiscigl: [clang tidy recommended pass by move](https://clang.llvm.org/extra/clang-tidy/checks/modernize/pass-by-value.html)
 copying from their reasoning

With move semantics added to the language and the standard library updated with move constructors added for many types it is now interesting to take an argument directly by value, instead of by const-reference, and then copy. This check allows the compiler to take care of choosing the best way to construct the copy.

The transformation is usually beneficial when the calling code passes an rvalue and assumes the move construction is a cheap operation.`

src/aws-cpp-sdk-core/include/smithy/tracing/ScopedTracer.h
sbiscigl: yeah fair, let me change

src/aws-cpp-sdk-core/include/smithy/tracing/RootTraceProbe.h
sbiscigl: good call, making it unique ptr so that the root probe will own it

SergeyRyabinin (approved):

DmitriyMusatkin (approved):

PR #2598, opened 2023-07-28
Add checksum param to transfer manager

sbiscigl:
*Description of changes:*

Configures transfer manager to use a default checksum algorithm of CRC32 for transfers. Supports using other checksums if specified.

*Check all that applies:*

- [x] Linux
- [x] Windows
- [ ] Android
- [x] MacOS
- [ ] IOS
- [ ] Other Platforms

### DIFF
--- a/src/aws-cpp-sdk-transfer/include/aws/transfer/TransferHandle.h
+++ b/src/aws-cpp-sdk-transfer/include/aws/transfer/TransferHandle.h
@@ -77,6 +77,8 @@ namespace Aws
                 bool IsLastPart() { return m_lastPart; }
                 void SetLastPart() { m_lastPart = true; }
 
+                Aws::String GetChecksum() const { return m_checksum; };
+                void SetChecksum(const Aws::String& checksum) { m_checksum = checksum; }
             private:
 
                 int m_partId;
@@ -90,6 +92,7 @@ namespace Aws
                 std::atomic<Aws::IOStream *> m_downloadPartStream;
                 std::atomic<unsigned char*> m_downloadBuffer;
                 bool m_lastPart;
+                Aws::String m_checksum;
         };
 
         using PartPointer = std::shared_ptr< PartState >;
--- a/src/aws-cpp-sdk-transfer/include/aws/transfer/TransferManager.h
+++ b/src/aws-cpp-sdk-transfer/include/aws/transfer/TransferManager.h
@@ -11,6 +11,7 @@
 #include <aws/s3/model/GetObjectRequest.h>
 #include <aws/s3/model/CreateMultipartUploadRequest.h>
 #include <aws/s3/model/UploadPartRequest.h>
+#include <aws/s3/model/CompletedPart.h>
 #include <aws/core/utils/threading/Executor.h>
 #include <aws/core/utils/memory/stl/AWSStreamFwd.h>
 #include <aws/core/utils/ResourceManager.h>
@@ -54,7 +55,7 @@ namespace Aws
             /**
              * When true, TransferManager will calculate the MD5 digest of the content being uploaded.
              * The digest is sent to S3 via an HTTP header enabling the service to perform integrity checks.
-             * This option is disabled by default.
+             * This option is disabled by default. Defer to checksumAlgorithm to use other checksum algorithms.
              */
             bool computeContentMD5;
             /**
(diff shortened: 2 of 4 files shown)


SergeyRyabinin (approved):

src/aws-cpp-sdk-transfer/include/aws/transfer/TransferHandle.h
```diff
@@ -77,6 +77,8 @@ namespace Aws
                 bool IsLastPart() { return m_lastPart; }
                 void SetLastPart() { m_lastPart = true; }
 
+                Aws::String GetChecksum() const { return m_checksum; };
+                void SetChecksum(const Aws::String& checksum) { m_checksum = checksum; }
```
DmitriyMusatkin, line 81: should it really have a setter? what happens if you set it mid transfer?

src/aws-cpp-sdk-transfer/include/aws/transfer/TransferManager.h
```diff
@@ -116,6 +117,13 @@ namespace Aws
              * key/val of map entries will be key/val of query strings.
              */
             Aws::Map<Aws::String, Aws::String> customizedAccessLogTag;
+
+            /**
+             * Set the Checksum Algorithm for the transfer manager to use for multipart
+             * upload. Defaults to CRC32. Will be overwritten to use MD5 if computeContentMD5
+             * is set to true.
+             */
+            Aws::S3::Model::ChecksumAlgorithm checksumAlgorithm = S3::Model::ChecksumAlgorithm::CRC32;
```
DmitriyMusatkin, line 126: you can set it to none to disable checksum?

src/aws-cpp-sdk-transfer/include/aws/transfer/TransferManager.h
```diff
@@ -328,6 +336,22 @@ namespace Aws
             void AddTask(std::shared_ptr<TransferHandle> handle);
             void RemoveTask(const std::shared_ptr<TransferHandle>& handle);
 
+            /**
+             * Gets the BytBuffer of a checksum on a stream based on the checksum algorithm passed.
```
DmitriyMusatkin, line 340: nit: typo in ByteBuffer

src/aws-cpp-sdk-transfer/source/transfer/TransferManager.cpp
```diff
@@ -1419,5 +1433,53 @@ namespace Aws
             auto handle = CreateUploadFileHandle(fileStream.get(), bucketName, keyName, contentType, metadata, context, fileName);
             return SubmitUpload(handle);
         }
+
+        Aws::Utils::ByteBuffer TransferManager::GetChecksumForAlgorithm(Aws::S3::Model::ChecksumAlgorithm algorithm,
```
DmitriyMusatkin, line 1437: is this really needed? if tm sets algo on UploadPart, shouldnt it force calculation of checksum during execution of that api and then you can retrieve calculated checksum? This will result in checksum being calculated twice

src/aws-cpp-sdk-transfer/include/aws/transfer/TransferHandle.h
sbiscigl: yeah fair

sbiscigl: This isnt on the public interface this is on "part state" that keeps track of one the parts being processed, we attach the checksum to the part so that when we call CompleteMultipart, we know the checksum of the part.

src/aws-cpp-sdk-transfer/include/aws/transfer/TransferManager.h
sbiscigl: Yeah there is specific logic for that where we dont

src/aws-cpp-sdk-transfer/source/transfer/TransferManager.cpp
sbiscigl: Yes we do because when we call `CompleteMultipartUpload` we need to provide it the checksum of each part uploaded. so we have to keep track of it. as far as it being calculated twice, yes that is unfortunate and we could likely refactor it to a virtual function and do a conditional calculation, but that can be a nice to have we can do later

sbiscigl: nope theres a easy around this, implementing

DmitriyMusatkin (approved):

PR #2604, opened 2023-08-01
Add cmake parameter to enforce tls version

sbiscigl:
*Description of changes:*

Adds two cmake parameters `USE_TLS_V2` and `USE_TLS_V3` that force the http client to use TLS 1.2 or 1.3. Currently they default to 1.2 per [AWS guidance](https://aws.amazon.com/blogs/security/tls-1-2-required-for-aws-endpoints/).

*Check all that applies:*

- [x] Linux
- [x] Windows
- [ ] Android
- [x] MacOS
- [ ] IOS
- [ ] Other Platforms


CMakeLists.txt
```diff
@@ -78,6 +80,12 @@ if (LEGACY_BUILD)
     if (DISABLE_INTERNAL_IMDSV1_CALLS)
         add_definitions(-DDISABLE_IMDSV1)
     endif ()
+    if (USE_TLS_V2)
+        add_definitions(-DENFORCE_TLS_V2)
```
SergeyRyabinin, line 84: hate to be that guy
but there is no TLS V2 or V3
There is V1_2 or V1_3 :-(

sbiscigl: hahaha yeah can change

SergeyRyabinin (approved):

src/aws-cpp-sdk-core/source/http/curl/CurlHttpClient.cpp
tmajkowski-ot, line 1: these changes have no effect with current versions of curl 8 since LIBCURL_VERSION_MINOR would be <34.
For curl version range checks LIBCURL_VERSION_NUM might be a better choice
https://curl.se/docs/versions.html

also, CURL_SSLVERSION_TLSv1_3 is available as of curl 7.52 according to curl documentation
https://curl.se/libcurl/c/CURLOPT_SSLVERSION.html

PR #2613, opened 2023-08-02
[Curl Client / Transfer Manager]: add error handling for failing local writes

grrtrr:
When downloading file data via the CurlHTTPClient or the TransferManager, there are no checks for failing local writes.

If a local write fails, it fails silently.

This caused difficult to debug corruption in several of our download applications. If a local disk was close to filling up, some files of the parallel download would be partially written out. Distinguishing partially written/corrupted files from healthy ones took time.

Avoid the problem by checking the stream status, and returning an error if any of the fstream operations fails.

*Check all that applies:*
  Code changes fixed missing error handling. We haven't had issues reported with failing local writes since introducing this change about 2 years ago.

- [X] Linux
- [ ] Windows
- [ ] Android
- [ ] MacOS
- [ ] IOS
- [ ] Other Platforms


jmklix:
Testing this PR

jmklix (approved):

SergeyRyabinin (approved):

PR #2615, opened 2023-08-02
[aws-cpp-sdk-s3-crt] support for the exponential back-off retry strategy

grrtrr:
The `S3CrtClient` not longer has a configurable retry strategy.

By default, the `S3CrtClient` uses the 'standard' AWS retry strategy with default parameters (5 maximum retries, scale factor of 25msec and default/full jitter options).

In porting from `S3Client` to `S3CrtClient` deployments, support for exponential back-off is necessary, in order to retry for a longer period.

This is based on the truncated exponential backoff in https://github.com/awslabs/aws-c-io/pull/587.

*Check all that applies:*

- [X] Linux
- [ ] Windows
- [ ] Android
- [ ] MacOS
- [ ] IOS
- [ ] Other Platforms


jmklix:
Can you add an enum flag to opt in to this new retry strategy?

jmklix:
It would look like this:
```cpp
enum class AwsS3CrtRetryStrategy
  {
    DEFAULT=0, /* defaults to standard now but may change */
    STANDARD,
    EXPONENTIAL_BACKOFF,
    NO_RETRY=-1,
  };
```

grrtrr:
> It would look like this:

Ok thanks. It may take a few weeks to implement this, I don't have time to get into that at the moment.

generated/src/aws-cpp-sdk-s3-crt/source/S3CrtClient.cpp
```diff
@@ -480,6 +492,7 @@ void S3CrtClient::init(const S3Crt::ClientConfiguration& config,
   {
     aws_tls_connection_options_clean_up(&nonConstTlsOptions);
   }
+  aws_retry_strategy_release(s3CrtConfig.retry_strategy);
```
grrtrr, line 495: `aws_retry_strategy_release` does nothing when given a `NULL` parameter, so safe to use with defaults.

grrtrr:
@jmklix - sorry it has taken a while. I have added an enum flag as requested, specifically
- used capital snake-case naming, for consistency with `US_EAST_1_REGIONAL_ENDPOINT_OPTION` in `include/aws/s3-crt/S3CrtClientConfiguration.h`,
- changed parameter names to match those of the updated https://github.com/awslabs/aws-c-io/pull/587.

DmitriyMusatkin (approved):

jmklix:
Thanks for taking the time to make this PR, but it's no longer necessary. We added a S3 CRT retry strategy config with this PR: https://github.com/aws/aws-sdk-cpp/pull/3110. Please let us know if there are any additional things that you would like added to this sdk.

Issue #2612, opened 2023-08-02
[aws-cpp-sdk-core]: allow disabling IMDS in ClientConfiguration default constructor


grrtrr:
### Problem Description

The `ClientConfiguration` and its sub-classes allow disabling IMDS lookup in all but the default constructor:
```c++
        struct AWS_CORE_API ClientConfiguration
        {
            ClientConfiguration(); // Does IMDS look-up by default, also in sub-classes.
            ClientConfiguration(const char* profileName, bool shouldDisableIMDS = false);
            explicit ClientConfiguration(bool useSmartDefaults, const char* defaultMode = "legacy", bool shouldDisableIMDS = false);
        }
```
In the base class, `profileName` can be set to `nullptr`, but not in the sub-classes: e.g. in `S3CrtClientConfiguration`, the `const char *inputProfileName` gets converted to a `std::string` (which does not like `nullptr`).

Hence the following use-case will always cause IMDS look-up:
```c++
        Aws::S3Crt::ClientConfiguration config;   // Default constructor
        config.scheme = defaults.scheme;
        config.endpointOverride = defaults.endpointOverride;
        config.connectTimeoutMs = 30000;
        // ...
```
This is because the default constructor hard-codes IMDS lookup:
```c++
// src/aws-cpp-sdk-core/source/client/ClientConfiguration.cpp
ClientConfiguration::ClientConfiguration()
{
    this->disableIMDS = false;  // <=== HERE
    // ...

    if (!this->disableIMDS &&
        region.empty() &&
        Aws::Utils::StringUtils::ToLower(Aws::Environment::GetEnv("AWS_EC2_METADATA_DISABLED").c_str()) != "true")
    {
        auto client = Aws::Internal::GetEC2MetadataClient();
        if (client)
        {
            region = client->GetCurrentRegion();  // <== IMDS look-up
        }
    }
    // ...
}
```
For an earlier record of the problems (slowness) that unnecessary IMDS look-up causes, please see https://github.com/aws/aws-sdk-cpp/issues/1440.

### What to do

For `ClientConfiguration`, `GenericClientConfiguration`, and `Aws::S3Crt::ClientConfiguration`, please provide a `shouldDisableIMDS` default argument like the 2 other constructors:
```c++
        struct AWS_CORE_API ClientConfiguration
        {
            explicit ClientConfiguration(bool shouldDisableIMDS = false) {
               this->disableIMDS = shouldDisableIMDS;
[...]

PR #2618, opened 2023-08-04
Fix issue 2612

yasminetalby:
Provide shouldDisableIMDS default argument for ClientConfiguration and GenericClientConfiguration constructors

*Issue #, if available:* #2612

*Description of changes:*

*Check all that applies:*

- [ ] Linux
- [ ] Windows
- [ ] Android
- [ ] MacOS
- [ ] IOS
- [ ] Other Platforms

### DIFF
--- a/src/aws-cpp-sdk-core/include/aws/core/client/ClientConfiguration.h
+++ b/src/aws-cpp-sdk-core/include/aws/core/client/ClientConfiguration.h
@@ -60,8 +60,11 @@ namespace Aws
           UseRequestCompression useRequestCompression=UseRequestCompression::ENABLE;
           size_t requestMinCompressionSizeBytes = 10240;
         };
-
-        struct DefaultClientConfiguration {
+         /**
+         * This structure is used to provide initial configuration values to the default ClientConfiguration constructor for the following parameter(s):
+          * - disableIMDS
+         */
+        struct ClientConfigurationInitValues {
             bool shouldDisableIMDS = false;
         };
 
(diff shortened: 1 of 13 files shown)


src/aws-cpp-sdk-core/include/aws/core/client/ClientConfiguration.h
```diff
@@ -67,7 +67,7 @@ namespace Aws
          */
         struct AWS_CORE_API ClientConfiguration
         {
-            ClientConfiguration();
+            ClientConfiguration(bool shouldDisableIMDS = false);
```
sbiscigl, line 70: Ok so as discussed this will not compile because

```
explicit ClientConfiguration(bool useSmartDefaults, 
  const char* defaultMode = "legacy", 
  bool shouldDisableIMDS = false);
```

resolves ambiguously to this. One way around this is that we could be creating a `DefaultConfiguration` struct that can be passed optionally to the default constructor that we can add things to in over time i.e.

```
struct DefaultClientConfiguration {
  bool shouldDisableIMDS = false;
};

struct ClientConfiguration {
  ...
  ClientConfiguration(const DefaultClientConfiguration &configuration = {})
  ...
}
```

then invoked though

```
ClientConfiguration::ClientConfiguration(const DefaultClientConfiguration &configuration)
{
    this->disableIMDS = configuration.shouldDisableIMDS;
    ...
}
```

this was we dont have to add more arguments each time we want to add something to the default constructor _and_ it doesnt resolve ambigiously

sbiscigl (approved):

SergeyRyabinin (approved):

Issue #2540, opened 2023-06-14
Aws::S3::Model::PutObjectRequest::SetChecksumSHA256 not behaving as expected


aphillips-opentext:
### Describe the bug

I am trying to make use of additional checksums in S3 as described here:
https://aws.amazon.com/getting-started/hands-on/amazon-s3-with-additional-checksums/?ref=docs_gateway/amazons3/checking-object-integrity.html

The page mentions that I can supply a precomputed checksum and the server will check it on upload. I added the checksum type and value to the put request and uploaded it and everything seemed to work; the web UI correctly displayed the checksum I had supplied. However I noticed that if I supply the wrong checksum, or even a string that isn't a checksum, the call still succeeds and the website still displays the correct checksum. This would suggest the SDK is not using what I supplied and calculating its own. Is this a bug or am I misunderstanding something?

### Expected Behavior

If I supply a bad checksum I expect the object upload to fail.

### Current Behavior

The upload succeeds.

### Reproduction Steps

```cpp
void UploadFile(const std::string& access_key_id, const std::string& secret_key, const std::string bucket_name, const std::string key_name, std::shared_ptr<std::fstream> file_stream)
{
  Aws::S3::S3ClientConfiguration config;
  config.scheme = Aws::Http::Scheme::HTTPS;
  config.region = "us-east-2";

  Aws::S3::S3Client client(Aws::Auth::AWSCredentials(access_key_id, secret_key), config, Aws::Client::AWSAuthV4Signer::PayloadSigningPolicy::RequestDependent, false);
  
  Aws::S3::Model::PutObjectRequest put_object_request;
  put_object_request.SetBucket(bucket_name);
  put_object_request.SetKey(key_name);
  put_object_request.SetBody(file_stream);
  put_object_request.SetChecksumAlgorithm(Aws::S3::Model::ChecksumAlgorithm::SHA256);
  put_object_request.SetChecksumSHA256("not_a_hash");

  auto outcome(client.PutObject(put_object_request));

  std::cout << "Put " << (outcome.IsSuccess() ? "succeeded" : "failed") << std::endl;
}
```

### Additional Information/Context

The SDK was built with curl and openssl.

### AWS CPP SDK version used

1.11.63

### Compiler and Version used

Visual Studio 2022

### Operating System and version

Windows 10 Enterprise

yasminetalby:
Hello @aphillips-opentext ,

Thank you very much for your submission.

[...]

PR #2644, opened 2023-08-29
Fix to allow user to provide their own checksum value when checksum algorithm is specified

yasminetalby:
*Issue #, if available:* #2540

*Description of changes:*
Fix to allow user to provide their own checksum value when checksum algorithm is specified.

*Check all that applies:*

- [ ] Linux
- [ ] Windows
- [ ] Android
- [ ] MacOS
- [ ] IOS
- [ ] Other Platforms

### DIFF
--- a/src/aws-cpp-sdk-core/include/aws/core/utils/crypto/Hash.h
+++ b/src/aws-cpp-sdk-core/include/aws/core/utils/crypto/Hash.h
@@ -48,6 +48,23 @@ namespace Aws
 
                 // when hashing streams, this is the size of our internal buffer we read the stream into
                 static const uint32_t INTERNAL_HASH_STREAM_BUFFER_SIZE = 8192;
+
+                void SetPrecalculatedHash(const Aws::String &precalculatedHash) {
+                    m_precalculatedHash = precalculatedHash;
+                    m_precalculatedHashSet = true;
+                }
+
+                bool isPrecalculatedHashSet() const {
+                    return m_precalculatedHashSet;
+                }
+
+                const Aws::String &GetPrecalculatedHash() const {
+                    return m_precalculatedHash;
+                }
+
+            private:
+                bool m_precalculatedHashSet = false;
+                Aws::String m_precalculatedHash;
             };
 
             /**
--- a/src/aws-cpp-sdk-core/include/aws/core/utils/crypto/Hash.h
+++ b/src/aws-cpp-sdk-core/include/aws/core/utils/crypto/Hash.h
@@ -50,7 +50,7 @@ namespace Aws
                 /**
                  * Return the base64 encoded checksum value
                  */
-                virtual Aws::String GetHashBase64(ByteBuffer m_checksum){
+                virtual Aws::String GetHashBase64(ByteBuffer& m_checksum){
                     Aws::String base64Encoded = HashingUtils::Base64Encode(m_checksum);
                     return base64Encoded;
                 }
--- a/src/aws-cpp-sdk-core/include/aws/core/utils/crypto/Hash.h
+++ b/src/aws-cpp-sdk-core/include/aws/core/utils/crypto/Hash.h
@@ -9,6 +9,7 @@
 #include <aws/core/utils/crypto/HashResult.h>
 #include <aws/core/utils/memory/stl/AWSStreamFwd.h>
 #include <aws/core/utils/memory/stl/AWSString.h>
+#include <aws/core/utils/HashingUtils.h>
 
 namespace Aws
 {
@@ -46,25 +47,18 @@ namespace Aws
                  */
                 virtual HashResult GetHash() = 0;
 
+                /**
+                 * Return the base64 encoded checksum value
+                 */
+                virtual Aws::String GetHashBase64(ByteBuffer m_checksum){
+                    Aws::String base64Encoded = HashingUtils::Base64Encode(m_checksum);
+                    return base64Encoded;
+                }
+
                 // when hashing streams, this is the size of our internal buffer we read the stream into
                 static const uint32_t INTERNAL_HASH_STREAM_BUFFER_SIZE = 8192;
 
-                void SetPrecalculatedHash(const Aws::String &precalculatedHash) {
-                    m_precalculatedHash = precalculatedHash;
-                    m_precalculatedHashSet = true;
-                }
 
-                bool isPrecalculatedHashSet() const {
-                    return m_precalculatedHashSet;
-                }
-
-                const Aws::String &GetPrecalculatedHash() const {
-                    return m_precalculatedHash;
-                }
-
-            private:
-                bool m_precalculatedHashSet = false;
-                Aws::String m_precalculatedHash;
             };
 
             /**
--- a/src/aws-cpp-sdk-core/include/aws/core/utils/crypto/PrecalculatedHash.h
+++ b/src/aws-cpp-sdk-core/include/aws/core/utils/crypto/PrecalculatedHash.h
@@ -20,7 +20,7 @@ namespace Aws {
                 virtual HashResult Calculate(const Aws::String &str) override;
                 HashResult Calculate(Aws::IStream &stream) override;
                 void Update(unsigned char *string, size_t bufferSize) override;
-                virtual Aws::String GetHashBase64(ByteBuffer m_checksum) override;
+                virtual Aws::String GetHashBase64(ByteBuffer& m_checksum) override;
                 HashResult GetHash() override;
 
             private:
--- a/src/aws-cpp-sdk-core/include/aws/core/utils/crypto/PrecalculatedHash.h
+++ b/src/aws-cpp-sdk-core/include/aws/core/utils/crypto/PrecalculatedHash.h
@@ -13,12 +13,9 @@
 namespace Aws {
     namespace Utils {
         namespace Crypto {
-            /**
-             * CRC32 hash implementation.
-             */
             class AWS_CORE_API PrecalculatedHash : public Hash {
             public:
-                PrecalculatedHash(const Aws::String &mHash);
+                PrecalculatedHash(const Aws::String &hash);
                 virtual ~PrecalculatedHash();
                 virtual HashResult Calculate(const Aws::String &str) override;
                 HashResult Calculate(Aws::IStream &stream) override;
@@ -27,7 +24,7 @@ namespace Aws {
                 HashResult GetHash() override;
 
             private:
-                Aws::String m_hash_string;
+                Aws::String m_hashString;
 
             };
         }
--- a/src/aws-cpp-sdk-core/include/aws/core/utils/crypto/PrecalculatedHash.h
+++ b/src/aws-cpp-sdk-core/include/aws/core/utils/crypto/PrecalculatedHash.h
@@ -1,0 +1,36 @@
+/**
+ * Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.
+ * SPDX-License-Identifier: Apache-2.0.
+ */
+
+#pragma once
+
+#include <aws/core/Core_EXPORTS.h>
+#include <aws/core/utils/crypto/Hash.h>
+#include <aws/core/utils/crypto/HashResult.h>
+#include "aws/core/utils/Outcome.h"
+
+namespace Aws {
+    namespace Utils {
+        namespace Crypto {
+            /**
+             * CRC32 hash implementation.
+             */
+            class AWS_CORE_API PrecalculatedHash : public Hash {
+            public:
+                PrecalculatedHash(const Aws::String &mHash);
+                virtual ~PrecalculatedHash();
+                virtual HashResult Calculate(const Aws::String &str) override;
+                HashResult Calculate(Aws::IStream &stream) override;
+                void Update(unsigned char *string, size_t bufferSize) override;
+                virtual Aws::String GetHashBase64(ByteBuffer m_checksum) override;
+                HashResult GetHash() override;
+
+            private:
+                HashResult m_hash;
+                Aws::String m_hash_string;
+
+            };
+        }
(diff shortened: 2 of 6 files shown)


src/aws-cpp-sdk-core/source/http/curl/CurlHttpClient.cpp
```diff
@@ -329,10 +329,14 @@ static size_t ReadBody(char* ptr, size_t size, size_t nmemb, void* userdata, boo
...
                 Aws::StringStream chunkedTrailer;
                 chunkedTrailer << "0\r\n";
-                if (request->GetRequestHash().second != nullptr)
-                {
+                if (request->GetRequestHash().second->isPrecalculatedHashSet()) {
+                    chunkedTrailer << "x-amz-checksum-" << request->GetRequestHash().first << ":"
+                                   << request->GetRequestHash().second->GetPrecalculatedHash() << "\r\n";
+                } else if (request->GetRequestHash().second != nullptr) {
+                    auto hash = request->GetRequestHash().second->GetHash().GetResult();
+                    std::cout << hash.GetUnderlyingData();
```
sbiscigl, line 337: looks like some debugging code slipped in here, should remove this and potentially inline the hash variable

src/aws-cpp-sdk-core/source/auth/signer/AWSAuthV4Signer.cpp
```diff
@@ -251,6 +257,7 @@ bool AWSAuthV4Signer::SignRequest(Aws::Http::HttpRequest& request, const char* r
         {
             payloadHash = STREAMING_UNSIGNED_PAYLOAD_TRAILER;
             Aws::String trailerHeaderValue = Aws::String("x-amz-checksum-") + request.GetRequestHash().first;
+            request.DeleteHeader(trailerHeaderValue.c_str());
```
sbiscigl, line 260: we should consider not serializing the header `x-amz-checksum-` to the header in the first place during codegen as we will always end up adding it one way or another. this is fine for now imo opinon, but removing the serialization could perhaps be a better long term solution

src/aws-cpp-sdk-core/include/aws/core/utils/crypto/Hash.h
```diff
@@ -48,6 +48,23 @@ namespace Aws
...
+                void SetPrecalculatedHash(const Aws::String &precalculatedHash) {
+                    m_precalculatedHash = precalculatedHash;
+                    m_precalculatedHashSet = true;
+                }
+
+                bool isPrecalculatedHashSet() const {
+                    return m_precalculatedHashSet;
+                }
+
+                const Aws::String &GetPrecalculatedHash() const {
```
sbiscigl, line 61: could make these methods inline

src/aws-cpp-sdk-core/source/client/AWSClient.cpp
```diff
@@ -774,61 +774,101 @@ void AWSClient::AppendHeaderValueToRequest(const std::shared_ptr<Aws::Http::Http
...
+    //check if user has provided the checksum algorithm
     Aws::String checksumAlgorithmName = Aws::Utils::StringUtils::ToLower(request.GetChecksumAlgorithmName().c_str());
+    bool checksumAlgorithmProvided = !checksumAlgorithmName.empty();
 
     // Request checksums
-    if (!checksumAlgorithmName.empty())
+    if (checksumAlgorithmProvided)
     {
+        // Check if user has provided a checksum value for the specified algorithm
+        Aws::String checksum_type = "x-amz-checksum-" + checksumAlgorithmName;
```
SergeyRyabinin, line 785: this file (and SDK) uses camelCase.
also consider adding const if this variable is not updated later

src/aws-cpp-sdk-core/source/auth/signer/AWSAuthV4Signer.cpp
```diff
@@ -238,7 +238,13 @@ bool AWSAuthV4Signer::SignRequest(Aws::Http::HttpRequest& request, const char* r
         if (request.GetRequestHash().second != nullptr)
         {
             Aws::String checksumHeaderKey = Aws::String("x-amz-checksum-") + request.GetRequestHash().first;
-            Aws::String checksumHeaderValue = HashingUtils::Base64Encode(request.GetRequestHash().second->Calculate(*(request.GetContentBody())).GetResult());
+            auto checksumHeaderValue = [](const std::shared_ptr<Utils::Crypto::Hash> &hash,
```
SergeyRyabinin, line 241: I know IIFE is cool, but KISS is also cool and ternary operator would work here too:

```
auto checksumHeaderValue = request.GetRequestHash().second ? 
    request.GetRequestHash().second : HashingUtils::Base64Encode(hash->Calculate(request.GetContentBody().get()).GetResult());
```

also, because we now dereference GetContentBody, it worth adding a check it is present (and trace error and return false if not)

SergeyRyabinin: or keep the IIFE, resulting ternary operator is also looking complicated.

DmitriyMusatkin, line 241: nit: lets backlog this and put a comment - we should not be computing checksum over the whole body and put it in header. we should keep a running checksum and compute it as we are streaming the body

src/aws-cpp-sdk-core/include/aws/core/utils/crypto/PrecalculatedHash.h
```diff
@@ -0,0 +1,36 @@
...
+#include <aws/core/Core_EXPORTS.h>
+#include <aws/core/utils/crypto/Hash.h>
+#include <aws/core/utils/crypto/HashResult.h>
+#include "aws/core/utils/Outcome.h"
+
+namespace Aws {
+    namespace Utils {
+        namespace Crypto {
+            /**
+             * CRC32 hash implementation.
```
sbiscigl, line 17: nit: you copied the comment as well

src/aws-cpp-sdk-core/source/utils/crypto/PrecalculatedHash.cpp
```diff
@@ -0,0 +1,43 @@
...
+ * SPDX-License-Identifier: Apache-2.0.
+ */
+
+
+#include <aws/core/utils/crypto/PrecalculatedHash.h>
+
+using namespace Aws::Utils::Crypto;
+
+PrecalculatedHash::PrecalculatedHash(const Aws::String &mHash): m_hash(ByteBuffer((unsigned char*)mHash.c_str(),mHash.length())){
+m_hash_string = mHash;
```
sbiscigl, line 12: add to the initializer list instead of assigning in the ctor and rename to m_hashString pls

src/aws-cpp-sdk-core/include/aws/core/utils/crypto/PrecalculatedHash.h
```diff
@@ -0,0 +1,36 @@
...
+             */
+            class AWS_CORE_API PrecalculatedHash : public Hash {
+            public:
+                PrecalculatedHash(const Aws::String &mHash);
```
sbiscigl, line 21: nit: call it it hash instead of mHash lol. It really doesnt matter but we call member variables m_variable name to disambiguate from non-members

src/aws-cpp-sdk-core/source/client/AWSClient.cpp
```diff
@@ -793,10 +794,9 @@ void AWSClient::AddChecksumToRequest(const std::shared_ptr<Aws::Http::HttpReques
         {
             if (request.IsStreaming() && checksumValueAndAlgorithmProvided)
             {
-                auto hash = Aws::MakeShared<Crypto::CRC32>(AWS_CLIENT_LOG_TAG);
-                auto precalculatedValue = request.GetHeaders().find(checksumType)->second;
-                hash->SetPrecalculatedHash(precalculatedValue);
-                httpRequest->SetRequestHash(checksumAlgorithmName, hash); }
+                auto hash = Aws::MakeShared<Crypto::PrecalculatedHash>(AWS_CLIENT_LOG_TAG,request.GetHeaders().find(checksumType)->second);
```
sbiscigl, line 797: we could likely factor all of this out into a single `if (request.IsStreaming() && checksumValueAndAlgorithmProvided)` statement instead of creating one for each checksum algo

src/aws-cpp-sdk-core/source/utils/crypto/PrecalculatedHash.cpp
```diff
@@ -0,0 +1,43 @@
...
+HashResult PrecalculatedHash::Calculate(const Aws::String& str)
+{
+    AWS_UNREFERENCED_PARAM(str);
+    return m_hash;
+}
+
+HashResult PrecalculatedHash::Calculate(Aws::IStream& stream)
+{
+    AWS_UNREFERENCED_PARAM(stream);
+    return m_hash;
```
sbiscigl, line 26: do we need `m_hash` at all anymore? imo whenever we want to return a byte buffer we should be calling `HashingUtils::Base64Decode(m_hashString)`  that way we only are keeping track of one member variable and the constructor becomes a lot cleaner as we are only using one string param and not casting it to a byte array.

also it might be incorrect as we would be creating a byte buffer of the base54 encoding and not the actual hash result

src/aws-cpp-sdk-core/include/aws/core/utils/crypto/Hash.h
```diff
@@ -46,8 +47,18 @@ namespace Aws
                  */
                 virtual HashResult GetHash() = 0;
 
+                /**
+                 * Return the base64 encoded checksum value
+                 */
+                virtual Aws::String GetHashBase64(ByteBuffer m_checksum){
+                    Aws::String base64Encoded = HashingUtils::Base64Encode(m_checksum);
```
sbiscigl, line 54: instead of passing m_checksum couldnt we do it without any parameters like

```
return HashingUtils::Base64Encode(GetHash());
```

src/aws-cpp-sdk-core/source/client/AWSClient.cpp
```diff
@@ -774,61 +775,76 @@ void AWSClient::AppendHeaderValueToRequest(const std::shared_ptr<Aws::Http::Http
 void AWSClient::AddChecksumToRequest(const std::shared_ptr<Aws::Http::HttpRequest>& httpRequest,
```
sennhann29, line 775: r-q7d1

sbiscigl (approved):

SergeyRyabinin (approved):

src/aws-cpp-sdk-core/source/auth/signer/AWSAuthV4Signer.cpp
```diff
@@ -238,7 +238,7 @@ bool AWSAuthV4Signer::SignRequest(Aws::Http::HttpRequest& request, const char* r
         if (request.GetRequestHash().second != nullptr)
         {
             Aws::String checksumHeaderKey = Aws::String("x-amz-checksum-") + request.GetRequestHash().first;
-            Aws::String checksumHeaderValue = HashingUtils::Base64Encode(request.GetRequestHash().second->Calculate(*(request.GetContentBody())).GetResult());
+            auto checksumHeaderValue = request.GetRequestHash().second->GetHashBase64(request.GetRequestHash().second->Calculate(*(request.GetContentBody())).GetResult());
```
DmitriyMusatkin, line 241: GetHashBase64 should not be taking result of calculate here. Just keep the old code here and let calculate return raw value for precomputed as described below

src/aws-cpp-sdk-core/source/http/curl/CurlHttpClient.cpp
```diff
@@ -329,10 +329,10 @@ static size_t ReadBody(char* ptr, size_t size, size_t nmemb, void* userdata, boo
             {
                 Aws::StringStream chunkedTrailer;
                 chunkedTrailer << "0\r\n";
-                if (request->GetRequestHash().second != nullptr)
-                {
+                if (request->GetRequestHash().second != nullptr) {
+                    auto hash = request->GetRequestHash().second->GetHash().GetResult();
```
DmitriyMusatkin, line 333: this hash can be changed to GetHashBase64()

src/aws-cpp-sdk-core/source/client/AWSClient.cpp
```diff
@@ -774,61 +775,76 @@ void AWSClient::AppendHeaderValueToRequest(const std::shared_ptr<Aws::Http::Http
...
+        const Aws::String checksumType = "x-amz-checksum-" + checksumAlgorithmName;
+        const HeaderValueCollection &headers = request.GetHeaders();
+        bool checksumValueAndAlgorithmProvided = headers.find(checksumType) != headers.end();
+
         // For non-streaming payload, the resolved checksum location is always header.
         // For streaming payload, the resolved checksum location depends on whether it is an unsigned payload, we let AwsAuthSigner decide it.
-        if (checksumAlgorithmName == "crc32")
+        if (request.IsStreaming() && checksumValueAndAlgorithmProvided)
+        {
+            auto hash = Aws::MakeShared<Crypto::PrecalculatedHash>(AWS_CLIENT_LOG_TAG,request.GetHeaders().find(checksumType)->second);
```
DmitriyMusatkin, line 794: I would switch a logic slightly here. Lets decode b64 that customer provided, make sure its valid b64, then construct precalculated hash and give it both b64 and raw formats. precalculated hash can then return b64/raw as specified by method.

we can probably remove the need to decode the b64, but that can be done along with the changes needed to avoid hashing the whole stream for signing purposes

jmklix:
merged here: https://github.com/aws/aws-sdk-cpp/pull/2826

PR #2684, opened 2023-09-26
cache credential chain results

sbiscigl:
*Description of changes:*

Right now by default the credentials provider chain retries each provider in order regardless if one succeeded before. This caches the last known credentials provider to provide credentials. If that cached provider fails to provide credentials it will fallback to the provider chain.

*Check all that applies:*

- [x] Linux
- [x] Windows
- [ ] Android
- [x] MacOS
- [ ] IOS
- [ ] Other Platforms

### DIFF
--- a/src/aws-cpp-sdk-core/include/aws/core/auth/AWSCredentialsProviderChain.h
+++ b/src/aws-cpp-sdk-core/include/aws/core/auth/AWSCredentialsProviderChain.h
@@ -46,8 +46,10 @@ namespace Aws
             void AddProvider(const std::shared_ptr<AWSCredentialsProvider>& provider) { m_providerChain.push_back(provider); }
 
 
-        private:            
+        private:
             Aws::Vector<std::shared_ptr<AWSCredentialsProvider> > m_providerChain;
+            std::shared_ptr<AWSCredentialsProvider> m_cachedProvider;
+            mutable Aws::Utils::Threading::ReaderWriterLock m_cachedProviderLock;
         };
 
         /**
--- a/src/aws-cpp-sdk-core/source/auth/AWSCredentialsProviderChain.cpp
+++ b/src/aws-cpp-sdk-core/source/auth/AWSCredentialsProviderChain.cpp
@@ -12,6 +12,7 @@
 #include <aws/core/utils/logging/LogMacros.h>
 
 using namespace Aws::Auth;
+using namespace Aws::Utils::Threading;
 
 static const char AWS_ECS_CONTAINER_CREDENTIALS_RELATIVE_URI[] = "AWS_CONTAINER_CREDENTIALS_RELATIVE_URI";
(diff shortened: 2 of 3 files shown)


src/aws-cpp-sdk-core/source/auth/AWSCredentialsProviderChain.cpp
```diff
@@ -21,11 +21,20 @@ static const char DefaultCredentialsProviderChainTag[] = "DefaultAWSCredentialsP
 
 AWSCredentials AWSCredentialsProviderChain::GetAWSCredentials()
 {
+    std::unique_lock<std::mutex> lock(m_providerChainLock);
```
SergeyRyabinin, line 24: Use semaphore class available in the SDK
(multi-read or single write)

src/aws-cpp-sdk-core/include/aws/core/auth/AWSCredentialsProviderChain.h
```diff
@@ -46,8 +46,10 @@ namespace Aws
             void AddProvider(const std::shared_ptr<AWSCredentialsProvider>& provider) { m_providerChain.push_back(provider); }
 
 
-        private:            
+        private:
             Aws::Vector<std::shared_ptr<AWSCredentialsProvider> > m_providerChain;
+            Aws::Crt::Optional<std::shared_ptr<AWSCredentialsProvider>> m_cachedProvider;
```
SergeyRyabinin, line 51: It is already a shared_ptr, I'm not sure Optional is needed here.
Of course the object is optional, but using `optional` type will create an additional allocation and de-reference

sbiscigl: good call

src/aws-cpp-sdk-core/source/auth/AWSCredentialsProviderChain.cpp
sbiscigl: Took a look at `Semaphore` and saw that `ReaderLockGuard` actually uses it and provides a more convenient API for what im trying to do. we actually already use it in [credentials provider](https://github.com/aws/aws-sdk-cpp/blob/main/src/aws-cpp-sdk-core/source/auth/AWSCredentialsProvider.cpp#L204-L219) in basically the same use case.

```
void ProfileConfigFileAWSCredentialsProvider::RefreshIfExpired()
{
    ReaderLockGuard guard(m_reloadLock);
    if (!IsTimeToRefresh(m_loadFrequencyMs))
    {
       return;
    }

    guard.UpgradeToWriterLock();
    if (!IsTimeToRefresh(m_loadFrequencyMs)) // double-checked lock to avoid refreshing twice
    {
        return;
    }

    Reload();
}
```

Let me know if you're ok with that

DmitriyMusatkin (approved):

SergeyRyabinin (approved):

PR #2689, opened 2023-09-29
User agent format update

SergeyRyabinin:
*Issue #, if available:*
n/a
*Description of changes:*
User agent format update
*Check all that applies:*

- [ ] Linux
- [ ] Windows
- [ ] Android
- [ ] MacOS
- [ ] IOS
- [ ] Other Platforms

### DIFF
--- a/src/aws-cpp-sdk-core/include/aws/core/client/AWSClient.h
+++ b/src/aws-cpp-sdk-core/include/aws/core/client/AWSClient.h
@@ -183,6 +183,7 @@ namespace Aws
              */
             virtual void SetServiceClientName(const Aws::String& name);
 
+            void AppendToUserAgent(const Aws::String& valueToAppend);
         protected:
             /**
              * Calls AttemptOneRequest until it either, succeeds, runs out of retries from the retry strategy,
@@ -338,11 +339,10 @@ namespace Aws
             std::shared_ptr<Aws::Utils::RateLimits::RateLimiterInterface> m_writeRateLimiter;
             std::shared_ptr<Aws::Utils::RateLimits::RateLimiterInterface> m_readRateLimiter;
             Aws::String m_userAgent;
-            bool m_customizedUserAgent;
             std::shared_ptr<Aws::Utils::Crypto::Hash> m_hash;
             long m_requestTimeoutMs;
             bool m_enableClockSkewAdjustment;
-            Aws::String m_serviceName;
+            Aws::String m_serviceName = "AWSBaseClient";
             Aws::Client::RequestCompressionConfig m_requestCompressionConfig;
             void AppendHeaderValueToRequest(
                 const std::shared_ptr<Http::HttpRequest> &request, String header,
--- a/src/aws-cpp-sdk-core/include/aws/core/client/RetryStrategy.h
+++ b/src/aws-cpp-sdk-core/include/aws/core/client/RetryStrategy.h
@@ -74,6 +74,15 @@ namespace Aws
              */
             virtual void RequestBookkeeping(const HttpResponseOutcome& /* httpResponseOutcome */) {}
             virtual void RequestBookkeeping(const HttpResponseOutcome& /* httpResponseOutcome */, const AWSError<CoreErrors>& /* lastError */) {}
+
+            /**
+             * Get return strategy name.
+             */
+            virtual const char* GetStrategyName() const
+            {
+                // SDK provided strategies will provide actual name
+                return "custom";
+            }
         };
 
         /**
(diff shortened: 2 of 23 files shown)


src/aws-cpp-sdk-core/include/aws/core/client/RetryStrategy.h
```diff
@@ -74,6 +74,15 @@ namespace Aws
...
             virtual void RequestBookkeeping(const HttpResponseOutcome& /* httpResponseOutcome */) {}
             virtual void RequestBookkeeping(const HttpResponseOutcome& /* httpResponseOutcome */, const AWSError<CoreErrors>& /* lastError */) {}
+
+            /**
+             * Get return strategy name.
+             */
+            virtual const char* GetStrategyName() const
+            {
+                // SDK provided strategies will provide actual name
+                return "custom";
```
sbiscigl, line 84: why not make pure virtual?

src/aws-cpp-sdk-core/include/aws/core/client/AWSClient.h
```diff
@@ -183,6 +183,7 @@ namespace Aws
              */
             virtual void SetServiceClientName(const Aws::String& name);
 
+            void AppendToUserAgent(const Aws::String& valueToAppend);
```
sbiscigl, line 186: should we make it a r-value?

src/aws-cpp-sdk-core/source/Version.cpp
```diff
@@ -36,15 +36,44 @@ namespace Version
...
 #endif
 #undef str
+#undef xstr
+  }
+
+  const char* GetCPPStandard()
+  {
+#define xstr(s) str(s)
+#define str(s) #s
+#if defined(CPP_STANDARD)
```
sbiscigl, line 55: i know `CPP_STANDARD` is used internally to [specify the standard flag](https://github.com/aws/aws-sdk-cpp/blob/05edc68675ba391e8669641ffa4e034ba691cb97/cmake/platform/linux.cmake#L38)

```
 list(APPEND CMAKE_REQUIRED_FLAGS "-std=c++${CPP_STANDARD}")
```

but do we know how this value intersects with the cmake arg [`CMAKE_CXX_STANDARD`](https://cmake.org/cmake/help/latest/variable/CMAKE_CXX_STANDARD.html)? This is likely perhaps outside of scope of this PR. But it is something we should likely reconcile at some point.

src/aws-cpp-sdk-core/source/internal/AWSHttpResourceClient.cpp
```diff
@@ -79,7 +79,10 @@ namespace Aws
         }
 
         AWSHttpResourceClient::AWSHttpResourceClient(const Aws::Client::ClientConfiguration& clientConfiguration, const char* logtag)
-        : m_logtag(logtag), m_retryStrategy(clientConfiguration.retryStrategy), m_httpClient(nullptr)
+        : m_logtag(logtag),
+          m_userAgent(Aws::Client::ComputeUserAgentString(&clientConfiguration)),
```
sbiscigl, line 83: why do we need to pass `clientConfiguration` as a raw ptr instead of just a normal lvalue ref?

src/aws-cpp-sdk-core/include/aws/core/client/RetryStrategy.h
SergeyRyabinin: To maintain backward compatibility and not break build if somebody inherits from this class.

src/aws-cpp-sdk-core/source/internal/AWSHttpResourceClient.cpp
SergeyRyabinin: Old `ComputeUserAgentString` does no accept any argument.
This is an approach to make it 
1. backward compatible
2. allow client config to be optional (in case this method is called in a place where there is no client config around).

src/aws-cpp-sdk-core/include/aws/core/client/AWSClient.h
SergeyRyabinin: I don't think it would bring much benefit here.
Unless you have an idea how it maybe more helpful here.

src/aws-cpp-sdk-core/source/Version.cpp
SergeyRyabinin: `__cplusplus` shall be a definitive answer, while CPP_STANDARD is a cmake argument that we allow to configure.
You are right we might need to revisit this later.

sbiscigl (approved):

PR #2698, opened 2023-10-05
SimpleStreamBuf data access

JBKacerovsky:
*Issue #, if available:*

*Description of changes:*

Added getters for private members m_buffer and m_bufferSize to SimpleStreamBuf.  In my project I needed to pass objects downloaded from s3 to another function as `char *` this minor change made accessing the data a bit more convenient than using SimpleStreamBuf::str() 

In benchmarks it is also marginally faster, but the difference is negligible.

*Check all that applies:*
- [ ] Added proper tests to cover this PR. (If tests are not applicable, explain.) Very minor change. Only exposes access to existing member variables without functional changes. 

- [X] Linux
- [ ] Windows
- [ ] Android
- [ ] MacOS
- [ ] IOS
- [ ] Other Platforms


src/aws-cpp-sdk-core/include/aws/core/utils/stream/SimpleStreamBuf.h
```diff
@@ -42,6 +42,12 @@ namespace Stream
 
             void swap(SimpleStreamBuf& rhs);
 
+            /// returns pointer to the underlying data buffer.
+            char* get_buffer_data();
```
SergeyRyabinin, line 46: thanks for your change!

Could you please use the styling convention used in this sdk? (i.e. the PascalCase, and call it `GetBufferData()`?)

Unless you wanted to make it compatible with some std:: stringbuf interface like [`basic_stringbuf`](https://en.cppreference.com/w/cpp/io/basic_stringbuf) (but I could not find if any std:: class has such method).

src/aws-cpp-sdk-core/include/aws/core/utils/stream/SimpleStreamBuf.h
```diff
@@ -42,6 +42,12 @@ namespace Stream
...
+
+            /// returns size of the underlying data buffer.
+            int get_buffer_size();
```
SergeyRyabinin, line 49: Could you please change it to
```
size_t GetBufferSize() const;
```

you can also make both methods inline.

PR #2703, opened 2023-10-09
Add Windows version checking before setting proxy flag

siwei-yuan:
*Issue #, if available:*

*Description of changes:*
The access type passed when opening a WinHTTP-session is set to `WINHTTP_ACCESS_TYPE_AUTOMATIC_PROXY` if the application using aws cpp sdk is built with >= Windows 8,1 SDK and have `allowSystemProxy` set to True. As a result, when running the application on Windows version older than 8.1, the WinHTTP-session cannot be successfully initiated. This change adds runtime version checking to determine the proper parameter to use.

*Check all that applies:*

- [ ] Linux
- [x] Windows
- [ ] Android
- [ ] MacOS
- [ ] IOS
- [ ] Other Platforms

### DIFF
--- a/src/aws-cpp-sdk-core/source/http/windows/WinHttpSyncHttpClient.cpp
+++ b/src/aws-cpp-sdk-core/source/http/windows/WinHttpSyncHttpClient.cpp
@@ -21,6 +21,7 @@
 #include <winhttp.h>
 #include <sstream>
 #include <iostream>
+#include <versionhelpers.h>
 
 using namespace Aws::Client;
 using namespace Aws::Http;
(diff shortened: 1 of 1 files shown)


src/aws-cpp-sdk-core/source/http/windows/WinHttpSyncHttpClient.cpp
```diff
@@ -52,10 +53,16 @@ WinHttpSyncHttpClient::WinHttpSyncHttpClient(const ClientConfiguration& config)
         << " request timeout " << config.requestTimeoutMs << ",and connect timeout " << config.connectTimeoutMs);
 
 #if defined(WINHTTP_ACCESS_TYPE_AUTOMATIC_PROXY)
-    DWORD winhttpFlags = config.allowSystemProxy ? WINHTTP_ACCESS_TYPE_AUTOMATIC_PROXY : WINHTTP_ACCESS_TYPE_NO_PROXY;
+    DWORD winhttpFlags;
+    if (config.allowSystemProxy) {
```
gurugubs, line 57: Logic might be easier to follow if we reorder the whole block of code:
```
DWORD winhttpFlags = config.allowSystemProxy ? WINHTTP_ACCESS_TYPE_DEFAULT_PROXY : WINHTTP_ACCESS_TYPE_NO_PROXY;
#if defined(WINHTTP_ACCESS_TYPE_AUTOMATIC_PROXY)
    if (config.allowSystemProxy && IsWindows8Point1OrGreater()) {
        winhttpFlags = WINHTTP_ACCESS_TYPE_AUTOMATIC_PROXY;
    }
#endif
```

SergeyRyabinin (approved):

Issue #2699, opened 2023-10-05
Aws::InitAPI(aws_options_) in two independent libraries may crash on OSX


levongh:
### Describe the bug

I do have a python binding library which is setting up an aws sdk during library init. And in the other hand in my python script, I am also using pyarrow library which crashes with the following stack trace.

```
Exiting Application
################################################################################
Stack trace:
################################################################################
1   libarrow.1300.dylib                 0x0000000125a09026 aws_fatal_assert + 70
2   libarrow.1300.dylib                 0x0000000125a08214 aws_mem_release + 52
3   libarrow.1300.dylib                 0x0000000125a0e945 cJSON_Delete + 101
4   libarrow.1300.dylib                 0x0000000125a0e92c cJSON_Delete + 76
5   libarrow.1300.dylib                 0x0000000125a00bb7 s_endpoints_ruleset_destroy + 23
6   libarrow.1300.dylib                 0x0000000125a1b1f2 aws_ref_count_release + 34
7   libarrow.1300.dylib                 0x0000000125a00c32 aws_endpoints_ruleset_release + 18
8   libarrow.1300.dylib                 0x00000001259fefd7 s_endpoints_rule_engine_destroy + 23
9   libarrow.1300.dylib                 0x0000000125a1b1f2 aws_ref_count_release + 34
10  libarrow.1300.dylib                 0x00000001259ff042 aws_endpoints_rule_engine_release + 18
11  libarrow.1300.dylib                 0x0000000125994aa1 _ZN3Aws3Crt9Endpoints10RuleEngineD1Ev + 17
12  api.cpython-310-darwin.so           0x000000013b1e4ecc _ZN3Aws2S38S3ClientD0Ev + 108
13  api.cpython-310-darwin.so           0x000000013aaafd86 api.cpython-310-darwin.so + 11033990
14  api.cpython-310-darwin.so           0x000000013aaafe2e api.cpython-310-darwin.so + 11034158
15  api.cpython-310-darwin.so           0x000000013a1bafd6 api.cpython-310-darwin.so + 1642454
16  api.cpython-310-darwin.so           0x000000013a140855 api.cpython-310-darwin.so + 1140821
17  api.cpython-310-darwin.so           0x000000013b7517d2 _ZN8pybind1117error_already_set23m_fetched_error_deleterEPNS_6detail25error_fetch_and_normalizeE + 64066
[...]

PR #2710, opened 2023-10-11
Add InitAPI call count to workaround multiple init mis-use

SergeyRyabinin:
*Issue #, if available:*
This is a workaround for issues where customer is not able to control underlying libraries using the C++ SDK, such as https://github.com/aws/aws-sdk-cpp/issues/2699
*Description of changes:*
Add reference count on and blocking on mutex on InitAPI-ShutdownAPI methods.
This does not really fix the root architectural issue, but at least it provides a some level of workaround.
All consequent calls to InitAPI only increases the SDK usage count, `options` argument is completely ignored on any subsequent call, and only the last ShutdownAPI will actually shut down the library.
*Check all that applies:*

- [x] Linux
- [x] Windows
- [ ] Android
- [ ] MacOS
- [ ] IOS
- [ ] Other Platforms

### DIFF
--- a/src/aws-cpp-sdk-core/source/Aws.cpp
+++ b/src/aws-cpp-sdk-core/source/Aws.cpp
@@ -26,8 +26,20 @@ namespace Aws
 {
     static const char* ALLOCATION_TAG = "Aws_Init_Cleanup";
 
+    static std::mutex s_initShutdownMutex;
+    static size_t s_initCount = 0;
+
     void InitAPI(const SDKOptions &options)
     {
+        std::unique_lock<std::mutex> lock(s_initShutdownMutex);
+        s_initCount += 1;
+        if(s_initCount != 1)
+        {
+            AWS_LOGSTREAM_ERROR(ALLOCATION_TAG, "AWS-SDK-CPP is already initialized " << s_initCount - 1 << " times. "
+                                                "Consequent calls to InitAPI are ignored.");
+            return;
+        }
+
 #ifdef USE_AWS_MEMORY_MANAGEMENT
         if(options.memoryManagementOptions.memoryManager)
         {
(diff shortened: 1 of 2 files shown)


src/aws-cpp-sdk-core/source/Aws.cpp
```diff
@@ -172,6 +184,20 @@ namespace Aws
 
     void ShutdownAPI(const SDKOptions& options)
     {
+        std::unique_lock<std::mutex> lock(s_initShutdownMutex);
+        if(s_initCount != 1)
+        {
+            if(!s_initCount) {
+                AWS_LOGSTREAM_ERROR(ALLOCATION_TAG, "Unable to ShutdownAPI of AWS-SDK-CPP: the SDK was not initialized.");
```
SergeyRyabinin, line 191: I have no idea how this one is going to be actually traced (without an initialized logger)
I don't want to add an assert here though.

src/aws-cpp-sdk-core/source/Aws.cpp
```diff
@@ -26,8 +26,20 @@ namespace Aws
...
     static const char* ALLOCATION_TAG = "Aws_Init_Cleanup";
 
+    static std::mutex s_initShutdownMutex;
+    static size_t s_initCount = 0;
+
     void InitAPI(const SDKOptions &options)
     {
+        std::unique_lock<std::mutex> lock(s_initShutdownMutex);
+        s_initCount += 1;
+        if(s_initCount != 1)
```
sbiscigl, line 36: could we use an atomic int instead of a mutex?

sbiscigl (approved):

src/aws-cpp-sdk-core/source/Aws.cpp
SergeyRyabinin: I also wanted to prevent simultaneous Init/Shutdown calls.

PR #2714, opened 2023-10-16
Fix awsmigrationhub build error

LilyWangLL:
*Issue #, if available:*

*Description of changes:*
Change project `AWSMigrationHub` name from `aws-cpp-sdk-AWSMigrationHub` to `aws-cpp-sdk-awsmigrationhub`, fix build error:
```
CMake Error at cmake/dependencies.cmake:16 (get_target_property):
  get_target_property() called with non-existent target
  "aws-cpp-sdk-awsmigrationhub".
Call Stack (most recent call first):
  cmake/dependencies.cmake:73 (compute_links)
  cmake/sdks.cmake:288 (sort_links)
  CMakeLists.txt:328 (add_sdks)

CMake Error at cmake/dependencies.cmake:84 (get_target_property):
  get_target_property() called with non-existent target
  "aws-cpp-sdk-awsmigrationhub".
Call Stack (most recent call first):
  cmake/sdks.cmake:288 (sort_links)
  CMakeLists.txt:328 (add_sdks)
```

I'm not sure if the directory name of the project `aws-cpp-sdk-AWSMigrationHub` also needs to be changed to lowercase. Can you help confirm? cc @sdavtaker

*Check all that applies:*

- [ ] Linux
- [x] Windows
- [ ] Android
- [x] MacOS
- [ ] IOS
- [ ] Other Platforms

### DIFF
--- a/generated/src/aws-cpp-sdk-AWSMigrationHub/CMakeLists.txt
+++ b/generated/src/aws-cpp-sdk-AWSMigrationHub/CMakeLists.txt
@@ -1,4 +1,4 @@
-add_project(aws-cpp-sdk-AWSMigrationHub "C++ SDK for the AWS AWSMigrationHub service" aws-cpp-sdk-core)
+add_project(aws-cpp-sdk-awsmigrationhub "C++ SDK for the AWS AWSMigrationHub service" aws-cpp-sdk-core)
 
 file(GLOB AWS_AWSMIGRATIONHUB_HEADERS
     "include/aws/AWSMigrationHub/*.h"


sbiscigl:
So with regard to how things are named and why inconsistencies exist, everything in the directory `generated` is generated by our [model codegen](https://github.com/aws/aws-sdk-cpp/tree/main/tools/code-generation/generator). When i say "model" that is [every single aws service provides us with a model file](https://github.com/aws/aws-sdk-cpp/tree/main/tools/code-generation/api-descriptions) and we use those to auto generate the directory names and class names. 

[aws migration hub specifically can be found here](https://github.com/aws/aws-sdk-cpp/blob/main/tools/code-generation/api-descriptions/AWSMigrationHub-2017-05-31.normal.json)

and in there we can see `"targetPrefix":"AWSMigrationHub",` which is why this is named the way it is. So if it were to change anywhere it would change here.

We however do not own this configuration and it is owned by the migration hub service and is provided to us. We try not to make one-off code generation changes because we then have to support a specific service by service customization going forward. We also cant change the name of this currently because people may already rely on it and changing it would create a backward incompatible change.

With the concern of backwards compatibility breakage, and the fact the model comes from upstream i'm against changing this to support a branching model of how we handle the name of the directory. That said I do understand why the inconsistency is not fun.

closing the PR but if you want to talk more about how code gen effects the names of things feel free to re-open and reply.

dg0yt:
> I think the bug is in the vcpkg port.

Now I can confirm that this is a vcpkg bug.

Issue #2717, opened 2023-10-17
CMake Config mode support for Curl


pr0g:
### Describe the bug

Right now it does not seem possible to use a custom build of Curl built with CMake when it is installed in Config mode.

When building aws-sdk-cpp, CURL will be found in Config mode, but the include directories and library variables are not populated.

```
...
-- Http client: Curl
-- Found CURL: /path/to/build/lib/cmake/CURL/CURLConfig.cmake (found version "8.2.1-DEV")  
--   Curl include directory: 
--   Curl library: 
...
```

This is because in Config mode the variables `${<library>_INCLUDE_DIRS}` and `${<library>_LIBRARIES}` are not populated (in this case `${CURL_INCLUDE_DIRS}` and `${CURL_LIBRARIES}`).

There is a workaround for this (please see this experimental draft PR - https://github.com/aws/aws-sdk-cpp/pull/2718) where we use the CMake `find_package` command to find Curl, and then link against the Curl target in `aws-sdk-cpp/src/aws-cpp-sdk-core/CMakeLists.txt` 

```cmake
# cmake/external_dependencies.cmake
...
  find_package(CURL)
  if(NOT CURL_FOUND)
      message(FATAL_ERROR "Could not find curl")
  else()
...
```

```cmake
# aws-cpp-sdk-core/CMakeLists.txt
...
target_link_libraries(
    ${PROJECT_NAME} PRIVATE 
    ${PLATFORM_DEP_LIBS} ${CLIENT_LIBS} ${CRYPTO_LIBS}
    ${AWS_SDK_ADDITIONAL_LIBRARIES}
    CURL::libcurl) # <-- added
...
```

I realize this is a super hacky change and is not fit for purpose as is, but it proves that this can work with a fairly minor modification.

### Expected Behavior

aws-sdk-cpp can be built when using a build of CURL in Config mode (see [CMake Finding packages](https://cmake.org/cmake/help/book/mastering-cmake/chapter/Finding%20Packages.html#finding-packages) for details).

### Current Behavior

Currently aws-sdk-cpp does not build when using a build of CURL in Config mode.

### Reproduction Steps

To reproduce this issue (tested on macOS) you can use this simple CMake project:

```cmake
# CMakeLists.txt

cmake_minimum_required(VERSION 3.22)

project(aws-cpp-sdk-external)

include(ExternalProject)
include(ProcessorCount)

ProcessorCount(CPU_COUNT)

# Custom build of OpenSSL
[...]

PR #2718, opened 2023-10-17
Experimental change to switch to using find_package and linking against CURL::libcurl target

pr0g:
#2717

*Description of changes:*

Experimental change accompanying issue #2717

*Check all that applies:*

_This PR is not intended to be merged as is, hence the 'draft' status, it is more for illustrative purposes to help discuss a solution for https://github.com/aws/aws-sdk-cpp/issues/2717_

- [ ] Linux
- [ ] Windows
- [ ] Android
- [x] MacOS
- [ ] IOS
- [ ] Other Platforms


pr0g:
Thanks for following up, would running all tests be a way to check if this could be updated to work for everyone? It could be made an opt-in behaviour to start with? The issue referenced is still a problem without a change such as this one.

jmklix:
Not necessarily, because there are additional tests that we run internally before releasing

pr0g:
Understood, hopefully I'll have some time to come back to this and will have a more robust proposal. Right now I'll just maintain [this fork](https://github.com/pr0g/aws-sdk-cpp/tree/experiment-remove-find-library) which solves the issues I was having with building and depending on the library.

PR #2763, opened 2023-11-20
Compute if shapes contain each other through their all ancestors too

SergeyRyabinin:
*Issue #, if available:*
n/a
*Description of changes:*
Current check for "mutually referenced" does not go beyond first level of shape composition, so use BFS to collect all and set intersection.

Also, memoize the result of shape names collection as this method is called quite many times.

*Check all that applies:*

- [ ] Linux
- [ ] Windows
- [ ] Android
- [ ] MacOS
- [ ] IOS
- [ ] Other Platforms

### DIFF
--- a/tools/code-generation/generator/src/main/java/com/amazonaws/util/awsclientgenerator/domainmodels/codegeneration/Shape.java
+++ b/tools/code-generation/generator/src/main/java/com/amazonaws/util/awsclientgenerator/domainmodels/codegeneration/Shape.java
@@ -10,6 +10,7 @@ import lombok.Builder;
 import lombok.Data;
 import lombok.NoArgsConstructor;
 
+import java.util.HashMap;
 import java.util.LinkedHashMap;
 import java.util.List;
 import java.util.Map;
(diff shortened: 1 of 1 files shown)


tools/code-generation/generator/src/main/java/com/amazonaws/util/awsclientgenerator/domainmodels/codegeneration/Shape.java
```diff
@@ -219,23 +220,49 @@ public String toString() {
       return "null";
     }
 
+    private void getAllChildrenShapes(Map<String, Shape> collected) {
+        if (members == null || members.isEmpty()) {
+            return;
+        }
+
+        Map<String, Shape> thisChildren = members.values().parallelStream()
+                .map(value -> value.getShape())
```
sbiscigl, line 229: nit: could be made
```suggestion
                .map(ShapeMember::getShape)
```

tools/code-generation/generator/src/main/java/com/amazonaws/util/awsclientgenerator/domainmodels/codegeneration/Shape.java
```diff
@@ -219,23 +220,49 @@ public String toString() {
...
+                .filter(shape -> !collected.containsKey(shape.getName()))
+                .map(memberShape -> memberShape.isList() ? memberShape.getListMember().getShape() : memberShape) //if references through a list container
+                .collect(Collectors.toMap(
+                        shape -> shape.getName(), shape -> shape,
```
sbiscigl, line 233: same nit
```suggestion
                        ShapeMember::getShape, shape -> shape,
```

tools/code-generation/generator/src/main/java/com/amazonaws/util/awsclientgenerator/domainmodels/codegeneration/Shape.java
```diff
@@ -219,23 +220,49 @@ public String toString() {
...
+    private Set<String> getThisAndAllChildrenShapeNames() {
+        if (thisAndAllChildrenShapeNamesMemoized != null && !thisAndAllChildrenShapeNamesMemoized.isEmpty()) {
+            // this method is called for at least shape count ^ 2 times
+            return thisAndAllChildrenShapeNamesMemoized;
+        }
+        Map<String, Shape> shapes = new HashMap<String, Shape>();
+
+        getAllChildrenShapes(shapes);
+
+        thisAndAllChildrenShapeNamesMemoized = new HashSet<>(shapes.keySet());
```
sbiscigl, line 253: nit: instead of creating then assigning it, we can just make it a instance of immutable set

```suggestion
      thisAndAllChildrenShapeNamesMemoized = ImmutableSet.<String>builder()
                .addAll(shapes.keySet()).
                add(getName())
                .build();
        return thisAndAllChildrenShapeNamesMemoized;
```

sbiscigl (approved):

PR #2778, opened 2023-12-04
Updates for General HTTP Credentials provider 

SergeyRyabinin:
(previously aka TaskRole credentials provider)

*Issue #, if available:*
aligning with other SDKs
*Description of changes:*

*Check all that applies:*

- [x] Linux
- [x] Windows
- [ ] Android
- [ ] MacOS
- [ ] IOS
- [ ] Other Platforms

### DIFF
--- a/src/aws-cpp-sdk-core/include/aws/core/auth/GeneralHTTPCredentialsProvider.h
+++ b/src/aws-cpp-sdk-core/include/aws/core/auth/GeneralHTTPCredentialsProvider.h
@@ -1,0 +1,120 @@
+/**
+ * Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.
+ * SPDX-License-Identifier: Apache-2.0.
+ */
+
+
+#pragma once
+
+#include <aws/core/auth/AWSCredentialsProvider.h>
+
+namespace Aws
+{
+    namespace Auth
+    {
+        /**
+        * General HTTP Credentials Provider (previously known as ECS credentials provider)
+        * implementation that loads credentials from an arbitrary HTTP(S) endpoint specified by the environment
+        * or loop back / Amazon ECS / Amazon EKS container host metadata services by default.
+        */
+        class AWS_CORE_API GeneralHTTPCredentialsProvider : public AWSCredentialsProvider
+        {
+        private:
+            static bool ShouldCreateGeneralHTTPProvider(const Aws::String& relativeUri, const Aws::String& absoluteUri, const Aws::String authToken);
+        public:
+            using ShouldCreateFunc = std::function<bool(const Aws::String& relativeUri, const Aws::String& absoluteUri, const Aws::String authToken)>;
+
+            /**
+             * Initializes the provider to retrieve credentials from a general http provided endpoint every 5 minutes or before it
+             * expires.
+             * @param relativeUri A path appended to the metadata service endpoint. OR
+             * @param absoluteUri The full URI to resolve to get credentials.
+             * @param authTokenFilePath A path to a file with optional authorization token passed to the URI via the 'Authorization' HTTP header.
+             * @param authToken An optional authorization token passed to the URI via the 'Authorization' HTTP header.
+             * @param refreshRateMs The number of milliseconds after which the credentials will be fetched again.
+             * @param ShouldCreateFunc
+             */
+            GeneralHTTPCredentialsProvider(const Aws::String& relativeUri,
+                                           const Aws::String& absoluteUri,
+                                           const Aws::String& authTokenFilePath = "",
+                                           const Aws::String& authToken = "",
+                                           long refreshRateMs = REFRESH_THRESHOLD,
+                                           ShouldCreateFunc shouldCreateFunc = ShouldCreateGeneralHTTPProvider);
+
+            /**
+             * Check if GeneralHTTPCredentialsProvider was initialized with allowed configuration
+             * @return true if provider configuration is valid
+             */
+            bool IsValid() const
+            {
+                if (!m_ecsCredentialsClient)
+                    return false;
+                if (!m_authTokenFilePath.empty())
+                    return !LoadTokenFromFile().empty();
+                return true;
+            }
+
+            /**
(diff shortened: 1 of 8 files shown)


src/aws-cpp-sdk-core/include/aws/core/auth/GeneralHTTPCredentialsProvider.h
```diff
@@ -0,0 +1,105 @@
...
+            static const char AWS_CONTAINER_AUTHORIZATION_TOKEN[];
+
+        protected:
+            void Reload() override;
+
+        private:
+            bool ExpiresSoon() const;
+            void RefreshIfExpired();
+
+            std::shared_ptr<Aws::Internal::ECSCredentialsClient> m_ecsCredentialsClient;
```
sbiscigl, line 95: nit: should we do something like

```
using GeneralHttpClient = ECSCredentialsClient;
```

like you down below now that after this is, this really doesnt have anything to do with ecs anymore?

src/aws-cpp-sdk-core/source/auth/GeneralHTTPCredentialsProvider.cpp
```diff
@@ -0,0 +1,204 @@
...
+    m_loadFrequencyMs(refreshRateMs)
+{
+    if (ShouldCreateGeneralHTTPProvider(relativeUri, absoluteUri, authToken))
+    {
+        AWS_LOGSTREAM_INFO(GEN_HTTP_LOG_TAG, "Creating GeneralHTTPCredentialsProvider with refresh rate " << refreshRateMs);
+        if (!relativeUri.empty()) {
+            m_ecsCredentialsClient = Aws::MakeShared<Aws::Internal::ECSCredentialsClient>(GEN_HTTP_LOG_TAG, relativeUri.c_str(), "", authToken.c_str());
+        } else if (!absoluteUri.empty()) {
+            m_ecsCredentialsClient = Aws::MakeShared<Aws::Internal::ECSCredentialsClient>(GEN_HTTP_LOG_TAG, "", absoluteUri.c_str(), authToken.c_str());
+        }
```
sbiscigl, line 110: nit: should we add an error or assertion that one of those urls has to exist?

src/aws-cpp-sdk-core/source/auth/GeneralHTTPCredentialsProvider.cpp
```diff
@@ -0,0 +1,204 @@
...
+        {
+            return true;
+        }
+
+        Aws::Http::URI absUri(absoluteUri);
+        const Aws::String& authority = absUri.GetAuthority();
+
+        // Otherwise, implementations MUST fail to resolve when the URI hostname does not satisfy any of the following conditions
+        // address is an ECS / EKS container host
+        if (authority == "169.254.170.2") // ECS container host
```
sbiscigl, line 60: nit: add constants for string that seem random
169.254.170.2 -> `ECS_CONTAINER_HOST`
169.254.170.23 -> `EKS_CONTAINER_HOST`
fd00:ec2::23 -> `EKS_IPV6`

src/aws-cpp-sdk-core/source/auth/GeneralHTTPCredentialsProvider.cpp
```diff
@@ -0,0 +1,204 @@
...
+                                                               const Aws::String& authToken,
+                                                               const Aws::String& authTokenFilePath,
+                                                               long refreshRateMs) :
+    m_authTokenFilePath(authTokenFilePath),
+    m_loadFrequencyMs(refreshRateMs)
+{
+    if (ShouldCreateGeneralHTTPProvider(relativeUri, absoluteUri, authToken))
+    {
+        AWS_LOGSTREAM_INFO(GEN_HTTP_LOG_TAG, "Creating GeneralHTTPCredentialsProvider with refresh rate " << refreshRateMs);
+        if (!relativeUri.empty()) {
```
sbiscigl, line 106: should we move this to a factory function and move the ecs client into the initializer list and the logic whether or not we should create a client into the factory function?

sbiscigl (approved):

src/aws-cpp-sdk-core/source/auth/GeneralHTTPCredentialsProvider.cpp
SergeyRyabinin: I thought about it and even implemented this way on the first attempt.
But 
1. it won't put these checks on existing use-cases.
2. or will have to perform these checks twice.

src/aws-cpp-sdk-core/include/aws/core/auth/GeneralHTTPCredentialsProvider.h
sbiscigl: more so im asking if we should create an alias for this now because the name "ecsClient" doesnt reflect that its not actually calling ECS nessecarily.

src/aws-cpp-sdk-core/source/auth/GeneralHTTPCredentialsProvider.cpp
```diff
@@ -0,0 +1,261 @@
...
+        const Aws::String& authority = absUri.GetAuthority();
+
+        // Otherwise, implementations MUST fail to resolve when the URI hostname does not satisfy any of the following conditions
+        if (IsAllowedIp(authority))
+        {
+            return true;
+        }
+
+        Aws::Crt::Io::HostResolver* pHostResolver = Aws::Crt::ApiHandle::GetOrCreateStaticDefaultHostResolver();
+        if (pHostResolver)
```
sbiscigl, line 103: should we create a configuration option to disable DNS lookup incase a existing customer is using this and DNS lookup break their use case.

SergeyRyabinin: added option to override the restriction to use HTTPs / pre-allowed addresses only.

sbiscigl (approved):

PR #2786, opened 2023-12-12
Rename fields that clash with built-in request methods

mswaanen:
*Issue #, if available:*

*Description of changes:*
We do this beacuse the C++ SDK has a design flaw: Model fields and request fields are mixed
 in a single object. Practically this means that if a request has a field called `headers`, the SDK will
generate a method called `GetHeaders`, which clashes with the `GetHeaders` method to get access to the
HTTP request headers.

This method will substitute `headers` for `headerValues` so we end up with a `GetHeaderValues`
method. This is still going to be confusing for the end user but it's the best I can think of right now.

*Check all that applies:*
- [ ] Added proper tests to cover this PR. *The whole shape renaming code seems untested.*

- [x] Linux
- [ ] Windows
- [ ] Android
- [ ] MacOS
- [ ] IOS
- [ ] Other Platforms

### DIFF
--- a/tools/code-generation/generator/src/main/java/com/amazonaws/util/awsclientgenerator/transform/C2jModelToGeneratorModelTransformer.java
+++ b/tools/code-generation/generator/src/main/java/com/amazonaws/util/awsclientgenerator/transform/C2jModelToGeneratorModelTransformer.java
@@ -14,6 +14,9 @@ import com.fasterxml.jackson.core.JsonProcessingException;
 import com.fasterxml.jackson.databind.JsonNode;
 import com.fasterxml.jackson.databind.ObjectMapper;
 import com.google.common.collect.ImmutableList;
+import com.google.common.collect.ImmutableMap;
+import com.google.common.collect.ImmutableSet;
+import lombok.Value;
 import org.apache.commons.lang.WordUtils;
 
 import java.util.*;
@@ -99,6 +102,21 @@ public class C2jModelToGeneratorModelTransformer {
         LEGACY_SERVICE_IDS.add("transcribestreaming");
         LEGACY_SERVICE_IDS.add("dynamodbstreams");
     }
+
+    /**
+     * Type representing what a member should be remapped to and services that it
+     * cannot be renamed to, to preserve backwards compat.
+     */
+    @Value
+    private static class MemberMapping {
+        String remappingName;
+        Set<String> servicesToSkip;
+    }
+
+    private static final Map<String, MemberMapping> RESERVED_REQUEST_MEMBER_MAPPING = ImmutableMap.of(
+        "body", new MemberMapping("requestBody", ImmutableSet.of("amplifyuibuilder", "apigateway", "apigateway2", "bedrock-runtime", "glacier", "repostspace")),
+        "headers", new MemberMapping("headerValues", ImmutableSet.of("apigateway"))
+    );
 
     /**
      * There was a bug with namespace collision detection where customers
@@ -665,6 +683,7 @@ public class C2jModelToGeneratorModelTransformer {
             newName = baseName + suffixIt.next();
 
             if (shape.getName().equals(newName)) {
+                renameReservedFields(shape);
                 return shape;
             }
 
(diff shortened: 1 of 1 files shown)


tools/code-generation/generator/src/main/java/com/amazonaws/util/awsclientgenerator/transform/C2jModelToGeneratorModelTransformer.java
```diff
@@ -703,10 +704,21 @@ Shape renameShape(Shape shape, String baseName, List<String> suffixOptions) {
 
         Shape cloned = cloneShape(shape);
         cloned.setName(newName);
+        renameReservedFields(cloned);
         shapes.put(newName, cloned);
         return cloned;
     }
 
+    void renameReservedFields(Shape shape) {
```
sbiscigl, line 712: While i am onboard with transforming the member `headers` because as you state, it will cause a collision, there are several things that will need to be dealt with. 

namely that [api-gateway already has a non-standard version of this](https://github.com/aws/aws-sdk-cpp/blob/main/tools/code-generation/generator/src/main/java/com/amazonaws/util/awsclientgenerator/generators/cpp/apigateway/APIGatewayRestJsonCppClientGenerator.java#L37-L41). so we have to preserve backwards compat for them while creating a standardized way going forward.

Also ideally we want to view this change as "configuration" so that we could simply add string values going forward without having to hardcode anything. 

taking a stab at what im talking about here what we could to is create set of values with replacement values because we know they will cause a collision, alongside service that we know we cannot make the change in, i.e.
```
/**
 * Typle representing what a member should be remapped to and services that it
 * cannot be renamed to, to preserve backwards compat.
 */
@Value
private static class MemberMapping {
    String remappingName;
    Set<String> servicesToSkip;
}

private static final Map<String, MemberMapping> RESERVED_REQUEST_MEMBER_MAPPING = ImmutableMap.of(
        "headers", new MemberMapping("headerValues", ImmutableSet.of("apigateway"))
);
```
which translate to "replace the member `headers` with `headerValues` exceptin the service apigateway because that service already has a customization".

we could then invoke it later on in your function

```
void renameReservedFields(Shape shape) {
    if (shape.getName().endsWith("Request")) {
        final Map<String, ShapeMember> members = shape.getMembers();
        RESERVED_REQUEST_MEMBER_MAPPING.entrySet().stream()
                .filter(reservedMapping -> members.containsKey(reservedMapping.getKey()))
                .filter(reservedMapping -> !reservedMapping.getValue().servicesToSkip.contains(c2jServiceModel.getServiceName()))
                .forEach(reservedMapping -> {
                    final ShapeMember member = members.get(reservedMapping.getKey());
                    renameShapeMember(shape,
                            reservedMapping.getKey(),
                            member.getShape().getName(),
                            reservedMapping.getValue().remappingName,
                            member.getShape().getName(),
                            false);
                });
    }
}
```

so that we dont need to ever update the code, just the configuration of what is being replaced. also note that we are now calling `renameShapeMember` instead of custom logic so that we follow the same renaming rules as used in other places. note that we will have to change that function a little bit to be

```
void renameShapeMember(Shape parentShape,
                        String originalMemberKey, String originalShapeName,
                        String newMemberKey, String newShapeName, boolean isPayload) {
    if (!shapes.containsKey(originalShapeName)) {
        throw new NoSuchElementException("Shape to rename was not found in all shapes: " + originalShapeName);
    }
    if (!parentShape.getMembers().containsKey(originalMemberKey)) {
        throw new NoSuchElementException("Requested to rename non-existent child shape key "
                + originalMemberKey + " of a Shape type " + originalShapeName);
    }
    if (!Objects.equals(originalShapeName, newShapeName)) {
        shapes.get(originalShapeName).setName(newShapeName);
        shapes.put(newShapeName, shapes.get(originalShapeName));
        shapes.remove(originalShapeName);
    }
    if (!Objects.equals(originalMemberKey, newMemberKey)) {
        parentShape.getMembers().put(newMemberKey, parentShape.getMembers().get(originalMemberKey));
        parentShape.RemoveMember(originalMemberKey);
    }
    if (isPayload)
    {
        parentShape.setPayload(newMemberKey);
    }
}
```

to account for whent he shapes/members have the same so ti is a no-op.

the full changes of what im [suggesting can be seen in this diff](https://gist.github.com/sbiscigl/e7de2e5d8cf04136585b08a08e217228)

let me know what you think.

mswaanen: That actually looks quite okay. I've updated the PR to be just that with the addition of adding a remap for body, as that's likely to be a collision in the future.

sbiscigl (commented):
Once the java compile errors and backwards compat issues are resovled, lgtm. also if you could please squash the commits

sbiscigl:
moved to [pull/2792](https://github.com/aws/aws-sdk-cpp/pull/2792) and was merged from there, resolving.

PR #2798, opened 2023-12-21
fix s3 copy constructor

sbiscigl:
*Description of changes:*

There is currently a bug where if you copy construct a S3Client it will use the default credential provider chain instead of a specified credential change. This changes the copy ctor to re-create the signers with previously provided credential provider.

*Check all that applies:*

- [x] Linux
- [x] Windows
- [ ] Android
- [x] MacOS
- [ ] IOS
- [ ] Other Platforms

### DIFF
--- a/src/aws-cpp-sdk-core/source/auth/signer-provider/AWSAuthSignerProviderBase.cpp
+++ b/src/aws-cpp-sdk-core/source/auth/signer-provider/AWSAuthSignerProviderBase.cpp
@@ -1,0 +1,12 @@
+/**
+* Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.
+ * SPDX-License-Identifier: Apache-2.0.
+ */
+#include <aws/core/auth/signer-provider/AWSAuthSignerProviderBase.h>
+#include <aws/core/auth/AWSCredentialsProviderChain.h>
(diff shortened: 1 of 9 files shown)


src/aws-cpp-sdk-core/source/auth/signer-provider/AWSAuthSignerProviderBase.cpp
```diff
@@ -0,0 +1,10 @@
+/**
+* Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.
+ * SPDX-License-Identifier: Apache-2.0.
+ */
+#include <aws/core/auth/signer-provider/AWSAuthSignerProviderBase.h>
+#include <aws/core/auth/AWSCredentialsProviderChain.h>
+
+std::shared_ptr<AWSCredentialsProvider> AWSAuthSignerProvider::GetCredentialsProvider() const {
+  return MakeShared<DefaultAWSCredentialsProviderChain>("AWSAuthSignerProvider");
```
sbiscigl, line 9: In a perfect world i would want to make the function pure virutal, but for backwards compat we can return the default credential provider.

DmitriyMusatkin (approved):

Issue #2814, opened 2024-01-17
Add SDKOption for default Credentials Provider Chain


MitchellKagerKodiak:
### Describe the feature

Every service client accepts an `AwsCredentialsProviderChain` that determines how credentials are resolved. It should be possible to configure a default during SDK initialization.

### Use Case

`STSProfileCredentialsProvider` does not exist in default chain [[150](https://github.com/aws/aws-sdk-cpp/issues/150)]. In certain codebases, it would be much more code-efficient to configure the credentials provider chain during initialization as opposed to passing a chain to construction of every client object.

### Proposed Solution

* Add a new `CredentialsProviderChainOptions` struct to `Aws.h`'s SDKOptions
* During `InitSDK`, check for presence of this struct's closure. If it is present, call some `SetDefaultCredentialsProviderChain` with the result of the closure. If it is not present, call `Set...Chain` method with `DefaultAWSCredentialsProviderChain` (current default)
* Update all occurrences of `DefaultAWSCredentialsProviderChain` across clients with `GetDefaultCredentialsProviderChain` (to resolve to configured value or pre-existing `DefaultAWSCredentialsProviderChain`).

### Other Information

posting a PR right now

### Acknowledgements

- [X] I may be able to implement this feature request
- [ ] This feature might incur a breaking change

jmklix:
Thanks for opening this feature request and creating a PR to add this option. We are reviewing your PR to see if we would like to accept this changes and change how the credentials work.

jmklix:
Thanks for taking the time and effort to make this PR. But we don't want to accept these changes, as we have plans to rework the Credentials Provider.

MitchellKagerKodiak:
hey @jmklix , have you had a chance to rework the credentials provider? I still have a use case where I want to be able to easily reconfigure the default credentials provider against dozens of processes and I don't want to have to manually modify every single place I am constructing SDK clients. my processes generally construct SDK clients indirectly via wrappers and it would be incredibly tedious to modify the signature of every wrapper to pass a credentials provider through. it would be much more convenient to have each client configure this when they SDK::InitApi.

Issue #150, opened 2016-04-06
Add support for assuming a role via a profile


ltn100:
Currently the C++ SDK supports using credentials from the `~/.aws/credentials` file, and different profiles can be selected by setting the `AWS_DEFAULT_PROFILE` environment variable. However, if the profile assumes a role (as [supported by the CLI](http://docs.aws.amazon.com/cli/latest/userguide/cli-roles.html)) then the access/secret keys become invalid (the `role_arn` and `source` are not currently supported).

An excerpt from the logs shows:

```
[DEBUG] 2016-04-06 14:15:34 ProfileConfigFileAWSCredentialsProvider [140635887626240] Found profile Dev
[TRACE] 2016-04-06 14:15:34 ProfileConfigFileAWSCredentialsProvider [140635887626240] Found property role_arnfor profile Dev
[TRACE] 2016-04-06 14:15:34 ProfileConfigFileAWSCredentialsProvider [140635887626240] Found property source_profilefor profile Dev
[INFO] 2016-04-06 14:15:34 ProfileConfigFileAWSCredentialsProvider [140635887626240] Access key for profile not found.
[INFO] 2016-04-06 14:15:34 ProfileConfigFileAWSCredentialsProvider [140635887626240] Secret key for profile not found.
[INFO] 2016-04-06 14:15:34 ProfileConfigFileAWSCredentialsProvider [140635887626240] Optional session token for profile not found.
```

JonathanHenson:
This is coming incredibly soon. Sorry for the delay.

clechasseur:
This remains open, but is this supposed to be implemented now? I'm having a problem that looks similar, i.e. my `.aws/credentials` file is set up with a `source_profile` and `role_arn`, but I always get `AccessDenied` when performing S3 operations through the C++ SDK, whereas the same operations succeed with the CLI.

pavanjains:
I am having the same problem, is this looked into?

mdashti:
Can you give an update on this feature? Is there any other way to authenticate users via `role_arn` (even if it should be done explicitly and not still available in the default auth chain)?

LantosDavid:
We wrote the following temporary workaround on our end:
```
// 'getProfileToUse' is copied from the ProfileConfigFileAWSCredentialsProvider constructor in the
// AWS C++ sdk (AWSCredentialsProvider.cpp)
Aws::String getProfileToUse(void)
{
[...]

PR #2815, opened 2024-01-17
Add SDKOption for default Credentials Provider Chain

MitchellKagerKodiak:
*Issue #, if available:* related to [150](https://github.com/aws/aws-sdk-cpp/issues/150); this is kind of a workaround whereby developer can set their default chain to include `STSProfileCredentialsProvider`. Just opened an [issue / feature request 2814](https://github.com/aws/aws-sdk-cpp/issues/2814).

*Description of changes:* If provided in `SDKOption`, initialization will resolve closure and set that value in static variable. If not provided, configure static variable with `DefaultAWSCredentialsProviderChain`. Most usages of `DefaultAWSCredentialsProviderChain` throughout clients has been replaced with `GetDefaultCredentialsProviderChain`.

I could use some help writing test cases, and I probably missed a couple places where I need to replace `DefaultAWSCredentialsProviderChain`.

I know this looks tedious to review; but, all of the logic changes are in `src/aws-cpp-sdk-core`. Everything outside of that is just changing existing usages of `DefaultAWSCredentialsProviderChain` => `GetDefaultCredentialsProviderChain` across the services' clients.

*Check all that applies:*

- [x] Linux
- [ ] Windows
- [ ] Android
- [ ] MacOS
- [ ] IOS
- [ ] Other Platforms

### DIFF
--- a/generated/src/aws-cpp-sdk-AWSMigrationHub/source/MigrationHubClient.cpp
+++ b/generated/src/aws-cpp-sdk-AWSMigrationHub/source/MigrationHubClient.cpp
@@ -59,7 +59,7 @@ MigrationHubClient::MigrationHubClient(const MigrationHub::MigrationHubClientCon
                                        std::shared_ptr<MigrationHubEndpointProviderBase> endpointProvider) :
   BASECLASS(clientConfiguration,
             Aws::MakeShared<AWSAuthV4Signer>(ALLOCATION_TAG,
-                                             Aws::MakeShared<DefaultAWSCredentialsProviderChain>(ALLOCATION_TAG),
+                                             GetDefaultCredentialsProviderChain(),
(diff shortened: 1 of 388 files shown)


jmklix:
Thanks for taking the time and effort to make this PR. But we don't want to accept these changes, as we have plans to rework the Credentials Provider.

MitchellKagerKodiak:
hey @jmklix , have you had a chance to rework the credentials provider? I still have a use case where I want to be able to easily reconfigure the default credentials provider against dozens of processes and I don't want to have to manually modify every single place I am constructing SDK clients. my processes generally construct SDK clients indirectly via wrappers and it would be incredibly tedious to modify the signature of every wrapper to pass a credentials provider through. it would be much more convenient to have each client configure this when they SDK::InitApi.

PR #2816, opened 2024-01-19
Fix transcribe tests and sample

SergeyRyabinin:
*Issue #, if available:*
many
*Description of changes:*
 - Do not close output stream on input stream close (fixes lost last frame);
 - Increase H2 encoder buffer size (fixes poor performance / jittering / frequent buffer underflows);
 - Do not block indefinitely on encoder iostream buffer underflow().
*Check all that applies:*

- [x] Linux
- [ ] Windows
- [ ] Android
- [x] MacOS
- [ ] IOS
- [ ] Other Platforms


src/aws-cpp-sdk-core/source/utils/stream/ConcurrentStreamBuf.cpp
```diff
@@ -67,19 +93,32 @@ namespace Aws
...
+                bool closeStream = false;
                 {
                     std::unique_lock<std::mutex> lock(m_lock);
-                    m_signal.wait(lock, [this]{ return m_backbuf.empty() == false || m_eof; });
 
-                    if (m_eof && m_backbuf.empty())
+                    if (!m_eofInput)
                     {
-                        return std::char_traits<char>::eof();
+                        m_signal.wait_for(lock, std::chrono::milliseconds(1000), [this] { return m_backbuf.empty() == false || m_eofInput; });
```
SergeyRyabinin, line 102: this is technically non-compliant with std::
underflow must fill get buffer or return EOF, 

but we may have nothing to fill and blocking will make things worse;
EOF will terminate the HTTP2 conversation too.

src/aws-cpp-sdk-core/include/aws/core/utils/DateTime.h
```diff
@@ -139,6 +139,11 @@ namespace Aws
             */
             Aws::String ToGmtString(const char* formatStr) const;
 
+            /**
+            * Convert dateTime to GMT time string using predefined format.
+            */
+            Aws::String ToGmtStringWithMs() const;
```
SergeyRyabinin, line 145: this is not used but I found it very useful for dev debug tracing

src/aws-cpp-sdk-core/source/utils/stream/ConcurrentStreamBuf.cpp
```diff
@@ -25,15 +27,106 @@ namespace Aws
...
+                    }
                 }
                 m_signal.notify_all();
             }
 
+            bool ConcurrentStreamBuf::WaitForDrain(int64_t timeoutMs)
+            {
+                const auto waitStarted = std::chrono::steady_clock::now();
+
+                // timed FlushPutArea
```
sbiscigl, line 71: would maybe be useful to create a "timed action" utility template. you do three timed actions here and im sure we will need to do something like that in future again

sbiscigl (approved):

DmitriyMusatkin (approved):

PR #2817, opened 2024-01-19
use curl in config mode if include and lib missing

sbiscigl:
*Issue #, if available:*

[issues/2717](https://github.com/aws/aws-sdk-cpp/issues/2717)

*Description of changes:*

As mentioned in the issue

```
Right now it does not seem possible to use a custom build of Curl built with CMake when it is installed in Config mode.
```

which is true, as of the moment if you build curl from source the only way you can is by building it with autoconf. This changes to fall back to use curl by configuration if we cannot find the needed include or lib dirs from `FindCURL`.

*Check all that applies:*

- [x] Linux
- [x] Windows
- [ ] Android
- [x] MacOS
- [ ] IOS
- [ ] Other Platforms

### DIFF
--- a/cmake/dependencies.cmake
+++ b/cmake/dependencies.cmake
@@ -12,6 +12,12 @@ function(compute_links lib)
         message(FATAL_ERROR "Circular dependency for library '${lib}'")
     endif()
 
+    #check if target is an alias and use that
+    get_property(aliased_target TARGET "${lib}" PROPERTY ALIASED_TARGET)
+    if(NOT "${aliased_target}" STREQUAL "")
+        set(lib "${aliased_target}")
+    endif()
+
     #interface libraries cannot be set
     get_target_property(target_type ${lib} TYPE)
     if ("INTERFACE_LIBRARY"  STREQUAL ${target_type})
--- a/cmake/external_dependencies.cmake
+++ b/cmake/external_dependencies.cmake
@@ -83,20 +83,28 @@ if(NOT NO_HTTP_CLIENT AND NOT USE_CRT_HTTP_CLIENT)
             include(FindCURL)
             if(NOT CURL_FOUND)
                 message(FATAL_ERROR "Could not find curl")
-            else()
-                message(STATUS "  Curl include directory: ${CURL_INCLUDE_DIRS}")
-                message(STATUS "  Curl library: ${CURL_LIBRARIES}")
             endif()
-            List(APPEND EXTERNAL_DEPS_INCLUDE_DIRS ${CURL_INCLUDE_DIRS})
+
+            # When built from source using cmake, curl does not include
+            # CURL_INCLUDE_DIRS or CURL_INCLUDE_DIRS so we need to use
+            # find_package to fix it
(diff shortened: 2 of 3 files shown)


cmake/external_dependencies.cmake
```diff
@@ -83,20 +83,29 @@ if(NOT NO_HTTP_CLIENT AND NOT USE_CRT_HTTP_CLIENT)
...
+            endif()
+
+            # When built from source using cmake, curl does not include
+            # CURL_INCLUDE_DIRS or CURL_INCLUDE_DIRS so we need to use
+            # find_package to fix it
+            if ("${CURL_INCLUDE_DIRS}" STREQUAL "" AND "${CURL_LIBRARIES}" STREQUAL "")
+                message(STATUS "Could not find curl include or library path, falling back to find with config.")
+                find_package(CURL)
+                set(CURL_LIBRARIES CURL::libcurl)
+                set(CLIENT_LIBS_ABSTRACT_NAME curl)
```
pr0g, line 95: nit: As this happens in both branches of the `if` it might be worth pulling it out as it always happens

cmake/dependencies.cmake
```diff
@@ -12,6 +12,12 @@ function(compute_links lib)
         message(FATAL_ERROR "Circular dependency for library '${lib}'")
     endif()
 
+    #check if target is a alias and use that
```
pr0g, line 15: uber nit: `a` -> `an`

cmake/external_dependencies.cmake
```diff
@@ -83,20 +83,29 @@ if(NOT NO_HTTP_CLIENT AND NOT USE_CRT_HTTP_CLIENT)
...
+            else ()
                 message(STATUS "  Curl include directory: ${CURL_INCLUDE_DIRS}")
                 message(STATUS "  Curl library: ${CURL_LIBRARIES}")
```
pr0g, line 98: suggestion: Is it worth moving this print (or at least `message(STATUS " Curl library: ${CURL_LIBRARIES}")`) after the if so in both configs we print the libraries?

pr0g (commented):
Changes LGTM! 😄 Thanks for adding support for this 🥳

SergeyRyabinin (approved):

cmake/external_dependencies.cmake
sbiscigl: i think the sentiment here is good but maybe we need to shuffle the messaging a little bit. 

`Curl library: CURL::libcurl` 

doesnt really make sense to a reader maybe we should change it to a single 

"Curl Target Link: ${CURL_LIBRARIES}"

which is more indicative of what is actually happening in the end. will update it to that, let me know what you think. ideally we should actually change the variable name `CURL_LIBRARIES` to something else but its referenced everywhere and unraveling it might lead to more issues than fixes.

pr0g: Good call 👍 Sounds like a plan

SergeyRyabinin (approved):

PR #2826, opened 2024-01-25
Checksum notes rebase

jmklix:
*Issue #, if available:*
https://github.com/aws/aws-sdk-cpp/pull/2644

*Description of changes:*
Fix checksum behavior when precalculated checksum is added

*Check all that applies:*

- [x] Linux
- [x] Windows
- [ ] Android
- [x] MacOS
- [ ] IOS
- [ ] Other Platforms

### DIFF
--- a/src/aws-cpp-sdk-core/include/aws/core/utils/crypto/Hash.h
+++ b/src/aws-cpp-sdk-core/include/aws/core/utils/crypto/Hash.h
@@ -47,14 +47,6 @@ namespace Aws
                  */
                 virtual HashResult GetHash() = 0;
 
-                /**
-                 * Return the base64 encoded checksum value
-                 */
-                virtual Aws::String GetHashBase64(ByteBuffer& m_checksum){
-                    Aws::String base64Encoded = HashingUtils::Base64Encode(m_checksum);
-                    return base64Encoded;
-                }
-
                 // when hashing streams, this is the size of our internal buffer we read the stream into
                 static const uint32_t INTERNAL_HASH_STREAM_BUFFER_SIZE = 8192;
             };
--- a/src/aws-cpp-sdk-core/include/aws/core/utils/crypto/Hash.h
+++ b/src/aws-cpp-sdk-core/include/aws/core/utils/crypto/Hash.h
@@ -9,6 +9,7 @@
 #include <aws/core/utils/crypto/HashResult.h>
 #include <aws/core/utils/memory/stl/AWSStreamFwd.h>
 #include <aws/core/utils/memory/stl/AWSString.h>
+#include <aws/core/utils/HashingUtils.h>
 
 namespace Aws
 {
@@ -45,6 +46,14 @@ namespace Aws
                  * Get the result in the current value
                  */
                 virtual HashResult GetHash() = 0;
+
+                /**
+                 * Return the base64 encoded checksum value
+                 */
+                virtual Aws::String GetHashBase64(ByteBuffer& m_checksum){
+                    Aws::String base64Encoded = HashingUtils::Base64Encode(m_checksum);
+                    return base64Encoded;
+                }
 
                 // when hashing streams, this is the size of our internal buffer we read the stream into
                 static const uint32_t INTERNAL_HASH_STREAM_BUFFER_SIZE = 8192;
--- a/src/aws-cpp-sdk-core/include/aws/core/utils/crypto/Hash.h
+++ b/src/aws-cpp-sdk-core/include/aws/core/utils/crypto/Hash.h
@@ -9,7 +9,6 @@
 #include <aws/core/utils/crypto/HashResult.h>
 #include <aws/core/utils/memory/stl/AWSStreamFwd.h>
 #include <aws/core/utils/memory/stl/AWSString.h>
-#include <aws/core/utils/HashingUtils.h>
 
 namespace Aws
 {
--- a/src/aws-cpp-sdk-core/include/aws/core/utils/crypto/PrecalculatedHash.h
+++ b/src/aws-cpp-sdk-core/include/aws/core/utils/crypto/PrecalculatedHash.h
@@ -8,11 +8,16 @@
 #include <aws/core/Core_EXPORTS.h>
 #include <aws/core/utils/crypto/Hash.h>
 #include <aws/core/utils/crypto/HashResult.h>
-#include "aws/core/utils/Outcome.h"
+#include <aws/core/utils/Outcome.h>
 
 namespace Aws {
     namespace Utils {
         namespace Crypto {
+            /**
+             * A hash implementation that stores a pre-calculated hash
+             * behind the Hash interface. It is a read through wrapper
+             * around the checksum value and no hashing are actually done.
+             */
             class AWS_CORE_API PrecalculatedHash : public Hash {
             public:
                 explicit PrecalculatedHash(const Aws::String &hash);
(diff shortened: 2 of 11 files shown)


src/aws-cpp-sdk-core/source/auth/signer/AWSAuthV4Signer.cpp
```diff
@@ -264,6 +265,7 @@ bool AWSAuthV4Signer::SignRequest(Aws::Http::HttpRequest& request, const char* r
         {
             payloadHash = STREAMING_UNSIGNED_PAYLOAD_TRAILER;
             Aws::String trailerHeaderValue = Aws::String("x-amz-checksum-") + request.GetRequestHash().first;
+            request.DeleteHeader(trailerHeaderValue.c_str());
```
DmitriyMusatkin, line 268: why is trailer being deleted?

src/aws-cpp-sdk-core/source/client/AWSClient.cpp
```diff
@@ -791,56 +792,69 @@ void AWSClient::AddChecksumToRequest(const std::shared_ptr<Aws::Http::HttpReques
...
+        const Aws::String checksumType = "x-amz-checksum-" + checksumAlgorithmName;
+        const HeaderValueCollection &headers = request.GetHeaders();
+        bool checksumValueAndAlgorithmProvided = headers.find(checksumType) != headers.end();
+
         // For non-streaming payload, the resolved checksum location is always header.
         // For streaming payload, the resolved checksum location depends on whether it is an unsigned payload, we let AwsAuthSigner decide it.
-        if (checksumAlgorithmName == "crc32")
+        if (request.IsStreaming() && checksumValueAndAlgorithmProvided)
+        {
+            auto hash = Aws::MakeShared<Crypto::PrecalculatedHash>(AWS_CLIENT_LOG_TAG,request.GetHeaders().find(checksumType)->second);
```
DmitriyMusatkin, line 807: nit: spacing inconsistent, do we need to run lint?

src/aws-cpp-sdk-core/include/aws/core/utils/crypto/Hash.h
```diff
@@ -46,6 +47,14 @@ namespace Aws
                  */
                 virtual HashResult GetHash() = 0;
 
+                /**
+                 * Return the base64 encoded checksum value
+                 */
+                virtual Aws::String GetHashBase64(ByteBuffer& m_checksum){
```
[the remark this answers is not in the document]
sbiscigl: removed

src/aws-cpp-sdk-core/source/client/AWSClient.cpp
sbiscigl: fixed spacing, we really should have a clang format file. We've discussed it multiple times and i think the only thing stopping us is the one time format the world event. Im very much pro a format the world event though. but we should start thinking about adding one sooner than later.

src/aws-cpp-sdk-core/source/auth/signer/AWSAuthV4Signer.cpp
sbiscigl: not a good name for the variable re-naming to "checksumHeaderValue". We are removing the checksum header from being sent as it is part of the trailer in this branch of logic. The bug beforehand was that a user would add the checksum to the request, it would serialized as a header AND calculated as part of the request. thus my comment on the [original PR](https://github.com/aws/aws-sdk-cpp/pull/2644)

> we should consider not serializing the header x-amz-checksum- to the header in the first place during codegen as we will always end up adding it one way or another if needed. this is fine for now imo opinon, but removing the serialization could perhaps be a better long term solution

so when sent there would be a header and a trailer, which is wrong.

src/aws-cpp-sdk-core/include/aws/core/utils/crypto/PrecalculatedHash.h
```diff
@@ -0,0 +1,32 @@
...
+
+#include <aws/core/Core_EXPORTS.h>
+#include <aws/core/utils/crypto/Hash.h>
+#include <aws/core/utils/crypto/HashResult.h>
+#include "aws/core/utils/Outcome.h"
+
+namespace Aws {
+    namespace Utils {
+        namespace Crypto {
+            class AWS_CORE_API PrecalculatedHash : public Hash {
```
SergeyRyabinin, line 16: please add docstring comments to this class 
and why it is needed, i.e. what is the difference with a regular hash?

src/aws-cpp-sdk-core/include/aws/core/utils/crypto/Hash.h
```diff
@@ -9,6 +9,7 @@
 #include <aws/core/utils/crypto/HashResult.h>
 #include <aws/core/utils/memory/stl/AWSStreamFwd.h>
 #include <aws/core/utils/memory/stl/AWSString.h>
+#include <aws/core/utils/HashingUtils.h>
```
SergeyRyabinin, line 12: why are you adding a header to a header while nothing is changed here?

src/aws-cpp-sdk-core/source/auth/signer/AWSAuthV4Signer.cpp
```diff
@@ -237,17 +237,18 @@ bool AWSAuthV4Signer::SignRequest(Aws::Http::HttpRequest& request, const char* r
             return false;
         }
 
-        if (request.GetRequestHash().second != nullptr)
+        Aws::String checksumHeaderKey = Aws::String("x-amz-checksum-") + request.GetRequestHash().first;
+        const auto headers = request.GetHeaders();
```
SergeyRyabinin, line 241: accidental copy of all request's headers creation.

src/aws-cpp-sdk-core/include/aws/core/utils/crypto/Hash.h
sbiscigl: because i removed code and forgot to remove header

src/aws-cpp-sdk-core/source/auth/signer/AWSAuthV4Signer.cpp
sbiscigl: ah can just use `HasHeader` actually, doing that instead

src/aws-cpp-sdk-core/source/client/AWSClient.cpp
```diff
@@ -791,56 +792,69 @@ void AWSClient::AddChecksumToRequest(const std::shared_ptr<Aws::Http::HttpReques
...
                               request.GetServiceSpecificParameters()->parameterMap.end();
 
-    // Request checksums
+    //Check if user has provided the checksum algorithm
     if (!checksumAlgorithmName.empty() && !shouldSkipChecksum)
     {
+        // Check if user has provided a checksum value for the specified algorithm
+        const Aws::String checksumType = "x-amz-checksum-" + checksumAlgorithmName;
+        const HeaderValueCollection &headers = request.GetHeaders();
+        bool checksumValueAndAlgorithmProvided = headers.find(checksumType) != headers.end();
```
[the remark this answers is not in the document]
sbiscigl: yes because its [serialized as such from the request](https://github.com/aws/aws-sdk-cpp/blob/main/generated/src/aws-cpp-sdk-s3/source/model/PutObjectRequest.cpp#L143-L174). which actually relates to dmitrys quesiton about removing the header when adding the trailing checksum. Serializing the checksum as part of the request as a header is one of the issues that got us here.

PR #2830, opened 2024-01-26
Add constructor overload to `STSProfileCredentialsProvider` where the client factory returns a shared pointer.

teo-tsirpanis:
*Issue #, if available:*

*Description of changes:*

The `STSProfileCredentialProvider` class has a constructor overload that accepts an `STSClient` factory function. The problem with this function is that it returns an `STSClient*`, making lifetime management tricky in certain cases.

This PR introduces an additional overload that accepts an factory that returns a `shared_ptr<STSClient>`. This will enable the client to be freed automatically when the credentials provider no longer needs it.

The existing constructor overload wraps the `STSClient*` into a shared pointer with a deleter that does nothing, maintaining compatibility. I also added an overload to disambiguate between creating a credential provider with a `nullptr` factory.

*Check all that applies:*
- [Χ] Checked if this PR will _not_ introduce cross-platform inconsistent behavior.
- [Χ] Checked if this PR would require a ReadMe/Wiki update.

- [ ] Linux
- [Χ] Windows
- [ ] Android
- [ ] MacOS
- [ ] IOS
- [ ] Other Platforms

### DIFF
--- a/src/aws-cpp-sdk-identity-management/source/auth/STSProfileCredentialsProvider.cpp
+++ b/src/aws-cpp-sdk-identity-management/source/auth/STSProfileCredentialsProvider.cpp
@@ -17,6 +17,12 @@ using namespace Aws::Auth;
 
 constexpr char CLASS_TAG[] = "STSProfileCredentialsProvider";
 
+template <typename T>
+struct NoOpDeleter
+{
+    void operator()(T*) {}
+};
+
 STSProfileCredentialsProvider::STSProfileCredentialsProvider()
     : STSProfileCredentialsProvider(GetConfigProfileName(), std::chrono::minutes(60)/*duration*/, nullptr/*stsClientFactory*/)
 {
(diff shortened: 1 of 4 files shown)


jmklix:
Can you update your PR to remove the .gitignore file change?

teo-tsirpanis:
It was a good change but sure. Done.

src/aws-cpp-sdk-identity-management/source/auth/STSProfileCredentialsProvider.cpp
```diff
@@ -27,8 +33,24 @@ STSProfileCredentialsProvider::STSProfileCredentialsProvider(const Aws::String&
...
+    m_reloadFrequency(std::chrono::minutes(std::max(int64_t(5), static_cast<int64_t>(duration.count()))) - std::chrono::minutes(5)),
+    m_stsClientFactory(nullptr)
+{
+}
+
 STSProfileCredentialsProvider::STSProfileCredentialsProvider(const Aws::String& profileName, std::chrono::minutes duration, const std::function<Aws::STS::STSClient*(const AWSCredentials&)> &stsClientFactory)
       : m_profileName(profileName),
+        m_duration(duration),
+        m_reloadFrequency(std::chrono::minutes(std::max(int64_t(5), static_cast<int64_t>(duration.count()))) - std::chrono::minutes(5)),
+        m_stsClientFactory([=](const auto& credentials) {return std::shared_ptr<Aws::STS::STSClient>(stsClientFactory(credentials), NoOpDeleter<Aws::STS::STSClient>()); })
```
teo-tsirpanis, line 48: _continuing discussion from https://github.com/aws/aws-sdk-cpp/pull/2839#discussion_r1793925901_

> should use MakeShared for allocator awareness?

@sbera87 because we cannot safely `delete` an arbitrary pointer we have to create the shared pointer with a special deleter that does nothing.

PR #2833, opened 2024-01-29
Avoid processing log statement if logger has been destructed

SergeyRyabinin:
in the middle of logging

*Issue #, if available:*

in case of heavy mis-use of basic-use such as in
```
TEST(AwsMemoryManagementTest, MultiInitParallel)
```

the logger may perform
```
m_syncData.m_queuedLogMessages.emplace_back(std::move(statement));
```
on already destructed self.
With custom memory management enabled, it will result in UB.

*Description of changes:*

*Check all that applies:*

- [ ] Linux
- [ ] Windows
- [ ] Android
- [ ] MacOS
- [ ] IOS
- [ ] Other Platforms

### DIFF
--- a/src/aws-cpp-sdk-core/include/aws/core/utils/logging/DefaultLogSystem.h
+++ b/src/aws-cpp-sdk-core/include/aws/core/utils/logging/DefaultLogSystem.h
@@ -58,12 +58,13 @@ namespace Aws
                 struct LogSynchronizationData
                 {
                 public:
-                    LogSynchronizationData() : m_stopLogging(false) {}
+                    LogSynchronizationData() : m_stopLogging(false), m_loggingThreadStopped(false) {}
 
                     std::mutex m_logQueueMutex;
                     std::condition_variable m_queueSignal;
                     Aws::Vector<Aws::String> m_queuedLogMessages;
-                    bool m_stopLogging;
+                    bool m_stopLogging = false;
+                    bool m_loggingThreadStopped = false;
 
                 private:
                     LogSynchronizationData(const LogSynchronizationData& rhs) = delete;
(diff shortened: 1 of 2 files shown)


src/aws-cpp-sdk-core/include/aws/core/utils/logging/DefaultLogSystem.h
```diff
@@ -63,7 +63,8 @@ namespace Aws
                     std::mutex m_logQueueMutex;
                     std::condition_variable m_queueSignal;
                     Aws::Vector<Aws::String> m_queuedLogMessages;
-                    bool m_stopLogging;
+                    bool m_stopLogging = false;
```
jeking3, line 66: This is redundant to the constructor at line 61.

src/aws-cpp-sdk-core/include/aws/core/utils/logging/DefaultLogSystem.h
```diff
@@ -63,7 +63,8 @@ namespace Aws
...
+                    bool m_loggingThreadStopped = false;
```
jeking3, line 67: Should this be initialized in the constructor instead?

src/aws-cpp-sdk-core/source/utils/logging/DefaultLogSystem.cpp
```diff
@@ -90,12 +96,27 @@ DefaultLogSystem::~DefaultLogSystem()
         m_syncData.m_queueSignal.notify_one();
     }
 
+    // explicitly wait for logging thread to finish
+    {
+        std::unique_lock<std::mutex> locker(m_syncData.m_logQueueMutex);
+        if (!m_syncData.m_loggingThreadStopped)
```
jeking3, line 102: Should this be "while" instead of "if" and not use a conditional wait_for?  It is not safe to leave the destructor until we know the thread is gone.

jeking3 (changes requested):

src/aws-cpp-sdk-core/source/utils/logging/DefaultLogSystem.cpp
SergeyRyabinin: there is .join for the thread later.
I just did not want to block user code if the boolean is never set.

sbiscigl (approved):

Issue #2862, opened 2024-02-20
Cannot build sdk on a Amazon Linux 2023 based Container after 1.11.211 Version Tag


davehines:
### Describe the bug

When I try to build the SDK after the 1.11.211 tag , the build fails on one of the tests with the following error 

`/usr/include/c++/11/bits/shared_ptr_base.h:979: std::__shared_ptr_access<_Tp, _Lp, <anonymous>, <anonymous> >::element_type& std::__shared_ptr_access<_Tp, _Lp, <anonymous>, <anonymous> >::operator*() const [with _Tp = Aws::Http::ServiceSpecificParameters; __gnu_cxx::_Lock_policy _Lp = __gnu_cxx::_S_atomic; bool <anonymous> = false; bool <anonymous> = false; std::__shared_ptr_access<_Tp, _Lp, <anonymous>, <anonymous> >::element_type = Aws::Http::ServiceSpecificParameters]: Assertion '_M_get() != nullptr' failed.`

If I build with run unit tests turned  off  `  -DAUTORUN_UNIT_TESTS=OFF`
the build succeeds , but my simple application fails with the exact same error

### Expected Behavior

Test should run and pass 

### Current Behavior

Error 
`/usr/include/c++/11/bits/shared_ptr_base.h:979: std::__shared_ptr_access<_Tp, _Lp, <anonymous>, <anonymous> >::element_type& std::__shared_ptr_access<_Tp, _Lp, <anonymous>, <anonymous> >::operator*() const [with _Tp = Aws::Http::ServiceSpecificParameters; __gnu_cxx::_Lock_policy _Lp = __gnu_cxx::_S_atomic; bool <anonymous> = false; bool <anonymous> = false; std::__shared_ptr_access<_Tp, _Lp, <anonymous>, <anonymous> >::element_type = Aws::Http::ServiceSpecificParameters]: Assertion '_M_get() != nullptr' failed.`

image of relevant part stack from gdb from our test application  which gives the 
![image](https://github.com/aws/aws-sdk-cpp/assets/2527679/0e4004db-297b-4994-b85c-41664b5ee2ad)

### Reproduction Steps

cmake ..   -DCMAKE_INSTALL_PREFIX=/usr/local/AWS_XXX -DCMAKE_BUILD_TYPE=Release - -DBUILD_ONLY="access-management;cognito-identity;core;dynamodb;ecs;events;iam;logs;s3;sqs;transfer;states;sts;xray"
make

[...]

PR #2870, opened 2024-02-22
change argument to avoid UB in pointer dreference

sbiscigl:
*Issue #, if available:*

[issues/2862](https://github.com/aws/aws-sdk-cpp/issues/2862)

*Description of changes:*

Fixes a UB issue when built with GCC and `_GLIBCXX_ASSERTIONS ` turned on. It is specifically that we are passing a nullptr to a function where it is not used, it is only used in the S3 client. Although it will not actually cause runtime UB a compiler will flag it as such, and we should fix it.

*Check all that applies:*

- [x] Linux
- [x] Windows
- [ ] Android
- [x] MacOS
- [ ] IOS
- [ ] Other Platforms

### DIFF
--- a/src/aws-cpp-sdk-core/include/aws/core/auth/signer/AWSAuthV4Signer.h
+++ b/src/aws-cpp-sdk-core/include/aws/core/auth/signer/AWSAuthV4Signer.h
@@ -165,7 +165,7 @@ namespace Aws
             */
             bool PresignRequest(Aws::Http::HttpRequest& request, const char* region, const char* serviceName, long long expirationInSeconds = 0) const override;
 
-            virtual Aws::Auth::AWSCredentials GetCredentials(const Aws::Http::ServiceSpecificParameters &serviceSpecificParameters) const;
+            virtual Aws::Auth::AWSCredentials GetCredentials(const std::shared_ptr<Aws::Http::ServiceSpecificParameters> &serviceSpecificParameters) const;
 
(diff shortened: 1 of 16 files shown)


SergeyRyabinin (approved):

src/aws-cpp-sdk-core/include/aws/core/auth/signer/AWSAuthV4Signer.h
```diff
@@ -165,7 +165,7 @@ namespace Aws
             */
             bool PresignRequest(Aws::Http::HttpRequest& request, const char* region, const char* serviceName, long long expirationInSeconds = 0) const override;
 
-            virtual Aws::Auth::AWSCredentials GetCredentials(const Aws::Http::ServiceSpecificParameters &serviceSpecificParameters) const;
+            virtual Aws::Auth::AWSCredentials GetCredentials(const std::shared_ptr<Aws::Http::ServiceSpecificParameters> serviceSpecificParameters) const;
```
SergeyRyabinin, line 168: I think that passing a shared_ptr by value is less efficient here than passing a reference to a shared_ptr

by value cons: memory allocation, shared_ptr atomic/locking refcount increase
by ref cons: double pointer deference in case it used later

SergeyRyabinin (approved):

SergeyRyabinin (approved):

PR #2885, opened 2024-03-11
Remove multi-thread synchronization from CRT logger wrapper

SergeyRyabinin:
*Issue #, if available:*

*Description of changes:*

*Check all that applies:*

- [x] Linux
- [x] Windows
- [ ] Android
- [ ] MacOS
- [ ] IOS
- [ ] Other Platforms

### DIFF
--- a/src/aws-cpp-sdk-core/source/utils/logging/CRTLogging.cpp
+++ b/src/aws-cpp-sdk-core/source/utils/logging/CRTLogging.cpp
@@ -24,11 +24,7 @@ namespace Utils
 namespace Logging {
 
 static std::shared_ptr<CRTLogSystemInterface> CRTLogSystem(nullptr);
-static Aws::Utils::Threading::ReaderWriterLock CRTLogSystemLock;
-
-static bool s_CRTLogsRedirectionIsSet(false);
 static aws_logger s_sdkCrtLogger;
-static aws_logger s_previousCrtLogger;
 
 static int s_aws_logger_redirect_log(
         struct aws_logger *logger,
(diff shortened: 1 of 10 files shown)


src/aws-cpp-sdk-core/source/utils/logging/CRTLogging.cpp
```diff
@@ -37,84 +33,53 @@ static int s_aws_logger_redirect_log(
...
-        va_list args;
-        va_start(args, format);
-        CRTLogSystem->Log(logLevel, subjectName, format, args);
-        va_end(args);
-        return AWS_OP_SUCCESS;
-    }
-    else if (s_CRTLogsRedirectionIsSet)
+    if (!CRTLogSystem)
     {
-        // SDK CRT logger is terminated, fallback to an original/previous CRT logger
```
SergeyRyabinin, line 54: I just thought about it more and realized that we don't control / own "original" logger, so better to not try to over-smart ourselves.

src/aws-cpp-sdk-core/source/utils/logging/DefaultCRTLogSystem.cpp
```diff
@@ -22,57 +22,36 @@ namespace Aws
...
-                va_end(tmp_args);
+                return Aws::Utils::Logging::LogLevel::Off;
+            }
 
-                Aws::OStringStream logStream;
-                if (requiredLength > 1)
-                {
-                    Array<char> outputBuff(requiredLength);
-#ifdef _WIN32
-                    vsnprintf_s(outputBuff.GetUnderlyingData(), requiredLength, _TRUNCATE, formatStr, args);
```
SergeyRyabinin, line 64: exactly the same logic is in the main SDK logger, so just call it instead.

src/aws-cpp-sdk-core/source/utils/logging/DefaultCRTLogSystem.cpp
```diff
@@ -22,57 +22,36 @@ namespace Aws
...
-                m_stopLogging(false),
-                m_logsProcessed(0)
+            DefaultCRTLogSystem::DefaultCRTLogSystem(LogLevel logLevel)
             {
+                AWS_UNREFERENCED_PARAM(logLevel); // will use one from the main SDK logger
             }
 
             DefaultCRTLogSystem::~DefaultCRTLogSystem()
             {
-                m_stopLogging = true;
```
sbiscigl, line 34: np: couldnt we just move this out to the header as `~DefaultCRTLogSystem() = default` then?

sbiscigl (approved):

src/aws-cpp-sdk-core/source/utils/logging/FormattedLogSystem.cpp
```diff
@@ -17,44 +17,56 @@
...
-            break;
+            return "[TRACE] ";
     }
+    return "[UNKNOWN] ";
+}
 
-    ss << DateTime::Now().CalculateGmtTimeWithMsPrecision() << " " << tag << " [" << std::this_thread::get_id() << "] ";
-
-    return ss.str();
+static Aws::String CreateLogPrefixLine(LogLevel logLevel, const char* tag, const size_t statementSize = 80)
```
SergeyRyabinin, line 43: so what happens here is the attempt to reduce number of memory allocations to 1 (from zero) per logging statement processing.
There are still additional allocations from `CalculateGmtTimeWithMsPrecision` and `<<std::this_thread::get_id` but the rest of allocations (and stack-allocated Aws::Array) are eliminated.

src/aws-cpp-sdk-core/source/utils/logging/FormattedLogSystem.cpp
```diff
@@ -17,44 +17,56 @@
...
+{
+    Aws::String prefix;
+    static const size_t THREAD_ID_LEN = [](){
+        Aws::StringStream strStr;
+        strStr << std::this_thread::get_id();
+        return strStr.str().size();
```
graebm, line 49: this runs only once per application instance, right?

Is it guaranteed that every thread's ID is the same length? If this were just an integer on some platform, I could imagine different threads having different length integers. I guess it's still safe, since this is just for a `reserve()` call...

Simpler to just have a hard-coded high guess? Or overestimate somewhere and just like, ` + 10` to prefixLen?

sbiscigl (approved):

sbiscigl (approved):

src/aws-cpp-sdk-core/source/utils/logging/CRTLogging.cpp
```diff
@@ -128,39 +97,33 @@ static struct aws_logger_vtable s_aws_logger_redirect_vtable = {
...
-    Aws::Utils::Threading::WriterLockGuard g(CRTLogSystemLock);
-    CRTLogSystem.reset();
+    if (aws_logger_get() == &s_sdkCrtLogger)
+    {
+        aws_logger_set(nullptr);
+    }
+    // GetLogSystem returns a raw pointer
+    // so this is a hack to let all other threads finish their log statement after getting a LogSystem pointer
+    // otherwise we would need to perform ref-counting on each logging statement
+    auto tmpLogger = std::move(CRTLogSystem);
```
jeking3, line 124: This approach would not be necessary if the logger's lifetime always exceeded the CRT.  We should be able to shutdown the CRT _and then_ call aws_logger_set(nullptr) and have no need for time-based quiesces which are ultimately a flawed approach (someone will always exceed the sleep time).  We would then get all the log messages from CRT shutdown.  We're shutting down CRT logging before CRT is stopped, and hooking CRT logging in after CRT is started.  There's a more fundamental problem here.

bjosv: I concur. I think we hit an issue with this when running tests with ASAN in CI (throttled/high CPU usage).

Issue #2538, opened 2023-06-14
sdk code is repeatedly calling curl_global_cleanup


ButcherOfBlaviken:
### Describe the bug

source file: aws-sdk-cpp-1.10.20/aws-cpp-sdk-core/source/http/curl/CurlHttpClient.cpp

Sorry my English is not very good, I hope you can understand what I mean.

For every Aws::InitAPI and Aws::ShutdownAPI pair, aws sdk calls CurlHttpClient::InitGlobalState() and CurlHttpClient::CleanupGlobalState(). In the InitGlobalState function there is a bool type flag variable isInit, because of it InitGlobalState only calls libcurl initialization API curl_global_init once, but in CleanupGlobalState function, this flag won't be unset, so CleanupGlobalState is calling libcurl cleanup API curl_global_cleanup without any restriction, according to libcurl, This function releases resources acquired by curl_global_init and application should call curl_global_cleanup once for each call you make to curl_global_init.

For my case I use both aws S3 and libcurl based https, and aws sdk code cleanup the global initialization for my libcurl based code and caused CURLE_OUT_OF_MEMORY error when I try to make a https POST.

### Expected Behavior

AWS sdk shall should call [curl_global_cleanup] once for each call it makes to [curl_global_init]

### Current Behavior

AWS sdk calls [curl_global_cleanup] only once but calls [curl_global_init] multi-times

### Reproduction Steps

always

### Possible Solution

void CurlHttpClient::CleanupGlobalState()
{
    if(true==isInit)
    {
        curl_global_cleanup();
        isInit=false;
    }
}

### Additional Information/Context

https://curl.se/libcurl/c/curl_global_cleanup.html

### AWS CPP SDK version used

aws-sdk-cpp-1.10.20

### Compiler and Version used

aarch64-linux-gnu gcc version 9.4.0

### Operating System and version

aarch64-linux-gnu

SergeyRyabinin:
Hi @ButcherOfBlaviken ,

SDK options struct has an option to disable calling this curl method, if you prefer to call it yourself:
https://github.com/aws/aws-sdk-cpp/blob/0175cb2dd1dfdae3fd5f49a984db03802786746d/src/aws-cpp-sdk-core/include/aws/core/Aws.h#L90-L94

At the same time, you are right with a proposed suggestion that we should check the flag value before attempting to `curl_global_cleanup`.
If you prefer, you could submit a PR with a change, otherwise we will follow up on it later.

Best regards,
[...]

PR #2890, opened 2024-03-13
only call curl_global_cleanup once

jmklix:
*Issue #, if available:*
https://github.com/aws/aws-sdk-cpp/issues/2538

*Description of changes:*
only call curl_glabal_cleanup once

*Check all that applies:*

- [x] Linux
- [x] Windows
- [ ] Android
- [x] MacOS
- [ ] IOS
- [ ] Other Platforms

### DIFF
--- a/src/aws-cpp-sdk-core/source/http/curl/CurlHttpClient.cpp
+++ b/src/aws-cpp-sdk-core/source/http/curl/CurlHttpClient.cpp
@@ -530,8 +530,11 @@ void CurlHttpClient::InitGlobalState()
 
 void CurlHttpClient::CleanupGlobalState()
 {
-    curl_global_cleanup();
-    isInit = false;
+    if (isInit)
+    {
+        curl_global_cleanup();
+        isInit = false;
+    }
 }
 
 Aws::String CurlInfoTypeToString(curl_infotype type)


src/aws-cpp-sdk-core/source/http/curl/CurlHttpClient.cpp
```diff
@@ -530,8 +530,11 @@ void CurlHttpClient::InitGlobalState()
 
 void CurlHttpClient::CleanupGlobalState()
 {
-    curl_global_cleanup();
-    isInit = false;
+    if (isInit)
```
sbiscigl, line 533: is this thread safe? should we put it behing a [unique_lock](https://en.cppreference.com/w/cpp/thread/unique_lock) ? we do things like that [elsewhere](https://github.com/aws/aws-sdk-cpp/blob/main/src/aws-cpp-sdk-core/include/aws/core/utils/ResourceManager.h#L123)

jmklix: Yes. isInit is atomic:
```cpp
std::atomic<bool> CurlHttpClient::isInit(false);
```

sbiscigl (approved):

PR #2928, opened 2024-04-17
Fix failure to call curl[c-ares]

jimwang118:
*Issue #, if available:*
After installing curl's feature c-ares on arm64-osx, there will be a problem that c-ares cannot be linked when installing aws-sdk-cpp[core,s3,transfer].
You can also refer to the specific questions of [36661](https://github.com/microsoft/vcpkg/issues/36661).
*Description of changes:*
Modified the curl calling method, using the module mode of cmake to call curl.
*Check all that applies:*

- [ ] Linux
- [X] Windows
- [ ] Android
- [X] MacOS
- [ ] IOS
- [ ] Other Platforms


cmake/external_dependencies.cmake
```diff
@@ -80,23 +80,13 @@ if(NOT NO_HTTP_CLIENT AND NOT USE_CRT_HTTP_CLIENT)
             set(BUILD_CURL 1)
             message(STATUS "  Building Curl as part of AWS SDK")
         else()
-            include(FindCURL)
+            find_package(CURL REQUIRED)
```
sbiscigl, line 83: we already call find`find_package` if `include(FindCURL)` doesnt find `CURL_INCLUDE_DIRS` and `CURL_LIBRARIES`. So my question to you is why is the incorrect value there?

the answer seems to be [in this comment](https://github.com/microsoft/vcpkg/issues/36661#issuecomment-1941721834)

> I believe vcpkg-cmake-wrapper.cmake has to be patched to account for $<LINK_ONLY:c-ares::cares>. In fact, this needs to be done for other transitive libraries used by curl, too.

it sounds like vcpkg explictly sets`CURL_LIBRARIES` to not include c-ares. We honor `CURL_LIBRARIES` as found by [`include(FindCURL)`](https://cmake.org/cmake/help/latest/module/FindCURL.html). Why cannot `CURL_LIBRARIES` be fixed to reflect the libraries that are needed?

PR #2930, opened 2024-04-17
Revert "Memory management integration with common runtime libraries."

SergeyRyabinin:
*Issue #, if available:*
CRT STL allocator being used in the SDK.
This forces CRT library to be initialized whenever SDK tries to use a custom allocator.
This prevents SDK from shutting down a CRT logger wrapper after shutting down the CRT library.

*Description of changes:*
apples and oranges;
separate the sheep from the goats;

*Check all that applies:*

- [x] Linux
- [x] Windows
- [ ] Android
- [x] MacOS
- [ ] IOS
- [ ] Other Platforms

### DIFF
--- a/src/aws-cpp-sdk-core/source/Aws.cpp
+++ b/src/aws-cpp-sdk-core/source/Aws.cpp
@@ -210,14 +210,14 @@ namespace Aws
 
         Aws::Config::CleanupConfigAndCredentialsCacheManager();
 
+        Aws::Client::CoreErrorsMapper::CleanupCoreErrorsMapper();
+        Aws::CleanupCrt();
+
         if (options.loggingOptions.logLevel != Aws::Utils::Logging::LogLevel::Off)
         {
             Aws::Utils::Logging::ShutdownCRTLogging();
             Aws::Utils::Logging::PushLogger(nullptr); // stops further logging but keeps old logger object alive
         }
-        Aws::Client::CoreErrorsMapper::CleanupCoreErrorsMapper();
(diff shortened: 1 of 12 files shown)


sbiscigl (approved):

jeking3 (approved):
This seems quite reasonable to me.  If one does not initialize the SDK one should not be trying to use it.

src/aws-cpp-sdk-core/source/Aws.cpp
```diff
@@ -210,14 +210,14 @@ namespace Aws
 
         Aws::Config::CleanupConfigAndCredentialsCacheManager();
 
+        Aws::Client::CoreErrorsMapper::CleanupCoreErrorsMapper();
```
jeking3, line 213: We should also be initializing the logger and installing it before we initialize the CRT, so we log the CRT startup log messages as well.  This will also prove the logger has no dependencies on the CRT, and then the mechanism(s) to cleanly shutdown with delays and quiesces around logging can all be removed.

jeking3 (commented):
We should also be initializing the logger and installing it before we initialize the CRT, so we log the CRT startup log messages as well.  This will also prove the logger has no dependencies on the CRT, and then the mechanism(s) to cleanly shutdown with delays and quiesces around logging can all be removed.

sbiscigl (approved):

DmitriyMusatkin (approved):

sbiscigl (approved):

PR #2460, opened 2023-04-24
Finish crt migration, DRAFT, DO NOT MERGE

JonathanHenson:
*Issue #, if available:*

*Description of changes:*

*Check all that applies:*
- [x ] Did a review by yourself.
- [x ] Added proper tests to cover this PR. (If tests are not applicable, explain.)
- [x ] Checked if this PR is a breaking (APIs have been changed) change. ( Struct sizes change AND the hash and cipher invariants have been updated.)
- [x ] Checked if this PR will _not_ introduce cross-platform inconsistent behavior.
- [ ] Checked if this PR would require a ReadMe/Wiki update.

Check which platforms you have built SDK on to verify the correctness of this PR.
- [x ] Linux
- [x ] Windows
- [ ] Android
- [x ] MacOS
- [ ] IOS
- [ ] Other Platforms

By submitting this pull request, I confirm that you can use, modify, copy, and redistribute this contribution, under the terms of your choice.

PR #2952, opened 2024-05-09
Remove crypto dependency from sdk

sbiscigl:
*Description of changes:*

Removes the direct dependency on libcrypto from the SDK and uses c-cal from crt instead simplifying the dependency set in the SDK.

Mostly borrowed from [pull/2460](https://github.com/aws/aws-sdk-cpp/pull/2460) with bug fixes.

*Check all that applies:*

- [x] Linux
- [x] Windows
- [ ] Android
- [x] MacOS
- [ ] IOS
- [ ] Other Platforms


cmake/external_dependencies.cmake
```diff
@@ -19,43 +19,9 @@ endif()
 
 
 # Encryption control
+# TODO: BYO Crypto is not implemented for CRT/Was not working in the latest version of the SDK.
 if(NOT NO_ENCRYPTION)
```
DmitriyMusatkin, line 23: hard fail if someone requests no_encryption for now?

src/aws-cpp-sdk-core/include/aws/core/utils/Array.h
```diff
@@ -77,29 +96,40 @@ namespace Aws
                 {
                     if(arr->m_size > 0 && arr->m_data)
                     {
-                        size_t arraySize = arr->m_size;
+                        size_t arraySize = arr->m_length;
+#ifdef _WIN32
+                        std::copy(arr->m_data.get(), arr->m_data.get() + arraySize, stdext::checked_array_iterator< T * >(m_data.get() + location, m_length));
```
DmitriyMusatkin, line 101: win throws error if + is unchecked? is there a way to write code that works for all platforms without ifdefs?

src/aws-cpp-sdk-core/include/aws/core/utils/base64/Base64.h
```diff
@@ -67,4 +67,3 @@ namespace Aws
         } // namespace Base64
     } // namespace Utils
 } // namespace Aws
-
```
DmitriyMusatkin, line 70: why? dont we want empty line at the end of files? lint is not flagging it?

src/aws-cpp-sdk-core/include/aws/core/utils/crypto/Cipher.h
```diff
@@ -16,7 +16,6 @@ namespace Aws
         namespace Crypto
         {
             static const size_t SYMMETRIC_KEY_LENGTH = 32;
-            static const size_t MIN_IV_LENGTH = 12;
```
DmitriyMusatkin, line 19: those are public constants? so removing them might be backwards incompatible?

src/aws-cpp-sdk-core/include/aws/core/utils/crypto/SecureRandom.h
```diff
@@ -4,6 +4,8 @@
  */
 #pragma once
 #include <type_traits>
+#include <memory>
```
DmitriyMusatkin, line 7: why?

src/aws-cpp-sdk-core/include/aws/core/utils/crypto/crt/CRTHash.h
```diff
@@ -0,0 +1,43 @@
...
+            class AWS_CORE_API CRTHash : public Hash
+            {
+            public:
+                ~CRTHash() override = default;
+                CRTHash(const CRTHash &) = delete;
+                CRTHash &operator=(const CRTHash &) = delete;
+                CRTHash(CRTHash &&) = default;
+                CRTHash &operator=(CRTHash &&) = default;
+
+                HashResult Calculate(const String &str) override;
```
DmitriyMusatkin, line 27: oof on string? that can have a lot of "fun" corner cases. is it possible to switch to a byte buf or its too late now?

src/aws-cpp-sdk-core/source/auth/signer/AWSAuthEventStreamV4Signer.cpp
```diff
@@ -118,15 +118,14 @@ bool AWSAuthEventStreamV4Signer::SignRequest(Aws::Http::HttpRequest& request, co
     AWS_LOGSTREAM_DEBUG(v4StreamingLogTag, "Canonical Request String: " << canonicalRequestString);
 
     //now compute sha256 on that request string
-    auto hashResult = m_hash.Calculate(canonicalRequestString);
-    if (!hashResult.IsSuccess())
+    auto sha256Digest = HashingUtils::CalculateSHA256(canonicalRequestString);
+    if (sha256Digest.GetLength() == 0)
```
DmitriyMusatkin, line 122: its still computing result, no? why does this have to change?

CMakeLists.txt
```diff
@@ -66,7 +65,6 @@ if (LEGACY_BUILD)
     option(BUILD_OPTEL_OTLP_BENCHMARKS "Enables building the benchmark tests with open telemetry OTLP clients" OFF)
     option(USE_TLS_V1_2 "Set http client to enforce TLS 1.2" ON)
     option(USE_TLS_V1_3 "Set http client to enforce TLS 1.3" OFF)
-    option(AWS_USE_CRYPTO_SHARED_LIBS "Forces FindCrypto to use a shared crypto library if found. regardless of the value of BUILD_SHARED_LIBS" OFF)
```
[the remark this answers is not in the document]
sbiscigl: both [CRT](https://github.com/awslabs/aws-c-cal/blob/main/CMakeLists.txt#L16) and [S2N](https://github.com/aws/s2n-tls/blob/main/CMakeLists.txt) have a equivalent but "different" arg, since they are the same value, we can safely remove this now and let the downstream variables be the location for further changes.

src/aws-cpp-sdk-core/include/aws/core/utils/base64/Base64.h
sbiscigl: we should have a clang-tidy task as part of our CI, and i would happily add one. I would suggest we dont add one as part of this change though but a follow up task

src/aws-cpp-sdk-core/include/aws/core/utils/crypto/Cipher.h
sbiscigl: good eye

src/aws-cpp-sdk-core/source/auth/signer/AWSAuthEventStreamV4Signer.cpp
sbiscigl: `CalculateSHA256` returns a byte buf, which according to the CRT API is empty if it "failed". since this utility function hides away cipher state we use the empty/non-empty signal to determine if it was successfully calculated.

src/aws-cpp-sdk-core/include/aws/core/utils/base64/Base64.h
sbiscigl: also i think the diff is messed up because im seeing a new line in the file locally

src/aws-cpp-sdk-core/include/aws/core/utils/crypto/SecureRandom.h
sbiscigl: non-unity builds fail because they are missing a include for `std::shared_ptr`, it would appear that this use to get that include transitively before from crypto impl, but now it no longer does.

src/aws-cpp-sdk-core/include/aws/core/utils/crypto/crt/CRTHash.h
sbiscigl: too late now, that is part of the interface for [hash that has long existed](https://github.com/aws/aws-sdk-cpp/blob/main/src/aws-cpp-sdk-core/include/aws/core/utils/crypto/Hash.h#L29-L32)

src/aws-cpp-sdk-core/include/aws/core/utils/crypto/SecureRandom.h
DmitriyMusatkin: we wrap shared ptr, shouldnt we include that instead?

src/aws-cpp-sdk-core/source/auth/signer/AWSAuthV4Signer.cpp
```diff
@@ -567,40 +560,38 @@ Aws::Utils::ByteBuffer AWSAuthV4Signer::ComputeHash(const Aws::String& secretKey
 {
     Aws::String signingKey(Aws::Auth::AWSAuthHelper::SIGNING_KEY);
     signingKey.append(secretKey);
-    auto hashResult = m_HMAC->Calculate(ByteBuffer((unsigned char*)simpleDate.c_str(), simpleDate.length()),
-            ByteBuffer((unsigned char*)signingKey.c_str(), signingKey.length()));
+    auto kDate = HashingUtils::CalculateSHA256HMAC(
+        ByteBuffer((unsigned char *) simpleDate.c_str(), simpleDate.length()),
```
DmitriyMusatkin, line 564: can we just add bytebuffer from string constructor?

src/aws-cpp-sdk-core/source/utils/HashingUtils.cpp
```diff
@@ -91,16 +90,17 @@ static ByteBuffer TreeHashFinalCompute(Aws::List<ByteBuffer>& input)
 
 ByteBuffer HashingUtils::CalculateSHA256TreeHash(const Aws::String& str)
 {
-    Sha256 hash;
     if (str.size() == 0)
     {
+        Sha256 hash;
```
DmitriyMusatkin, line 95: do we even need object for this anymore? can we just make it a static helper?

src/aws-cpp-sdk-core/source/utils/crypto/crt/CRTSecureRandomBytes.cpp
```diff
@@ -0,0 +1,21 @@
...
+namespace Aws
+{
+    namespace Utils
+    {
+        namespace Crypto
+        {
+            void CRTSecureRandomBytes::GetBytes(unsigned char *buffer, std::size_t bufferSize)
+            {
+                auto outputBuf = Crt::ByteBufFromEmptyArray(buffer, bufferSize);
+                (void)Crt::Crypto::GenerateRandomBytes(outputBuf, bufferSize);
```
DmitriyMusatkin, line 17: assert on output?

src/aws-cpp-sdk-core/source/utils/crypto/crt/CRTSymmetricCipher.cpp
```diff
@@ -0,0 +1,97 @@
...
+            CryptoBuffer CRTSymmetricCipher::EncryptBuffer(const CryptoBuffer &unEncryptedData)
+            {
+                auto resultBuffer = Crt::ByteBufInit(get_aws_allocator(), Crt::Crypto::AES_256_CIPHER_BLOCK_SIZE);
+                Crt::ByteCursor toEncrypt = Crt::ByteCursorFromArray(unEncryptedData.GetUnderlyingData(), unEncryptedData.GetLength());
+
+                if (m_cipher.Encrypt(toEncrypt, resultBuffer))
+                {
+                    return {std::move(resultBuffer)};
+                }
+                Crt::ByteBufDelete(resultBuffer);
```
DmitriyMusatkin, line 41: how does error handling work here? we just return empty cryptoBuffer and nothing else?

src/aws-cpp-sdk-core/include/aws/core/utils/Array.h
SergeyRyabinin: I guess this is a side-effect of the rebase, this stdext usage has been recently removed from the SDK: https://github.com/aws/aws-sdk-cpp/pull/2726

src/aws-cpp-sdk-core/include/aws/core/utils/crypto/SecureRandom.h
sbiscigl: we wrap `make_shared` but not actually `shared_ptr` i.e where we [define make shared](https://github.com/aws/aws-sdk-cpp/blob/main/src/aws-cpp-sdk-core/include/aws/core/utils/memory/stl/AWSAllocator.h#L104)

```
std::shared_ptr<T> MakeShared(const char* allocationTag, 
  ArgTypes&&... args)
{
  return std::allocate_shared<T, Aws::Allocator<T>>(
    Aws::Allocator<T>(), std::forward<ArgTypes>(args)...);
}
```

we actually return a std pointer with a allocator, the only reason a unique wrapper exists is because [`make_unique`](https://en.cppreference.com/w/cpp/memory/unique_ptr/make_unique) slipped into c++14 and not 11....which still frustrates me

src/aws-cpp-sdk-core/source/auth/signer/AWSAuthV4Signer.cpp
sbiscigl: ByteBuffer is an Array, but yeah that would be a good convenience function let me add that

src/aws-cpp-sdk-core/source/utils/crypto/crt/CRTSecureRandomBytes.cpp
sbiscigl: good call

src/aws-cpp-sdk-core/source/utils/crypto/crt/CRTSymmetricCipher.cpp
sbiscigl: yeah thats all from the original [API](https://github.com/aws/aws-sdk-cpp/blob/main/src/aws-cpp-sdk-core/include/aws/core/utils/crypto/Cipher.h#L96-L119) which heavily relied on the `virtual operator bool() const { return Good(); }` operator to check if the cipher was in a erroronus state or not. which is why i change in this PR to catpure the state of being finished or ready. The error handling here would be to check the ciphers boolean operator before using the return value per the original API.

src/aws-cpp-sdk-core/include/aws/core/utils/Array.h
```diff
@@ -39,6 +40,7 @@ namespace Aws
              */
             Array(size_t arraySize = 0) :
                 m_size(arraySize),
+                m_length(arraySize),
```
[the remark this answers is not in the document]
sbiscigl: yeah renaming to match CRT naming

src/aws-cpp-sdk-core/include/aws/core/utils/Array.h
sbiscigl: yeah removing again

src/aws-cpp-sdk-core/source/utils/HashingUtils.cpp
sbiscigl: so in theory yes, we could make a static helper function but it would require a major re-write of of how hashing works. specifically in Sha256 we separate implementation using pimpl using a [unique ptr to a Hash type](https://github.com/aws/aws-sdk-cpp/blob/main/src/aws-cpp-sdk-core/include/aws/core/utils/crypto/Sha256.h#L69). because of this pimpl indirection we dont know the underlying type and invoke it using its virtual function. We could of course remove this pimpl indirection but the delta of that change would prove to be much more than and just declaring a temp hash var and using the vtable. IMO it would create a diff that would make this change more risky than it already is.

DmitriyMusatkin: we could just have a static method in Sha256 that creates an instance and calls hash on it, instead of doing it all over the place? or have oneshot variant that resets state to working after hash? but whatever, not something to lose sleep over.

DmitriyMusatkin (approved):

src/aws-cpp-sdk-core/include/aws/core/utils/Array.h
```diff
@@ -220,6 +262,29 @@ namespace Aws
...
+            {
+                m_data.reset(other.buffer);
+                other.capacity = 0;
+                other.len = 0;
+                other.allocator = nullptr;
+            }
+
+            CryptoBuffer& operator=(Crt::ByteBuf&& other) noexcept
+            {
+                m_capacity = other.len;
```
SergeyRyabinin, line 279: add a check that other != parent of this
(i.e. `static_cast<Crt::ByteBuf*>(this) != &other`) or smth like this
otherwise you risk erasing the parent

src/aws-cpp-sdk-core/include/aws/core/utils/base64/Base64.h
SergeyRyabinin: probably the original file was created on windows, thus \r\n line ending has changed to just \n.

src/aws-cpp-sdk-core/include/aws/core/utils/crypto/crt/CRTHash.h
SergeyRyabinin: Please add a todo to deprecate it, possibly a deprecation macro warning, but likely just a todo.

src/aws-cpp-sdk-core/include/aws/core/utils/crypto/crt/CRTHash.h
```diff
@@ -0,0 +1,43 @@
...
+
+                HashResult Calculate(IStream &stream) override;
+
+                void Update(unsigned char *string, size_t bufferSize) override;
+
+                HashResult GetHash() override;
+
+                // @private but made public for the sake of make_shared.
```
SergeyRyabinin, line 35: have you tried to fix perfect forwarding in MakeShared / make_shared?

src/aws-cpp-sdk-core/source/utils/HashingUtils.cpp
```diff
@@ -22,20 +22,19 @@ using namespace Aws::Utils;
 using namespace Aws::Utils::Base64;
 using namespace Aws::Utils::Crypto;
 
-// internal buffers are fixed-size arrays, so this is harmless memory-management wise
-static Aws::Utils::Base64::Base64 s_base64;
```
SergeyRyabinin, line 26: does it have any affect on the performance?

src/aws-cpp-sdk-core/source/utils/crypto/CryptoBuf.cpp
```diff
@@ -289,27 +289,24 @@ namespace Aws
...
 
-                    if (m_cipher)
+                    if(cryptoBuf.GetLength())
                     {
-                        if(cryptoBuf.GetLength())
+                        //allow mid block decryption. We have to decrypt it, but we don't have to write it to the stream.
+                        //the assumption here is that tellp() will always be 0 or >= 16 bytes. The block offset should only
+                        //be the offset of the first block read.
+                        size_t len = cryptoBuf.GetLength();
+                        size_t blockOffset = m_stream.tellp() > m_blockOffset ? 0 : m_blockOffset;
```
SergeyRyabinin, line 298: https://en.cppreference.com/w/cpp/io/basic_ostream/tellp
> If fail()==true, returns pos_type(-1)

I'd suggest checking first that the stream is in a valid state before comparing against the m_blockOffset (although it is a `int16_t`, still I'd prefer to check the stream state)

src/aws-cpp-sdk-core/source/utils/crypto/crt/CRTHash.cpp
```diff
@@ -0,0 +1,90 @@
...
+                {
+                    resultBuffer.SetLength(m_hash.DigestSize());
+                    return resultBuffer;
+                }
+
+                return false;
+            }
+
+            Crypto::HashResult Crypto::CRTHash::Calculate(IStream &stream) {
+                stream.seekg(0, stream.beg);
```
SergeyRyabinin, line 31: check the stream state before / after rewinding it.

src/aws-cpp-sdk-core/source/utils/crypto/crt/CRTHash.cpp
```diff
@@ -0,0 +1,90 @@
...
+                    while (stream.good())
+                    {
+                        stream.read((char *)streamBuffer, Aws::Utils::Crypto::Hash::INTERNAL_HASH_STREAM_BUFFER_SIZE);
+                        std::streamsize bytesRead = stream.gcount();
+                        if (bytesRead > 0)
+                        {
+                            auto curToHash = Crt::ByteCursorFromArray(streamBuffer, static_cast<size_t>(bytesRead));
+                            if (!m_hash.Update(curToHash))
+                            {
+                                // log here.
```
SergeyRyabinin, line 43: log here?

src/aws-cpp-sdk-core/source/utils/crypto/crt/CRTHash.cpp
```diff
@@ -0,0 +1,90 @@
...
+                    if (!stream.eof())
+                    {
+                        return false;
+                    }
+
+                    ByteBuffer resultBuffer(m_hash.DigestSize());
+                    auto outBuffer = Crt::ByteBufFromEmptyArray(resultBuffer.GetUnderlyingData(), resultBuffer.GetSize());
+                    if (!m_hash.Digest(outBuffer))
+                    {
+                        //log
```
SergeyRyabinin, line 58: log

src/aws-cpp-sdk-core/source/utils/crypto/crt/CRTSecureRandomBytes.cpp
```diff
@@ -0,0 +1,24 @@
...
+    namespace Utils
+    {
+        namespace Crypto
+        {
+            void CRTSecureRandomBytes::GetBytes(unsigned char *buffer, std::size_t bufferSize)
+            {
+                auto outputBuf = Crt::ByteBufFromEmptyArray(buffer, bufferSize);
+                if (!Crt::Crypto::GenerateRandomBytes(outputBuf, bufferSize))
+                {
+                    AWS_UNREACHABLE();
```
SergeyRyabinin, line 19: log?

tests/aws-cpp-sdk-mediastore-data-integration-tests/MediaStoreDataTest.cpp
```diff
@@ -109,7 +109,7 @@ namespace
         static Aws::String GetTestContainerName()
         {
             // "-" is not allowed in container name, convert uuid to its hex.
-            static const std::string suffix = HashingUtils::HexEncode(ByteBuffer(Aws::Utils::UUID::RandomUUID())).c_str();
+            static const std::string suffix = HashingUtils::HexEncode(ByteBuffer(Aws::Utils::UUID::RandomUUID().operator ByteBuffer())).c_str();
```
SergeyRyabinin, line 112: why did you have to change this?
it seems to be an API backward incompatible change.

src/aws-cpp-sdk-core/include/aws/core/utils/Array.h
sbiscigl: `CryptoBuffer != Crt::ByteBuf`  we actually _want_ to delete the parent unconditionally as we are taking ownership of the underlying buffer from the bytebuf. since this is a move operation `other` is invalid after this operation and `CryptoBuffer` should always delete parent, and it cannot move from self as they are different types.

src/aws-cpp-sdk-core/source/utils/HashingUtils.cpp
sbiscigl: i think the idea here was to use CRT base 64, which would be static on their end but that didnt happen, so ill revert it back to this

tests/aws-cpp-sdk-mediastore-data-integration-tests/MediaStoreDataTest.cpp
sbiscigl: This is because of @DmitriyMusatkin 's request changed for [adding a string to array ctor.](https://github.com/aws/aws-sdk-cpp/pull/2952#discussion_r1607114545). UUID is a special class in that it has a 
[operator Aws::String() const](https://github.com/aws/aws-sdk-cpp/blob/main/src/aws-cpp-sdk-core/include/aws/core/utils/UUID.h#L36)
_AND_
[operator ByteBuffer() const ](https://github.com/aws/aws-sdk-cpp/blob/main/src/aws-cpp-sdk-core/include/aws/core/utils/UUID.h#L40)

so when we convert a UUID to byte buf the ctor is ambigious now because there is a ctor with both string and byte buf. Is this a breaking API change? _sort of_? if a class has a operator defined for both String and ByteBuf which i think we can rationalize in making this change as we are the only people likely doing that.

src/aws-cpp-sdk-core/include/aws/core/utils/crypto/crt/CRTHash.h
sbiscigl: discussed offline, removing comment, its just a move ctor from the underlying type

SergeyRyabinin (approved):

PR #2954, opened 2024-05-10
Avoid excessive reply buffer copy in WinHTTP

SergeyRyabinin:
*Issue #, if available:*

Unlike in lib curl,

WinHTTP expects a pointer + size to where Windows is going to write us a reply:
```
WINHTTPAPI BOOL WinHttpReadData(
  [in]  HINTERNET hRequest,
  [out] LPVOID    lpBuffer,
  [in]  DWORD     dwNumberOfBytesToRead,
  [out] LPDWORD   lpdwNumberOfBytesRead
);
```
https://learn.microsoft.com/en-us/windows/win32/api/winhttp/nf-winhttp-winhttpreaddata
while C++'s iostream (it the central point of requests/response handling in the SDK from the initial design) manipulates C++'s streams:
```
// inserts blocks of characters 
basic_ostream& write( const char_type* s, std::streamsize count );
```
https://en.cppreference.com/w/cpp/io/basic_iostream
and can't provide raw pre-allocated buffer that WinHTTP wants.

Should WinHTTP provide std::iostream interface - this would not be an issue.

*Description of changes:*

Create a hack-y class to access raw pointers of the underlying buffer of the streambuffer to perform writes directly to the `pptr` pointer.

*Check all that applies:*

- [ ] Linux
- [ ] Windows
- [ ] Android
- [ ] MacOS
- [ ] IOS
- [ ] Other Platforms

### DIFF
--- a/src/aws-cpp-sdk-core/include/aws/core/utils/stream/StreamBufProtectedWriter.h
+++ b/src/aws-cpp-sdk-core/include/aws/core/utils/stream/StreamBufProtectedWriter.h
@@ -1,0 +1,120 @@
+
+/**
+ * Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.
+ * SPDX-License-Identifier: Apache-2.0.
+ */
+
+#pragma once
+
+#include <aws/core/Core_EXPORTS.h>
+#include <aws/core/utils/Array.h>
+#include <streambuf>
+#include <functional>
+
+namespace Aws
+{
+    namespace Utils
+    {
+        namespace Stream
+        {
+            /**
+             * This is a wrapper to perform a hack to write directly to the put area of the underlying streambuf
+             */
+            class AWS_CORE_API StreamBufProtectedWriter : public std::streambuf
+            {
+            public:
+                StreamBufProtectedWriter() = delete;
+
+                using WriterFunc = std::function<bool(char* dst, uint64_t dstSz, uint64_t& read)>;
+
+                static uint64_t WriteToBuffer(Aws::IOStream& ioStream, const WriterFunc& writerFunc)
+                {
+                    uint64_t totalRead = 0;
+
+                    while (true)
+                    {
+                        StreamBufProtectedWriter* pBufferCasted = static_cast<StreamBufProtectedWriter*>(ioStream.rdbuf());
+                        bool bufferPresent = pBufferCasted && pBufferCasted->pptr() && (pBufferCasted->pptr() < pBufferCasted->epptr());
+                        uint64_t read = 0;
+                        bool success = false;
+                        if (bufferPresent)
+                        {
+                            // have access to underlying put ptr.
+                            success = WriteDirectlyToPtr(pBufferCasted, writerFunc, read);
+                        }
+                        else
+                        {
+                            // can't access underlying buffer, stream buffer maybe be customized to not use put ptr.
+                            // or underlying put buffer is simply not initialized yet.
+                            success = WriteWithHelperBuffer(ioStream, writerFunc, read);
+                        }
+                        totalRead += read;
+                        if (!success)
+                        {
+                            break;
+                        }
+
+                        if (pBufferCasted && pBufferCasted->pptr() && (pBufferCasted->pptr() >= pBufferCasted->epptr()))
+                        {
+                            if(!ForceOverflow(ioStream, writerFunc))
+                            {
+                                break;
+                            } else {
+                                totalRead++;
+                            }
+                        }
+                    }
+                    return totalRead;
+                }
+            protected:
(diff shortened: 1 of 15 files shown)


src/aws-cpp-sdk-core/include/aws/core/utils/stream/StreamBufProtectedWriter.h
```diff
@@ -0,0 +1,101 @@
...
+                        }
+
+                        dstBegin = pBufferCasted->pptr();
+                        assert(pBufferCasted->epptr() > dstBegin);
+                        dstSz = pBufferCasted->epptr() - dstBegin;
+                        assert(dstSz);
+                    }
+
+                    read = 0;
+                    while (writerFunc(dstBegin, dstSz, read) && read > 0)
```
sbiscigl, line 78: instead of having two while loops they seem to have similar bodies, seems like we could use a do while here to avoid dry

SergeyRyabinin: there are 1 while loop. While it is a bit ugly to have similar conditions, it looks to have a bit more clear logic.

src/aws-cpp-sdk-core/include/aws/core/utils/stream/StreamBufProtectedWriter.h
```diff
@@ -0,0 +1,116 @@
...
+                        {
+                            if(!ForceOverflow(pBufferCasted, writerFunc))
+                            {
+                                return totalRead;
+                            } else {
+                                totalRead++;
+                            }
+                        }
+                    }
+                    return totalRead;
```
DmitriyMusatkin, line 66: under what conditions are you getting to this line? only way out of while is to return. is this expected or while is missing a break somewhere?

DmitriyMusatkin (approved):

src/aws-cpp-sdk-core/include/aws/core/utils/stream/StreamBufProtectedWriter.h
sbiscigl: same as dimitry also if the line is "unreachable in theory but we need to return something" we should assert use `AWS_UNREACHABLE` as a invariant

src/aws-cpp-sdk-core/include/aws/core/utils/stream/StreamBufProtectedWriter.h
```diff
@@ -0,0 +1,121 @@
...
+                {
+                    char dstChar;
+                    uint64_t read = 0;
+                    if (writerFunc(&dstChar, 1, read) && read > 0)
+                    {
+                        ioStream.write(&dstChar, 1);
+                        if (ioStream.fail()) {
+                            AWS_LOGSTREAM_ERROR("StreamBufProtectedWriter", "Failed to write 1 byte (eof: "
+                                    << ioStream.eof() << ", bad: " << ioStream.bad() << ")");
+                            return 0;
```
sbiscigl, line 80: np: please return false/true instead of int

tests/benchmark/benchmark.cpp
```diff
@@ -10,6 +10,13 @@
 #include <metric/CloudWatchMetrics.h>
 
 int main(int argc, char *argv[]) {
+    if (1 == argc ||
+        2 == argc && (std::string(argv[1]) == "-h" || std::string(argv[1]) == "--help" ))
```
sbiscigl, line 14: NICE!

sbiscigl (approved):

src/aws-cpp-sdk-core/include/aws/core/utils/stream/StreamBufProtectedWriter.h
SergeyRyabinin: changed to break + proper return

src/aws-cpp-sdk-core/include/aws/core/utils/stream/StreamBufProtectedWriter.h
```diff
@@ -0,0 +1,120 @@
...
+                static uint64_t WriteWithHelperBuffer(Aws::IOStream& ioStream, const WriterFunc& writerFunc, uint64_t& read)
+                {
+                    char tmpBuf[1024];
+                    uint64_t tmpBufSz = sizeof(tmpBuf);
+
+                    if(writerFunc(tmpBuf, tmpBufSz, read) && read > 0)
+                    {
+                        ioStream.write(tmpBuf, read);
+                        if (ioStream.fail()) {
+                            AWS_LOGSTREAM_ERROR("StreamBufProtectedWriter", "Failed to write " << tmpBufSz
```
sfod, line 96: nit: shouldn't it be `"Failed to write " << read`?

sfod (approved):

PR #2974, opened 2024-05-28
Call  `find_dependency(CURL)` if the Curl client is enabled.

teo-tsirpanis:
*Description of changes:*

This PR updates the exported config file to call `find_dependency(CURL)` if the Curl client had been enabled when building the SDK.

*Check all that applies:*

- [ ] Linux
- [ ] Windows
- [ ] Android
- [ ] MacOS
- [ ] IOS
- [ ] Other Platforms


jmklix:
Why exactly do you need this added to the cofig file? This looks like an ok change, but we want to make sure we understand what your use case is

teo-tsirpanis:
`aws-cpp-sdk-core` sometimes depends on the `CURL::libcurl` target but until now, that target is not guaranteed to exist. By finding Curl when we find the AWS SDK, we guarantee that the Curl targets exist if needed.

PR #2982, opened 2024-05-31
Add fallback to std::getenv in Aws::Environment::GetEnv for Windows

fwosar:
*Description of changes:*
In order to avoid a warning in Visual Studio, on Windows the GetEnv implementation that uses `_dupenv_s` to read environment variables. `_dupenv_s` is a non-standard function that does not exist in non-Microsoft compilers, like GCC and clang compiling against MinGW for example.

This PR adds a fallback for non-Microsoft compilers that uses `std::getenv` instead. 

*Check all that applies:*

Since this is an alternative implementation that gets selected at compile time, compiling any existing tests with an applicable toolchain will exercise this fallback method already.

*Check which platforms you have built SDK on to verify the correctness of this PR:*

- [x] Linux
- [x] Windows
- [ ] Android
- [ ] MacOS
- [ ] IOS
- [x] Other Platforms

Compiled on MinGW as well using GCC and clang.


SergeyRyabinin (approved):
Thank you for submitting this change!
We will merge it soon.

SergeyRyabinin:
Thank for submitting this pull request!

PR #2988, opened 2024-06-04
Fixed 'aws-c-common' dependency subdirectory in main AWS CMakeLists.txt.

lermy3d:
*Description of changes:*

The wrong path was used to add the 'aws-c-common' dependency since the currently used sub directory path for the 'aws-c-common' was missing the 'CMakeLists.txt' after cloning and downloading the 'aws-c-common' dependency. Let's also highlight the inconsistency with the previous list append command in line 216, so we maintained the path hierarchy from the list command and consider the add_subdirectory command an incomplete path for this command.

*Check all that applies:*
- [y] Did a review by yourself.
- [y] Added proper tests to cover this PR. (If tests are not applicable, explain.)
- [n] Checked if this PR is a breaking (APIs have been changed) change.
- [n] Checked if this PR will _not_ introduce cross-platform inconsistent behavior.
- [n] Checked if this PR would require a ReadMe/Wiki update.

- [n] Linux
- [x] Windows
- [n] Android
- [n] MacOS
- [n] IOS
- [n] Other Platforms

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -220,7 +220,7 @@ if (LEGACY_BUILD)
         set(BUILD_TESTING OFF CACHE BOOL "Disable all tests in dependencies.")
         # TODO: Use same BUILD_SHARED_LIBS for Aws Common Runtime dependencies.
         set(CRT_BUILD_SHARED_LIBS ${BUILD_SHARED_LIBS})
-        add_subdirectory(crt/aws-crt-cpp)
+        add_subdirectory(crt/aws-crt-cpp/crt/aws-c-common)
         set(BUILD_TESTING ${BUILD_TESTING_PREV})
     else ()
         # This is required in order to append /lib/cmake to each element in CMAKE_PREFIX_PATH


sbiscigl:
I don't think this is right. what that is doing is adding the subdirectory of the [aws-crt-cpp](https://github.com/aws/aws-sdk-cpp/tree/main/crt) which is the [c++ language binding](https://github.com/awslabs/aws-crt-cpp) to the c runtime library [AWS Common Runtime](https://docs.aws.amazon.com/sdkref/latest/guide/common-runtime.html).

In that project is where [aws-c-common is added](https://github.com/awslabs/aws-crt-cpp/blob/main/CMakeLists.txt#L83).

This change actually results in a build error
```
In file included from /.../src/aws-cpp-sdk-core/include/aws/core/utils/memory/stl/AWSString.h:10:
/.../src/aws-cpp-sdk-core/include/aws/core/utils/memory/stl/AWSAllocator.h:13:10: fatal error: 'aws/crt/StlAllocator.h' file not found
#include <aws/crt/StlAllocator.h>
         ^~~~~~~~~~~~~~~~~~~~~~~~
1 error generated.
```
which is because this change no longer adds the aws-crt-cpp as mentioned which is needed by the SDK and is the main point of entry to making calls to the CRT

lermy3d:
Could you share what is the dependency hierarchy on your machine? Cuz one I download the "AWS SDK C++" there is nothing on the aws/crt folder, after inspecting what is in the main aws/CMakeLists.txt on line 216 we can see that it is expecting the aws-c-common repository to be on: list(APPEND CMAKE_MODULE_PATH "${CMAKE_CURRENT_LIST_DIR}/crt/aws-crt-cpp/crt/aws-c-common/cmake")

The code snippet I am referting to:
    # build third-party targets
    if (BUILD_DEPS)
        set(CMAKE_INSTALL_RPATH "$ORIGIN")
        list(APPEND CMAKE_MODULE_PATH "${CMAKE_CURRENT_LIST_DIR}/crt/aws-crt-cpp/crt/aws-c-common/cmake")

sbiscigl:
it sounds like you did a shallow clone and not a recursive submodule clone. [referencing the readme](https://github.com/aws/aws-sdk-cpp?tab=readme-ov-file#building-from-source)

> Clone this repository with submodules
> git clone --recurse-submodules https://github.com/aws/aws-sdk-cpp

lermy3d:
> it sounds like you did a shallow clone and not a recursive submodule clone. [referencing the readme](https://github.com/aws/aws-sdk-cpp?tab=readme-ov-file#building-from-source)
> 
> > Clone this repository with submodules
> > git clone --recurse-submodules https://github.com/aws/aws-sdk-cpp

Yes! Thank I think that is what happened! I am gonna close this, I appreciate the clarification!

PR #3016, opened 2024-06-26
Receive, parse, and present http2 connection headers to the user callback

SergeyRyabinin:
*Issue #, if available:*
impossible to receive Session Id from Transcribe streaming session
*Description of changes:*
After HTTP Client has received headers for a connection
Call user handler on "headersReceived"
Add a default "headersReceived" handler to event stream operation requests and handlers to parse headers as "initial-response" event.

Also small refactoring on code gen headers de-serialization to a model object.
pending CIs and full code regen checks.
*Check all that applies:*

- [x] Linux
- [x] Windows
- [ ] Android
- [x] MacOS
- [ ] IOS
- [ ] Other Platforms

### DIFF
--- a/src/aws-cpp-sdk-core/include/aws/core/AmazonWebServiceRequest.h
+++ b/src/aws-cpp-sdk-core/include/aws/core/AmazonWebServiceRequest.h
@@ -119,38 +119,51 @@ namespace Aws
          * Set the response stream factory.
          */
         void SetResponseStreamFactory(const Aws::IOStreamFactory& factory) { m_responseStreamFactory = factory; }
+
+        ///@{
+        /**
+         * Sets the closure for headers received event.
+         */
+        inline virtual void SetHeadersReceivedEventHandler(const Aws::Http::HeadersReceivedEventHandler& headersReceivedEventHandler) { m_onHeadersReceived = headersReceivedEventHandler; }
+        inline virtual void SetHeadersReceivedEventHandler(Aws::Http::HeadersReceivedEventHandler&& headersReceivedEventHandler) { m_onHeadersReceived = std::move(headersReceivedEventHandler); }
+        ///@}
+
+        ///@{
         /**
          * Register closure for data received event.
          */
         inline virtual void SetDataReceivedEventHandler(const Aws::Http::DataReceivedEventHandler& dataReceivedEventHandler) { m_onDataReceived = dataReceivedEventHandler; }
+        inline virtual void SetDataReceivedEventHandler(Aws::Http::DataReceivedEventHandler&& dataReceivedEventHandler) { m_onDataReceived = std::move(dataReceivedEventHandler); }
+        ///@}
+
+        ///@{
         /**
-         * register closure for data sent event
+         * Register closure for data sent event
          */
         inline virtual void SetDataSentEventHandler(const Aws::Http::DataSentEventHandler& dataSentEventHandler) { m_onDataSent = dataSentEventHandler; }
+        inline virtual void SetDataSentEventHandler(Aws::Http::DataSentEventHandler&& dataSentEventHandler) { m_onDataSent = std::move(dataSentEventHandler); }
+        ///@}
+
+        ///@{
         /**
          * Register closure for  handling whether or not to continue a request.
          */
         inline virtual void SetContinueRequestHandler(const Aws::Http::ContinueRequestHandler& continueRequestHandler) { m_continueRequest = continueRequestHandler; }
-        /**
-         * Register closure for data received event.
-         */
-        inline virtual void SetDataReceivedEventHandler(Aws::Http::DataReceivedEventHandler&& dataReceivedEventHandler) { m_onDataReceived = std::move(dataReceivedEventHandler); }
-        /**
-         * register closure for data sent event
-         */
-        inline virtual void SetDataSentEventHandler(Aws::Http::DataSentEventHandler&& dataSentEventHandler) { m_onDataSent = std::move(dataSentEventHandler); }
-        /**
-         * Register closure for handling whether or not to cancel a request.
-         */
         inline virtual void SetContinueRequestHandler(Aws::Http::ContinueRequestHandler&& continueRequestHandler) { m_continueRequest = std::move(continueRequestHandler); }
+        ///@}
+
+        ///@{
         /**
          * Register closure for notification that a request is being retried
          */
         inline virtual void SetRequestRetryHandler(const RequestRetryHandler& handler) { m_requestRetryHandler = handler; }
-        /**
-         * Register closure for notification that a request is being retried
-         */
         inline virtual void SetRequestRetryHandler(RequestRetryHandler&& handler) { m_requestRetryHandler = std::move(handler); }
+        ///@}
+
+        /**
+         * get closure for headers received event.
+         */
+        inline virtual const Aws::Http::HeadersReceivedEventHandler& GetHeadersReceivedEventHandler() const { return m_onHeadersReceived; }
         /**
          * get closure for data received event.
          */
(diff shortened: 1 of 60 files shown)


src/aws-cpp-sdk-core/source/http/curl/CurlHttpClient.cpp
```diff
@@ -191,6 +191,12 @@ static size_t WriteData(char* ptr, size_t size, size_t nmemb, void* userdata)
         }
 
         HttpResponse* response = context->m_response;
+        auto& headersHandler = context->m_request->GetHeadersReceivedEventHandler();
```
SergeyRyabinin, line 194: to be honest, this is the wackiest part of this PR (the rest is quite nice).

However,
libcurl does not provide a way to receive HTTP2 headers in a raw hpack H2 event format.
libcurl always transforms HTTP2 headers into http 1.1 like headers.

Also libcurl does not have a callback to notify when all headers are received or when a single payload-less H2 frame handling is done. (technically we receive H2 frame with "connection" headers, but libcurl presents it to us like a regular http1.1 response headers).

So here, 
it is not "on headers received" but "headers maybe received but now you are receiving a payload" which is quite ugly/wacky, but it is the best I managed to get from libCurl.

The alternative would be: using a different HTTP client or [CURLOPT_CONNECT_ONLY](https://curl.se/libcurl/c/CURLOPT_CONNECT_ONLY.html) (essentially asking libcurl to not perform http2->http1.1 headers transform and implementing a significant part of http protocol ourselves).

tools/code-generation/generator/src/main/java/com/amazonaws/util/awsclientgenerator/domainmodels/codegeneration/ServiceModel.java
```diff
@@ -61,6 +61,16 @@ public boolean hasOnlyBearerAuth() {
...
     }
 
+    public Operation getOperationForRequestShapeName(String requestShapeName) {
+        for (Map.Entry<String, Operation> opEntry : operations.entrySet()) {
+            Operation op = opEntry.getValue();
+            if (op.getRequest() != null && op.getRequest().getShape().getName() == requestShapeName) {
+                return op;
+            }
+        }
+        return null;
```
sbiscigl, line 71: shouldnt ever return nullptr in java, should return [optional](https://docs.oracle.com/javase/8/docs/api/java/util/Optional.html) and deal with presence or absence of optional.

sbiscigl (approved):

tools/code-generation/generator/src/main/resources/com/amazonaws/util/awsclientgenerator/velocity/cpp/RequestEventStreamHandlerHeader.vm
```diff
@@ -38,6 +39,7 @@ namespace Model
 
     class ${operation.name}Handler : public Aws::Utils::Event::EventStreamHandler
     {
+        typedef std::function<void(const ${operation.name}InitialResponse&)> ${operation.name}ConnectionResponseCallback;
```
DmitriyMusatkin, line 42: its returning the same structure as initial response? should we just merge the 2 callbacks? do users ever need to know the difference?

SergeyRyabinin: technically it is still 2 replies.
I've just re-used the same structure for convenience (of us and the user).

ConnectionResponse is an equivalent of java sdk event stream operation level response such as [StartStreamTranscriptionResponse](https://sdk.amazonaws.com/java/api/latest/software/amazon/awssdk/services/transcribestreaming/model/StartStreamTranscriptionResponse.html). Ideally, we should refactor our event stream API to provide "operation level response too" (instead of NoResult), but it is backward incompatible. Additionally, our NoResult is presented to the user callback AFTER the streaming has completed, while java sdk Operation Response is presented at the beginning of streaming session.
Technically this is not a event stream streaming event.

while initial-response is a service event

While they technically serve the same purpose, the underlying mechanism is quite different and we don't know at the codegen time which one will be actually used by a AWS Service.
(I guess all will reply on HTTP level and in addition initial-response can be also sent).

DmitriyMusatkin: I understand the difference in whats going on in background, it just seems pretty counter intuitive from users perspective to separate them - on some apis you have to check one callback and on other apis you check a different one. This is just a rant and maybe an oversight with original design of event streams. From user perspective i would want to receive initial response data once without carrying whether its part of http response or event stream event.

Anyways, can we at least doc in more detail? i guess from our perspective we would not know which api uses which, but we can be explicit in handler that you should expect both and verify doc for which one is used.

SergeyRyabinin: Updated to use single callback for both. This callback may be called more than once though, depending on a server behavior.
Docstring is updated too to reflect this.

sbiscigl (approved):

src/aws-cpp-sdk-core/include/aws/core/AmazonWebServiceRequest.h
```diff
@@ -119,38 +119,51 @@ namespace Aws
...
         /**
          * Register closure for notification that a request is being retried
          */
         inline virtual void SetRequestRetryHandler(const RequestRetryHandler& handler) { m_requestRetryHandler = handler; }
+        inline virtual void SetRequestRetryHandler(RequestRetryHandler&& handler) { m_requestRetryHandler = std::move(handler); }
+        ///@}
+
         /**
-         * Register closure for notification that a request is being retried
+         * get closure for headers received event.
```
DmitriyMusatkin, line 164: nit: uppercase

src/aws-cpp-sdk-core/include/aws/core/utils/event/EventHeader.h
```diff
@@ -299,6 +301,36 @@ namespace Aws
...
+                        case EventHeaderType::BYTE:
+                            return Utils::StringUtils::to_string(GetEventHeaderValueAsByte());
+                        case EventHeaderType::INT16:
+                            return Utils::StringUtils::to_string(GetEventHeaderValueAsInt16());
+                        case EventHeaderType::INT32:
+                            return Utils::StringUtils::to_string(GetEventHeaderValueAsInt32());
+                        case EventHeaderType::INT64:
+                            return Utils::StringUtils::to_string(GetEventHeaderValueAsInt64());
+                        case EventHeaderType::BYTE_BUF:
+                            return Aws::String(reinterpret_cast<char*>(GetEventHeaderValueAsBytebuf().GetUnderlyingData()), GetEventHeaderValueAsBytebuf().GetLength());
```
DmitriyMusatkin, line 320: byte buf provides no guarantees to be null terminated. this seems dangerous

DmitriyMusatkin (approved):

src/aws-cpp-sdk-core/include/aws/core/utils/event/EventHeader.h
SergeyRyabinin: That's why the string is constructed from ptr + length, not just null-terminated pointer.

DmitriyMusatkin: oh yeah, you are right, missed the final part. ignore this comment. 
we have internal methods to get a proper c_str from buf internally, but i dont think we expose those in bindings

PR #3024, opened 2024-07-02
S3 200 error handling

sbera87:
*Issue #, if available:*

*Description of changes:*

*Check all that applies:*

- [x] Linux
- [x] Windows
- [ ] Android
- [ ] MacOS
- [ ] IOS
- [ ] Other Platforms


src/aws-cpp-sdk-core/source/client/AWSClient.cpp
```diff
@@ -1162,3 +1166,44 @@ void AWSClient::AppendRecursionDetectionHeader(std::shared_ptr<Aws::Http::HttpRe
...
+    Aws::String message{"Error in body of the response"};
+    //extract error message and code in the body
+    auto& body = httpResponse->GetResponseBody();
+    auto responseCode = httpResponse->GetResponseCode();
+    auto readPointer = body.tellg();
+    XmlDocument doc = XmlDocument::CreateFromXmlStream(body);
+    Aws::String bodyError;
+    auto coreErrorCode = CoreErrors::INTERNAL_FAILURE;
+    
+    if (doc.WasParseSuccessful() &&
```
SergeyRyabinin, line 1184: this logic should be applied only on S3 Client.

I'd suggest to add this logic into generated [S3ErrorMarshaller](https://github.com/aws/aws-sdk-cpp/blob/main/generated/src/aws-cpp-sdk-s3/include/aws/s3/S3ErrorMarshaller.h#L16)

The base class has a [virtual method `Marshall`](https://github.com/aws/aws-sdk-cpp/blob/a95671dc38b2c61a18f5398d1e549278e1fbe440/src/aws-cpp-sdk-core/source/client/AWSErrorMarshaller.cpp#L94) that could be overwritten.

src/aws-cpp-sdk-core/source/client/AWSClient.cpp
```diff
@@ -582,12 +582,16 @@ HttpResponseOutcome AWSClient::AttemptOneRequest(const std::shared_ptr<Aws::Http
...
-    if (DoesResponseGenerateError(httpResponse) || request.HasEmbeddedError(httpResponse->GetResponseBody(), httpResponse->GetHeaders()))
+    if (DoesResponseGenerateError(httpResponse) )
     {
         AWS_LOGSTREAM_DEBUG(AWS_CLIENT_LOG_TAG, "Request returned error. Attempting to generate appropriate error codes from response");
         auto error = BuildAWSError(httpResponse);
         return HttpResponseOutcome(std::move(error));
     }
+    else if(request.HasEmbeddedError(httpResponse->GetResponseBody(), httpResponse->GetHeaders()))
+    {
+        return HttpResponseOutcome(std::move(BuildAWSErrorFromResponseBody(httpResponse)) );
```
sbiscigl, line 593: gives a clang warning `error: moving a temporary object prevents copy elision [-Werror,-Wpessimizing-move]`

should update this to use a initializer list

```cpp
return {BuildAWSErrorFromResponseBody(httpResponse)};
```

sbera87: ack

tests/testing-resources/include/aws/testing/mocks/aws/client/MockAWSClient.h
```diff
@@ -41,8 +43,14 @@ class AmazonWebServiceRequestMock : public Aws::AmazonWebServiceRequest
     virtual const char* GetServiceRequestName() const override { return "AmazonWebServiceRequestMock"; }
     virtual bool HasEmbeddedError(Aws::IOStream& body, const Aws::Http::HeaderValueCollection& header) const override {
         (void) header;
+        
+        auto readPointer = body.tellg();
```
sbera87, line 47: This fixes the body of response parsing issue existing in test cases. The pointer in the stream was being moved to the end which caused subsequent parsing to fail.

tests/testing-resources/include/aws/testing/mocks/aws/client/MockAWSClient.h
```diff
@@ -61,7 +69,7 @@ class CountedRetryStrategy : public Aws::Client::DefaultRetryStrategy
           m_attemptedRetries(0)
       {}
     CountedRetryStrategy(long maxRetires)
-        : Aws::Client::DefaultRetryStrategy(maxRetires <= 0 ? (std::numeric_limits<long>::max)() : maxRetires),
+        : Aws::Client::DefaultRetryStrategy(maxRetires < 0 ? (std::numeric_limits<long>::max)() : maxRetires),
```
sbera87, line 72: This allows to try explicit tests without retry . There are times when we want to test the workflow in the first iteration only and not retry (which changes the error codes)

tests/testing-resources/include/aws/testing/mocks/aws/client/MockAWSClient.h
```diff
@@ -97,12 +105,21 @@ class MockAWSErrorMarshaller : public Aws::Client::AWSErrorMarshaller
...
+
+     MockAWSClient(const Aws::Client::ClientConfiguration& config, 
+                  const std::shared_ptr<Aws::Client::AWSErrorMarshaller>& errorMarshaller
+    ) : AWSClient(  config,
+                    Aws::MakeShared<Aws::Client::AWSAuthV4Signer>("MockAWSClient",
+                    Aws::MakeShared<Aws::Auth::SimpleAWSCredentialsProvider>("MockAWSClient", GetMockAccessKey(),
                     GetMockSecretAccessKey()), "service", config.region.empty() ? Aws::Region::US_EAST_1 : config.region),
-            Aws::MakeShared<MockAWSErrorMarshaller>("MockAWSClient")),
-        m_countedRetryStrategy(std::static_pointer_cast<CountedRetryStrategy>(config.retryStrategy)) { }
+                    errorMarshaller
```
sbera87, line 115: Just adding delegate constructor and leveraging it in passing different error marshallers for different test use cases

src/aws-cpp-sdk-core/source/client/AWSXmlClient.cpp
```diff
@@ -186,9 +186,19 @@ AWSError<CoreErrors> AWSXMLClient::BuildAWSError(const std::shared_ptr<Http::Htt
...
+        if(!httpResponse->GetResponseBody())
+        {
+            ss << "No response body.";
+        }
+        else
+        {
+            Aws::String xmlString((Aws::IStreamBufIterator(httpResponse->GetResponseBody())), Aws::IStreamBufIterator());
+
+            ss <<xmlString;
+            ss << " current pointer location in stream = "<<httpResponse->GetResponseBody().tellp();
```
sbiscigl, line 200: instead of returning the pointer location lets just stilll return `No response body.` and log out a debug statement. The message consumed as a error message in a response could be somewhat confusing.

tools/code-generation/generator/src/main/resources/com/amazonaws/util/awsclientgenerator/velocity/cpp/StreamRequestSource.vm
```diff
@@ -76,6 +76,32 @@ Aws::Http::HeaderValueCollection ${typeInfo.className}::GetRequestSpecificHeader
 
 }
 #end
+
```
sbiscigl, line 79: ah this is is what is causing the whitespace change, put it inside the `#if`. a empty line in  a ` vm` file will result in a new line in the code that is generated.

tools/code-generation/generator/src/main/resources/com/amazonaws/util/awsclientgenerator/velocity/cpp/s3/PutBucketNotificationConfigurationRequest.vm
```diff
@@ -83,4 +83,30 @@ Aws::Http::HeaderValueCollection ${typeInfo.className}::GetRequestSpecificHeader
   return headers;
 }
 #end
+
```
sbiscigl, line 86: same thing here no new white space change outside of the if statement

generated/src/aws-cpp-sdk-s3/source/S3ErrorMarshaller.cpp
```diff
@@ -96,4 +102,39 @@ Aws::String S3ErrorMarshaller::ExtractEndpoint(const AWSError<CoreErrors>& error
   }
 
   return {};
+}
+
+AWSError<Aws::Client::CoreErrors> S3ErrorMarshaller::Marshall(const Aws::Http::HttpResponse& httpResponse) const
```
sbiscigl, line 107: hm this isnt gunna work, by this logic the error would _always_ be in the body and we would never deserialize for a normal http error. i think you have to adopt your logic to look something like

```cpp
if (httpResponse.GetResponseCode() == HttpResponseCode::OK)
{
    // look in body
} else
{
    // run what is in XmlErrorMarshaller
}
```

tests/aws-cpp-sdk-s3-unit-tests/S3UnitTests.cpp
```diff
@@ -16,8 +17,29 @@ using namespace Aws::S3::Model;
 
 const char* ALLOCATION_TAG = "S3UnitTest";
 
+class S3UnitTest_S3EmbeddedErrorTestNonOKResponse_Test;
+
+class S3TestClient : public S3Client
```
sbera87, line 22: This is a quick way of adding a test wrapper on top of S3 client class to be able to make unit test related changes

tests/aws-cpp-sdk-s3-unit-tests/S3UnitTests.cpp
```diff
@@ -185,20 +207,60 @@ TEST_F(S3UnitTest, S3EmbeddedErrorTest) {
...
   _mockHttpClient->AddResponseToReturn(mockResponse);
 
   const auto response = _s3Client->CopyObject(request);
+
+  EXPECT_FALSE(response.IsSuccess());
+}
+
+
+//Set http error and error in body in a way to hit generic xml error marshaller, which sets the exception name
+TEST_F(S3UnitTest, S3EmbeddedErrorTestNonOKResponse) {
```
sbera87, line 235: This test case improves code coverage and tests corner case when there is a error body in response but http response code is not ok as well as the client is not S3. This defaults to generic error marshaller then.

tests/aws-cpp-sdk-s3-unit-tests/S3UnitTests.cpp
```diff
@@ -185,20 +207,60 @@ TEST_F(S3UnitTest, S3EmbeddedErrorTest) {
     .WithBucket("testBucket")
     .WithKey("testKey")
     .WithCopySource("testSource");
-
+  
   auto mockRequest = Aws::MakeShared<Standard::StandardHttpRequest>(ALLOCATION_TAG, "mockuri", HttpMethod::HTTP_GET);
   mockRequest->SetResponseStreamFactory([]() -> IOStream* {
-    return Aws::New<StringStream>(ALLOCATION_TAG, R"(<?xml version="1.0" encoding="UTF-8"?><Error><Code>SlowDown</Code><Message>Please reduce your request rate.</Message></Error>)", std::ios_base::in | std::ios_base::binary);
+    const std::string mockResponseString = "\n         <?xml version=\"1.0\" encoding=\"UTF-8\"?>\n\n         <Error>\n          <Code>InternalError</Code>\n          <Message>We encountered an internal error. Please try again.</Message>\n          <RequestId>656c76696e6727732072657175657374</RequestId>\n          <HostId>Uuag1LuByRx9e6j5Onimru9pO4ZVKnJ2Qz7/C1NPcfTWAtRPfTaOFg==</HostId>\n         </Error>\n    ";
```
sbera87, line 213: add a retryable error

tests/aws-cpp-sdk-s3-unit-tests/S3UnitTests.cpp
```diff
@@ -16,8 +17,29 @@ using namespace Aws::S3::Model;
 
 const char* ALLOCATION_TAG = "S3UnitTest";
 
+class S3UnitTest_S3EmbeddedErrorTestNonOKResponse_Test;
```
sbera87, line 20: Forward declaration of test class for friend class in google test suite

tools/code-generation/generator/src/main/resources/com/amazonaws/util/awsclientgenerator/velocity/cpp/s3/S3ErrorMarshallerSource.vm
```diff
@@ -96,4 +101,43 @@ Aws::String ${metadata.classNamePrefix}ErrorMarshaller::ExtractEndpoint(const AW
   }
 
   return {};
+}
+AWSError<Aws::Client::CoreErrors> ${metadata.classNamePrefix}ErrorMarshaller::Marshall(const Aws::Http::HttpResponse& httpResponse) const
+{
+  //if response code not ok, use error marshaller
+  if(httpResponse.GetResponseCode() != HttpResponseCode::OK)
+  {
+    return XmlErrorMarshaller::Marshall(httpResponse);
```
sbera87, line 110: for any error code other than ok, fall back to generic xml error marshaller

generated/src/aws-cpp-sdk-s3-crt/source/S3CrtErrorMarshaller.cpp
```diff
@@ -96,4 +101,43 @@ Aws::String S3CrtErrorMarshaller::ExtractEndpoint(const AWSError<CoreErrors>& er
...
+}
+AWSError<Aws::Client::CoreErrors> S3CrtErrorMarshaller::Marshall(const Aws::Http::HttpResponse& httpResponse) const
+{
+  //if response code not ok, use error marshaller
+  if(httpResponse.GetResponseCode() != HttpResponseCode::OK)
+  {
+    return XmlErrorMarshaller::Marshall(httpResponse);
+  }
+
+  Aws::String message{"Error in body of the response"};
```
SergeyRyabinin, line 113: why do we need this initial string value?

generated/src/aws-cpp-sdk-s3-crt/source/S3CrtErrorMarshaller.cpp
```diff
@@ -96,4 +101,43 @@ Aws::String S3CrtErrorMarshaller::ExtractEndpoint(const AWSError<CoreErrors>& er
...
+  Aws::String bodyError;
+
+  if (doc.WasParseSuccessful() &&
+      !doc.GetRootElement().IsNull() && 
+      doc.GetRootElement().GetName() == Aws::String("Error")) 
+  {        
+      auto messageNode = doc.GetRootElement().FirstChild("Message") ;
+      if(!messageNode.IsNull())
+      {
+          message = messageNode.GetText();
```
SergeyRyabinin, line 127: > GetText
> Get the inner text of the node (potentially includes other nodes)

Please check that the xml node does not have any children (to avoid accidental tracing of wrong message). The best case would be to check if the node contains only text value, but our XML wrapper seems to not provide such api.

generated/src/aws-cpp-sdk-s3-crt/source/S3CrtErrorMarshaller.cpp
```diff
@@ -96,4 +101,43 @@ Aws::String S3CrtErrorMarshaller::ExtractEndpoint(const AWSError<CoreErrors>& er
...
+      }
+      auto codeNode = doc.GetRootElement().FirstChild("Code") ;
+      if(!codeNode.IsNull())
+      {
+          bodyError = codeNode.GetText();
+      }
+  }
+  body.seekg(readPointer);
```
SergeyRyabinin, line 135: nit: I think it would be better to rewind the stream right after parsing (anyway parsing is done by copying the stream into another string).

generated/src/aws-cpp-sdk-s3-crt/source/model/AbortMultipartUploadRequest.cpp
```diff
@@ -27,6 +26,28 @@ AbortMultipartUploadRequest::AbortMultipartUploadRequest() :
 {
 }
 
+bool AbortMultipartUploadRequest::HasEmbeddedError(Aws::IOStream &body,
+  const Aws::Http::HeaderValueCollection &header) const
+{
+  // Header is unused
+  (void) header;
```
SergeyRyabinin, line 33: nit but please change: `AWS_UNREFERENCED_PARAM(header);` or just remove the header variable name from the signature.

generated/src/aws-cpp-sdk-s3-crt/source/S3CrtErrorMarshaller.cpp
```diff
@@ -96,4 +101,43 @@ Aws::String S3CrtErrorMarshaller::ExtractEndpoint(const AWSError<CoreErrors>& er
...
+  //if response code not ok, use error marshaller
+  if(httpResponse.GetResponseCode() != HttpResponseCode::OK)
+  {
+    return XmlErrorMarshaller::Marshall(httpResponse);
+  }
+
+  Aws::String message{"Error in body of the response"};
+  //extract error message and code in the body
+  auto& body = httpResponse.GetResponseBody();
+  auto readPointer = body.tellg();
```
SergeyRyabinin, line 116: check if the stream is valid, i.e. [`good`](https://en.cppreference.com/w/cpp/io/basic_ios/good) before doing operations on it.

src/aws-cpp-sdk-core/source/client/AWSXmlClient.cpp
```diff
@@ -186,9 +186,18 @@ AWSError<CoreErrors> AWSXMLClient::BuildAWSError(const std::shared_ptr<Http::Htt
...
         Aws::StringStream ss;
-        ss << "No response body.";
+        
+        if(!httpResponse->GetResponseBody())
+        {
+            ss << "No response body.";
+        }
+        else
+        {
+            Aws::String xmlString((Aws::IStreamBufIterator(httpResponse->GetResponseBody())), Aws::IStreamBufIterator());
```
SergeyRyabinin, line 197: some debug tracing here, please remove or update

generated/src/aws-cpp-sdk-s3-crt/source/model/CreateBucketRequest.cpp
```diff
@@ -34,6 +33,28 @@ CreateBucketRequest::CreateBucketRequest() :
 {
 }
 
+bool CreateBucketRequest::HasEmbeddedError(Aws::IOStream &body,
+  const Aws::Http::HeaderValueCollection &header) const
+{
+  // Header is unused
+  (void) header;
+
+  auto readPointer = body.tellg();
```
SergeyRyabinin, line 42: not very clear why a similar logic is here and in ErrorMarshaller (well I guess I understand why, because one answers if there is such error and another parses and error to the object).

nit: would be nice to extract this logic into some common function in S3 Client to avoid copy-pasted generated code.

generated/src/aws-cpp-sdk-s3-crt/source/model/AbortMultipartUploadRequest.cpp
sbera87: ack

generated/src/aws-cpp-sdk-s3-crt/source/model/CreateBucketRequest.cpp
sbera87: While I agree, this cleanup can be done as part of a separate PR after further review of XML error parsing. The XML parser does a few more steps than this . As of now, we can keep S3 handling separate before redirecting to a common API

generated/src/aws-cpp-sdk-s3-crt/source/S3CrtErrorMarshaller.cpp
sbera87: Ack. added an extra check if have children. But in my opinion this would still be ok because as long as the first root element is error , we will indicate error in response body. The specific code is obtained from parsing the exact message child node which is where the additional check will help

generated/src/aws-cpp-sdk-s3-crt/source/S3CrtErrorMarshaller.cpp
sbera87: ack, in this case it won't matter since the steam was already parsed and string copied over already by this point. but I can move it to the end of the function.

SergeyRyabinin (approved):

Issue #3057, opened 2024-07-26
Invalid Region Constant in Aws::Region (AP_NORTHEAST_4) & Missing Region Constant (AP_SOUTHEAST_4)


James-02:
### Describe the bug

Within `Aws::Region` there is a constant `AP_NORTHEAST_4`, which translates to `ap-northeast-4`, however this is not an actual region available in AWS: https://docs.aws.amazon.com/general/latest/gr/s3.html

### Expected Behavior

The `AP_NORTHEAST_4` constant should actually be `AP_SOUTHEAST_4`, mapping to `ap-southeast-4`

### Current Behavior

N/A

### Reproduction Steps

```C++
#include <aws/core/Aws.h>

std::cout << Aws::Region::AP_NORTHEAST_4 << std::endl;

```

### Possible Solution

Rename the invalid constant to `AP_SOUTHEAST_4` and it's string value to `ap-southeast-4`

### AWS CPP SDK version used

1.11.352

### Compiler and Version used

gcc (GCC) 11.4.1 20231218 (Red Hat 11.4.1-3)

### Operating System and version

Rocky Linux 9.3

sbiscigl:
Yep thats what we get for hand editing it, created a [PR](https://github.com/aws/aws-sdk-cpp/pull/3059) to code generate the enum based on the regions defined in `parititons.json`. Will merge on monday pending any action items on the PR. Thanks for the issue!

PR #3059, opened 2024-07-26
codegen aws regions enums based on partitions.json

sbiscigl:
*Issue #, if available:*

[issues/3057](https://github.com/aws/aws-sdk-cpp/issues/3057)

*Description of changes:*

codegen `Region.h` enum from partitions.json instead of hand editing to prevent potential errors from hand editing adding new regions.

*Check all that applies:*

- [x] Linux
- [x] Windows
- [x] Android
- [x] MacOS
- [ ] IOS
- [ ] Other Platforms

### DIFF
--- a/src/aws-cpp-sdk-core/include/aws/core/Region.h
+++ b/src/aws-cpp-sdk-core/include/aws/core/Region.h
@@ -6,7 +6,6 @@
 #pragma once
 
 #include <aws/core/Core_EXPORTS.h>
-
 #include <aws/core/utils/memory/stl/AWSString.h>
 
 namespace Aws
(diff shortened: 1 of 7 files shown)


SergeyRyabinin (approved):

src/aws-cpp-sdk-core/include/aws/core/Region.h
```diff
@@ -16,44 +15,48 @@ namespace Aws
      */
     namespace Region
     {
-        // AWS_GLOBAL is a pseudo region that can be used to tell SDK to use the service global endpoint if there is any.
```
DmitriyMusatkin, line 19: do you want to keep that comment about global?

DmitriyMusatkin (approved):

src/aws-cpp-sdk-core/include/aws/core/Region.h
sbiscigl: nah now that theres several global regions, and putting the global region as the first entry of the enum adds a bit of weird custom logic that makes this change a bit messy.

also [existing AWS documentation on this exists that users could refer to](https://docs.aws.amazon.com/sdkref/latest/guide/feature-region.html).

> aws-global specifies the global endpoint for services that support a separate global endpoint in addition to Regional endpoints, such as AWS Security Token Service (AWS STS) and Amazon Simple Storage Service (Amazon S3).

PR #3060, opened 2024-07-30
fix noisy log when MD5 checksum is specified

sbiscigl:
*Description of changes:*

Fixes a problem where the following code would emit the log

> Checksum algorithm: md5is not supported by SDK.

even though the checksum was set and is supported. The log is incorrect and should not be logged in this case as MD5 was set and not calculated.

```
#include <aws/core/Aws.h>
#include <aws/s3/S3Client.h>
#include <aws/core/utils/HashingUtils.h>
#include <aws/s3/model/PutObjectRequest.h>

using namespace Aws;
using namespace Aws::S3;
using namespace Aws::Utils;

const char * ALLOCATION_TAG = "test";

int main()
{
    SDKOptions options;
    options.loggingOptions.logLevel = Utils::Logging::LogLevel::Info;
    InitAPI(options);
    {
        S3Client client{};
        auto request = PutObjectRequest().WithBucket("bucket_name").WithKey("key");
        const String body{"a test string"};
        std::shared_ptr<Aws::IOStream> inputData = Aws::MakeShared<Aws::StringStream>(ALLOCATION_TAG,
          body,
          std::ios_base::in | std::ios_base::binary);
        request.SetBody(inputData);
        request.SetContentMD5(HashingUtils::Base64Encode(HashingUtils::CalculateMD5(body)));
        assert(client.PutObject(request).IsSuccess());
    }
    ShutdownAPI(options);
    return 0;
}
```

*Check all that applies:*

- [x] Linux
- [x] Windows
- [x] Android
- [x] MacOS
- [ ] IOS
- [ ] Other Platforms

### DIFF
--- a/src/aws-cpp-sdk-core/source/client/AWSClient.cpp
+++ b/src/aws-cpp-sdk-core/source/client/AWSClient.cpp
@@ -866,9 +866,9 @@ void AWSClient::AddChecksumToRequest(const std::shared_ptr<Aws::Http::HttpReques
         {
             httpRequest->SetHeaderValue(Http::CONTENT_MD5_HEADER, HashingUtils::Base64Encode(HashingUtils::CalculateMD5(*(GetBodyStream(request)))));
         }
-        else
+        else if (headers.find(CONTENT_MD5_HEADER) == headers.end())
         {
-            AWS_LOGSTREAM_WARN(AWS_CLIENT_LOG_TAG, "Checksum algorithm: " << checksumAlgorithmName << "is not supported by SDK.");
+            AWS_LOGSTREAM_WARN(AWS_CLIENT_LOG_TAG, "Checksum algorithm: " << checksumAlgorithmName << " is not supported by SDK.");
         }
     }
 


src/aws-cpp-sdk-core/source/client/AWSClient.cpp
```diff
@@ -866,7 +866,7 @@ void AWSClient::AddChecksumToRequest(const std::shared_ptr<Aws::Http::HttpReques
         {
             httpRequest->SetHeaderValue(Http::CONTENT_MD5_HEADER, HashingUtils::Base64Encode(HashingUtils::CalculateMD5(*(GetBodyStream(request)))));
         }
-        else
+        else if (headers.find(CONTENT_MD5_HEADER) == headers.end())
         {
             AWS_LOGSTREAM_WARN(AWS_CLIENT_LOG_TAG, "Checksum algorithm: " << checksumAlgorithmName << "is not supported by SDK.");
```
jeking3, line 871: Add a space before `"is` as well.

src/aws-cpp-sdk-core/source/client/AWSClient.cpp
```diff
@@ -866,7 +866,7 @@ void AWSClient::AddChecksumToRequest(const std::shared_ptr<Aws::Http::HttpReques
         {
```
jeking3, line 866: I'm concerned there is some confusion about checksums and MD5.  MD5 is part of the S3 API (PutObject, for example) and is _not_ a checksum algorithm.  CRC32, CRC64, SHA1, and SHA256 are.

https://docs.aws.amazon.com/AmazonS3/latest/API/API_PutObject.html

DmitriyMusatkin: this has some extra info - https://aws.amazon.com/blogs/aws/new-additional-checksum-algorithms-for-amazon-s3/.

Unfortunately, historically how md5 (contentMD5 header) works has not been documented particularly well and there are differences in what scenarios sdks compute it. In general, MD5 is the original checksumming algo that s3 has (and its used purely for integrity purposes and s3 does not rely on any security properties of md5) and other algoes, like crc32/sha1/etc... were added as an alternative to MD5 - meaning if they are set, md5 is not computed anymore.

DmitriyMusatkin (approved):

src/aws-cpp-sdk-core/source/client/AWSClient.cpp
jeking3: Please consider the following before making changes in this area: other cloud object storage systems like GCS, Alibaba, Seagate, Oracle; and on-premises systems like IBM, Dell Isilon all implement the S3 API and still require the use of Content-MD5.  If S3 does not want to use MD5 any more, fine, but removing Content-MD5 from the SDK breaks the whole ecosystem.

SergeyRyabinin (approved):

PR #3064, opened 2024-07-31
Disable endpoint discovery on the dynamodb by default

SergeyRyabinin:
*Issue #, if available:*
Disable endpoint discovery on the dynamodb by default as it causes a lot of pain.
It is a feature on a deprecation path and the service team suggests to not use it.
It is there purely for backward compatibility.

Still, 2 clients of Timestream would require endpoint discovery enabled by default (because it is required, otherwise the API call would fail)
*Description of changes:*
Get rid of intermediate template GenericClientConfiguration, there are only 3 service clients using the endpoint discovery, so they will generate their service-specific ClientConfiguration classes.

*Check all that applies:*

- [ ] Linux
- [ ] Windows
- [ ] Android
- [ ] MacOS
- [ ] IOS
- [ ] Other Platforms

### DIFF
--- a/generated/src/aws-cpp-sdk-dynamodb/source/DynamoDBClientConfiguration.cpp
+++ b/generated/src/aws-cpp-sdk-dynamodb/source/DynamoDBClientConfiguration.cpp
@@ -1,0 +1,74 @@
+﻿/**
+ * Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.
+ * SPDX-License-Identifier: Apache-2.0.
+ */
+
+#include <aws/dynamodb/DynamoDBClientConfiguration.h>
+
+namespace Aws
+{
+namespace DynamoDB
+{
+
+
+bool IsEndpointDiscoveryEnabled(const Aws::String& endpointOverride, const Aws::String &profileName)
+{
+  bool enabled = false;
+
+  if (!endpointOverride.empty())
+  {
+    enabled = false;
+  }
+  else
+  {
+    static const char* AWS_ENABLE_ENDPOINT_DISCOVERY_ENV_KEY = "AWS_ENABLE_ENDPOINT_DISCOVERY";
+    static const char* AWS_ENABLE_ENDPOINT_DISCOVERY_PROFILE_KEY = "endpoint_discovery_enabled";
+    static const char* AWS_EP_DISCOVERY_ENABLED = "true";
+    static const char* AWS_EP_DISCOVERY_DISABLED = "false";
+    static const char* DEFAULT_VALUE_FOR_DYNAMODB = AWS_EP_DISCOVERY_DISABLED;
+
+    Aws::String configVal = Client::ClientConfiguration::LoadConfigFromEnvOrProfile(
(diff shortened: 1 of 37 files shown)


generated/src/aws-cpp-sdk-dynamodb/source/DynamoDBClientConfiguration.cpp
```diff
@@ -0,0 +1,70 @@
...
+    static const char* AWS_ENABLE_ENDPOINT_ENABLED = "true";
+    static const char* AWS_ENABLE_ENDPOINT_DISABLED = "false";
+
+    Aws::String enableEndpointDiscovery = Client::ClientConfiguration::LoadConfigFromEnvOrProfile(
+        AWS_ENABLE_ENDPOINT_DISCOVERY_ENV_KEY, profileName, AWS_ENABLE_ENDPOINT_DISCOVERY_PROFILE_KEY,
+        {AWS_ENABLE_ENDPOINT_ENABLED, AWS_ENABLE_ENDPOINT_DISABLED}, AWS_ENABLE_ENDPOINT_ENABLED);
+
+    if (AWS_ENABLE_ENDPOINT_DISABLED == enableEndpointDiscovery)
+    {
+      // enabled by default unless explicitly disabled in ENV, profile config file, or programmatically later
```
DmitriyMusatkin, line 34: dont we want disabled by default unless explicitly enabled in profile or config. or maybe im misreading what this means

generated/src/aws-cpp-sdk-dynamodb/source/DynamoDBClientConfiguration.cpp
```diff
@@ -0,0 +1,70 @@
...
+    enabled = false;
+  }
+  else
+  {
+    static const char* AWS_ENABLE_ENDPOINT_DISCOVERY_ENV_KEY = "AWS_ENABLE_ENDPOINT_DISCOVERY";
+    static const char* AWS_ENABLE_ENDPOINT_DISCOVERY_PROFILE_KEY = "AWS_ENABLE_ENDPOINT_DISCOVERY";
+    static const char* AWS_ENABLE_ENDPOINT_ENABLED = "true";
+    static const char* AWS_ENABLE_ENDPOINT_DISABLED = "false";
+
+    Aws::String enableEndpointDiscovery = Client::ClientConfiguration::LoadConfigFromEnvOrProfile(
```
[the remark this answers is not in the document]
SergeyRyabinin: The profile file can change and we must reload it.
The env variable technically won't change after the app launch, but it can be changed during the runtime by the application itself.
So it works as it worked before, and it is client-dependent.

generated/src/aws-cpp-sdk-dynamodb/source/DynamoDBClientConfiguration.cpp
SergeyRyabinin: I've just moved some code blocks around, the comment does not match the logic.
Thanks for noticing, let me update it.

sbiscigl (approved):

DmitriyMusatkin (approved):

PR #3067, opened 2024-08-01
Fix build on smithy client with test actually allocating a client

SergeyRyabinin:
*Issue #, if available:*

*Description of changes:*

*Check all that applies:*

- [ ] Linux
- [ ] Windows
- [ ] Android
- [ ] MacOS
- [ ] IOS
- [ ] Other Platforms

### DIFF
--- a/src/aws-cpp-sdk-core/include/smithy/client/AwsSmithyClient.h
+++ b/src/aws-cpp-sdk-core/include/smithy/client/AwsSmithyClient.h
@@ -31,22 +31,19 @@ namespace client
     class AwsSmithyClientT : public AwsSmithyClientBase
     {
     public:
-        AwsSmithyClientT(Aws::Client::ClientConfiguration& clientConfig, const Aws::String& serviceName,
+        AwsSmithyClientT(const ServiceClientConfigurationT& clientConfig, const Aws::String& serviceName,
             const std::shared_ptr<Aws::Http::HttpClient>& httpClient,
             const std::shared_ptr<Aws::Client::AWSErrorMarshaller>& errorMarshaller,
-            const ServiceClientConfigurationT& m_client_config,
             const std::shared_ptr<EndpointProviderT> endpointProvider,
-            const std::shared_ptr<ServiceAuthSchemeResolverT>& m_auth_scheme_resolver,
-            const Aws::UnorderedMap<Aws::String, AuthSchemesVariantT>& m_auth_schemes)
+            const std::shared_ptr<ServiceAuthSchemeResolverT>& authSchemeResolver,
+            const Aws::UnorderedMap<Aws::String, AuthSchemesVariantT>& authSchemes)
             : AwsSmithyClientBase(clientConfig, serviceName, httpClient, errorMarshaller),
-              m_clientConfig(m_client_config),
+              m_clientConfig(clientConfig),
               m_endpointProvider(endpointProvider),
-              m_authSchemeResolver(m_auth_scheme_resolver),
-              m_authSchemes(m_auth_schemes)
+              m_authSchemeResolver(authSchemeResolver),
+              m_authSchemes(authSchemes)
         {
-            if (ServiceNameT) {
-                m_serviceName = ServiceNameT;
-            }
+            m_serviceName = ServiceNameT;
         }
 
         virtual ~AwsSmithyClientT() = default;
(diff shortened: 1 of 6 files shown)


sbera87 (commented):
Please add type traits to add some must have checks on the types

src/aws-cpp-sdk-core/include/smithy/client/AwsSmithyClient.h
```diff
@@ -31,22 +31,19 @@ namespace client
     class AwsSmithyClientT : public AwsSmithyClientBase
     {
     public:
-        AwsSmithyClientT(Aws::Client::ClientConfiguration& clientConfig, const Aws::String& serviceName,
```
sbera87, line 34: This should be explicit to avoid usage of default constructor. 
Also, should client be copyable, assignable, moveable?  Interface classes need more scrutiny

src/aws-cpp-sdk-core/include/smithy/client/AwsSmithyClient.h
```diff
@@ -31,22 +31,19 @@ namespace client
...
+        AwsSmithyClientT(const ServiceClientConfigurationT& clientConfig, const Aws::String& serviceName,
             const std::shared_ptr<Aws::Http::HttpClient>& httpClient,
             const std::shared_ptr<Aws::Client::AWSErrorMarshaller>& errorMarshaller,
-            const ServiceClientConfigurationT& m_client_config,
             const std::shared_ptr<EndpointProviderT> endpointProvider,
-            const std::shared_ptr<ServiceAuthSchemeResolverT>& m_auth_scheme_resolver,
-            const Aws::UnorderedMap<Aws::String, AuthSchemesVariantT>& m_auth_schemes)
+            const std::shared_ptr<ServiceAuthSchemeResolverT>& authSchemeResolver,
+            const Aws::UnorderedMap<Aws::String, AuthSchemesVariantT>& authSchemes)
```
sbera87, line 39: add type traits check to ensure its indeed a variant type

src/aws-cpp-sdk-core/include/smithy/client/AwsSmithyClient.h
```diff
@@ -70,28 +67,32 @@ namespace client
...
                     for (const auto& serviceParam : serviceParams->parameterMap) {
-                        identityParams.insert({serviceParam.first, serviceParam.second});
+                        identityParams.additionalProperties.insert({serviceParam.first, serviceParam.second});
                     }
                 }
             }
             Aws::Vector<AuthSchemeOption> authSchemeOptions = m_authSchemeResolver->resolveAuthScheme(identityParams);
 
-            auto authSchemeOptionIt = std::find_first_of(authSchemeOptions.begin(), authSchemeOptions.end(),
+            auto authSchemeOptionIt = std::find_if(authSchemeOptions.begin(), authSchemeOptions.end(),
```
sbera87, line 95: why are we asserting "assert(authSchemeOptionIt != authSchemeOptions.end());" when the else case of auth scheme not found we return AWS error. In such a case we wont ever return error then. 
Assertions are not good and should be handled gracefully with proper error

src/aws-cpp-sdk-core/include/smithy/client/AwsSmithyClient.h
```diff
@@ -70,28 +67,32 @@ namespace client
             assert(m_authSchemeResolver);
```
sbera87, line 67: shouldn't assert but handle gracefully with error outcome. Unless the asserts are stripped out in the final build this causes bad API user experience IMO

src/aws-cpp-sdk-core/include/smithy/client/AwsSmithyClient.h
```diff
@@ -70,28 +67,32 @@ namespace client
...
                 }
                 const auto& serviceParams = ctx.m_pRequest->GetServiceSpecificParameters();
                 if (serviceParams) {
                     for (const auto& serviceParam : serviceParams->parameterMap) {
-                        identityParams.insert({serviceParam.first, serviceParam.second});
+                        identityParams.additionalProperties.insert({serviceParam.first, serviceParam.second});
                     }
                 }
             }
             Aws::Vector<AuthSchemeOption> authSchemeOptions = m_authSchemeResolver->resolveAuthScheme(identityParams);
```
sbera87, line 93: I know the Smithy architecture says list, of auth scheme options but maybe we could have used a map of string to auth options to avoid linear lookup.

src/aws-cpp-sdk-core/include/smithy/client/AwsSmithyClient.h
```diff
@@ -31,22 +31,19 @@ namespace client
...
     public:
-        AwsSmithyClientT(Aws::Client::ClientConfiguration& clientConfig, const Aws::String& serviceName,
+        AwsSmithyClientT(const ServiceClientConfigurationT& clientConfig, const Aws::String& serviceName,
             const std::shared_ptr<Aws::Http::HttpClient>& httpClient,
             const std::shared_ptr<Aws::Client::AWSErrorMarshaller>& errorMarshaller,
-            const ServiceClientConfigurationT& m_client_config,
             const std::shared_ptr<EndpointProviderT> endpointProvider,
-            const std::shared_ptr<ServiceAuthSchemeResolverT>& m_auth_scheme_resolver,
-            const Aws::UnorderedMap<Aws::String, AuthSchemesVariantT>& m_auth_schemes)
+            const std::shared_ptr<ServiceAuthSchemeResolverT>& authSchemeResolver,
```
sbera87, line 38: why not const shared_ptr<const T> p . The interface should clearly indicate the intention. Non const means I can change the contests inside the class

src/aws-cpp-sdk-core/include/smithy/client/AwsSmithyClient.h
```diff
@@ -70,28 +67,32 @@ namespace client
...
-                        identityParams.insert({epParam.GetName(), epParam.GetBoolValueNoCheck()});
+                        identityParams.additionalProperties.insert({epParam.GetName(), epParam.GetBoolValueNoCheck()});
                     else
                         assert(!"Unknown endpoint parameter!");
                 }
                 const auto& serviceParams = ctx.m_pRequest->GetServiceSpecificParameters();
                 if (serviceParams) {
                     for (const auto& serviceParam : serviceParams->parameterMap) {
-                        identityParams.insert({serviceParam.first, serviceParam.second});
+                        identityParams.additionalProperties.insert({serviceParam.first, serviceParam.second});
```
sbera87, line 89: Please use emplace which uses move

sbera87 (commented):
The idea is ok but there are several coding practices issues I have pointed out which will make code less prone to bugs

src/aws-cpp-sdk-core/include/smithy/client/AwsSmithyClientBase.h
```diff
@@ -77,7 +77,7 @@ namespace client
         using SelectAuthSchemeOptionOutcome = Aws::Utils::Outcome<AuthSchemeOption, AWSError>;
         using ResolveEndpointOutcome = Aws::Utils::Outcome<Aws::Endpoint::AWSEndpoint, AWSError>;
 
-        AwsSmithyClientBase(Aws::Client::ClientConfiguration& clientConfig,
+        AwsSmithyClientBase(const Aws::Client::ClientConfiguration& clientConfig,
                             Aws::String serviceName,
```
sbera87, line 81: pass by const ref  for strings is a good practice. This avoids double copy

src/aws-cpp-sdk-core/include/smithy/client/common/AwsSmithyRequestSigning.h
```diff
@@ -71,10 +71,10 @@ namespace smithy
                 assert(!"Auth scheme has not been found for a given auth option!");
                 return false;
             }
-            AuthSchemesVariantT authScheme = *authSchemeIt;
+            AuthSchemesVariantT authScheme = authSchemeIt->second;
```
sbera87, line 74: Consider using auto here to make code more future proof

PR #3073, opened 2024-08-07
Sigv4a support

sbera87:
*Issue #, if available:*

*Description of changes:*

*Check all that applies:*

- [ ] Linux
- [ ] Windows
- [ ] Android
- [ ] MacOS
- [ ] IOS
- [ ] Other Platforms


SergeyRyabinin (approved):

src/aws-cpp-sdk-core/include/smithy/client/common/AwsSmithyRequestSigning.h
```diff
@@ -51,7 +52,8 @@ namespace smithy
                                      false/*retryable*/));
             }
 
-            AuthSchemesVariantT authScheme = *authSchemeIt;
+            const AuthSchemesVariantT& authScheme = authSchemeIt->second;
+            std::cout<<"FOUND variant"<<std::endl;
```
sbiscigl, line 56: remove std::cout call

src/aws-cpp-sdk-core/include/smithy/identity/auth/built-in/SigV4aAuthScheme.h
```diff
@@ -0,0 +1,53 @@
...
+#include <smithy/identity/signer/built-in/SigV4Signer.h>
+
+
+namespace smithy {
+    constexpr char SIGV4A[] = "aws.auth#sigv4a";
+    
+
+    class SigV4aAuthScheme : public AuthScheme<AwsCredentialIdentityBase>
+    {
+        //constexpr char AUTH_SCHEME_NAME[] = "SigV4aAuthScheme";
```
sbiscigl, line 22: remove commented code

src/aws-cpp-sdk-core/include/smithy/identity/auth/built-in/SigV4aAuthSchemeResolver.h
```diff
@@ -0,0 +1,25 @@
...
+
+namespace smithy {
+    template<typename ServiceAuthSchemeParametersT = DefaultAuthSchemeResolverParameters>
+    class SigV4aAuthSchemeResolver : public AuthSchemeResolverBase<ServiceAuthSchemeParametersT>
+    {
+    public:
+        using ServiceAuthSchemeParameters = ServiceAuthSchemeParametersT;
+        virtual ~SigV4aAuthSchemeResolver() = default;
+
+        Aws::Vector<AuthSchemeOption> resolveAuthScheme(const ServiceAuthSchemeParameters& identityProperties) override
```
sbiscigl, line 19: this is something i left on another review, maybe not addressed in this review -- should we be using `DefaultAuthSchemeResolverParameters` by default or another object for passing these params? seems like we dont actually initialize everything that is uses

src/aws-cpp-sdk-core/include/smithy/identity/signer/built-in/SigV4Signer.h
```diff
@@ -6,7 +6,7 @@
 
 #include <smithy/identity/signer/AwsSignerBase.h>
 #include <smithy/identity/identity/AwsCredentialIdentityBase.h>
-#include <smithy/identity/auth/built-in/SigV4AuthScheme.h>
+//#include <smithy/identity/auth/built-in/SigV4AuthScheme.h>
```
sbiscigl, line 9: remove commented import

src/aws-cpp-sdk-core/include/smithy/identity/signer/built-in/SigV4Signer.h
```diff
@@ -18,24 +18,39 @@ namespace smithy {
...
         }
 
-        SigningFutureOutcome sign(std::shared_ptr<HttpRequest> httpRequest, const AwsCredentialIdentityBase& identity, SigningProperties properties, const AdditionalParameters& additionalParameters) override
+        SigningFutureOutcome sign(std::shared_ptr<HttpRequest> httpRequest, const AwsCredentialIdentityBase& identity, SigningProperties properties) override
         {
-            AWS_UNREFERENCED_PARAM(additionalParameters);
+            const char ALLOCATION_TAG[] = "AwsSigV4Signer";
 
-            Aws::Auth::AWSCredentials legacyCreds(identity.accessKeyId(), identity.secretAccessKey(), *identity.sessionToken(), *identity.expiration());
+            std::shared_ptr<Aws::Auth::AWSCredentials> legacyCreds_sp;
```
sbiscigl, line 33: why are we creating a shared pointer instead of using a stack variable? the only place you use it, you de-reference it to a raw value

tests/aws-cpp-sdk-core-tests/smithy/client/SmithyClientTest.cpp
```diff
@@ -8,9 +8,17 @@
 #include <smithy/identity/auth/built-in/SigV4AuthScheme.h>
 
 
+#include <smithy/identity/auth/built-in/SigV4aAuthSchemeResolver.h>
+#include <smithy/identity/auth/built-in/SigV4aAuthScheme.h>
 #include <aws/core/client/ClientConfiguration.h>
 
 #include "aws/core/endpoint/EndpointProviderBase.h"
```
sbiscigl, line 15: looks like this was missed somewhere before, please clean up using `<>`

tests/aws-cpp-sdk-core-tests/smithy/client/SmithyClientTest.cpp
```diff
@@ -19,16 +27,162 @@ class SmithyClientTest : public Aws::Testing::AwsCppSdkGTestSuite {
...
+};
+
+
+
+TEST_F(SmithyClientTest, testSigV4) {
+
+    const char ALLOCATION_TAG[] = "SmithyClientTest";
+
+    Aws::Client::ClientConfiguration clientConfig;
+    std::shared_ptr<Aws::Http::HttpClient> httpClient = Aws::Http::CreateHttpClient(clientConfig);
```
sbiscigl, line 82: should we be creating a real http client or using the [mock http client](MockHttpClient)? we likely dont want to be creating a real http client for unit tests.

tests/aws-cpp-sdk-core-tests/smithy/client/SmithyClientTest.cpp
```diff
@@ -19,16 +27,162 @@ class SmithyClientTest : public Aws::Testing::AwsCppSdkGTestSuite {
...
+    ctx.m_pRequest = nullptr;
+
+    auto res = ptr->SelectAuthSchemeOption(ctx);
+
+    EXPECT_EQ(res.IsSuccess(), true);
+    std::cout<<"selected scheme id="<<res.GetResult().schemeId<<std::endl;
+    EXPECT_EQ(res.GetResult().schemeId, key);
+
+    //for s3 client as an example
+    Aws::String uri{"https://objecttest-cb93079d-24a0-4862-8ef2-88dae12cd549.s3.amazonaws.com"};
```
sbiscigl, line 123: lets use a test string instead of one generated by integration tests

tests/aws-cpp-sdk-core-tests/smithy/client/SmithyClientTest.cpp
sbera87: This is created for integration test to see where it actually fails in the workflow. Some parts are yet to be validated. I agree for unit tests we should use mock

tests/aws-cpp-sdk-core-tests/smithy/client/SmithyClientTest.cpp
sbera87: I can, but my ultimate goal is to test the entire auth workflow where i figured it needs a real valid endpoint for the integration test

tests/aws-cpp-sdk-core-tests/smithy/client/SmithyClientTest.cpp
sbera87: ack

src/aws-cpp-sdk-core/include/smithy/identity/signer/built-in/SigV4Signer.h
sbera87: ack

src/aws-cpp-sdk-core/include/smithy/identity/auth/built-in/SigV4aAuthSchemeResolver.h
sbera87: yes, at the least. we should add explicit constructor to fill mandatory fields
I also think that instead of letting authschemes pass different parameter types, make it pass a base polymorphic type with generic interface to fetch fields from a map of field to variant
This will unify the interface and much easier to even autogen when we have common interface signature

src/aws-cpp-sdk-core/include/smithy/client/common/AwsSmithyRequestSigning.h
sbera87: ack. oversight on my part

tests/aws-cpp-sdk-core-tests/smithy/client/SmithyClientTest.cpp
sbiscigl: yes but _this is a real bucket_ if you want to integration test a existing bucket, you have to create/destroy it as part of the test

src/aws-cpp-sdk-core/include/smithy/identity/signer/built-in/SigV4Signer.h
sbiscigl: no why does this need to be a pointer? it looks like a stack variable could just as easily be used, pointer semantics arent required here

tests/aws-cpp-sdk-core-tests/smithy/client/SmithyClientTest.cpp
sbera87: I would like this to be an integration test to create an actual signature as i want to test the auth using the aws credentials api calls

src/aws-cpp-sdk-core/include/smithy/identity/signer/built-in/SigV4Signer.h
sbera87: this will work. Ack

tests/aws-cpp-sdk-core-tests/smithy/client/SmithyClientTest.cpp
sbera87: ack

src/aws-cpp-sdk-core/include/smithy/identity/signer/built-in/SigV4Signer.h
sbiscigl: cpp 11 doesnt promise copy elision but for a string most modern compilers would do it, i would say prefer stack allocation over heap allocation whenever possible

tests/aws-cpp-sdk-core-tests/smithy/client/SmithyClientTest.cpp
sbiscigl: an HTTP client is not required for a signature, only to send a request, testing signing in this context would be a unit tests, when you actually call a aws service it would be a integration test

tests/aws-cpp-sdk-core-tests/smithy/client/SmithyClientTest.cpp
sbera87: This is just a mock string now, i used it to preserve the uri format

src/aws-cpp-sdk-core/include/smithy/identity/resolver/built-in/DefaultAwsCredentialIdentityResolver.h
```diff
@@ -9,33 +9,51 @@
 #include <aws/core/auth/AWSCredentials.h>
 #include <aws/core/auth/AWSCredentialsProviderChain.h>
 
+namespace{
+    constexpr char ALLOC_ID[] = "DefaultAwsCredentialIdentityResolver";
```
sbiscigl, line 13: why is this in the anonymous namespace and not in the smithy namespace?

src/aws-cpp-sdk-core/include/smithy/identity/signer/built-in/SigV4aSigner.h
```diff
@@ -0,0 +1,202 @@
...
+    class AwsSigV4aSigner : public AwsSignerBase<AwsCredentialIdentityBase> {
+        
+    public:
+        using SigV4aAuthSchemeParameters = smithy::DefaultAuthSchemeResolverParameters;
+        explicit AwsSigV4aSigner(const SigV4aAuthSchemeParameters& parameters)
+            :  m_parameters{parameters}
+        {
+        }
+
+        SigningFutureOutcome sign(std::shared_ptr<HttpRequest> httpRequest, const AwsCredentialIdentityBase& identity, SigningProperties properties) override
```
sbera87, line 42: please note that this uses legacy implementation just refactored and cleaned up a bit

src/aws-cpp-sdk-core/include/smithy/identity/signer/built-in/SigV4aSigner.h
```diff
@@ -0,0 +1,202 @@
...
+        {
+            
+            auto signPayloadIt = properties.find("SignPayload");
+            bool signPayload = signPayloadIt != properties.end() ? signPayloadIt->second.get<Aws::String>() == "true" : false;
+
+            assert(httpRequest);
+            assert(identity.sessionToken().has_value());
+            assert(identity.expiration().has_value());
+
+            auto &request = *httpRequest;
```
sbiscigl, line 52: why are we de-referencing instead of just using the shared ptr passed in? if its passed in we should pass it as a shared ptr to avoid ownership issues

src/aws-cpp-sdk-core/include/smithy/identity/signer/built-in/SigV4aSigner.h
```diff
@@ -0,0 +1,202 @@
...
+        }
+
+        SigningFutureOutcome sign(std::shared_ptr<HttpRequest> httpRequest, const AwsCredentialIdentityBase& identity, SigningProperties properties) override
+        {
+            
+            auto signPayloadIt = properties.find("SignPayload");
+            bool signPayload = signPayloadIt != properties.end() ? signPayloadIt->second.get<Aws::String>() == "true" : false;
+
+            assert(httpRequest);
+            assert(identity.sessionToken().has_value());
```
sbiscigl, line 49: expiration and session token are not required

src/aws-cpp-sdk-core/include/smithy/identity/resolver/built-in/DefaultAwsCredentialIdentityResolver.h
```diff
@@ -9,33 +9,51 @@
...
     /**
      * A smithy SigV4 AWS Credentials resolver wrapper on top of legacy SDK Credentials provider
      * TODO: refactor into own signer using smithy design
      */
     class DefaultAwsCredentialIdentityResolver : public AwsCredentialIdentityResolver {
+    protected:
+
+        mutable std::shared_ptr<Aws::Auth::AWSCredentialsProviderChain> legacyChain_sp;
+
+        //mutable Aws::Auth::DefaultAWSCredentialsProviderChain legacyCredentialsProvider{};
```
sbiscigl, line 27: remove commented code, also why did you move this member to the top of the class defintion?

src/aws-cpp-sdk-core/include/smithy/identity/signer/built-in/SigV4aSigner.h
```diff
@@ -0,0 +1,202 @@
...
+ * Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.
+ * SPDX-License-Identifier: Apache-2.0.
+ */
+
+#pragma once
+
+#include <smithy/identity/signer/AwsSignerBase.h>
+#include <smithy/identity/identity/AwsCredentialIdentityBase.h>
+
+//#include <aws/core/auth/signer/AWSAuthV4Signer.h>
```
sbiscigl, line 11: remove commented include

src/aws-cpp-sdk-core/include/smithy/identity/signer/built-in/SigV4aSigner.h
```diff
@@ -0,0 +1,202 @@
...
+
+                return SigningError(Aws::Client::CoreErrors::MEMORY_ALLOCATION, "", "Failed to get Auth configureation", false);
+            }
+
+            std::shared_ptr<Aws::Crt::Http::HttpRequest> crtHttpRequest = request.ToCrtHttpRequest();
+
+            auto sigv4HttpRequestSigner = Aws::MakeShared<Aws::Crt::Auth::Sigv4HttpRequestSigner>(v4AsymmetricLogTag);
+            
+            sigv4HttpRequestSigner->SignRequest(crtHttpRequest, awsSigningConfig,
+                [&request, &success, this](const std::shared_ptr<Aws::Crt::Http::HttpRequest>& signedCrtHttpRequest, int errorCode) {
```
sbiscigl, line 76: If this is the way it was before, we had a bug, whoopsies.

we're going to need synchronization here if you plan on returning based on checking "success" per the [documentation of CRT](https://github.com/awslabs/aws-crt-cpp/blob/52dea442dd7a495c3c691268d99eb5f5765c7cee/include/aws/crt/auth/Signing.h#L78C1-L86C42)

```
/**
 * Signs an http request based on the signing implementation and supplied configuration
 * @param request http request to sign
 * @param config base signing configuration.  Actual type should match the configuration expected
 * by the signer implementation
 * @param completionCallback completion function to invoke when signing has completed or failed
 * @return true if the signing process was kicked off, false if there was a synchronous failure.
 */
virtual bool SignRequest(
```

so we return true if the signing process was kicked off, not that _it finished_

we need to add checks for:
* that we only attempt to read success after we know the callback has run.
* if Sign encountered a synchronous failure.

src/aws-cpp-sdk-core/include/smithy/identity/signer/built-in/SigV4aSigner.h
```diff
@@ -0,0 +1,202 @@
...
+            
+            return success? SigningFutureOutcome(std::move(httpRequest)) : SigningError(Aws::Client::CoreErrors::MEMORY_ALLOCATION, "", "Failed to sign the request with sigv4", false);
+        }
+
+
+
+        virtual ~AwsSigV4aSigner() {};
+    protected:
+
+        bool createAwsSigningConfig(
```
sbiscigl, line 116: why are we returning a bool, seems like we should returning a `Outcome<Aws::Crt::Auth::AwsSigningConfig, SomeError>`. if this was the way it was before might as well clean it up now.

tests/aws-cpp-sdk-core-tests/smithy/client/SmithyClientTest.cpp
sbiscigl: still valid in this revision, lets make this a fake bucket name in the URI

tests/aws-cpp-sdk-core-tests/smithy/client/SmithyClientTest.cpp
```diff
@@ -8,27 +8,232 @@
...
+    
+    std::cout<<"Final auth="<<res2.GetResult()->GetAwsAuthorization()<<std::endl;
+
+    EXPECT_EQ(res2.GetResult()->GetSigningAccessKey(), "dummyAccessId");
+
+
+}
+
+
+TEST_F(SmithyClientTest, testSigV4a) {
```
sbiscigl, line 186: this test is beautiful

src/aws-cpp-sdk-core/include/smithy/identity/resolver/built-in/DefaultAwsCredentialIdentityResolver.h
sbera87: this is in a .h file, and I think there are other strings with same name, this tag is used only internal to that resolver

src/aws-cpp-sdk-core/include/smithy/identity/signer/built-in/SigV4aSigner.h
sbera87: They are required for sigv4a. expiration is automatically assigned in constructor

src/aws-cpp-sdk-core/include/smithy/identity/signer/built-in/SigV4aSigner.h
sbera87: This is used in a helper function which doesn't care about ownership

src/aws-cpp-sdk-core/include/smithy/identity/signer/built-in/SigV4aSigner.h
sbera87: ack

src/aws-cpp-sdk-core/include/smithy/identity/signer/built-in/SigV4aSigner.h
sbera87: This is an internal helper, the output is the reference parameter. The real outcome is one step above

src/aws-cpp-sdk-core/include/smithy/identity/signer/built-in/SigV4aSigner.h
sbera87: checking

tests/aws-cpp-sdk-core-tests/smithy/client/SmithyClientTest.cpp
sbera87: ack, let me stub out keeping the format

src/aws-cpp-sdk-core/include/smithy/identity/signer/built-in/SigV4aSigner.h
sbera87: I am not sure why "so we return true if the signing process was kicked off, not that it finished" is true because the function is a sync one and I see that we do return the status on completion. Happy to discuss if I am wrong

src/aws-cpp-sdk-core/include/smithy/identity/resolver/built-in/DefaultAwsCredentialIdentityResolver.h
sbiscigl: > I think there are other strings with same name
 
 then lets name it something else instead of creating a anon namespace

src/aws-cpp-sdk-core/include/smithy/identity/signer/built-in/SigV4aSigner.h
sbera87: sigv4HttpRequestSigner->SignRequest takes a callback which gets invoked at the end of the sync workflow and updates success variable , after which we return

src/aws-cpp-sdk-core/include/smithy/identity/signer/built-in/SigV4aSigner.h
sbiscigl: i know the outcome is in a reference parameter, im saying instead of doing that we should create it here and return it via RVO/NVRO in a outcome

src/aws-cpp-sdk-core/include/smithy/identity/signer/built-in/SigV4aSigner.h
sbiscigl: if its a shared pointer, we should pass the shared pointer to the helper function, we should not be ignoring that it is not, especially since it is a private helper function

sbiscigl (approved):

PR #3088, opened 2024-08-26
Add a configuration to avoid hot-loop on input polling in streaming

SergeyRyabinin:
*Issue #, if available:*

*Description of changes:*

*Check all that applies:*

- [ ] Linux
- [ ] Windows
- [ ] Android
- [ ] MacOS
- [ ] IOS
- [ ] Other Platforms

### DIFF
--- a/src/aws-cpp-sdk-core/source/http/curl/CurlHttpClient.cpp
+++ b/src/aws-cpp-sdk-core/source/http/curl/CurlHttpClient.cpp
@@ -423,9 +423,9 @@ static size_t SeekBody(void* userdata, curl_off_t offset, int origin)
     return CURL_SEEKFUNC_OK;
 }
 #if LIBCURL_VERSION_NUM >= 0x072000 // 7.32.0
-static int CurlProgressCallback(void *userdata, curl_off_t, curl_off_t, curl_off_t, curl_off_t)
+int CurlHttpClient::CurlProgressCallback(void *userdata, curl_off_t, curl_off_t, curl_off_t, curl_off_t)
 #else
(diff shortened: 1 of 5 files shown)


src/aws-cpp-sdk-core/source/http/curl/CurlHttpClient.cpp
```diff
@@ -436,16 +436,33 @@ static int CurlProgressCallback(void *userdata, double, double, double, double)
...
+        // This is a spin pause-unpause in case of no data provided by a customer callback
+        // But otherwise curl will slow down the transfer and start calling as at frequency of 1s
+        //   see https://curl.se/mail/lib-2020-07/0046.html
+        // we should use multi handle or another HTTP client in the future to avoid this
+        curl_easy_pause(context->m_curlHandle, CURLPAUSE_CONT);
+    }
+    else
+    {
+        char output[1];
+        if (ioStream->readsome(output, 1) > 0)
```
sbiscigl, line 456: could we use [peek](https://en.cppreference.com/w/cpp/io/basic_istream/peek)? that way we dont have to do any unwinding/rewinding?

sbiscigl (approved):

src/aws-cpp-sdk-core/source/http/curl/CurlHttpClient.cpp
SergeyRyabinin: ....because of my hacks to make custom IOStream work as an async non-closable stream
peak will always return something ('a' or 'z'), while readsome will return only if there is something to read.

PR #3091, opened 2024-08-27
Swap dynamo client to use a smithy base class

sbiscigl:
*Description of changes:*

Per the work to move to a more common architecture between different AWS SDKs. This moves the Dynamo DBClient to use the smithy identity based flow instead of the legacy workflow,

Examples other SDKs updating their flow:
* [Java V2 change](https://github.com/aws/aws-sdk-java-v2/pull/3764)
* [JS V3 example](https://github.com/aws/aws-sdk-js-v3/pull/6299)

*Check all that applies:*

- [x] Linux
- [x] Windows
- [x] Android
- [x] MacOS
- [ ] IOS
- [ ] Other Platforms

### DIFF
--- a/generated/src/aws-cpp-sdk-dynamodb/include/aws/dynamodb/DynamoDBClient.h
+++ b/generated/src/aws-cpp-sdk-dynamodb/include/aws/dynamodb/DynamoDBClient.h
@@ -6,15 +6,18 @@
 #pragma once
 #include <aws/dynamodb/DynamoDB_EXPORTS.h>
 #include <aws/core/client/ClientConfiguration.h>
-#include <aws/core/client/AWSClient.h>
 #include <aws/core/client/AWSClientAsyncCRTP.h>
-#include <aws/core/utils/json/JsonSerializer.h>
 #include <aws/dynamodb/DynamoDBServiceClientModel.h>
+#include <smithy/client/AwsSmithyClient.h>
+#include <smithy/identity/auth/built-in/SigV4AuthSchemeResolver.h>
+#include <smithy/identity/auth/built-in/SigV4AuthScheme.h>
+#include <smithy/client/serializer/JsonOutcomeSerializer.h>
 
 namespace Aws
 {
 namespace DynamoDB
 {
+  AWS_DYNAMODB_API extern const char SERVICE_NAME[];
   /**
    * <fullname>Amazon DynamoDB</fullname> <p>Amazon DynamoDB is a fully managed NoSQL
    * database service that provides fast and predictable performance with seamless
@@ -33,10 +36,16 @@ namespace DynamoDB
    * Zones in an Amazon Web Services Region, providing built-in high availability and
    * data durability.</p>
    */
-  class AWS_DYNAMODB_API DynamoDBClient : public Aws::Client::AWSJsonClient, public Aws::Client::ClientWithAsyncTemplateMethods<DynamoDBClient>
+  class AWS_DYNAMODB_API DynamoDBClient : smithy::client::AwsSmithyClientT<Aws::DynamoDB::SERVICE_NAME,
+      Aws::DynamoDB::DynamoDBClientConfiguration,
+      smithy::SigV4AuthSchemeResolver<>,
+      Aws::Crt::Variant<smithy::SigV4AuthScheme>,
+      DynamoDBEndpointProviderBase,
+      smithy::client::JsonOutcomeSerializer,
+      smithy::client::JsonOutcome>,
+    Aws::Client::ClientWithAsyncTemplateMethods<DynamoDBClient>
   {
     public:
-      typedef Aws::Client::AWSJsonClient BASECLASS;
       static const char* GetServiceName();
       static const char* GetAllocationTag();
 
@@ -2275,9 +2284,12 @@ namespace DynamoDB
       friend class Aws::Client::ClientWithAsyncTemplateMethods<DynamoDBClient>;
       void init(const DynamoDBClientConfiguration& clientConfiguration);
 
+      void OptionallyUpdateDescribeEndpointsCache(Aws::Endpoint::AWSEndpoint& resolvedEndpoint,
+        const Aws::String& operationName,
+        const Aws::String& endpointKey,
+        const Aws::DynamoDB::Model::DescribeEndpointsRequest& endpointRequest,
+        bool enforceDiscovery) const;
       mutable Aws::Utils::ConcurrentCache<Aws::String, Aws::String> m_endpointsCache;
-      DynamoDBClientConfiguration m_clientConfiguration;
-      std::shared_ptr<DynamoDBEndpointProviderBase> m_endpointProvider;
   };
 
 } // namespace DynamoDB
--- a/generated/src/aws-cpp-sdk-dynamodb/source/DynamoDBClient.cpp
+++ b/generated/src/aws-cpp-sdk-dynamodb/source/DynamoDBClient.cpp
@@ -4,19 +4,17 @@
  */
 
 #include <aws/core/utils/Outcome.h>
-#include <aws/core/auth/AWSAuthSigner.h>
 #include <aws/core/client/CoreErrors.h>
 #include <aws/core/client/RetryStrategy.h>
 #include <aws/core/http/HttpClient.h>
-#include <aws/core/http/HttpResponse.h>
 #include <aws/core/http/HttpClientFactory.h>
 #include <aws/core/auth/AWSCredentialsProviderChain.h>
-#include <aws/core/utils/json/JsonSerializer.h>
 #include <aws/core/utils/memory/stl/AWSStringStream.h>
 #include <aws/core/utils/threading/Executor.h>
 #include <aws/core/utils/DNS.h>
 #include <aws/core/utils/logging/LogMacros.h>
 #include <aws/core/utils/logging/ErrorMacros.h>
+
 #include <aws/core/platform/Environment.h>
 
 #include <aws/dynamodb/DynamoDBClient.h>
@@ -82,6 +80,9 @@
 
 #include <smithy/tracing/TracingUtils.h>
 
+#include <smithy/identity/resolver/built-in/SimpleAwsCredentialIdentityResolver.h>
+#include <smithy/identity/resolver/built-in/DefaultAwsCredentialIdentityResolver.h>
+#include <smithy/identity/resolver/built-in/AwsCredentialsProviderIdentityResolver.h>
 
 using namespace Aws;
 using namespace Aws::Auth;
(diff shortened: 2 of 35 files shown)


generated/src/aws-cpp-sdk-dynamodb/include/aws/dynamodb/DynamoDBClient.h
```diff
@@ -33,10 +36,16 @@ namespace DynamoDB
    * Zones in an Amazon Web Services Region, providing built-in high availability and
    * data durability.</p>
    */
-  class AWS_DYNAMODB_API DynamoDBClient : public Aws::Client::AWSJsonClient, public Aws::Client::ClientWithAsyncTemplateMethods<DynamoDBClient>
+  class AWS_DYNAMODB_API DynamoDBClient : smithy::client::AwsSmithyClientT<Aws::DynamoDB::SERVICE_NAME,
```
SergeyRyabinin, line 39: might worth adding common wrappers such as

```
    template<const char* ServiceNameT,
             typename ServiceClientConfigurationT,
             typename EndpointProviderT>
    class AwsJsonSigV4AuthSmithyClient : public AwsSmithyClientT<ServiceNameT, ServiceClientConfigurationT, smithy::SigV4AuthSchemeResolver<>, Aws::Crt::Variant<smithy::SigV4AuthScheme>, EndpointProviderT, smithy::client::JsonOutcomeSerializer, smithy::client::JsonOutcome>
```

to reduce the amount of generated boilerplate template arguments for the majority of services.

Or keep it as is.

generated/src/aws-cpp-sdk-dynamodb/include/aws/dynamodb/DynamoDBClient.h
```diff
@@ -2257,7 +2266,6 @@ namespace DynamoDB
       mutable Aws::Utils::ConcurrentCache<Aws::String, Aws::String> m_endpointsCache;
       DynamoDBClientConfiguration m_clientConfiguration;
       std::shared_ptr<Aws::Utils::Threading::Executor> m_executor;
```
SergeyRyabinin, line 2268: please don't forget to use m_executor from the config.

SergeyRyabinin (approved):

generated/src/aws-cpp-sdk-dynamodb/include/aws/dynamodb/DynamoDBClient.h
sbiscigl: yep will remove as part of next iteration with codegen

generated/src/aws-cpp-sdk-dynamodb/include/aws/dynamodb/DynamoDBClient.h
sbiscigl: let me play around with it in the code generation revision, im tempted to keep it as is because i dont want `XmlClient` or `JsonClient` to exist as they had in the past but this would likely make sense from a boiler plate standpoint and since its just a alias for something else it makes sense.

tools/code-generation/generator/src/main/java/com/amazonaws/util/awsclientgenerator/generators/cpp/CppClientGenerator.java
```diff
@@ -666,6 +666,68 @@ protected Set<String> getOperationsToRemove(){
         return new HashSet<String>();
     }
 
+    protected SdkFileEntry generateClientSmithyHeaderFile(final ServiceModel serviceModel) {
+        Template template = velocityEngine.getTemplate("com/amazonaws/util/awsclientgenerator/velocity/cpp/smithy/SmithyClientHeader.vm", StandardCharsets.UTF_8.name());
+
+        VelocityContext context = createContext(serviceModel);
+        context.put("CppViewHelper", CppViewHelper.class);
+        context.put("RequestlessOperations", requestlessOperations);
+        context.put("AuthSchemeResolver", "SigV4AuthSchemeResolver");
```
sbiscigl, line 675: I think we may need to code gen AuthSchemeResolvers per client. i.e. from [smithy](https://smithy.io/2.0/spec/authentication-traits.html#auth-trait)

>11.7. auth trait
Summary
Defines the priority ordered authentication schemes supported by a service or operation. When applied to a service, it defines the default authentication schemes of every operation in the service. When applied to an operation, it defines the list of all authentication schemes supported by the operation, overriding any auth trait specified on a service.
Trait selector
:is(service, operation)
>Service or operation shapes
>Value type
This trait contains a priority ordered list of unique string values that reference authentication scheme shape IDs defined on a service shape.

from my understand the [auth list](https://github.com/aws/aws-sdk-cpp/blob/main/tools/code-generation/api-descriptions/dynamodb-2012-08-10.normal.json#L15) _is_ the scheme provider 

from the SRA book:
```
The AuthSchemeResolver accepts data from the request (e.g. request members) and client (e.g. AWS region) and returns a list of authentication scheme options that the SDK should use when authenticating that request. Each authentication scheme option includes the selected scheme ID and properties for use when generating the identity and request signature. The SDK will choose the first authentication scheme in the result list that it supports. An authentication scheme is considered “supported” if the SDK recognizes the selected scheme ID and the SDK has an identity resolver for the selected scheme ID. If the identity resolver fails after the scheme ID is chosen, the request fails and does not move on to the next authentication scheme.
```

so in my head the "AuthSchemeResolver" is this list and at run time we select the first one we support.


for now though since we are doing only one client -- i think we can hardcode it.

generated/src/aws-cpp-sdk-dynamodb/include/aws/dynamodb/DynamoDBClient.h
sbiscigl: so spent some time messing around with this and im not opposed to it however see my [other comment](https://github.com/aws/aws-sdk-cpp/pull/3091/files#r1752638598). if we have to codegen either `AuthSchemeResolver` or the variant of `AuthSchemes` then we would have to make a unique class for each combination. yes sigv4 is the most common and captures most if not all of our use cases, but if more are added we will have to continue creating more and more exotic combinations of AuthSchemeResolver` and `AuthSchemes`.

src/aws-cpp-sdk-core/include/smithy/client/AwsSmithyClientBase.h
```diff
@@ -148,6 +148,8 @@ namespace client
                                       std::shared_ptr<Aws::Http::HttpResponse> httpResponse) const;
 
         inline virtual const char* GetServiceClientName() const { return m_serviceName.c_str(); }
+        inline virtual std::shared_ptr<Aws::Http::HttpClient> GetHttpClient() { return m_httpClient; }
```
SergeyRyabinin, line 151: return const ref
returning object will increase the shared_ptr use ref and the code actually using it won't be able to determine if this http client got only a single owner

sbiscigl: I did this because this is how we currently handle it in [AwsClient](https://github.com/aws/aws-sdk-cpp/blob/main/src/aws-cpp-sdk-core/include/aws/core/client/AWSClient.h#L172) and its how you added [GetHttpClient](https://github.com/aws/aws-sdk-cpp/commit/9e8e71d3f2b263e6194a00a08c39f227b7d56970#diff-579705b19505b3ce9d73df235130af7e4455c41396dd18f00956c64cfa51beb8) when you made the PR for improving shutdown. I had to add this method when doing the rebase. Im not opposed to making this change but we will have to update [AWSClientAsyncCRTP](https://github.com/aws/aws-sdk-cpp/blob/main/src/aws-cpp-sdk-core/include/aws/core/client/AWSClientAsyncCRTP.h#L90-L124) where we call shutdown to create a new shutdown method that could be a template specialization for what `GetHttpClient` returns whether its a pointer or ref.

but curious if we should make it const ref also for AwsClient so we dont have to diverge the codepath for smithy client vs aws client shutdown.

SergeyRyabinin: I meant to add const ref as in AwsClient you linked
```
const std::shared_ptr<Aws::Http::HttpClient>&
```
vs
```
std::shared_ptr<Aws::Http::HttpClient>
```
Returning shared_ptr by value will increase it is internal reference count.
I had to perform such check in CRTP
```
            if (pClient->GetHttpClient().use_count() == 1)
            {
                pClient->DisableRequestProcessing();
            }
```
to be sure that the service client is not (accidentally) shallow-copied and the shutdown would not accidentally block some other client.
Unfortunately we did not enforce any policy on (default implicit shallow) copy-constructors, so have to rely on use_count.

src/aws-cpp-sdk-core/include/smithy/client/serializer/JsonOutcomeSerializer.h
```diff
@@ -0,0 +1,99 @@
...
+        using AWSError = Aws::Client::AWSError<CoreErrors>;
+        using JsonValue = Aws::Utils::Json::JsonValue;
+        using HttpResponseOutcome = Aws::Utils::Outcome<std::shared_ptr<Aws::Http::HttpResponse>, AWSError>;
+        using JsonOutcome = Aws::Utils::Outcome<Aws::AmazonWebServiceResult<JsonValue>, AWSError>;
+        using TelemetryProvider = components::tracing::TelemetryProvider;
+
+        class JsonOutcomeSerializer
+        {
+        public:
+            explicit JsonOutcomeSerializer(const std::shared_ptr<TelemetryProvider> telemetryProvider)
```
SergeyRyabinin, line 28: you'd prefer pass-by-value-and-move or pass-by-const-reference-and-copy

passing by const value will create an unnecessary temporary copy of telemetryProvider shared_ptr (which is relatively somewhat expensive).

src/aws-cpp-sdk-core/include/smithy/client/AwsSmithyClientBase.h
sbiscigl: ahhhhhhh, thank you, makes sense

src/aws-cpp-sdk-core/include/smithy/client/serializer/JsonOutcomeSerializer.h
sbiscigl: yeah youre right i think i missed this somewhere in doing the change, good eye

SergeyRyabinin (approved):

generated/src/aws-cpp-sdk-dynamodb/source/DynamoDBClient.cpp
```diff
@@ -3117,60 +1546,66 @@ UpdateTimeToLiveOutcome DynamoDBClient::UpdateTimeToLive(const UpdateTimeToLiveR
...
+        this->UpdateEndpointWithDescribeEndpoints(resolvedEndpoint, request.GetServiceRequestName(), endpointKey, endpointRequest, false);
+      }));
     },
     TracingUtils::SMITHY_CLIENT_DURATION_METRIC,
     *meter,
     {{TracingUtils::SMITHY_METHOD_DIMENSION, request.GetServiceRequestName()}, {TracingUtils::SMITHY_SERVICE_DIMENSION, this->GetServiceClientName()}});
 }
 
+
+void DynamoDBClient::UpdateEndpointWithDescribeEndpoints(Aws::Endpoint::AWSEndpoint& resolvedEndpoint,
```
DmitriyMusatkin, line 1570: nit: maybe something to indicate that its based on config? something like OptionallyUpdateDescribeEndpointsCache?

DmitriyMusatkin (approved):

generated/src/aws-cpp-sdk-dynamodb/source/DynamoDBClient.cpp
sbiscigl: good call

generated/src/aws-cpp-sdk-dynamodb/include/aws/dynamodb/DynamoDBClient.h
```diff
@@ -6,15 +6,18 @@
...
+#include <smithy/client/AwsSmithyClient.h>
+#include <smithy/identity/auth/built-in/SigV4AuthSchemeResolver.h>
+#include <smithy/identity/auth/built-in/SigV4AuthScheme.h>
+#include <smithy/client/serializer/JsonOutcomeSerializer.h>
 
 namespace Aws
 {
 namespace DynamoDB
 {
+  AWS_DYNAMODB_API extern const char SERVICE_NAME[];
```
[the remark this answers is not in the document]
sbiscigl: good call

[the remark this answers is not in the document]
sbiscigl: this actually leads to gcc errors about odr

> aws-sdk-cpp/generated/src/aws-cpp-sdk-dynamodb/include/aws/dynamodb/DynamoDBClient.h:39:26: error: 'Aws::DynamoDB::DynamoDBClient' has a field 'Aws::DynamoDB::DynamoDBClient::<anonymous>' whose type uses the anonymous namespace [-Werror=subobject-linkage]
--
1223 | class AWS_DYNAMODB_API DynamoDBClient : smithy::client::AwsSmithyClientT<Aws::DynamoDB::SERVICE_NAME,
1224 | ^~~~~~~~~~~~~~
1225 | aws-sdk-cpp/generated/src/aws-cpp-sdk-dynamodb/include/aws/dynamodb/DynamoDBClient.h:39:26: error: 'Aws::DynamoDB::DynamoDBClient' has a base 'smithy::client::AwsSmithyClientT<(& Aws::DynamoDB::SERVICE_NAME), Aws::DynamoDB::DynamoDBClientConfiguration, smithy::SigV4AuthSchemeResolver<>, Aws::Crt::Variant<smithy::SigV4AuthScheme>, Aws::Endpoint::EndpointProviderBase<Aws::DynamoDB::DynamoDBClientConfiguration, Aws::DynamoDB::Endpoint::DynamoDBBuiltInParameters, Aws::Endpoint::ClientContextParameters>, smithy::client::JsonOutcomeSerializer, Aws::Utils::Outcome<Aws::AmazonWebServiceResult<Aws::Utils::Json::JsonValue>, Aws::Client::AWSError<Aws::Client::CoreErrors> > >' whose type uses the anonymous namespace [-Werror=subobject-linkage]
1226 | cc1plus: all warnings being treated as errors
1227 | [618/2688] Building CXX object generated/src/a

we ran into other things like this, and just [looking around](https://stackoverflow.com/questions/37722850/how-to-silence-whose-type-uses-the-anonymous-namespace-werror-gcc-version-4), i think it boils down to "we cant have the deifnition in a header"

PR #3102, opened 2024-09-04
Smithy Bearer Auth  

sbera87:
*Issue #, if available:*

*Description of changes:*

*Check all that applies:*

- [x] Linux
- [ ] Windows
- [ ] Android
- [ ] MacOS
- [ ] IOS
- [ ] Other Platforms

### DIFF
--- a/src/aws-cpp-sdk-core/include/smithy/identity/auth/built-in/BearerTokenAuthScheme.h
+++ b/src/aws-cpp-sdk-core/include/smithy/identity/auth/built-in/BearerTokenAuthScheme.h
@@ -22,7 +22,7 @@ class BearerTokenAuthScheme : public AuthScheme<AwsBearerTokenIdentityBase>
     explicit BearerTokenAuthScheme(
         std::shared_ptr<AwsCredentialIdentityResolverT> identityResolver,
         const Aws::String &serviceName, const Aws::String &region)
-        : AuthScheme(smithy::BearerTokenAuthSchemeOption::BEARER_SCHEME),
+        : AuthScheme("smithy.api#HTTPBearerAuth"),
           m_identityResolver{identityResolver},
           m_signer{Aws::MakeShared<smithy::BearerTokenSigner>(
               "BearerTokenAuthScheme", serviceName, region)}
--- a/src/aws-cpp-sdk-core/include/smithy/identity/auth/built-in/BearerTokenAuthSchemeOption.h
+++ b/src/aws-cpp-sdk-core/include/smithy/identity/auth/built-in/BearerTokenAuthSchemeOption.h
@@ -10,11 +10,8 @@ namespace smithy
 struct BearerTokenAuthSchemeOption
 {
     static AuthSchemeOption bearerTokenAuthSchemeOption;
-    static char BEARER_SCHEME[];
 };
 
-char BearerTokenAuthSchemeOption::BEARER_SCHEME[] = "smithy.api#HTTPBearerAuth";
-
 AuthSchemeOption BearerTokenAuthSchemeOption::bearerTokenAuthSchemeOption =
-    AuthSchemeOption(BEARER_SCHEME);
+    AuthSchemeOption("smithy.api#HTTPBearerAuth");
 } // namespace smithy
--- a/tests/aws-cpp-sdk-core-tests/smithy/client/SmithyClientTest.cpp
+++ b/tests/aws-cpp-sdk-core-tests/smithy/client/SmithyClientTest.cpp
@@ -271,7 +271,9 @@ TEST_F(SmithyClientTest, bearer)
 
     Aws::UnorderedMap<Aws::String, SigVariant> authSchemesMap;
 
-    Aws::String key{smithy::BearerTokenAuthSchemeOption::BEARER_SCHEME};
+    Aws::String key{
+        smithy::BearerTokenAuthSchemeOption::bearerTokenAuthSchemeOption
+            .schemeId};
 
     std::shared_ptr<smithy::AwsBearerTokenIdentityResolver> resolver =
         Aws::MakeShared<TestAwsBearerTokenIdentityResolver>(ALLOCATION_TAG);


src/aws-cpp-sdk-core/include/smithy/identity/auth/built-in/BearerTokenAuthScheme.h
```diff
@@ -0,0 +1,61 @@
...
+
+#include <smithy/identity/auth/AuthScheme.h>
+#include <smithy/identity/auth/built-in/BearerTokenAuthSchemeOption.h>
+#include <smithy/identity/identity/AwsBearerTokenIdentityBase.h>
+#include <smithy/identity/resolver/AwsBearerTokenIdentityResolver.h>
+#include <smithy/identity/signer/built-in/BearerTokenSigner.h>
+
+namespace smithy
+{
+constexpr char BEARER_SIGNER[] = "Bearer";
```
sbiscigl, line 15: was thinking about this the other day, do we really need this in its own constant or can we just pass the string in the parent ctor i.e.
```
AuthScheme("Bearer")
```
we avoid some namespace pollution that way

src/aws-cpp-sdk-core/include/smithy/identity/auth/built-in/BearerTokenAuthSchemeOption.h
```diff
@@ -0,0 +1,18 @@
...
+
+namespace smithy
+{
+struct BearerTokenAuthSchemeOption
+{
+    static AuthSchemeOption bearerTokenAuthSchemeOption;
+};
+
+AuthSchemeOption BearerTokenAuthSchemeOption::bearerTokenAuthSchemeOption =
+    AuthSchemeOption("Bearer");
```
sbiscigl, line 17: where did you get this string btw? in the smithy docs its `smithy.api#HTTPBearerAuth`

src/aws-cpp-sdk-core/include/smithy/identity/resolver/AwsBearerTokenIdentityResolver.h
```diff
@@ -4,16 +4,92 @@
...
+#include <aws/core/auth/bearer-token-provider/AWSBearerTokenProviderBase.h>
+#include <aws/core/auth/bearer-token-provider/SSOBearerTokenProvider.h>
+#include <smithy/identity/identity/AwsBearerTokenIdentity.h>
 #include <smithy/identity/resolver/AwsIdentityResolverBase.h>
 
-#include <smithy/identity/identity/AwsBearerTokenIdentity.h>
+namespace smithy
+{
+
+static const char SSO_DEFAULT_BEARER_TOKEN_PROVIDER_CHAIN_LOG_TAG[] =
```
sbiscigl, line 15: same comment as above, maybe we should just inline these to avoid pollution

src/aws-cpp-sdk-core/include/smithy/identity/auth/built-in/BearerTokenAuthSchemeOption.h
sbera87: https://github.com/aws/aws-sdk-cpp/blob/af8d1fb3a246ea96e28c4739e992e5e7fe5dc0c1/tools/code-generation/generator/src/main/java/com/amazonaws/util/awsclientgenerator/transform/C2jModelToGeneratorModelTransformer.java#L587

this is the signer we use.

src/aws-cpp-sdk-core/include/smithy/identity/auth/built-in/BearerTokenAuthScheme.h
sbera87: I had raised this before (during autha) but we agreed putting this in namespace smithy to comply with what Sergey originally proposed. We do pass it in the base class constructor. I can simply pass the raw string .
Whatever you finalize. I am ok with either
``` explicit BearerTokenAuthScheme(
        std::shared_ptr<AwsCredentialIdentityResolverT> identityResolver,
        const Aws::String &serviceName, const Aws::String &region)
        : AuthScheme(BEARER_SIGNER), m_identityResolver{identityResolver},
          m_signer{Aws::MakeShared<smithy::BearerTokenSigner>(
              "BearerTokenAuthScheme", serviceName, region)}
```

src/aws-cpp-sdk-core/include/smithy/identity/auth/built-in/BearerTokenAuthSchemeOption.h
sbera87: Are we moving to a new signer name? Coz, that's not what legacy signer name is

src/aws-cpp-sdk-core/include/smithy/identity/auth/built-in/BearerTokenAuthScheme.h
sbiscigl: raw string is better in my opinion thing about it, one less thing in the namespace that is actually internal to us

src/aws-cpp-sdk-core/include/smithy/identity/auth/built-in/BearerTokenAuthSchemeOption.h
sbiscigl: why do you think that is the name of the signer? look at what exists for the SigV4AuthSchemeOption
https://github.com/aws/aws-sdk-cpp/blob/main/src/aws-cpp-sdk-core/include/smithy/identity/auth/built-in/SigV4AuthSchemeOption.h#L15

and SigV4aAuthSchemeOption
https://github.com/aws/aws-sdk-cpp/blob/main/src/aws-cpp-sdk-core/include/smithy/identity/auth/built-in/SigV4aAuthSchemeOption.h#L15

this is the AuthScheme name, which as in the SRA reference is named as `smithy.api#HTTPBearerAuth`. You can find the exact same in the scheme id for java for bearer auth
https://github.com/aws/aws-sdk-java-v2/blob/5253ae375004b0f51036657eda6fc539f9cd2035/core/http-auth/src/main/java/software/amazon/awssdk/http/auth/scheme/BearerAuthScheme.java#L36

src/aws-cpp-sdk-core/include/smithy/identity/resolver/AwsBearerTokenIdentityResolver.h
sbera87: I can put it in the class scope to make it better

src/aws-cpp-sdk-core/include/smithy/identity/auth/built-in/BearerTokenAuthSchemeOption.h
```diff
@@ -0,0 +1,20 @@
...
+{
+struct BearerTokenAuthSchemeOption
+{
+    static AuthSchemeOption bearerTokenAuthSchemeOption;
+    static char BEARER_SCHEME[];
+};
+
+char BearerTokenAuthSchemeOption::BEARER_SCHEME[] = "smithy.api#HTTPBearerAuth";
+
+AuthSchemeOption BearerTokenAuthSchemeOption::bearerTokenAuthSchemeOption =
```
sbiscigl, line 18: why not inline this like in [sigv4](https://github.com/aws/aws-sdk-cpp/blob/main/src/aws-cpp-sdk-core/include/smithy/identity/auth/built-in/SigV4AuthSchemeOption.h#L15) and [sigv4a](https://github.com/aws/aws-sdk-cpp/blob/main/src/aws-cpp-sdk-core/include/smithy/identity/auth/built-in/SigV4aAuthSchemeOption.h#L15), what makes this auth scheme different where a client needs this symbol and couldnt access it through `BearerTokenAuthSchemeOption::bearerTokenAuthSchemeOption::schemeId`

sbiscigl (approved):

src/aws-cpp-sdk-core/include/smithy/identity/auth/built-in/BearerTokenAuthSchemeOption.h
sbera87: I can change that but the main reason for having the variable is usage from other places, so that way we just refer to one constant instead of copies elsewhere. I think we should extend this to the other auths too if you like that

sbiscigl: we can refer to it other places using `BearerTokenAuthSchemeOption::bearerTokenAuthSchemeOption::schemeId` without needing the static const char

Issue #2594, opened 2023-07-26
Retry settings are not reflected in s3CrtClient


MaZvnGpGm9:
### Describe the bug

To change the "max_retries" of the number of retries from 5, I set retryStrategy in Aws::S3Crt::ClientConfiguration and passed it to a new instance of Aws::S3Crt::S3CrtClient. However, the retry count does not change.

### Expected Behavior

The retry strategy set in retryStrategy is enabled, and the number of retries changes.

### Current Behavior

The default retry strategy and number of retries are always applied.

### Reproduction Steps

Create an instance of Aws::S3Crt::S3CrtClient with retryStrategy and attempt an operation that requires many retries.

### AWS CPP SDK version used

1.11.61

### Compiler and Version used

gcc 8.3.1

### Operating System and version

RHEL8

jmklix:
Adding all of the same settings in the s3CrtClient was overlooked when it was made. Looking into how complicated this would be to get working

jmklix:
Changing this to a feature request, because this is not how the s3CrtClient was initially designed. Please let us know if you would like this feature added by 👍 the above post.

Manjunathagopi:
Hi,
Can someone explain how this retry strategy works, i configured timeout configs to 100ms and retry strategy to NO_RETRY. But still i am seeing requests are not getting cancelled and time to get object is more than 1 second.

sample code snippet.
   ```
 Aws::S3Crt::ClientConfiguration config;
    config.httpRequestTimeoutMs = 100;
    config.connectTimeoutMs = 100;
    config.requestTimeoutMs = 100;  
    config.crtRetryStrategyConfig.crtRetryStrategyType = 
    Aws::S3Crt::S3CrtClientConfiguration::CrtRetryStrategyConfig::CrtRetryStrategyType::NO_RETRY;
    config.crtRetryStrategyConfig.config.maxRetries = 2;
    config.crtRetryStrategyConfig.config.scaleFactorMs = 500;
    config.crtRetryStrategyConfig.config.maxBackoffSecs = 2;
    config.throughputTargetGbps = 1;
    config.partSize = 10*1024*1024;
    s3_crt_client = Aws::New<Aws::S3Crt::S3CrtClient>("test", config);
```

SergeyRyabinin:
Hello,

The support for retry strategy configuration on S3 CRT client has been added. I'm closing this issue as resolved.

[...]

PR #3110, opened 2024-09-12
S3 CRT retry strategy config

SergeyRyabinin:
*Issue #, if available:*
https://github.com/aws/aws-sdk-cpp/issues/2594
*Description of changes:*
Better late than never, please sorry all for a long wait.
Update S3 CRT client config to provide different options to specify the retry strategy.
Also introduces the pattern factory into the S3 CRT generated client config which can be extended to other parameters as well.
*Check all that applies:*

- [ ] Linux
- [ ] Windows
- [ ] Android
- [ ] MacOS
- [ ] IOS
- [ ] Other Platforms


generated/src/aws-cpp-sdk-s3-crt/include/aws/s3-crt/S3CrtClientConfiguration.h
```diff
@@ -62,6 +65,80 @@ namespace Aws
...
+
+            /** Client bootstrap used for common staples such as event loop group, host resolver, etc..
+             *  If this is nullptr, SDK will create a default one for you.
+             */
+            std::shared_ptr<Aws::Crt::Io::ClientBootstrap> clientBootstrap;
+
+            /** Size of parts the files will be downloaded or uploaded in. Useful for Put/GetObject APIs
+             *  defaults to 5MB, if user set it to be less than 5MB, SDK will set it to 5MB.
+             */
+            size_t partSize = 5 * 1024 * 1024;
```
DmitriyMusatkin, line 95: we should probably prefer 8mb. 5mb is known to have weird perf

generated/src/aws-cpp-sdk-s3-crt/include/aws/s3-crt/S3CrtClientConfiguration.h
```diff
@@ -62,6 +65,80 @@ namespace Aws
...
+
+            /** TLS Options to be used for each connection.
+             *  If scheme is Https and tlsConnectionOptions is null, SDK will create a default one for you.
+             */
+            std::shared_ptr<Aws::Crt::Io::TlsConnectionOptions> tlsConnectionOptions;
+
+            /* Throughput target in Gbps that we are trying to reach. Normally it's the NIC's throughput */
+            double throughputTargetGbps = 2.0;
```
DmitriyMusatkin, line 103: i know its an old setting, but 10 is probably better. and those low values behave a bit weird due to rounding that currently happens in crt (i.e. there is no difference between any values you set below 4)

generated/src/aws-cpp-sdk-s3-crt/include/aws/s3-crt/S3CrtClientConfiguration.h
```diff
@@ -62,6 +65,80 @@ namespace Aws
...
+
+            /** Control the maximum memory used by downloads.
+             *  When set to a value > 0, the SDK uses flow control to bring the memory usage very close
+             *  to the specified window. Without this cap, memory usage grows proportional to file size.
+             */
+            size_t downloadMemoryUsageWindow = 0;
```
DmitriyMusatkin, line 109: hmm, im not exactly sure how this is bound out, but we should also add https://github.com/awslabs/aws-c-s3/blob/main/include/aws/s3/s3_client.h#L449. this is preferable way to limit mem usage. and flow control should only be used in advanced cases.

sbiscigl (approved):

Issue #3125, opened 2024-09-25
SIGABRT in Aws::S3Crt::Model::GetObjectRequest when partSize is configured to 5GB


rohansuri:
### Describe the bug

I want to turn off S3CrtClient's default behaviour of doing multi-range GETs and multi-part PUTs. So I set
Aws::S3Crt::ClientConfiguration.partSize set to 5GB. So that only for objects greater than 5GB, will a multi-part PUT will happen (5GB is chosen because that is the size limit of a single PUT call). Otherwise, I want only a single PUT/GET to happen for objects lesser than 5GB.

However, my application is crashing. As per the stack trace, it seems the client is automatically doing a ranged GET of 5GB in size and trying to allocate a buffer of that size, which results in an assertion failing as the buffer pool limit is only 2GB, resulting in SIGABRT. 

Here's the stack trace:
> Fatal error condition occurred in /Users/rohan/cb/aws-sdk-cpp/crt/aws-crt-cpp/crt/aws-c-s3/source/s3_buffer_pool.c:258: size <= buffer_pool->mem_limit
Exiting Application
################################################################################
Stack trace:
################################################################################
1   libaws-c-common.1.0.0.dylib         0x000000010189dcec aws_backtrace_print + 200
2   libaws-c-common.1.0.0.dylib         0x0000000101860318 aws_fatal_assert + 104
3   libaws-c-s3.1.0.0.dylib             0x00000001011de210 aws_s3_buffer_pool_reserve + 204
4   libaws-c-s3.1.0.0.dylib             0x00000001011d59f8 s_s3_auto_ranged_get_update + 1360
5   libaws-c-s3.1.0.0.dylib             0x00000001011f4ba4 aws_s3_meta_request_update + 220
6   libaws-c-s3.1.0.0.dylib             0x00000001011e660c aws_s3_client_update_meta_requests_threaded + 348
7   libaws-c-s3.1.0.0.dylib             0x00000001011e9b08 s_s3_client_process_work_default + 1312
8   libaws-c-s3.1.0.0.dylib             0x00000001011ea5d0 s_s3_client_process_work_task + 324
9   libaws-c-common.1.0.0.dylib         0x00000001018a6760 aws_task_run + 308
10  libaws-c-common.1.0.0.dylib         0x00000001018a6e7c s_run_all + 512
11  libaws-c-common.1.0.0.dylib         0x00000001018a7ac8 aws_task_scheduler_run_all + 88
[...]

PR #3126, opened 2024-09-25
add configuration for MPU threshold for CRT

sbiscigl:
*Issue #, if available:*

[issues/3125](https://github.com/aws/aws-sdk-cpp/issues/3125)

*Description of changes:*

Introduces a configuration option `multipartUploadThreshold` to match the [crt's configuration option](https://github.com/awslabs/aws-c-s3/blob/main/include/aws/s3/s3_client.h#L432-L443).

*Check all that applies:*

- [x] Linux
- [x] Windows
- [ ] Android
- [x] MacOS
- [ ] IOS
- [ ] Other Platforms

### DIFF
--- a/generated/src/aws-cpp-sdk-s3-crt/source/S3CrtClient.cpp
+++ b/generated/src/aws-cpp-sdk-s3-crt/source/S3CrtClient.cpp
@@ -381,6 +381,7 @@ void S3CrtClient::init(const S3Crt::ClientConfiguration& config,
 
   static const size_t DEFAULT_PART_SIZE = 8 * 1024 * 1024; // 8MB
   s3CrtConfig.part_size = config.partSize < DEFAULT_PART_SIZE ? DEFAULT_PART_SIZE : config.partSize;
+  s3CrtConfig.multipart_upload_threshold = config.multipartUploadThreshold;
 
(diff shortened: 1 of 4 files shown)


generated/src/aws-cpp-sdk-s3-crt/source/S3CrtClient.cpp
```diff
@@ -381,6 +381,10 @@ void S3CrtClient::init(const S3Crt::ClientConfiguration& config,
 
   static const size_t DEFAULT_PART_SIZE = 8 * 1024 * 1024; // 8MB
   s3CrtConfig.part_size = config.partSize < DEFAULT_PART_SIZE ? DEFAULT_PART_SIZE : config.partSize;
+  static const uint64_t FIVE_MB = 5 * 1024 * 1024;
+  s3CrtConfig.multipart_upload_threshold = config.multipartUploadThreshold.has_value() ?
+    config.multipartUploadThreshold.value() :
+    std::max(s3CrtConfig.part_size, FIVE_MB);
```
DmitriyMusatkin, line 387: dont set default and let crt default? passing 0 should be fine

sbiscigl: I wanted to have a bit of fun with with the logic too. But yeah makes sense will update

DmitriyMusatkin (approved):

PR #3144, opened 2024-10-11
fixes S3 200 issue with correct error returned

sbiscigl:
*Description of changes:*

There is a rather peculiar behaviour in S3 where S3 can return a [200 OK with a error in the body](https://repost.aws/knowledge-center/s3-resolve-200-internalerror). We made a change to address this operation on [many APIs in july](https://github.com/aws/aws-sdk-cpp/pull/3024/files#diff-afd33f1c79a2c69ace5b2cfb4fdf9047dfdf8a2b854dec9bd8099a546bc1a769). There was a issue where while a error would be returned it wouldn't be actually translated to a error type instead being unknown

```
Aws::Client::AWSError<Aws::S3::S3Errors> = {Aws::Client::AWSError<Aws::S3::S3Errors>} 
 m_errorType = {Aws::S3::S3Errors} UNKNOWN
 m_exceptionName = {Aws::String} ""
 m_message = {Aws::String} "Error in body of the response"
 m_remoteHostIpAddress = {Aws::String} ""
 m_requestId = {Aws::String} ""
 m_responseHeaders = {Aws::Http::HeaderValueCollection} size=0
 m_responseCode = {Aws::Http::HttpResponseCode} REQUEST_NOT_MADE
 m_errorPayloadType = {Aws::Client::ErrorPayloadType} NOT_SET
 m_xmlPayload = {Aws::Utils::Xml::XmlDocument} 
 m_jsonPayload = {Aws::Utils::Json::JsonValue} 
 m_retryableType = {Aws::Client::RetryableType} NOT_RETRYABLE
```

this change correct a error where `!codeNode.HasChildren()` was used in the logic to create the error when it was not needed. This correctly produces the error object, and more importantly marks it as retryable

```
Aws::Client::AWSError<Aws::S3::S3Errors> = {Aws::Client::AWSError<Aws::S3::S3Errors>} 
 m_errorType = {Aws::S3::S3Errors} INTERNAL_FAILURE
 m_exceptionName = {Aws::String} ""
 m_message = {Aws::String} "We encountered an internal error. Please try again."
 m_remoteHostIpAddress = {Aws::String} ""
 m_requestId = {Aws::String} ""
 m_responseHeaders = {Aws::Http::HeaderValueCollection} size=0
 m_responseCode = {Aws::Http::HttpResponseCode} REQUEST_NOT_MADE
 m_errorPayloadType = {Aws::Client::ErrorPayloadType} NOT_SET
 m_xmlPayload = {Aws::Utils::Xml::XmlDocument} 
 m_jsonPayload = {Aws::Utils::Json::JsonValue} 
 m_retryableType = {Aws::Client::RetryableType} RETRYABLE
```

*Check all that applies:*

- [x] Linux
- [x] Windows
- [x] Android
- [x] MacOS
- [ ] IOS
- [ ] Other Platforms

### DIFF
--- a/generated/src/aws-cpp-sdk-s3/source/S3ErrorMarshaller.cpp
+++ b/generated/src/aws-cpp-sdk-s3/source/S3ErrorMarshaller.cpp
@@ -127,6 +127,7 @@ AWSError<Aws::Client::CoreErrors> S3ErrorMarshaller::Marshall(const Aws::Http::H
   XmlDocument doc = XmlDocument::CreateFromXmlStream(body);
   body.seekg(readPointer);
   Aws::String bodyError;
+  Aws::String requestId;
 
   if (doc.WasParseSuccessful() &&
       !doc.GetRootElement().IsNull() && 
@@ -134,20 +135,27 @@ AWSError<Aws::Client::CoreErrors> S3ErrorMarshaller::Marshall(const Aws::Http::H
   {        
       //check if the first node fetched has no children
       auto messageNode = doc.GetRootElement().FirstChild("Message") ;
-      if(!messageNode.IsNull() && !messageNode.HasChildren())
+      if(!messageNode.IsNull())
       {
           message = messageNode.GetText();
       }
       auto codeNode = doc.GetRootElement().FirstChild("Code") ;
-      if(!codeNode.IsNull() && !codeNode.HasChildren())
+      if(!codeNode.IsNull())
       {
           bodyError = codeNode.GetText();
+      }
+
+      auto requestIdNode = doc.GetRootElement().FirstChild("RequestId") ;
+      if(!requestIdNode.IsNull())
+      {
+        requestId = requestIdNode.GetText();
       }
   }
 
   auto error = FindErrorByName(bodyError.c_str());
 
   error.SetMessage(message);
+  error.SetRequestId(requestId);
 
   return error;
 
--- a/generated/src/aws-cpp-sdk-s3-crt/source/S3CrtErrorMarshaller.cpp
+++ b/generated/src/aws-cpp-sdk-s3-crt/source/S3CrtErrorMarshaller.cpp
@@ -127,6 +127,7 @@ AWSError<Aws::Client::CoreErrors> S3CrtErrorMarshaller::Marshall(const Aws::Http
   XmlDocument doc = XmlDocument::CreateFromXmlStream(body);
   body.seekg(readPointer);
   Aws::String bodyError;
+  Aws::String requestId;
 
   if (doc.WasParseSuccessful() &&
       !doc.GetRootElement().IsNull() && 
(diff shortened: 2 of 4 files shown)


generated/src/aws-cpp-sdk-s3/source/S3ErrorMarshaller.cpp
```diff
@@ -127,27 +127,35 @@ AWSError<Aws::Client::CoreErrors> S3ErrorMarshaller::Marshall(const Aws::Http::H
...
   Aws::String bodyError;
+  Aws::String requestId;
 
   if (doc.WasParseSuccessful() &&
       !doc.GetRootElement().IsNull() && 
       doc.GetRootElement().GetName() == Aws::String("Error")) 
   {        
       //check if the first node fetched has no children
       auto messageNode = doc.GetRootElement().FirstChild("Message") ;
-      if(!messageNode.IsNull() && !messageNode.HasChildren())
```
sbera87, line 137: previously we were ensuring leaf node but now we just get from the root node.
If say the text message is empty, do we need to recurse to find the first non empty message?

generated/src/aws-cpp-sdk-s3/source/S3ErrorMarshaller.cpp
```diff
@@ -127,27 +127,35 @@ AWSError<Aws::Client::CoreErrors> S3ErrorMarshaller::Marshall(const Aws::Http::H
...
+      if(!messageNode.IsNull())
       {
           message = messageNode.GetText();
       }
       auto codeNode = doc.GetRootElement().FirstChild("Code") ;
-      if(!codeNode.IsNull() && !codeNode.HasChildren())
+      if(!codeNode.IsNull())
```
sbera87, line 143: same question for the error code, do we need to recurse down the tree to find first valid (non-empty) text

generated/src/aws-cpp-sdk-s3/source/S3ErrorMarshaller.cpp
```diff
@@ -127,27 +127,35 @@ AWSError<Aws::Client::CoreErrors> S3ErrorMarshaller::Marshall(const Aws::Http::H
...
       {
           bodyError = codeNode.GetText();
       }
+
+      auto requestIdNode = doc.GetRootElement().FirstChild("RequestId") ;
```
sbera87, line 148: is this more for logging?

sbiscigl: no it needs to be returned so that the user has the request ID so that they can provide it to AWS support for inquiring about issues

sbera87 (approved):

SergeyRyabinin (approved):

generated/src/aws-cpp-sdk-s3/source/S3ErrorMarshaller.cpp
sbiscigl: > previously we were ensuring leaf node

no we weren't otherwise this would have worked. the child of a empty node `HasChildren` is true with a empty value. i.e.

```cpp
#include <aws/core/Aws.h>
#include <aws/core/utils/xml/XmlSerializer.h>

using namespace Aws;
using namespace Aws::Utils::Xml;

auto main() -> int
{
    SDKOptions options;
    InitAPI(options);
    {
        const auto doc = XmlDocument::CreateFromXmlString("<?xml version=\"1.0\" encoding=\"UTF-8\"?><Error><Code>InvalidAction</Code><Message>We encountered an internal error. Please try again.</Message><RequestId>656c76696e6727732072657175657374</RequestId><HostId>Uuag1LuByRx9e6j5Onimru9pO4ZVKnJ2Qz7/C1NPcfTWAtRPfTaOFg==</HostId></Error>");
        assert(doc.WasParseSuccessful());
        auto node = doc.GetRootElement().FirstChild("Message");
        std::cout << node.GetText() << std::endl;
        if (node.HasChildren())
        {
            const auto thechild = node.FirstChild();
            std::cout << thechild.GetText() << std::endl;
        }
    }
    Aws::ShutdownAPI(options);
}

```

generated/src/aws-cpp-sdk-s3/source/S3ErrorMarshaller.cpp
sbiscigl: same response as other question, this wasn't validating leaf, leaf has a empty child

PR #3169, opened 2024-10-31
add formatting check only on changed lines

sbera87:
*Issue #, if available:*
V1506947992

*Description of changes:*

*Check all that applies:*
- [ x] Did a review by yourself.
- [ ] Added proper tests to cover this PR. (If tests are not applicable, explain.)
- [ ] Checked if this PR is a breaking (APIs have been changed) change.
- [ ] Checked if this PR will _not_ introduce cross-platform inconsistent behavior.
- [ ] Checked if this PR would require a ReadMe/Wiki update.

Check which platforms you have built SDK on to verify the correctness of this PR.
- [ ] Linux
- [ ] Windows
- [ ] Android
- [ ] MacOS
- [ ] IOS
- [ ] Other Platforms

By submitting this pull request, I confirm that you can use, modify, copy, and redistribute this contribution, under the terms of your choice.

PR #3177, opened 2024-11-05
Add support for CRC64, refactor CRC32

sbiscigl:
*Description of changes:*

Adds a crypto utility class for CRC64 from CRT, refactors CRC32 related crypto factories to avoid DRY-ing.

*Check all that applies:*

- [x] Linux
- [x] Windows
- [x] Android
- [x] MacOS
- [ ] IOS
- [ ] Other Platforms


src/aws-cpp-sdk-core/source/utils/crypto/factory/Factories.cpp
```diff
@@ -45,6 +46,12 @@ static std::shared_ptr<HashFactory>& GetCRC32CFactory()
     return s_CRC32CFactory;
 }
 
+static std::shared_ptr<HashFactory>& GetCRC64Factory()
+{
+    static std::shared_ptr<HashFactory> s_CRC64Factory(nullptr);
```
SergeyRyabinin, line 51: how does it work with our custom memory management?

src/aws-cpp-sdk-core/include/aws/core/utils/crypto/CRC.h
```diff
@@ -0,0 +1,93 @@
...
+                    if (currentPos == std::ios::pos_type(-1))
+                    {
+                        currentPos = 0;
+                        stream.clear();
+                    }
+
+                    stream.seekg(0, Aws::IStream::beg);
+
+                    uint8_t streamBuffer[Aws::Utils::Crypto::Hash::INTERNAL_HASH_STREAM_BUFFER_SIZE];
+                    while (stream.good())
```
SergeyRyabinin, line 59: trace an error if stream is in a non-recoverable failed state after this loop

src/aws-cpp-sdk-core/source/utils/crypto/factory/Factories.cpp
sbiscigl: its gets assigned when we call its factory, it follows the same call path as CRC32 does. i.e.

[create accessor for static factory](https://github.com/aws/aws-sdk-cpp/blob/main/src/aws-cpp-sdk-core/source/utils/crypto/factory/Factories.cpp#L36-L40)
```cpp
static std::shared_ptr<HashFactory>& GetCRC32Factory()
{
    static std::shared_ptr<HashFactory> s_CRC32Factory(nullptr);
    return s_CRC32Factory;
}
```

then is assigned during the [init crypto call](https://github.com/aws/aws-sdk-cpp/blob/main/src/aws-cpp-sdk-core/source/utils/crypto/factory/Factories.cpp#L549-L552)
```cpp
if(!GetCRC32Factory())
{
  GetCRC32Factory() = Aws::MakeShared<DefaultCRC32Factory>(s_allocationTag);
}
```

SergeyRyabinin: OMG, okay.

src/aws-cpp-sdk-core/include/aws/core/utils/crypto/CRC.h
```diff
@@ -0,0 +1,99 @@
...
+namespace Aws
+{
+    namespace Utils
+    {
+        namespace Crypto
+        {
+            template <typename HashT>
+            ByteBuffer ConvertToBuffer(HashT value)
+            {
+                Aws::Utils::ByteBuffer buffer(sizeof(HashT));
```
SergeyRyabinin, line 20: please add a static assert https://en.cppreference.com/w/cpp/types/is_integral
```
static_assert(std::is_integral(HashT)::value, "Unsupported type!");
```

SergeyRyabinin (approved):

src/aws-cpp-sdk-core/source/utils/crypto/factory/Factories.cpp
sbiscigl: yeah changing this pattern and avoiding static state would be a larger change than this, but I agree something that should be done as part of some future "eliminate static state" task

tests/aws-cpp-sdk-core-tests/utils/crypto/HashTest.cpp
```diff
@@ -118,3 +119,25 @@ TEST_F(HashUpdateTest, TestCRC32CUpdate)
...
+
+TEST_F(HashUpdateTest, TestCRC64Update)
+{
+    Crypto::CRC64 hash;
+
+    unsigned char buffer[Crypto::Hash::INTERNAL_HASH_STREAM_BUFFER_SIZE];
+    memset(buffer, 0, Crypto::Hash::INTERNAL_HASH_STREAM_BUFFER_SIZE);
+
+    size_t remainingBufferSize = 26214400;
+    while (remainingBufferSize > 0)
```
sbera87, line 131: while (remainingBufferSize) less verbose

src/aws-cpp-sdk-core/source/utils/crypto/CRC64.cpp
```diff
@@ -0,0 +1,33 @@
...
+{
+    return m_hashImpl->Calculate(stream);
+}
+
+void CRC64::Update(unsigned char* buffer, size_t bufferSize)
+{
+    m_hashImpl->Update(buffer, bufferSize);
+}
+
+HashResult CRC64::GetHash()
```
sbera87, line 30: get apis typically should be marked const for correctness

src/aws-cpp-sdk-core/include/aws/core/utils/crypto/CRC.h
```diff
@@ -0,0 +1,101 @@
...
+                        currentPos = 0;
+                        stream.clear();
+                    }
+
+                    stream.seekg(0, Aws::IStream::beg);
+
+                    uint8_t streamBuffer[Aws::Utils::Crypto::Hash::INTERNAL_HASH_STREAM_BUFFER_SIZE];
+                    while (stream.good())
+                    {
+                        stream.read(reinterpret_cast<char*>(streamBuffer), Aws::Utils::Crypto::Hash::INTERNAL_HASH_STREAM_BUFFER_SIZE);
```
sbera87, line 64: better to use 
while (stream.read(reinterpret_cast<char*>(streamBuffer), Aws::Utils::Crypto::Hash::INTERNAL_HASH_STREAM_BUFFER_SIZE)) {
to handle end of stream errors more cleanly

src/aws-cpp-sdk-core/include/aws/core/utils/crypto/CRC.h
```diff
@@ -0,0 +1,101 @@
...
+                {
+                    Aws::Crt::ByteCursor byteCursor = Aws::Crt::ByteCursorFromArray(reinterpret_cast<const uint8_t*>(str.data()), str.size());
+                    m_runningChecksum = CRTChecksumFuncT(byteCursor, m_runningChecksum);
+                    return ByteBufferFuncT(m_runningChecksum);
+                };
+
+                HashResult Calculate(Aws::IStream& stream) override
+                {
+                    auto currentPos = stream.tellg();
+                    if (currentPos == std::ios::pos_type(-1))
```
sbera87, line 53: maybe use stream.eof() ?

sbera87:
to format code , please use the following cmd
`git diff -U0 --no-color main jmes -- . ':!generated/' | python3 /opt/homebrew/opt/llvm/share/clang/clang-format-diff.py -p1 -i`

src/aws-cpp-sdk-core/source/utils/crypto/CRC64.cpp
```diff
@@ -0,0 +1,33 @@
...
+    return m_hashImpl->Calculate(str);
+}
+
+HashResult CRC64::Calculate(Aws::IStream& stream)
+{
+    return m_hashImpl->Calculate(stream);
+}
+
+void CRC64::Update(unsigned char* buffer, size_t bufferSize)
+{
```
sbera87, line 26: null check? 
also constness of the ptr,  unsigned char* const buffer ?

tests/aws-cpp-sdk-core-tests/utils/crypto/HashTest.cpp
```diff
@@ -118,3 +119,25 @@ TEST_F(HashUpdateTest, TestCRC32CUpdate)
     Aws::String hexHash = HashingUtils::HexEncode(digest);
     ASSERT_STREQ("fb5b991d", hexHash.c_str());
 }
+
+TEST_F(HashUpdateTest, TestCRC64Update)
+{
+    Crypto::CRC64 hash;
+
+    unsigned char buffer[Crypto::Hash::INTERNAL_HASH_STREAM_BUFFER_SIZE];
```
sbera87, line 127: can use 
unsigned char buffer[Crypto::Hash::INTERNAL_HASH_STREAM_BUFFER_SIZE] = {};
to accomplish the same more concisely

sbiscigl:
> to format code , please use the following cmd `git diff -U0 --no-color main <your branch> -- . ':!generated/' | python3 /opt/homebrew/opt/llvm/share/clang/clang-format-diff.py -p1 -i`

you didnt commit a `.clang-format` in [pull/3169](https://github.com/aws/aws-sdk-cpp/pull/3169). once we add a clang format file we can adhere to that check, but right now since there is no format file, we are not formatting against any known format

src/aws-cpp-sdk-core/source/utils/crypto/CRC64.cpp
sbiscigl: cant because the [original API is not](https://github.com/aws/aws-sdk-cpp/blob/main/src/aws-cpp-sdk-core/include/aws/core/utils/crypto/Hash.h#L22-L51), and we cannot change it without breaking backwards compatibilty.

src/aws-cpp-sdk-core/source/utils/crypto/CRC64.cpp
sbiscigl: cant because the [original API is not](https://github.com/aws/aws-sdk-cpp/blob/main/src/aws-cpp-sdk-core/include/aws/core/utils/crypto/Hash.h#L22-L51), and we cannot change it without breaking backwards compatibilty.

sbera87:
> > to format code , please use the following cmd `git diff -U0 --no-color main <your branch> -- . ':!generated/' | python3 /opt/homebrew/opt/llvm/share/clang/clang-format-diff.py -p1 -i`
> 
> you didnt commit a `.clang-format` in [pull/3169](https://github.com/aws/aws-sdk-cpp/pull/3169). once we add a clang format file we can adhere to that check, but right now since there is no format file, we are not formatting against any known format

using the clang-format-diff specifying the style doesn't need the clang_format file.  On can provide the file for customizations but its not needed per documentation

sbiscigl:
> using the clang-format-diff specifying the style doesn't need the clang_format file. On can provide the file for customizations but its not needed per documentation

we're formatting against a unknown default format then, im against this check until we add a defined file, otherwise we dont know what style we are formatting against as none is specified.

src/aws-cpp-sdk-core/include/aws/core/utils/crypto/CRC.h
sbiscigl: this actually causes the integration tests to fail, so there is a behavior difference between `stream.good` and `if(stream.read)`

sbera87 (approved):

PR #3187, opened 2024-11-12
make smoke tests Java 8 compatible

sbera87:
*Issue #, if available:*

*Description of changes:
Previous iteration of changes are based on Java 16 which is not compatible with cdk. Hence, this change is necessary.

*Check all that applies:*

- [x] Linux
- [ ] Windows
- [ ] Android
- [x] MacOS
- [ ] IOS
- [ ] Other Platforms

### DIFF
--- a/tools/code-generation/smithy/codegen/cpp-smoke-tests-codegen/src/main/java/com/amazonaws/util/awsclientsmithygenerator/generators/CppImportContainer.java
+++ b/tools/code-generation/smithy/codegen/cpp-smoke-tests-codegen/src/main/java/com/amazonaws/util/awsclientsmithygenerator/generators/CppImportContainer.java
@@ -7,12 +7,14 @@
 
  import software.amazon.smithy.codegen.core.ImportContainer;
  import software.amazon.smithy.codegen.core.Symbol;
+
  import java.util.Collections;
  import java.util.HashSet;
  import java.util.Set;
  import java.util.Map;
+ import java.util.HashMap;
 
- public final class CppImportContainer implements ImportContainer {
+public final class CppImportContainer implements ImportContainer {
 
      private final String c2jNamespace;
      private final Set<String> coreHeaders;
@@ -41,21 +43,19 @@
             "algorithm"
             );
 
-        this.unitTestHeaders = Set.of(
-            "aws/testing/AwsCppSdkGTestSuite.h",
-            "aws/testing/AwsTestHelpers.h"
-            );
+        this.unitTestHeaders = new HashSet<>();
+         this.unitTestHeaders.add("aws/testing/AwsCppSdkGTestSuite.h");
+         this.unitTestHeaders.add("aws/testing/AwsTestHelpers.h");
 
-        //this will be added to based upon datastructures found
-        containerHeaderMap = Map.ofEntries(
-            Map.entry("Aws::String", "aws/core/utils/memory/stl/AWSString.h"),
-            Map.entry("Aws::Map", "aws/core/utils/memory/stl/AWSMap.h"),
-            Map.entry("Aws::Utils::DateTime", "aws/core/utils/DateTime.h"),
-            Map.entry("Aws::Utils::Document","aws/core/utils/Document.h"),
-            Map.entry("Aws::Utils::ByteBuffer","aws/core/utils/Array.h")
-        );
-        
-        dynamicHeaders.add(String.format("aws/%s/%sClient.h", c2jNamespace, clientNamespace));
+         //this will be added to based upon datastructures found
+        containerHeaderMap = new HashMap<>();
+         containerHeaderMap.put("Aws::String", "aws/core/utils/memory/stl/AWSString.h");
+         containerHeaderMap.put("Aws::Map", "aws/core/utils/memory/stl/AWSMap.h");
+         containerHeaderMap.put("Aws::Utils::DateTime", "aws/core/utils/DateTime.h");
+         containerHeaderMap.put("Aws::Utils::Document", "aws/core/utils/Document.h");
+         containerHeaderMap.put("Aws::Utils::ByteBuffer", "aws/core/utils/Array.h");
+
+         dynamicHeaders.add(String.format("aws/%s/%sClient.h", c2jNamespace, clientNamespace));
         
     }
 
--- a/tools/code-generation/smithy/codegen/cpp-smoke-tests-codegen/src/main/java/com/amazonaws/util/awsclientsmithygenerator/generators/CppImportContainer.java
+++ b/tools/code-generation/smithy/codegen/cpp-smoke-tests-codegen/src/main/java/com/amazonaws/util/awsclientsmithygenerator/generators/CppImportContainer.java
@@ -5,6 +5,8 @@
 
  package com.amazonaws.util.awsclientsmithygenerator.generators;
 
+ import com.google.common.collect.ImmutableMap;
+ import com.google.common.collect.ImmutableSet;
  import software.amazon.smithy.codegen.core.ImportContainer;
  import software.amazon.smithy.codegen.core.Symbol;
 
@@ -12,7 +14,6 @@
  import java.util.HashSet;
  import java.util.Set;
  import java.util.Map;
- import java.util.HashMap;
 
 public final class CppImportContainer implements ImportContainer {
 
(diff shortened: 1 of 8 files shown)


tools/code-generation/smithy/codegen/cpp-smoke-tests-codegen/src/main/java/com/amazonaws/util/awsclientsmithygenerator/generators/SmokeTestsCMakeWriter.java
```diff
@@ -19,58 +19,55 @@ public String generate()
...
+                folderNamespace + "-smoke-tests\n" +
+                "\"Tests for the AWS " + projectNameCaps + " C++ SDK\"\n" +
+                "testing-resources\n" +
+                "aws-cpp-sdk-" + folderNamespace + "\n" +
+                "aws-cpp-sdk-core\n" +
+                ")"
         );
 
-        write(""" 
-            file(GLOB AWS_$L_GENERATED_SMOKE_TEST_SRC
```
sbera87, line 33: Multi line strings not supported on Java 8. Hence needed refactoring

tools/code-generation/smithy/codegen/cpp-smoke-tests/build.gradle.kts
```diff
@@ -56,44 +58,49 @@ tasks.register("generate-smithy-build") {
...
                     // as part of the actual build task done by the smithy gradle plugin.
                     .assemble().result.get();
-            val services = model.shapes(ServiceShape::class.javaObjectType).sorted().toList();
+
+            val services = model.shapes(ServiceShape::class.java).sorted().toList();
+
             if (services.size != 1) {
-                throw Exception("There must be exactly one service in each aws model file, but found " +
-                        "${services.size} in ${file.name}: ${services.map { it.id }}");
+                println("Unexpected number of service shapes detected: ${services.size} while processing file: ${file.name}. ")
```
sbera87, line 72: deliberately not throwing exception to allow other services to move on

tools/code-generation/smithy/codegen/cpp-smoke-tests-codegen/src/main/java/com/amazonaws/util/awsclientsmithygenerator/generators/SmithyC2JNamespaceMap.java
```diff
@@ -17,9 +17,16 @@ public class SmithyC2JNamespaceMap {
 
     private SmithyC2JNamespaceMap(String jsonString){
 
-        Type mapType = new TypeToken<Map<String, String>>(){}.getType();
-        Gson gson = new Gson();
-        this.SMITHY_C2J_SERVICE_NAME_MAP = gson.fromJson(jsonString, mapType);
+        if(jsonString.isEmpty())
```
sbera87, line 20: This is a fix when no mapping is provided

tools/code-generation/smithy/codegen/cpp-smoke-tests-codegen/src/main/java/com/amazonaws/util/awsclientsmithygenerator/generators/CppImportContainer.java
```diff
@@ -41,21 +43,19 @@ public CppImportContainer(String namespace) {
             "algorithm"
             );
 
-        this.unitTestHeaders = Set.of(
-            "aws/testing/AwsCppSdkGTestSuite.h",
-            "aws/testing/AwsTestHelpers.h"
-            );
+        this.unitTestHeaders = new HashSet<>();
```
sbiscigl, line 46: Should use guavas `ImmubtableSet.of` and add the [guava dep](https://mvnrepository.com/artifact/com.google.guava/guava) if needed, would be nice but not required

tools/code-generation/smithy/codegen/cpp-smoke-tests-codegen/src/main/java/com/amazonaws/util/awsclientsmithygenerator/generators/CppImportContainer.java
```diff
@@ -41,21 +43,19 @@ public CppImportContainer(String namespace) {
...
-            Map.entry("Aws::String", "aws/core/utils/memory/stl/AWSString.h"),
-            Map.entry("Aws::Map", "aws/core/utils/memory/stl/AWSMap.h"),
-            Map.entry("Aws::Utils::DateTime", "aws/core/utils/DateTime.h"),
-            Map.entry("Aws::Utils::Document","aws/core/utils/Document.h"),
-            Map.entry("Aws::Utils::ByteBuffer","aws/core/utils/Array.h")
-        );
-        
-        dynamicHeaders.add(String.format("aws/%s/%sClient.h", c2jNamespace, clientNamespace));
+         //this will be added to based upon datastructures found
+        containerHeaderMap = new HashMap<>();
```
sbiscigl, line 51: Similar as above should use guavas 'ImmutableMap.of' and add the dep

tools/code-generation/smithy/codegen/cpp-smoke-tests-codegen/src/main/java/com/amazonaws/util/awsclientsmithygenerator/generators/CppSymbolVisitor.java
```diff
@@ -5,6 +5,7 @@
 
  package com.amazonaws.util.awsclientsmithygenerator.generators;
 
+ import lombok.val;
```
sbiscigl, line 8: Dont use Lombok val, spell out the complete type

tools/code-generation/smithy/codegen/cpp-smoke-tests-codegen/src/main/java/com/amazonaws/util/awsclientsmithygenerator/generators/SmithyC2JNamespaceMap.java
sbiscigl: when would there be no mapping file, dont we always pass it one? shouldn't this be an exception?

tools/code-generation/smithy/codegen/cpp-smoke-tests/build.gradle.kts
sbiscigl: we should catch and throw issues when we find them, and leave them up to the caller, i.e. the CI script to ignore. for instance here, why are there more than 1 service shape? what is wrong with the smithy files that allows this to be the case?

tools/code-generation/smithy/codegen/cpp-smoke-tests-codegen/src/main/java/com/amazonaws/util/awsclientsmithygenerator/generators/SmithyC2JNamespaceMap.java
sbera87: This should work with or without mapping. Mapping is just an override. Codegen script passes mapping but it doesnt stop anyone from just using gradle build without one for testing (where no service needs to be remapped) will cause the issue

tools/code-generation/smithy/codegen/cpp-smoke-tests/build.gradle.kts
sbera87: yes, but it affects all the other services that don't have issue. I can still throw it but it is going to show incomplete builds in pipeline if even one service has an issue

tools/code-generation/smithy/codegen/cpp-smoke-tests-codegen/src/main/java/com/amazonaws/util/awsclientsmithygenerator/generators/SmithyC2JNamespaceMap.java
sbiscigl: yeah thats fair

tools/code-generation/smithy/codegen/cpp-smoke-tests/build.gradle.kts
sbiscigl: the doesn't have anything to do with the pipeline, or CI at all, at pure level, if the code generator fails, it should bubble up the error to the caller, and the caller is responsible for handling it. in this case the CI pipeline.

> for instance here, why are there more than 1 service shape? what is wrong with the smithy files that allows this to be the case?

you didnt answer this question -- why do we need this and is this a data problem. what services cause this problem and why.

tools/code-generation/smithy/codegen/cpp-smoke-tests-codegen/src/main/java/com/amazonaws/util/awsclientsmithygenerator/generators/CppSymbolVisitor.java
sbera87: ok, that was IDE suggestion. will change

SergeyRyabinin (approved):

PR #3190, opened 2024-11-13
Cleanup code generation templates by removing if endpointRules checks

SergeyRyabinin:
*Issue #, if available:*
n/a
*Description of changes:*
Tested by regenerating all and seeing no diff in generated
*Check all that applies:*

- [x] Linux
- [x] Windows
- [ ] Android
- [ ] MacOS
- [ ] IOS
- [ ] Other Platforms


sbera87 (commented):
I gather from the changes that we now will consider
serviceModel.endpointRules to always be present and  welcome this set of changes for cleanup .  I have one question, can we add a check in Java codegen to indeed throw an exception if serviceModel.endpointRules  is not met so that all this removed code would never need to be triggered.

sbera87 (commented):
lgtm

PR #3191, opened 2024-11-14
increase gradle jvm heap size

sbera87:
*Issue #, if available:*

*Description of changes:*
Codegen uses Gradle . Gradle build uses a separate process called the Gradle Daemon, which is an instance of the JVM dedicated to executing Gradle builds. This daemon has its own heap settings. This is necessary to set increased heap memory so that cdk can run gradle build and generate code for all services

*Check all that applies:*

- [x] Linux
- [ ] Windows
- [ ] Android
- [x] MacOS
- [ ] IOS
- [ ] Other Platforms

### DIFF
--- a/tools/code-generation/smithy/codegen/gradle.properties
+++ b/tools/code-generation/smithy/codegen/gradle.properties
@@ -1,1 +1,1 @@
-#org.gradle.jvmargs=-Xmx2g
+org.gradle.jvmargs=-Xmx6g


tools/code-generation/smithy/codegen/gradle.properties
```diff
@@ -1 +1 @@
-#org.gradle.jvmargs=-Xmx2g
\ No newline at end of file
+org.gradle.jvmargs=-Xmx6g
```
sbera87, line 1: This file is used by gradle build for configuration

SergeyRyabinin (approved):

sbiscigl (approved):

PR #3200, opened 2024-11-19
Fix #3167: avoid infinite wait on failure to init curl handle acquire

### LINKED ISSUE #3167: AWS::Http::CurlHandleContainer will cause the service to be permanently unavailable
### Describe the bug

The DestroyCurlHandle function will fail to rebuild the curl connection, but at this time m_poolSize has not changed, which may result in no connection available in the end. For example: there are N connections in the connection pool, and there are currently 2N concurrent requests. Unfortunately, the N threads that have obtained the connection have problems when making http calls, and DestroyCurlHandle fails to rebuild the curl connection. At this time, there is no connection in the connection pool, but m_poolSize is equal to m_maxPoolSize, so the blocked N threads cannot be awakened, and the new working thread will also be blocked, causing the service to be permanently unavailable.
2. If function AWS::Utils::ExclusiveOwnershipResourceManager::ShutdownAndWait  encounters the above situation, it will also be permanently blocked because it does not meet the condition of _resources.size() == resourceCount.

### Regression Issue

- [ ] Select this option if this issue appears to be a regression.

### Expected Behavior

none

### Current Behavior

I haven't actually encountered it, but it's a potential risk.

### Reproduction Steps

none

### Possible Solution

none

none

### AWS CPP SDK version used

1.11.438

### Compiler and Version used

 4.8.5 

linux

SergeyRyabinin:
*Issue #, if available:*
https://github.com/aws/aws-sdk-cpp/issues/3167
*Description of changes:*

*Check all that applies:*

- [ ] Linux
- [ ] Windows
- [ ] Android
- [ ] MacOS
- [ ] IOS
- [ ] Other Platforms

### DIFF
--- a/src/aws-cpp-sdk-core/include/aws/core/utils/ResourceManager.h
+++ b/src/aws-cpp-sdk-core/include/aws/core/utils/ResourceManager.h
@@ -75,6 +75,34 @@ namespace Aws
             }
 
             /**
+             * Returns a resource with exclusive ownership or a nullptr.
+             * If resource is available within the wait timeout then the resource is returned.
+             * otherwise (if the timeout has expired or container is shutdown) the nullptr is returned.
+             * You must call Release on the resource when you are finished.
+             * This method is enabled only for pointer RESOURCE_TYPE type.
+             *
+             * @return instance of RESOURCE_TYPE, nullptr if the resource manager is being shutdown
+             */
+            template <typename std::enable_if<std::is_pointer<RESOURCE_TYPE>::value>::type* = nullptr>
+            RESOURCE_TYPE TryAcquire(const uint64_t timeoutMs) {
+              std::unique_lock<std::mutex> locker(m_queueLock);
+              bool hasResource = m_shutdown.load() || !m_resources.empty();
+              if (!hasResource) {
+                hasResource = m_semaphore.wait_for(locker, std::chrono::milliseconds(timeoutMs),
+                                                   [&]() { return m_shutdown.load() || !m_resources.empty(); });
+              }
+
+              if (m_shutdown || !hasResource) {
+                return nullptr;
+              }
+
+              RESOURCE_TYPE resource = m_resources.back();
+              m_resources.pop_back();
+
+              return resource;
+            }
+
+            /**
              * Returns whether or not resources are currently available for acquisition
              *
              * @return true means that at this instant some resources are available (though another thread may grab them from under you),
@@ -122,6 +150,7 @@ namespace Aws
             {
                 std::unique_lock<std::mutex> locker(m_queueLock);
                 m_shutdown = true;
+                m_semaphore.notify_all();
 
                 //wait for all acquired resources to be released.
                 while (m_resources.size() < resourceCount)
--- a/src/aws-cpp-sdk-core/source/http/curl/CurlHandleContainer.cpp
+++ b/src/aws-cpp-sdk-core/source/http/curl/CurlHandleContainer.cpp
@@ -32,6 +32,7 @@ CurlHandleContainer::~CurlHandleContainer()
         AWS_LOGSTREAM_DEBUG(CURL_HANDLE_CONTAINER_TAG, "Cleaning up " << handle);
         curl_easy_cleanup(handle);
     }
+    m_poolSize = 0;
 }
 
 CURL* CurlHandleContainer::AcquireCurlHandle()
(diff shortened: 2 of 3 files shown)


src/aws-cpp-sdk-core/include/aws/core/utils/ResourceManager.h
```diff
@@ -74,6 +74,31 @@ namespace Aws
...
+             *
+             * @return instance of RESOURCE_TYPE, nullptr if the resource manager is being shutdown
+             */
+            template<typename std::enable_if<std::is_pointer<RESOURCE_TYPE>::value>::type * = nullptr>
+            RESOURCE_TYPE TryAcquire(const uint64_t timeoutMs) {
+              std::unique_lock<std::mutex> locker(m_queueLock);
+              const bool isNotEmpty = m_semaphore.wait_for(locker, std::chrono::milliseconds(timeoutMs),
+                                                           [&]() { return m_shutdown.load() || !m_resources.empty(); });
+
+              if (m_shutdown || !isNotEmpty) {
```
sbiscigl, line 92: nit `!` and `isNot` is kind of confusing could we just use `isEmpty` instead?

src/aws-cpp-sdk-core/source/http/curl/CurlHandleContainer.cpp
```diff
@@ -44,7 +45,20 @@ CURL* CurlHandleContainer::AcquireCurlHandle()
         CheckAndGrowPool();
     }
 
-    CURL* handle = m_handleContainer.TryAcquire();
+    // TODO: 1.12: start to fail instead of infinite loop
+    CURL* handle = nullptr;
+    bool errorLogged = false;  // avoid log explosion on legacy app behavior
+    while (!handle && m_poolSize) {
+      constexpr unsigned long DEFAULT_TIMEOUT = 1000l;
+      handle = m_handleContainer.TryAcquire(std::max(m_connectTimeout, DEFAULT_TIMEOUT));
```
sbiscigl, line 53: i think we should use something else besides m_connectTimeout  as that maps one to one with `CURLOPT_CONNECTTIMEOUT_MS`, and we are sort of tying those together. this has its own unique meaning of "aquisition from the pool time", not "DNS lookup time". maybe create a static const/maybe add a config var if this would touch several clients.

sbiscigl (approved):

src/aws-cpp-sdk-core/include/aws/core/utils/ResourceManager.h
sbera87: wait_for returns an enum. 
https://en.cppreference.com/w/cpp/thread/condition_variable/wait_for
Better to use this .
enum class cv_status {
    no_timeout,  => 0
    timeout    => 1
};

doesm't that mean return on !isNotEmpty is actually returning on no timeout. Is that the intention?

sbera87: Also, isNotEmpty can be renamed to isTimeOut for clarity

sbera87: Can we add a test case to test the timeout scenario, which should be easy to replicate

SergeyRyabinin: No it doesn't return an enum. Check the spec you linked for the (2) with predicate.

sbera87: that's right. looks ok then.  if you can add a test case that will be great

SergeyRyabinin: It is a really hard to test this piece of code: curl wrapper is hard-coded to a real c curl APIs, handle container is private and cannot be replaced.
I have a 1y old draft of curl wrapper refactoring: https://github.com/aws/aws-sdk-cpp/tree/sr/curlMulti2 which I still want to have merged one day. I can add curl wrapper unit tests there.

sbera87 (approved):

src/aws-cpp-sdk-core/include/aws/core/utils/ResourceManager.h
sbera87: cool, thanks

SergeyRyabinin: Added a test on ResourceManager template class.

src/aws-cpp-sdk-core/source/http/curl/CurlHandleContainer.cpp
SergeyRyabinin: Removed usage of m_connectTimeout, and just set 1s of wait time.

sbiscigl (approved):

sbera87 (approved):

PR #3201, opened 2024-11-20
Remove build-time detection of supported WinINet/WinHTTP protocol versions.

teo-tsirpanis:
*Issue #, if available:*

*Description of changes:*

This PR removes build-time detection of WinHTTP and WinINet support for HTTP/2 and HTTP/3. In its place, we always attempt setting the HTTP version, ignoring errors (on WinHTTP in certin cases we try both enabling both HTTP/2 and 3, and then only 2).

The main reason for this is to make cross-compilation from x64 to ARM64 easier. Besides that, the built binaries will automatically take advantage of the capabilities of the system's HTTP stack even if they are built on an older machine.

*Check all that applies:*

- [ ] Linux
- [X] Windows
- [ ] Android
- [ ] MacOS
- [ ] IOS
- [ ] Other Platforms

### DIFF
--- a/src/aws-cpp-sdk-core/source/http/windows/WinHttpSyncHttpClient.cpp
+++ b/src/aws-cpp-sdk-core/source/http/windows/WinHttpSyncHttpClient.cpp
@@ -40,61 +40,54 @@ using namespace Aws::Utils::Logging;
 #ifndef WINHTTP_OPTION_HTTP_PROTOCOL_USED
     static const DWORD WINHTTP_OPTION_HTTP_PROTOCOL_USED = 134;
 #endif
+#ifndef WINHTTP_PROTOCOL_FLAG_HTTP2
+    static const DWORD WINHTTP_PROTOCOL_FLAG_HTTP2 = 0x1;
+#endif
+#ifndef WINHTTP_PROTOCOL_FLAG_HTTP3
+    static const DWORD WINHTTP_PROTOCOL_FLAG_HTTP3 = 0x2;
+#endif
 
-DWORD ConvertHttpVersionToWinHttpVersion(const Aws::Http::Version version)
+std::initializer_list<DWORD> GetWinHttpVersionsToTry(const Aws::Http::Version version)
 {
     if (version == Version::HTTP_VERSION_NONE)
     {
         /* WinHTTP http version None maps to HTTP1.1, however, libCurl None maps to "let http client decide", sticking to libCurl behavior here (use highest) */
-#if defined(WINHTTP_HAS_H3)
-        return WINHTTP_PROTOCOL_FLAG_HTTP3;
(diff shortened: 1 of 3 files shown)


src/aws-cpp-sdk-core/source/http/windows/WinHttpSyncHttpClient.cpp
```diff
@@ -40,61 +40,54 @@ using namespace Aws::Utils::Logging;
...
 #endif
+#ifndef WINHTTP_PROTOCOL_FLAG_HTTP2
+    static const DWORD WINHTTP_PROTOCOL_FLAG_HTTP2 = 0x1;
+#endif
+#ifndef WINHTTP_PROTOCOL_FLAG_HTTP3
+    static const DWORD WINHTTP_PROTOCOL_FLAG_HTTP3 = 0x2;
+#endif
 
-DWORD ConvertHttpVersionToWinHttpVersion(const Aws::Http::Version version)
+std::initializer_list<DWORD> GetWinHttpVersionsToTry(const Aws::Http::Version version)
```
SergeyRyabinin, line 50: I don't think that initializer_list works in such context (return by value from a method), it is probably an undefined behavior here.
I'd suggest returning a const reference to a local static const object, such as
```
const std::initializer_list<DWORD>& GetWinHttpVersionsToTry(const Aws::Http::Version version) {
...
    else if (version == Version::HTTP_VERSION_3)
    {
      static const std::initializer_list<DWORD> HTTP123{ WINHTTP_PROTOCOL_FLAG_HTTP3 | WINHTTP_PROTOCOL_FLAG_HTTP2, WINHTTP_PROTOCOL_FLAG_HTTP2};
      return HTTP123;
    }
```
or use some other approach.

Please check the cpp reference:

> Copying a std::initializer_list does not copy the [backing array](https://en.cppreference.com/w/cpp/language/list_initialization#List-initializing_std::initializer_list) of the corresponding initializer list.

https://en.cppreference.com/w/cpp/utility/initializer_list

see also: https://stackoverflow.com/questions/15286450/lifetime-of-a-stdinitializer-list-return-value

SergeyRyabinin:
Hi @teo-tsirpanis , thank you a lot for submitting this pull request.

While the intention is clear and very good, I have a concern about std::initializer_list usage.
Please check the code review comment.
Best regards,

src/aws-cpp-sdk-core/source/http/windows/WinHttpSyncHttpClient.cpp
teo-tsirpanis: Done.

SergeyRyabinin (approved):

SergeyRyabinin (approved):

PR #3204, opened 2024-11-21
Support query compatible trait only for json protocol

sbera87:
*Issue #, if available:*

*Description of changes:*

*Check all that applies:*

- [x] Linux
- [ ] Windows
- [ ] Android
- [ ] MacOS
- [ ] IOS
- [ ] Other Platforms


src/aws-cpp-sdk-core/source/internal/AWSHttpResourceClient.cpp
```diff
@@ -602,12 +602,16 @@ namespace Aws
         SSOCredentialsClient::SSOCredentialsClient(const Aws::Client::ClientConfiguration& clientConfiguration, Aws::Http::Scheme scheme, const Aws::String& region)
                 : AWSHttpResourceClient(clientConfiguration, SSO_RESOURCE_CLIENT_LOG_TAG)
         {
-            SetErrorMarshaller(Aws::MakeUnique<Aws::Client::JsonErrorMarshaller>(SSO_RESOURCE_CLIENT_LOG_TAG));
+          (Aws::MakeUnique<Aws::Client::JsonErrorMarshaller>(
```
SergeyRyabinin, line 605: ???
why this file is even changed?

src/aws-cpp-sdk-core/include/aws/core/client/AWSClient.h
```diff
@@ -356,6 +358,7 @@ namespace Aws
             Aws::String m_serviceName = "AWSBaseClient";
             Aws::Client::RequestCompressionConfig m_requestCompressionConfig;
             Aws::Vector<std::shared_ptr<smithy::interceptor::Interceptor>> m_interceptors;
+            Aws::Http::HeaderValueCollection m_featureHeaders;
```
SergeyRyabinin, line 361: I guess smithy client needs to be updated as well.

sbera87: Ok, I will extend this to smithy client

.github/workflows/clang-format.yml
```diff
@@ -56,7 +56,7 @@ jobs:
         run: |
           clang-format --version
           if [ -s diff_output.patch ]; then
-            python3 clang-format-diff.py -p1 -style=file:.clang-format < diff_output.patch > formatted_differences.patch 2> error.log || true
+            python3 clang-format-diff.py -iregex '.*\.(cpp|cc|c\+\+|cxx|c|h|hh|hpp)' -p1 -style=file:.clang-format < diff_output.patch > formatted_differences.patch 2> error.log || true
```
sbera87, line 59: without this just the java file fails clang format in the pipeline, locally , no matter which version

src/aws-cpp-sdk-core/include/smithy/client/AwsSmithyClientBase.h
```diff
@@ -158,15 +158,17 @@ namespace client
         virtual SelectAuthSchemeOptionOutcome SelectAuthSchemeOption(const AwsSmithyClientAsyncRequestContext& ctx) const = 0;
         virtual SigningOutcome SignRequest(std::shared_ptr<HttpRequest> httpRequest, const AuthSchemeOption& targetAuthSchemeOption) const = 0;
         virtual bool AdjustClockSkew(HttpResponseOutcome& outcome, const AuthSchemeOption& authSchemeOption) const = 0;
+        void SetFeatureHeaders(Aws::Http::HeaderValueCollection&& headers);
```
SergeyRyabinin, line 161: while I think it is okay to add such field and member variable to the old AwsClient,
I can't say the same about new Smithy client.
On smithy client, we tried to avoid any service-specific customization from polluting the main request handling logic. In reality, only few generated clients will use this feature.

I'd suggest to insert this request header in the common service generated request class, such as
https://github.com/aws/aws-sdk-cpp/blob/main/generated/src/aws-cpp-sdk-sqs/include/aws/sqs/SQSRequest.h#L33
but there might be other options.

src/aws-cpp-sdk-core/include/smithy/client/AwsSmithyClientBase.h
```diff
@@ -160,13 +160,13 @@ namespace client
         virtual bool AdjustClockSkew(HttpResponseOutcome& outcome, const AuthSchemeOption& authSchemeOption) const = 0;
 
     protected:
-        Aws::UniquePtr<Aws::Client::ClientConfiguration> m_clientConfig;
```
sbera87, line 163: will remove this file

src/aws-cpp-sdk-core/source/internal/AWSHttpResourceClient.cpp
```diff
@@ -602,41 +602,33 @@ namespace Aws
         SSOCredentialsClient::SSOCredentialsClient(const Aws::Client::ClientConfiguration& clientConfiguration, Aws::Http::Scheme scheme, const Aws::String& region)
                 : AWSHttpResourceClient(clientConfiguration, SSO_RESOURCE_CLIENT_LOG_TAG)
         {
-            SetErrorMarshaller(Aws::MakeUnique<Aws::Client::JsonErrorMarshaller>(SSO_RESOURCE_CLIENT_LOG_TAG));
+          (Aws::MakeUnique<Aws::Client::JsonErrorMarshaller>(SSO_RESOURCE_CLIENT_LOG_TAG));
 
-            m_endpoint = buildEndpoint(scheme, region, "portal.sso.", "federation/credentials");
```
sbera87, line 607: will remove this file. no longer needed but was formatted by formatter already in one commit

tools/code-generation/generator/src/main/resources/com/amazonaws/util/awsclientgenerator/velocity/cpp/AbstractServiceRequestHeader.vm
```diff
@@ -39,6 +39,9 @@ namespace ${serviceNamespace}
         headers.emplace(Aws::Http::HeaderValuePair(Aws::Http::CONTENT_TYPE_HEADER, ${CppViewHelper.computeRequestContentType($metadata)} ));
 #if($metadata.acceptHeader)
         headers.emplace(Aws::Http::HeaderValuePair(Aws::Http::ACCEPT_HEADER, "${metadata.acceptHeader}"));
+#end
+#if ($metadata.awsQueryCompatible)
+        headers.emplace(Aws::Http::HeaderValuePair(Aws::Http::X_AMZN_QUERY_MODE,"True"));
```
SergeyRyabinin, line 44: the specification uses lowercase "true" here.
while http headers must be case-insensitive, I'd prefer to be consistent "just in case".

tools/code-generation/generator/src/main/resources/com/amazonaws/util/awsclientgenerator/velocity/cpp/json/JsonErrorMarshallerHeader.vm
```diff
@@ -21,6 +21,9 @@ namespace ${serviceNamespace}
 class ${CppViewHelper.computeExportValue($metadata.classNamePrefix)} $className : public Aws::Client::JsonErrorMarshaller
 {
 public:
+#if ($metadata.awsQueryCompatible)
+  explicit $className(bool queryCompatibilityMode) : Aws::Client::JsonErrorMarshaller(queryCompatibilityMode){}
```
sbiscigl, line 25: Two comments:

first:

you need `:public Aws::Client::JsonErrorMarshaller` otherwise i dont think this would actually work. i.e. in the [aws client constuctor](https://github.com/aws/aws-sdk-cpp/blob/main/src/aws-cpp-sdk-core/include/aws/core/client/AWSJsonClient.h#L40) 

```cpp
AWSJsonClient(const Aws::Client::ClientConfiguration& configuration,
  const std::shared_ptr<Aws::Client::AWSAuthSigner>& signer,
  const std::shared_ptr<AWSErrorMarshaller>& errorMarshaller);
```

since we are missing `public` and `is a` inheritance, i dont think this would end up compiling.

second:

I think that we can make this code a little bit more maintainable using a decorator instead of a boolean. This also forces the query compatible decision to compile time and not runtime. i.e. we could do

```cpp
class JsonErrorMarshallerQueryCompatible: public JsonErrorMarshaller {
  public:
    AWSError<CoreErrors> Marshall(const Aws::Http::HttpResponse& response) const override {
       //do special query compatible stuff
    }

AWSError<CoreErrors> BuildAWSError(const std::shared_ptr<Http::HttpResponse>& httpResponse) const override {
       //do special query compatible stuff
    }
};
```

then at code generation time we could do 

```
  explicit $className : public Aws::Client::JsonErrorMarshallerQueryCompatible
```

and avoid passing around true and falses

sbiscigl (approved):

PR #3207, opened 2024-11-27
add missing http verbs in http spec

sbiscigl:
*Description of changes:*

Adds missing verbs from the [http spec](https://datatracker.ietf.org/doc/html/rfc9110.html#section-9.3) specifically `CONNECT`, `OPTIONS`, and `TRACE`

*Check all that applies:*

- [x] Linux
- [x] Windows
- [x] Android
- [x] MacOS
- [ ] IOS
- [ ] Other Platforms

### DIFF
--- a/src/aws-cpp-sdk-core/include/aws/core/http/HttpTypes.h
+++ b/src/aws-cpp-sdk-core/include/aws/core/http/HttpTypes.h
@@ -20,43 +20,37 @@ namespace Aws
         /**
          * Models Http methods.
          */
-        enum class HttpMethod
-        {
-            HTTP_GET,
-            HTTP_POST,
-            HTTP_DELETE,
-            HTTP_PUT,
-            HTTP_HEAD,
-            HTTP_PATCH
-        };
+    enum class HttpMethod {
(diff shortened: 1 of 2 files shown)


SergeyRyabinin (approved):

src/aws-cpp-sdk-core/include/aws/core/http/HttpTypes.h
```diff
@@ -20,43 +20,37 @@ namespace Aws
...
-        };
-
-        namespace HttpMethodMapper
-        {
-            /**
-             * Gets the string value of an httpMethod.
-             */
-            AWS_CORE_API const char* GetNameForHttpMethod(HttpMethod httpMethod);
-        } // namespace HttpMethodMapper
+    enum class HttpMethod {
```
sbera87, line 23: This looks good. Food for thought:  Maybe in future, we can introduce a wrapped enum macro which can not only generate the enum class but also have a toString method for the concerned enum class. Then we can get the enum string and do contextual parsing as needed.

sbera87 (approved):

PR #3220, opened 2024-12-11
Fix compilation error for older gcc 

sbera87:
*Issue #, if available:*

*Description of changes:*

*Check all that applies:*

- [ ] Linux
- [ ] Windows
- [ ] Android
- [ ] MacOS
- [ ] IOS
- [ ] Other Platforms


src/aws-cpp-sdk-core/include/aws/core/client/AWSErrorMarshaller.h
```diff
@@ -91,11 +91,10 @@ namespace Aws
 
         protected:
          const Aws::Utils::Json::JsonValue& GetJsonPayloadFromError(const AWSError<CoreErrors>&) const;
+         AWSError<CoreErrors> MarshallHelper(const Aws::String& exceptionName, const Aws::String& message) const;
```
sbiscigl, line 94: so something i cant help but notice is that that `JsonErrorMarshallerQueryCompatible::Marshall` and `JsonErrorMarshaller::Marshall` are more or less copy and pasted code, which is why you have to invoke the parent class of `JsonErrorMarshaller`. instead we should try to make this more DRY and not have copy and pasted code and this eliminates the need for the child class to access the parents class, parent class. i.e. we could make a inherited callback.

```cpp
struct Error{};
struct Response{};

class JsonSerializer {
public:
  Error Marshall(Response response) {
    auto error = ParseResponse(response);
    inheritedCallbacks(error, response);
    return error;
  }

protected:
  std::function<void(Error& error, Response& response)> inheritedCallbacks{};
private:
  Error ParseResponse(Response response);
};

class JsonQueryCompatible : public JsonSerializer {
public:
  JsonQueryCompatible() {
    inheritedCallbacks = [](Error& error, Response& response) {
      //custom error parsing
    };
  }
}
```

this way we avoid parsing something twice, and we contain the new code to _ONLY_ be the bit about query compatible traits

SergeyRyabinin (approved):

sbiscigl (approved):

PR #3221, opened 2024-12-11
Fix script injection on variables

sbera87:
*Issue #, if available:*

*Description of changes:*

*Check all that applies:*

- [x] Linux
- [ ] Windows
- [ ] Android
- [ ] MacOS
- [ ] IOS
- [ ] Other Platforms


sbiscigl (approved):

.github/workflows/clang-format.yml
```diff
@@ -30,12 +30,21 @@ jobs:
       - name: Get the base and head branches
         id: fetchstep
         run: |
-          echo "BASE_BRANCH=${{ github.event.pull_request.base.ref }}" >> $GITHUB_ENV
-          echo "HEAD_BRANCH=${{ github.event.pull_request.head.ref }}" >> $GITHUB_ENV
```
SergeyRyabinin, line 34: I don't think it is with the github recommended approach:
https://docs.github.com/en/actions/security-for-github-actions/security-guides/security-hardening-for-github-actions#using-an-intermediate-environment-variable

sbera87: It is good to have env variables but when I used BASE_BRANCH as env var, it didnt work for me citing that this can't be an env variable
${{ github.event.pull_request.base.ref }}
So, I used quotes on local variables as the fix for script injection

sbera87: Turns out BASE_BRANCH:${{ github.event.pull_request.base.ref }} is different from
BASE_BRANCH: ${{ github.event.pull_request.base.ref }} where the spacing matters. Going to try with the env variable ,if check works then all good now

SergeyRyabinin (approved):

PR #3222, opened 2024-12-11
avoid double read from steam from curl when stream is small

sbiscigl:
*Description of changes:*

Fixes a corner case in [aws chunked streaming](https://docs.aws.amazon.com/AmazonS3/latest/API/sigv4-streaming-trailers.html) that lead to performance problems. we recently fixed a bug in aws chunked streaming that fixed it in curl. There was however a performance oversight. on a chunk less that the default chunk size, it would require two stream reads from curl instead of one. requiring two stream reads increased latency very slightly for smaller objects.

*Check all that applies:*

- [x] Linux
- [x] Windows
- [x] Android
- [x] MacOS
- [ ] IOS
- [ ] Other Platforms

### DIFF
--- a/src/aws-cpp-sdk-core/include/aws/core/utils/stream/AwsChunkedStream.h
+++ b/src/aws-cpp-sdk-core/include/aws/core/utils/stream/AwsChunkedStream.h
@@ -30,30 +30,25 @@ class AwsChunkedStream {
 
   size_t BufferedRead(char *dst, size_t amountToRead) {
     assert(dst != nullptr);
-    if (dst == nullptr) {
-      AWS_LOGSTREAM_ERROR("AwsChunkedStream", "dst is null");
-    }
 
-    // the chunk has ended and cannot be read from
-    if (m_chunkEnd) {
-      return 0;
-    }
-
-    // If we've read all of the underlying stream write the checksum trailing header
-    // the set that the chunked stream is over.
-    if (m_stream->eof() && !m_stream->bad() && (m_chunkingStream->eof() || m_chunkingStream->peek() == EOF)) {
-      return writeTrailer(dst, amountToRead);
-    }
-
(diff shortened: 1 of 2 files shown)


SergeyRyabinin (approved):

SergeyRyabinin (approved):

src/aws-cpp-sdk-core/include/aws/core/utils/stream/AwsChunkedStream.h
```diff
@@ -30,30 +30,28 @@ class AwsChunkedStream {
...
+      // Try to read in a 64K chunk, if we cant we know the stream is over
+      m_stream->read(m_data.GetUnderlyingData(), DataBufferSize);
+      size_t bytesRead = static_cast<size_t>(m_stream->gcount());
 
-    // If we've read all of the underlying stream write the checksum trailing header
-    // the set that the chunked stream is over.
-    if (m_stream->eof() && !m_stream->bad() && (m_chunkingStream->eof() || m_chunkingStream->peek() == EOF)) {
-      return writeTrailer(dst, amountToRead);
-    }
+      if (bytesRead > 0) {
```
sbera87, line 41: it would be good to have less branching if possible. is this check really needed? bytesRead is a size_t cant be negative

sbiscigl: it can be zero however

sbiscigl: in theory it could be zero though i think

sbera87: yea, we can do without the check

SergeyRyabinin (approved):

PR #3225, opened 2024-12-16
SigV4A auth selection update

SergeyRyabinin:
*Issue #, if available:*
specification update
*Description of changes:*
c2j models now have an `auth` list metadata that represents a list of supported Auth Schemes on the service client.
SDKs must rely on this trait to generate a SigV4-Asymmetric signer instead of Endpoint Rules. 
(with the exception of 4 service clients at the moment whose service models are non-compliant on this regard
S3
CloudFront KeyValueStore
SESv2
EventBridge)

tested by generating all and seeing no diff is generated (as expected now).

*Check all that applies:*

- [ ] Linux
- [ ] Windows
- [ ] Android
- [ ] MacOS
- [ ] IOS
- [ ] Other Platforms


tools/code-generation/generator/src/main/java/com/amazonaws/util/awsclientgenerator/generators/cpp/CppClientGenerator.java
```diff
@@ -208,6 +208,37 @@ protected void addRequestIdToResults(final ServiceModel serviceModel) {
                 });
     }
 
+    private static Set<String> servicesMissingMultiAuthMRAPTrait = new HashSet<>();
```
sbiscigl, line 211: nit: could just do
```java
private static Set<String> servicesMissingMultiAuthMRAPTrait = ImmutableSet.of(
  "S3",
  "S3-CRT",
  "CloudFront KeyValueStore",
  "SESv2",
  "EventBridge",
)
```

intead of having to have a static block, imo avoiding the static block is nice

sbiscigl (approved):

tools/code-generation/generator/src/main/java/com/amazonaws/util/awsclientgenerator/generators/cpp/CppClientGenerator.java
```diff
@@ -208,6 +208,37 @@ protected void addRequestIdToResults(final ServiceModel serviceModel) {
...
+
+    private void CheckAndEnableSigV4A(final ServiceModel serviceModel, VelocityContext context) {
+        List<String> c2jAuthList = serviceModel.getMetadata().getAuth();
+        String serviceId = serviceModel.getMetadata().getServiceId();
+        if (c2jAuthList != null && c2jAuthList.contains("aws.auth#sigv4a") ||
+             servicesMissingMultiAuthMRAPTrait.contains(serviceId)) {
+            context.put("multiRegionAccessPointSupported", true);
+        }
+        // todo: remove these checks later
+        if (serviceModel.getEndpointRules().contains("\"sigv4a\"") &&
```
sbera87, line 228: besides endpoint rules, I think we should also check in operations. at operation level there can be authscheme in "auth" field (example cognito idp)

tools/code-generation/generator/src/main/java/com/amazonaws/util/awsclientgenerator/generators/cpp/CppClientGenerator.java
SergeyRyabinin: updated, ty

tools/code-generation/generator/src/main/java/com/amazonaws/util/awsclientgenerator/generators/cpp/CppClientGenerator.java
SergeyRyabinin: updated, thank you.
although, I'm not sure we need to check for noauth or bearer per operation.
This block is for enabling MRAP flag only, not for c2j model validation.

sbera87 (approved):
lgtm

PR #3235, opened 2025-01-02
Allow cmake option to remove tests that depend on DNS lookup

sbiscigl:
*Description of changes:*

Creates a cmake option that allows the disabling of of tests that are dependent on DNS behavior. Some customers are using a custom HTTP client that does not perform DNS as expected that will result in the tests behavior not working as expected. This allows for customers to opt out of running these tests, but defaulting to keeping them on to find potential issues in HTTP clients.

*Check all that applies:*

- [x] Linux
- [x] Windows
- [x] Android
- [x] MacOS
- [ ] IOS
- [ ] Other Platforms

### DIFF
--- a/.cspell.json
+++ b/.cspell.json
@@ -57,7 +57,7 @@
     "MMSYSERR",
     "WAVEFORMATEX",
     "Unprepare",
-    "DDISABLE_IMDSV1",
+    "DDISABLE",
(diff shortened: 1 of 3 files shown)


sbera87 (approved):
lgtm

SergeyRyabinin (approved):

PR #3236, opened 2025-01-02
copyable smithy client changes

sbera87:
*Issue #, if available:*

*Description of changes:*

*Check all that applies:*

- [ ] Linux
- [ ] Windows
- [ ] Android
- [ ] MacOS
- [ ] IOS
- [ ] Other Platforms


src/aws-cpp-sdk-core/include/smithy/client/AwsSmithyClientBase.h
```diff
@@ -119,10 +123,76 @@ namespace client
...
+        {
+            if (this != &target)
+            {
+                m_clientConfig = target.m_clientConfig;
+                m_serviceName = target.m_serviceName;
+                m_userAgent = target.m_userAgent;
+                m_httpClient = Aws::Http::CreateHttpClient(*target.m_clientConfig);
+
+                // marshaller could be of different types
+                if (auto jsonMarshaller = std::dynamic_pointer_cast<Aws::Client::JsonErrorMarshaller>(m_errorMarshaller)) {
```
sbiscigl, line 137: not really a fan of the the use of `dynamic_pointer_cast` here, I think this is a symptom of the fact that we shouldnt be declaring a custom copy constructor here, but we should be in `AwsSmithyClient.h`, because at that point in time we know what that type of error marshaller exists.

Also this will break the cmake arg [ENABLE_RTTI](https://github.com/aws/aws-sdk-cpp/blob/a94edca824ecbe8984a70e3e67707bb77d9580c8/CMakeLists.txt#L40) if we turn it to false, which some of our customers do.

The custom copy constructor and assignment operator can call down to `AwsSmithyClientBase` functions that take these things as arguments.

src/aws-cpp-sdk-core/include/smithy/client/AwsSmithyClientBase.h
```diff
@@ -119,10 +123,76 @@ namespace client
...
+                m_errorMarshaller = Aws::MakeShared<Aws::Client::XmlErrorMarshaller>(ALLOC_TAG);
+            }
+            else if (auto jsonQueryCompatibleMarshaller = std::dynamic_pointer_cast<Aws::Client::JsonErrorMarshallerQueryCompatible>(m_errorMarshaller)) {
+                m_errorMarshaller = Aws::MakeShared<Aws::Client::JsonErrorMarshallerQueryCompatible>(ALLOC_TAG);
+            }
+
+            m_interceptors = {Aws::MakeShared<ChecksumInterceptor>(ALLOC_TAG)};
+        }
+
+        AwsSmithyClientBase(AwsSmithyClientBase&& target):
```
sbiscigl, line 172: why not `= default`?

src/aws-cpp-sdk-core/include/smithy/client/AwsSmithyClientBase.h
```diff
@@ -119,10 +123,76 @@ namespace client
...
+            m_serviceName{std::move(target.m_serviceName)},           
+            m_userAgent{std::move(target.m_userAgent)},             
+            m_httpClient{std::move(target.m_httpClient)},    
+            m_errorMarshaller{std::move(target.m_errorMarshaller)},     
+            m_interceptors{std::move(target.m_interceptors)}
+        {
+
+        }
+
+        AwsSmithyClientBase& operator=(AwsSmithyClientBase&& target)
```
sbiscigl, line 183: same as the other comment, why not `= default`?

sbera87: yea, makes sense . will change

src/aws-cpp-sdk-core/include/smithy/client/AwsSmithyClientBase.h
sbera87: will change it

src/aws-cpp-sdk-core/include/aws/core/client/AWSClientAsyncCRTP.h
```diff
@@ -75,6 +75,29 @@ namespace Client
             return *this;
         }
 
+        ClientWithAsyncTemplateMethods& operator=(ClientWithAsyncTemplateMethods&& other)
```
sbiscigl, line 78: why are we adding a copy ctor for AsyncCRTP?  this is unused in the smith code path?

src/aws-cpp-sdk-core/include/aws/core/utils/Cache.h
```diff
@@ -27,6 +27,31 @@ namespace Aws
             explicit Cache(size_t initialSize = 1000) : m_maxSize(initialSize)
             {
             }
+            
+            //this can't be default as compiler wont implicitly create copy constructor for const qualified members
+            Cache(const Cache& other):m_entries{other.m_entries},m_maxSize{other.m_maxSize}{
```
sbiscigl, line 32: we are _NOT_ making anything with a cache copy constructible, if the client is not copy constructible from before the smithy client migration, it does not need to be copy constructible afterwards.

src/aws-cpp-sdk-core/include/aws/core/utils/ConcurrentCache.h
```diff
@@ -19,6 +19,33 @@ namespace Aws
         public:
             explicit ConcurrentCache(size_t size = 1000) : m_cache(size) { }
 
+
+            ConcurrentCache(const ConcurrentCache& other) : m_cache(other.m_cache) {
```
sbiscigl, line 23: same comment as above, we are not retroactively making structures copy constructible, just client that are copy constructible we will keep that way. so anything that has a cache, is not copy constructible and we dont plan on changing that.

tools/code-generation/generator/src/main/resources/com/amazonaws/util/awsclientgenerator/velocity/cpp/smithy/SmithyClientHeader.vm
```diff
@@ -21,7 +21,7 @@
 \#include <smithy/identity/auth/built-in/SigV4AuthSchemeResolver.h>
 \#include <smithy/identity/auth/built-in/SigV4AuthScheme.h>
 \#include <smithy/client/serializer/${serializer}.h>
-
+\#include <aws/${metadata.projectName}/${metadata.classNamePrefix}ErrorMarshaller.h>
```
sbiscigl, line 24: please add  the newline back before the includes and namespace for readability

src/aws-cpp-sdk-core/include/aws/core/client/AWSClientAsyncCRTP.h
sbera87: It is one of the classes in multi inheritance that is used.

`  class AWS_DYNAMODB_API DynamoDBClient : smithy::client::AwsSmithyClientT<Aws::DynamoDB::SERVICE_NAME,
      Aws::DynamoDB::DynamoDBClientConfiguration,
      smithy::SigV4AuthSchemeResolver<>,
      Aws::Crt::Variant<smithy::SigV4AuthScheme>,
      DynamoDBEndpointProviderBase,
      smithy::client::JsonOutcomeSerializer,
      smithy::client::JsonOutcome,
      Aws::Client::DynamoDBErrorMarshaller>,
    Aws::Client::ClientWithAsyncTemplateMethods<DynamoDBClient>`

src/aws-cpp-sdk-core/include/aws/core/utils/Cache.h
sbera87: ok, will remove this

src/aws-cpp-sdk-core/include/aws/core/utils/ConcurrentCache.h
sbera87: ok, will remove this one

tools/code-generation/generator/src/main/resources/com/amazonaws/util/awsclientgenerator/velocity/cpp/smithy/SmithyClientHeader.vm
sbera87: shall do

sbiscigl: > please add the newline back before the includes and namespace for readability

let me be more explicit

this is not desired

```cpp
#include <smithy/identity/auth/built-in/SigV4AuthSchemeResolver.h>
#include <smithy/client/serializer/JsonOutcomeSerializer.h>

#include <aws/dynamodb/DynamoDBErrorMarshaller.h>
namespace Aws
{
```

this is desired
```
#include <smithy/identity/auth/built-in/SigV4AuthSchemeResolver.h>
#include <smithy/client/serializer/JsonOutcomeSerializer.h>
#include <aws/dynamodb/DynamoDBErrorMarshaller.h>

namespace Aws
{
```

sbera87: got it

generated/src/aws-cpp-sdk-dynamodb/include/aws/dynamodb/DynamoDBClient.h
```diff
@@ -42,13 +43,42 @@ namespace DynamoDB
...
+      }
+
+      DynamoDBClient& operator=(DynamoDBClient&& other)
+      {
+        if(this == &other) 
+        {
+            return *this;
+        }
+        AwsSmithyClientT::operator=(std::move(other));
+        Aws::Client::ClientWithAsyncTemplateMethods<DynamoDBClient>::operator=(std::move(other));
```
sbiscigl, line 78: dont think you can call std::move on a r value twice [without entering UB land](https://godbolt.org/z/x7rWsvnvr).

generated/src/aws-cpp-sdk-dynamodb/include/aws/dynamodb/DynamoDBClient.h
```diff
@@ -42,13 +43,42 @@ namespace DynamoDB
...
+        {
+            return *this;
+        }
+        AwsSmithyClientT::operator=(other);
+        Aws::Client::ClientWithAsyncTemplateMethods<DynamoDBClient>::operator=(other);
+        init(m_clientConfiguration);
+        return *this;
+      }
+
+      DynamoDBClient& operator=(DynamoDBClient&& other)
```
sbiscigl, line 71: do we even need a custom move ctor/assignment operator? thinking that `default` would be enough because the target would be the new owner of the all of the resources, that are already intiialized.

thinking default for move should be enough

src/aws-cpp-sdk-core/include/smithy/client/AwsSmithyClient.h
```diff
@@ -29,26 +29,62 @@ namespace client
...
               m_authSchemeResolver(authSchemeResolver),
               m_authSchemes(authSchemes),
-              m_serializer(Aws::MakeUnique<SerializerT>(ServiceNameT, m_clientConfiguration.telemetryProvider))
+              m_serializer(Aws::MakeShared<SerializerT>(ServiceNameT, m_clientConfiguration.telemetryProvider))
         {
             m_serviceName = ServiceNameT;
         }
 
+        AwsSmithyClientT(const AwsSmithyClientT& other):
+            AwsSmithyClientBase(Aws::MakeShared<ServiceClientConfigurationT>(ServiceNameT, other.m_clientConfig), other.m_serviceName, Aws::Http::CreateHttpClient(*other.m_clientConfig), Aws::MakeShared<MarshallerT>(other.m_serviceName.c_str())),
```
sbiscigl, line 58: for the sake of consistency `MakeShared` should be called with the same log tag, i.e.
`Aws::MakeShared<ServiceClientConfigurationT>(ServiceNameT, other.m_clientConfig)`
vs
`Aws::MakeShared<MarshallerT>(other.m_serviceName.c_str())`

`other.m_serviceName` should be the same `ServiceNameT` but since this is the class that is calling `MakeShared` lets make it consistent to `ServiceNameT` for both

src/aws-cpp-sdk-core/include/smithy/client/AwsSmithyClientBase.h
```diff
@@ -65,6 +67,33 @@ namespace client
     /* Non-template base client class that contains main Aws Client Request pipeline logic */
     class SMITHY_API AwsSmithyClientBase
     {
+    private:
```
sbiscigl, line 70: lets move private definitions after public per the [google style guide](https://google.github.io/styleguide/cppguide.html#Declaration_Order)

> A class definition should usually start with a public: section, followed by protected:, then private:. Omit sections that would be empty.

src/aws-cpp-sdk-core/include/smithy/client/AwsSmithyClientBase.h
```diff
@@ -65,6 +67,33 @@ namespace client
...
+        const char* ALLOC_TAG{"AwsSmithyClientBase"};
+
+
+        void deepCopyClientConfiguration(const AwsSmithyClientBase& target)
+        {
+            //first create copy from target
+            m_clientConfig = Aws::MakeShared<Aws::Client::ClientConfiguration>(target.m_serviceName.c_str(), *target.m_clientConfig);
```
sbiscigl, line 77: please make `Aws::MakeShared` with `m_serviceName` and not target. they should be the same, but lets be consistent.

sbiscigl, line 77: i dont think this is going to work the way we want it to because of slicing here we declare a pointer of `ClientConfiguration` not `ServiceClientConfigurationT`. in `AwsSmithyClientT` we call it as such

```
m_clientConfiguration{*static_cast<ServiceClientConfigurationT*>(AwsSmithyClientBase::m_clientConfig.get())},
```

since this is _NOT_ a `ServiceClientConfigurationT` we enter UB territory.

trying to think of a creative solution around this let me talk to you offline about this

SergeyRyabinin (approved):

src/aws-cpp-sdk-core/include/smithy/client/AwsSmithyClientBase.h
```diff
@@ -191,6 +117,50 @@ namespace client
...
+          m_clientConfig = std::move(clientConfig);
+          m_serviceName = serviceName;
+          m_httpClient = std::move(httpClient);
+          m_errorMarshaller = std::move(errorMarshaller);
+          m_interceptors = Aws::Vector<std::shared_ptr<interceptor::Interceptor>>{Aws::MakeShared<ChecksumInterceptor>("AwsSmithyClientBase")};
+          init();
+        }
+
+        void init() {
+          if (!m_clientConfig->retryStrategy)
```
sbera87, line 134: imo,  we should remove these checks and do an unconditional invocation of the functors because when we create the shared ptr from another, it does a shallow copy of the functors too and then it won't reinitialize

src/aws-cpp-sdk-core/include/smithy/client/AwsSmithyClientBase.h
```diff
@@ -191,6 +117,50 @@ namespace client
...
+          {
+            assert(m_clientConfig->configFactories.retryStrategyCreateFn);
+            m_clientConfig->retryStrategy = m_clientConfig->configFactories.retryStrategyCreateFn();
+          }
+          if (!m_clientConfig->executor)
+          {
+            assert(m_clientConfig->configFactories.executorCreateFn);
+            m_clientConfig->executor = m_clientConfig->configFactories.executorCreateFn();
```
sbera87, line 142: m_isInitialized should be set true since you have remove the client level init (auto generated)

sbiscigl: It is not needed because it is set in the constructor of `ClientWithAsyncTemplateMethods` which we invoke when we create a dynamo client. Before it was redundant so Im trying to remove redundancy with removing that init call.

```cpp
class AWS_DYNAMODB_API DynamoDBClient : 
  smithy::client::AwsSmithyClientT<Aws::DynamoDB::SERVICE_NAME,
      Aws::DynamoDB::DynamoDBClientConfiguration,
      smithy::SigV4AuthSchemeResolver<>,
      Aws::Crt::Variant<smithy::SigV4AuthScheme>,
      DynamoDBEndpointProviderBase,
      smithy::client::JsonOutcomeSerializer,
      smithy::client::JsonOutcome,
      Aws::Client::DynamoDBErrorMarshaller>,
  Aws::Client::ClientWithAsyncTemplateMethods<DynamoDBClient>
{
  ...
}

template <typename AwsServiceClientT>
class ClientWithAsyncTemplateMethods
{
  public:
    ClientWithAsyncTemplateMethods(): 
      m_isInitialized(true),
      m_operationsProcessed(0)
    {
      ...
     }
  ...
}
```

src/aws-cpp-sdk-core/include/smithy/client/AwsSmithyClientBase.h
sbiscigl: I dont think we can do that. consider the code

```cpp
#include <aws/core/Aws.h>
#include <aws/core/client/AdaptiveRetryStrategy.h>
#include <aws/dynamodb/DynamoDBClient.h>

using namespace Aws;
using namespace Aws::DynamoDB;

namespace {
 const char* LOG_TAG = "TAG";
}

auto main() -> int {
  SDKOptions options;
  options.loggingOptions.logLevel = Aws::Utils::Logging::LogLevel::Debug;
  InitAPI(options);
  {
      Client::ClientConfiguration configuration{};
      configuration.retryStrategy = Aws::MakeShared<Client::AdaptiveRetryStrategy>(LOG_TAG);
      DynamoDBClient client{};
  }
  ShutdownAPI(options);
  return 0;
}
```

if we remove the check we would unconditionally write over the users specifed retry strategy with the default factory.

So we need to check for the existence of a user specified entity on the configuration before invoking the factory

sbera87: In the above example
`DynamoDBClient client`
This calls the constructor which is supposed to do the initialization of shared ptr members because in the very beginning sp is default nullptr.
Now say we copy the client which will invoke the deepcopy method and first it will shallow copy the shared ptr members and then wont reinitialize it using the factory functions. This means 2 clients will have same copy of shared_ptr to retry strategy for example. But in reality each client should have its own separate retry strategy. 

`if (!m_clientConfig->retryStrategy)` will otherwise always be false on a copy.

sbiscigl: ooooo good catch let me push a update

SergeyRyabinin (approved):

sbera87:
lgtm

sbiscigl (approved):

SergeyRyabinin (approved):

PR #3238, opened 2025-01-07
update user agent to 2.1

sbiscigl:
*Description of changes:*

Updates the user agent string to be compliant with the 2.1 revision. all  SDKs will eventually update to this version. For example go v2 already has in [this commit](https://github.com/aws/aws-sdk-go-v2/pull/2682/commits/dc6e9d915fd343555499e12aa40a4fed4e5b9031).

example before:
```
user-agent: 
  aws-sdk-cpp/1.11.479 
  ua/2.0 
  md/aws-crt#0.19.7 
  os/Darwin/23.6.0 
  md/arch#arm64 
  lang/c++#C++11 
  md/Clang#15.0.0 
  api/S3
```

example after:
```
user-agent: 
  aws-sdk-cpp/1.11.479 
  ua/2.1 
  api/S3 
  os/Darwin#23.6.0 
  lang/c++#C++11 
  md/aws-crt#0.19.7 
  md/arch#arm64 
  md/Clang#15.0.0 
  m/D
```

*Check all that applies:*

- [x] Linux
- [x] Windows
- [x] Android
- [x] MacOS
- [ ] IOS
- [ ] Other Platforms

### DIFF
--- a/src/aws-cpp-sdk-core/include/aws/core/AmazonWebServiceRequest.h
+++ b/src/aws-cpp-sdk-core/include/aws/core/AmazonWebServiceRequest.h
@@ -8,6 +8,7 @@
 #include <aws/core/Core_EXPORTS.h>
 
 #include <aws/core/client/RequestCompression.h>
+#include <aws/core/client/UserAgent.h>
 #include <aws/core/auth/AWSAuthSigner.h>
 #include <aws/core/client/CoreErrors.h>
 #include <aws/core/endpoint/EndpointParameter.h>
@@ -203,6 +204,18 @@ namespace Aws
         virtual Aws::Client::CompressionAlgorithm
         GetSelectedCompressionAlgorithm(Aws::Client::RequestCompressionConfig) const { return Aws::Client::CompressionAlgorithm::NONE; }
 
+        /**
+         * Adds a used feature to the user agent string for the request.
+         * @param feature the feature to be added in the user agent string.
+         */
+        void AddUserAgentFeature(Aws::Client::UserAgentFeature feature) { m_userAgentFeatures.insert(feature); }
+
+        /**
+         * Gets all features that would be included in the requests user agent string.
+         * @return a set of features that will be included in the user agent associated with this request.
+         */
+        Aws::Set<Aws::Client::UserAgentFeature> GetUserAgentFeatures() const { return m_userAgentFeatures; }
+
     protected:
         /**
          * Default does nothing. Override this to convert what would otherwise be the payload of the
@@ -221,6 +234,7 @@ namespace Aws
         RequestSignedHandler m_onRequestSigned;
         RequestRetryHandler m_requestRetryHandler;
         mutable std::shared_ptr<Aws::Http::ServiceSpecificParameters> m_serviceSpecificParameters;
+        Aws::Set<Client::UserAgentFeature> m_userAgentFeatures;
     };
 
 } // namespace Aws
--- a/src/aws-cpp-sdk-core/include/aws/core/client/UserAgent.h
+++ b/src/aws-cpp-sdk-core/include/aws/core/client/UserAgent.h
@@ -1,0 +1,45 @@
+/**
+ * Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.
+ * SPDX-License-Identifier: Apache-2.0.
+ */
+#pragma once
+#include <aws/core/client/ClientConfiguration.h>
+#include <aws/core/client/RetryStrategy.h>
+#include <aws/core/utils/memory/stl/AWSSet.h>
+
+namespace Aws {
+namespace Client {
+
+enum class UserAgentFeature {
+  RETRY_MODE_LEGACY,
+  RETRY_MODE_STANDARD,
+  RETRY_MODE_ADAPTIVE,
+  S3_TRANSFER,
+  S3_CRYPTO_V1N,
+  S3_CRYPTO_V2,
+};
+
+class AWS_CORE_API UserAgent {
+ public:
+  explicit UserAgent(const ClientConfiguration& clientConfiguration, const Aws::String& retryStrategyName, const Aws::String& apiName);
+  Aws::String SerializeWithFeatures(const Aws::Set<UserAgentFeature>& features) const;
+  void SetApiName(const Aws::String& apiName) { m_api = apiName; }
+  void AddLegacyFeature(const Aws::String& legacyFeature);
+
+ private:
+  const Aws::String m_sdkVersion;
+  const Aws::String m_userAgentVersion;
+  Aws::String m_api;
+  const Aws::String m_crtVersion;
+  const Aws::String m_osVersion;
+  const Aws::String m_archName;
+  const Aws::String m_cppVersion;
+  const Aws::String m_compilerMetadata;
+  const Aws::String m_retryStrategyName;
+  const Aws::String m_execEnv;
+  const Aws::String m_appId;
+  const Aws::String m_customizations;
+  Aws::Set<UserAgentFeature> m_features;
+};
+}  // namespace Client
+}  // namespace Aws
--- a/src/aws-cpp-sdk-core/include/smithy/client/AwsSmithyClientBase.h
+++ b/src/aws-cpp-sdk-core/include/smithy/client/AwsSmithyClientBase.h
@@ -9,6 +9,7 @@
 #include <smithy/tracing/TelemetryProvider.h>
 #include <smithy/interceptor/Interceptor.h>
 #include <smithy/client/features/ChecksumInterceptor.h>
+#include <smithy/client/features/UserAgentInterceptor.h>
 
 #include <aws/crt/Variant.h>
 #include <aws/core/client/ClientConfiguration.h>
@@ -85,11 +86,12 @@ namespace client
 
         AwsSmithyClientBase(Aws::UniquePtr<Aws::Client::ClientConfiguration>&& clientConfig,
                             Aws::String serviceName,
+                            Aws::String serviceUserAgentName,
                             std::shared_ptr<Aws::Http::HttpClient> httpClient,
                             std::shared_ptr<Aws::Client::AWSErrorMarshaller> errorMarshaller) :
           m_clientConfig(std::move(clientConfig)),
           m_serviceName(std::move(serviceName)),
-          m_userAgent(),
+          m_serviceUserAgentName(std::move(serviceUserAgentName)),
           m_httpClient(std::move(httpClient)),
           m_errorMarshaller(std::move(errorMarshaller)),
           m_interceptors{Aws::MakeShared<ChecksumInterceptor>("AwsSmithyClientBase")}
@@ -100,11 +102,12 @@ namespace client
         AwsSmithyClientBase(const AwsSmithyClientBase& other,
                             Aws::UniquePtr<Aws::Client::ClientConfiguration>&& clientConfig,
                             Aws::String serviceName,
+                            Aws::String serviceUserAgentName,
                             std::shared_ptr<Aws::Http::HttpClient> httpClient,
                             std::shared_ptr<Aws::Client::AWSErrorMarshaller> errorMarshaller) :
           m_clientConfig(std::move(clientConfig)),
           m_serviceName(std::move(serviceName)),
-          m_userAgent(),
+          m_serviceUserAgentName(std::move(serviceUserAgentName)),
           m_httpClient(std::move(httpClient)),
           m_errorMarshaller(std::move(errorMarshaller)),
           m_interceptors{Aws::MakeShared<ChecksumInterceptor>("AwsSmithyClientBase")}
@@ -181,7 +184,7 @@ namespace client
 
         std::shared_ptr<Aws::Client::ClientConfiguration> m_clientConfig;
         Aws::String m_serviceName;
-        Aws::String m_userAgent;
+        Aws::String m_serviceUserAgentName;
 
         std::shared_ptr<Aws::Http::HttpClient> m_httpClient;
         std::shared_ptr<Aws::Client::AWSErrorMarshaller> m_errorMarshaller;
--- a/src/aws-cpp-sdk-core/include/smithy/client/features/UserAgentInterceptor.h
+++ b/src/aws-cpp-sdk-core/include/smithy/client/features/UserAgentInterceptor.h
@@ -1,0 +1,46 @@
+/**
+ * Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.
+ * SPDX-License-Identifier: Apache-2.0.
+ */
+#pragma once
+
+#include <aws/core/client/ClientConfiguration.h>
+#include <aws/core/client/RetryStrategy.h>
+#include <aws/core/client/UserAgent.h>
+#include <smithy/interceptor/Interceptor.h>
+
+#include <utility>
+
+namespace smithy {
+namespace client {
+class UserAgentInterceptor : public interceptor::Interceptor {
+ public:
+  explicit UserAgentInterceptor(const Aws::Client::ClientConfiguration& configuration,
+    const Aws::String& retryStrategyName,
+    const Aws::String& apiName)
+      : m_userAgent{configuration, retryStrategyName, apiName} {};
+  UserAgentInterceptor(const UserAgentInterceptor& other) = delete;
+  UserAgentInterceptor(UserAgentInterceptor&& other) noexcept = default;
+  UserAgentInterceptor& operator=(const UserAgentInterceptor& other) = delete;
+  UserAgentInterceptor& operator=(UserAgentInterceptor&& other) noexcept = delete;
+
+  ModifyRequestOutcome ModifyBeforeSigning(interceptor::InterceptorContext& context) override {
+    auto transmitRequest = context.GetTransmitRequest();
+    assert(transmitRequest);
+    transmitRequest->SetUserAgent(m_userAgent.SerializeWithFeatures(context.GetModeledRequest().GetUserAgentFeatures()));
+    return transmitRequest;
+  }
+
+  ModifyResponseOutcome ModifyBeforeDeserialization(interceptor::InterceptorContext& context) override {
+    return context.GetTransmitResponse();
+  }
+
+  void SetApiName(const Aws::String& apiName) { m_userAgent.SetApiName(apiName); }
+
+  void AddLegacyFeaturesToUserAgent(const Aws::String& valueToAppend) { m_userAgent.AddLegacyFeature(valueToAppend); }
+
+ private:
+  Aws::Client::UserAgent m_userAgent;
+};
+}  // namespace client
+}  // namespace smithy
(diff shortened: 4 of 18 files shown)


src/aws-cpp-sdk-core/source/client/ClientConfiguration.cpp
```diff
@@ -222,14 +230,6 @@ void setLegacyClientConfigurationParameters(ClientConfiguration& clientConfig)
             client->SetEndpoint(ec2MetadataServiceEndpoint);
         }
     }
-
-    clientConfig.appId = clientConfig.LoadConfigFromEnvOrProfile(
```
sbiscigl, line 226: this actually didnt work before because if region was present in a [env var or profile configuration](https://github.com/aws/aws-sdk-cpp/blob/main/src/aws-cpp-sdk-core/source/client/ClientConfiguration.cpp#L195-L212) it would short circuit and not evaluate this

src/aws-cpp-sdk-core/source/platform/linux-shared/OSVersionInfo.cpp
```diff
@@ -48,7 +48,7 @@ Aws::String ComputeOSVersionString()
     if(success >= 0)
     {
         Aws::StringStream ss;
-        ss << name.sysname << "/" << name.release;
```
sbiscigl, line 51: this was actually broken before too
`os/Darwin/23.6.0` -> `os/Darwin#23.6.0`
which is actually the correct format and actually how CLI has it in their user agent string
```
aws-cli/2.16.5 
  md/awscrt#0.20.11 
  ua/2.0 
  os/linux#5.10.230-202.885.amzn2int.x86_64 
  md/arch#x86_64 
  lang/python#3.11.8 
  md/pyimpl#CPython 
  cfg/retry-mode#standard 
  md/installer#exe 
  md/distrib#amzn.2 
  md/prompt#off 
  md/command#ec2.describe-instances
```

src/aws-cpp-sdk-core/source/client/UserAgent.cpp
```diff
@@ -0,0 +1,166 @@
...
+#include <numeric>
+#include <utility>
+
+using namespace Aws::Client;
+
+namespace {
+const char* LOG_TAG = "UserAgent";
+
+const std::pair<UserAgentFeature, const char*> BUSINESS_METRIC_MAPPING[] = {
+    {UserAgentFeature::RETRY_MODE_LEGACY, "D"}, {UserAgentFeature::RETRY_MODE_STANDARD, "E"}, {UserAgentFeature::RETRY_MODE_ADAPTIVE, "F"},
```
SergeyRyabinin, line 22: please don't format like this, newline for each entry is okay
also, you may accidentally create an extra empty feature by using a TS-style extra last comma in the initializer list

src/aws-cpp-sdk-core/source/client/UserAgent.cpp
```diff
@@ -0,0 +1,166 @@
...
+  const auto* const metric =
+      std::find_if(std::begin(BUSINESS_METRIC_MAPPING), std::end(BUSINESS_METRIC_MAPPING),
+                   [feature](const std::pair<UserAgentFeature, const char*>& pair) -> bool { return pair.first == feature; });
+  if (metric == std::end(BUSINESS_METRIC_MAPPING)) {
+    AWS_LOGSTREAM_ERROR(LOG_TAG, "business metric mapping not found for feature");
+    return {};
+  }
+  return metric->second;
+}
+const std::pair<const char*, UserAgentFeature> RETRY_FEATURE_MAPPING[] = {
```
SergeyRyabinin, line 36: please add a blank line here

src/aws-cpp-sdk-core/include/aws/core/client/UserAgent.h
```diff
@@ -0,0 +1,45 @@
...
+ private:
+  Aws::String m_sdkVersion;
+  Aws::String m_userAgentVersion;
+  Aws::String m_api;
+  Aws::String m_crtVersion;
+  Aws::String m_osVersion;
+  Aws::String m_archName;
+  Aws::String m_cppVersion;
+  Aws::String m_compilerMetadata;
+  std::shared_ptr<Aws::Client::RetryStrategy> m_retryStrategy;
```
SergeyRyabinin, line 38: I'm not really sure we need a whole shared pointer the the actual retryStrategy used
It would make sense if we were also to report some retry metrics, but it appears we only report strategy name.
It is a kinda nitpick to think about.

src/aws-cpp-sdk-core/source/client/ClientConfiguration.cpp
SergeyRyabinin: this a not nice side-effect within `LoadConfigFromEnvOrProfile`, I'd rather fix the method as well.

src/aws-cpp-sdk-core/include/aws/core/AmazonWebServiceRequest.h
```diff
@@ -203,6 +204,9 @@ namespace Aws
         virtual Aws::Client::CompressionAlgorithm
         GetSelectedCompressionAlgorithm(Aws::Client::RequestCompressionConfig) const { return Aws::Client::CompressionAlgorithm::NONE; }
 
+        void AddUserAgentFeature(Aws::Client::UserAgentFeature feature) { m_userAgentFeatures.insert(feature); }
```
SergeyRyabinin, line 207: please add a doc string to these methods

src/aws-cpp-sdk-core/source/client/UserAgent.cpp
sbiscigl: i actually had it that way before running clang format on the file, will do one perline (i like it), we can talk about the clang format update again soon

src/aws-cpp-sdk-core/include/aws/core/client/UserAgent.h
sbiscigl: so the reason i do this is because `GetStrategyName` is [declared virtual](https://github.com/aws/aws-sdk-cpp/blob/main/src/aws-cpp-sdk-core/include/aws/core/client/RetryStrategy.h#L81-L85) and i was trying to stick by the "[Dont call virtual methods in constructors](https://wiki.sei.cmu.edu/confluence/display/cplusplus/OOP50-CPP.+Do+not+invoke+virtual+functions+from+constructors+or+destructors)" guidance so i defer the call to when we actually assemble the User Agent. the retry strategy should be well defined when we call it in the AwsClient constructor, but i just wanted to avoid it all together.

I actaully had it the other way to begin with and when back on it, could go either way, lemme know if you prefer one way or the other.

src/aws-cpp-sdk-core/source/client/UserAgent.cpp
```diff
@@ -0,0 +1,166 @@
...
+  Aws::StringStream userAgentValue;
+  userAgentValue << SDK_VERSION << METADATA_DELIMINATOR << m_sdkVersion << RWS;
+  if (!m_customizations.empty()) {
+    userAgentValue << m_customizations << RWS;
+  }
+  userAgentValue << UA_VERSION << METADATA_DELIMINATOR << m_userAgentVersion << RWS;
+  if (!m_api.empty()) {
+    userAgentValue << API_NAME << METADATA_DELIMINATOR << m_api << RWS;
+  }
+  userAgentValue << OS_VERSION << METADATA_DELIMINATOR << m_osVersion << RWS;
```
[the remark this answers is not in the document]
sbiscigl: ah this makes it pretty clean

src/aws-cpp-sdk-core/source/client/ClientConfiguration.cpp
sbiscigl: not sure if I 100 percent understand but it doesnt have to do with `LoadConfigFromEnvOrProfile` it had to do with the [return calls in client config](https://github.com/aws/aws-sdk-cpp/blob/main/src/aws-cpp-sdk-core/source/client/ClientConfiguration.cpp#L195-L212). but yeah i should fix this function moving all region code to its own function.

src/aws-cpp-sdk-core/include/aws/core/client/UserAgent.h
sbiscigl: talked offline -- we only need to follow this when its a virtual method of the class itself we're constructing, todo - move back to using name

SergeyRyabinin (approved):

src/aws-cpp-sdk-core/include/smithy/client/features/UserAgentInterceptor.h
```diff
@@ -0,0 +1,41 @@
...
+class UserAgentInterceptor : public interceptor::Interceptor {
+ public:
+  explicit UserAgentInterceptor(const Aws::Client::ClientConfiguration& configuration, const Aws::String& retryStrategyName)
+      : m_userAgent{configuration, retryStrategyName} {};
+
+  ModifyRequestOutcome ModifyBeforeSigning(interceptor::InterceptorContext& context) override {
+    const auto userAgentString = m_userAgent.SerializeWithFeatures(context.GetModeledRequest().GetUserAgentFeatures());
+    auto transmitRequest = context.GetTransmitRequest();
+    assert(transmitRequest);
+    transmitRequest->SetUserAgent(userAgentString);
```
sbera87, line 25: nit: inline 
transmitRequest->SetUserAgent(m_userAgent.SerializeWithFeatures(context.GetModeledRequest().GetUserAgentFeatures()) )

src/aws-cpp-sdk-core/source/client/UserAgent.cpp
```diff
@@ -0,0 +1,180 @@
...
+const char* EXEC_ENV = "exec-env";
+const char* APP_ID = "app";
+const char* BUSINESS_METRICS = "m";
+}  // namespace
+
+UserAgent::UserAgent(const ClientConfiguration& clientConfiguration, const Aws::String& retryStrategyName)
+    : m_sdkVersion{FilterUserAgentToken(Version::GetVersionString())},
+      m_userAgentVersion{USER_AGENT_VERSION},
+      m_crtVersion{FilterUserAgentToken(AWS_CRT_CPP_VERSION)},
+      m_osVersion{FilterUserAgentToken(Aws::OSVersionInfo::ComputeOSVersionString().c_str())},
```
sbera87, line 70: maybe add empty string checks here. What happens if a field is not found, do we leave a new line ? What's the convention

src/aws-cpp-sdk-core/include/smithy/client/AwsSmithyClientBase.h
```diff
@@ -115,8 +116,14 @@ namespace client
                 assert(m_clientConfig->configFactories.telemetryProviderCreateFn);
                 m_clientConfig->telemetryProvider = m_clientConfig->configFactories.telemetryProviderCreateFn();
             }
-
-            m_userAgent = Aws::Client::ComputeUserAgentString(m_clientConfig.get());
+            if (m_clientConfig && m_clientConfig->retryStrategy) {
+              auto userAgentInterceptor = Aws::MakeShared<UserAgentInterceptor>("AwsSmithyClientBase", *m_clientConfig,
+                                                                                m_clientConfig->retryStrategy->GetStrategyName());
+              if (userAgentInterceptor) {
```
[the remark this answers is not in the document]
sbiscigl: in the original AwsClient we have a public method called [SetServiceName](https://github.com/aws/aws-sdk-cpp/blob/main/src/aws-cpp-sdk-core/include/aws/core/client/AWSClient.h#L187-L192) which mutably sets the service name in the user agent.

```
/**
   * service client name is part of userAgent.
   * We need to update userAgent after setting this name if it's not pre-set by customer.
   * Note: this API should only be called in your extended class constructors.
   */
virtual void SetServiceClientName(const Aws::String& name);
```

which is a public API we provide and cannot break. this call exists to preserve that backwards compatibility during the smithy client migration.

we can add it as a optional parameter to the constructor to avoid this logic which is nice, but making it mandated breaks backwards compat so im against it as its own constructor

src/aws-cpp-sdk-core/source/client/UserAgent.cpp
sbiscigl: we do empty string checks on most, but missed a few, adding

src/aws-cpp-sdk-core/source/client/UserAgent.cpp
```diff
@@ -0,0 +1,180 @@
...
+  // metrics
+  Aws::Vector<Aws::String> encodedMetrics{};
+
+  // add retry encoded value
+  if (!m_retryStrategyName.empty()) {
+    const auto* const retryFeature = std::find_if(
+        std::begin(RETRY_FEATURE_MAPPING), std::end(RETRY_FEATURE_MAPPING),
+        [this](const std::pair<const char*, UserAgentFeature>& pair) -> bool { return strcmp(pair.first, FilterUserAgentToken(m_retryStrategyName.c_str()).c_str()) == 0; });
+    if (retryFeature != std::end(RETRY_FEATURE_MAPPING)) {
+      encodedMetrics.emplace_back(BusinessMetricForFeature(retryFeature->second));
```
sbera87, line 134: this can be empty string. If so, will that affect the end result, or should it  throw an exception

sbiscigl: > this can be empty string. If so, will that affect the end result

no we check for emptiness a few lines below this

src/aws-cpp-sdk-core/include/aws/core/client/UserAgent.h
```diff
@@ -0,0 +1,45 @@
...
+
+class AWS_CORE_API UserAgent {
+ public:
+  explicit UserAgent(const ClientConfiguration& clientConfiguration, const Aws::String& retryStrategyName);
+  Aws::String SerializeWithFeatures(const Aws::Set<UserAgentFeature>& features) const;
+  void SetApiName(const Aws::String& apiName) { m_api = apiName; }
+  void AddLegacyFeature(const Aws::String& legacyFeature);
+
+ private:
+  Aws::String m_sdkVersion;
```
sbera87, line 30: shouldn't the members be const since noone should change once set?

sbiscigl: most can, but some can mutate

src/aws-cpp-sdk-core/source/client/UserAgent.cpp
```diff
@@ -0,0 +1,187 @@
...
+      m_api(apiName),
+      m_crtVersion{FilterUserAgentToken(AWS_CRT_CPP_VERSION)},
+      m_osVersion{FilterUserAgentToken(Aws::OSVersionInfo::ComputeOSVersionString().c_str())},
+      m_archName{FilterUserAgentToken(Aws::OSVersionInfo::ComputeOSVersionArch().c_str())},
+      m_cppVersion{FilterUserAgentToken(Version::GetCPPStandard())},
+      m_compilerMetadata{FilterUserAgentToken(Version::GetCompilerVersionString())},
+      m_retryStrategyName{retryStrategyName},
+      m_execEnv{FilterUserAgentToken(Aws::Environment::GetEnv(EXEC_ENV_VAR).c_str())},
+      m_appId{FilterUserAgentToken(clientConfiguration.appId.c_str())}
+#if defined(AWS_USER_AGENT_CUSTOMIZATION)
```
sbera87, line 80: personally I would not use macros unless necessary
an alternative perhaps . Not a blocker just food for thought
`#if defined(AWS_USER_AGENT_CUSTOMIZATION)
constexpr const char* awsUserAgentCustomization = AWS_USER_AGENT_CUSTOMIZATION;
#else
constexpr const char* awsUserAgentCustomization = "";
#endif`

and then , 
`m_customizations{FilterUserAgentToken(awsUserAgentCustomization)}
`

sbiscigl: [this is copied existing code](https://github.com/aws/aws-sdk-cpp/blob/37857690d32a42fe6f2573d41b29fb45cf470eef/src/aws-cpp-sdk-core/source/client/ClientConfiguration.cpp#L110-L116), that we cannot remove for backwards compatible reasons.

SergeyRyabinin:
```
constexpr const char* awsUserAgentCustomization = AWS_USER_AGENT_CUSTOMIZATION;
```
won't compile because, considering macro defined as `-DmyCustomUserAgentToken`, will be substituted by preprocessor to
```
constexpr const char* awsUserAgentCustomization = myCustomUserAgentToken; /* not a string literal! */
```

double macro is a standard approach here to convert macro value to a string literal.

src/aws-cpp-sdk-core/source/client/UserAgent.cpp
sbera87: i couldn't find but I see this logic
`std::accumulate(encodedMetrics.begin(), encodedMetrics.end(), Aws::String{},
                                      [](const Aws::String& acc, const Aws::String& value) -> Aws::String {
                                        return acc.empty() ? value : acc + "," + value;
                                      });`
                                      
   So if metrics is {"A","","B"} has an empty string it can show up as "A,,B". If it's acceptable then its ok. Else maybe some mechanism to catch it early on

src/aws-cpp-sdk-core/include/smithy/client/features/UserAgentInterceptor.h
```diff
@@ -0,0 +1,42 @@
...
+  ModifyResponseOutcome ModifyBeforeDeserialization(interceptor::InterceptorContext& context) override {
+    return context.GetTransmitResponse();
+  }
+
+  void SetApiName(const Aws::String& apiName) { m_userAgent.SetApiName(apiName); }
+
+  void AddLegacyFeaturesToUserAgent(const Aws::String& valueToAppend) { m_userAgent.AddLegacyFeature(valueToAppend); }
+
+ private:
+  Aws::Client::UserAgent m_userAgent;
```
sbera87, line 39: Note that with the presence of const members m_userAgent won't have default copy constructor / assignment operator etc. This will make UserAgentInterceptor not copyable. As long as that is fine.

sbiscigl: thats fine, its accessed via a pointer

src/aws-cpp-sdk-core/source/client/UserAgent.cpp
sbera87: I was going by the existing definition which is quoted. Without that yes it will fail. Yes, usually we do double macro for cmdline flags . this was just food for thought  for the internal ones that we set up /control as strings like this in cmake
`    set(AWS_USER_AGENT_CUSTOMIZATION "" CACHE STRING "User agent extension")
`

sbera87 (approved):
Approving with comments on empty str checks , const correctness, copy constructibility . Functionally looks ok to me.  Revisions look good

src/aws-cpp-sdk-core/source/client/UserAgent.cpp
sbiscigl: good catch refactoring some of the logic out to lambda to avoid doing the same thing twice

SergeyRyabinin (approved):

Issue #3244, opened 2025-01-14
endpointOverride with bare IPv6 address fails valid domain check


adamsb6:
### Describe the bug

I've found that with a bare IPv4 address I'm able to set endpointOverride and successfully make requests against a non-AWS S3 endpoint.

However, with a bare IPv6 address, we get this error:
```
CopyFileAllAtOnce file pair 0:13675366 failed after 1 attempts. 0 operation 0:13675367 failed after 1 attempts. Operation error message: S3Exception:S3 exception: error msg: Invalid DNS Label found in URI host  S3ErrorCode:14 Retriable:false IsWrite:false Message:path: s3://ai/foo.txt ApiCall: S3: HeadObject() HL: getObjectMetadata()
```

I found that the problem arises here: https://github.com/aws/aws-sdk-cpp/blob/main/src/aws-cpp-sdk-core/source/utils/DNS.cpp#L43

The endpoint gets passed into the function. We check if it has any `.` characters, and if not, fail the `IsValidHost()` function. Bare IPv4 addresses pass this test. But bare IPv6 addresses, lacking `.` characters, fail the test.

I also found that if I make an /etc/hosts entry for the IPv6 address I want to use and pass that instead, the client works. So connectivity works if we don't block ourselves with this test.

Can you rework it so that we also pass this test if the endpoint is a bare IPv6 address?

When searching to see if this had already been raised, I saw this issue has come up before a couple of other times without being diagnosed:
https://github.com/aws/aws-sdk-cpp/issues/1857
https://github.com/aws/aws-sdk-cpp/issues/3054

### Regression Issue

- [ ] Select this option if this issue appears to be a regression.

### Expected Behavior

I should be able to set an IPv6 address as my endpoint, and it should work.

### Current Behavior

The client fails out with a message that the endpoint has an invalid DNS label.

### Reproduction Steps

I wasn't trying to reach ::1, but you should be able to easily repro by instantiating a client with a config that has endpointOverride set to `https://[::1]:80`.

### Possible Solution

Fix the cited check to allow valid IPv6 addresses to pass the check.

### AWS CPP SDK version used

1.11.328

### Compiler and Version used

clang 17

### Operating System and version

CentOS 9

sbera87:
I confirm your observation and am working on a fix

sbera87:
For reference:https://github.com/awslabs/aws-c-common/pull/1180

mihir-robin:
Hi @sbera87 !

[...]

PR #3250, opened 2025-01-15
Ipv6

sbera87:
*Issue #, if available:*
https://github.com/aws/aws-sdk-cpp/issues/3244
*Description of changes:*

*Check all that applies:*

- [ ] Linux
- [ ] Windows
- [ ] Android
- [ ] MacOS
- [ ] IOS
- [ ] Other Platforms


src/aws-cpp-sdk-core/source/utils/DNS.cpp
```diff
@@ -40,6 +40,65 @@ namespace Aws
             return true;
         }
 
+        bool isValidIPv6Segment(const std::string& segment) {
+            if (segment.empty() || segment.length() > 4){
+                return false;
+            }
+            return std::all_of(segment.begin(), segment.end(), isxdigit);
```
SergeyRyabinin, line 47: > Like all other functions from [<cctype>](https://en.cppreference.com/w/cpp/header/cctype), the behavior of std::isxdigit is undefined if the argument's value is neither representable as unsigned char nor equal to [EOF](http://en.cppreference.com/w/cpp/io/c). To use these functions safely with plain chars (or signed chars), the argument should first be converted to unsigned char...

https://en.cppreference.com/w/cpp/string/byte/isxdigit

sbera87: Thanks for the comment. I have raised a change to address the corner case.

src/aws-cpp-sdk-core/source/utils/DNS.cpp
```diff
@@ -49,7 +50,7 @@ namespace Aws
                 return false;
             }
 
-            return !std::any_of(labels.begin(), labels.end(), [](const Aws::String& label){ return !IsValidDnsLabel(label); });
+            return !std::any_of(labels.begin(), labels.end(), [](const Aws::String& label){ return !IsValidDnsLabel(label); }) || aws_host_utils_is_ipv6(aws_byte_cursor_from_c_str(host.c_str()), false);
```
sbiscigl, line 53: two questions:

* why are we using the C function `aws_byte_cursor_from_c_str ` instead of the cpp binding [`ByteBufFromCString`](https://github.com/awslabs/aws-crt-cpp/blob/main/include/aws/crt/Types.h#L56). when using the CRT we should prefer C++ binding in the CRT as opposed to direct C calls.

* Why not create a CRT C++ binding for `aws_host_utils_is_ipv6` instead of just calling the C function? We should create small and reusable bindings when we can. Calling the CRT C directly is something that we want to try to avoid.

sbera87: updated with the binding

SergeyRyabinin (approved):

sbiscigl (approved):

PR #3251, opened 2025-01-15
Refactor and simplify logic in sdksCommon cmake script

SergeyRyabinin:
*Issue #, if available:*
cmake logic depends on c2j file list in `tools/code-generation/api-descriptions`
This does not make much sense anymore since
1. generated clients are all located under `generated/src` source tree
2. c2j file renaming logic is extracted and duplicated within the code generation script [run_code_generation.py](https://github.com/aws/aws-sdk-cpp/blob/main/tools/scripts/run_code_generation.py), so additional logic of datetime extraction, c2j filename renaming and etc is not actually required in the cmake script.

i.e. cmake can rely solely on the source tree

tested by building both full main and this branch and by comparing the cmake output and resulting build/install directories, i.e.
```
tree ~/sdk-install-main-clean/ > mainTree.log
tree ~/sdk-install-cmake-refactor/ > refactorTree.log
diff mainTree.log refactorTree.log
```

*Description of changes:*

*Check all that applies:*

- [ ] Linux
- [ ] Windows
- [ ] Android
- [ ] MacOS
- [ ] IOS
- [ ] Other Platforms

### DIFF
--- a/cmake/sdksCommon.cmake
+++ b/cmake/sdksCommon.cmake
@@ -1,22 +1,11 @@
 # Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.
 # SPDX-License-Identifier: Apache-2.0.
-#
 
-# helper function that that gives primitive map functionality by treating a colon as the key-value separator, while the list semi-colon separates pairs
-# to use, pass the list-map in as a third parameter (see helper functions below)
-function(get_map_element KEY VALUE_VAR)
-    foreach(ELEMENT_PAIR ${ARGN})
-        STRING(REGEX REPLACE "([^:]+):.*" "\\1" KEY_RESULT ${ELEMENT_PAIR})
-        if(${KEY_RESULT} STREQUAL ${KEY} )
-            STRING(REGEX REPLACE "[^:]+:(.*)" "\\1" VALUE_RESULT ${ELEMENT_PAIR})
-            set(${VALUE_VAR} "${VALUE_RESULT}" PARENT_SCOPE)
-            return()
-        endif()
-    endforeach()
-    set(${VALUE_VAR} "" PARENT_SCOPE)
-endfunction(get_map_element)
-
-function(get_multimap_element KEY VALUE_VAR)
+# Return all values for a key from a flatmap
+# I.e. given a flatmap list of items such as "key1:val1,val2,val3;key2:val2;key1:valDoNotLoseMe" and "key1" as arguments
+# when called such as "get_flatmap_value(${TEST_NAME} TEMP_VAR ${TEST_DEPENDENCY_LIST})"
+# will return "val1,val2,val3,valDoNotLoseMe"
+function(get_flatmap_value KEY VALUE_VAR)
     set(${VALUE_VAR} "" PARENT_SCOPE)
     set(RETURN_LIST "")
     foreach(ELEMENT_PAIR ${ARGN})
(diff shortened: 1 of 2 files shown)


sbiscigl (approved):

cmake/sdksCommon.cmake
```diff
@@ -80,69 +60,21 @@ function(get_sdks_depending_on SDK_NAME DEPENDING_SDKS_VAR)
...
-            LIST(APPEND C2J_SPECIAL_NAME_LIST "awstransfer:transfer")
-            set(svc "awstransfer")
-        elseif ("${svc}" STREQUAL "transcribe-streaming")
-            LIST(APPEND C2J_SPECIAL_NAME_LIST "transcribestreaming:transcribe-streaming")
-            set(svc "transcribestreaming")
-        elseif ("${svc}" STREQUAL "streams.dynamodb")
-            LIST(APPEND C2J_SPECIAL_NAME_LIST "dynamodbstreams:streams.dynamodb")
-            set (svc "dynamodbstreams")
-        elseif("${svc}" MATCHES "\\.")
-            string(REPLACE "." ";" nameParts ${svc})
```
sbera87, line 113: These special name mapping is also captured in https://github.com/aws/aws-sdk-cpp/blob/main/tools/code-generation/smithy/codegen/smithy2c2j_service_map.json
Please just ensure the original naming is retained otherwise smoke tests wont build because they are built using the cmake library name which mimics the folder name 
example: https://github.com/aws/aws-sdk-cpp/blob/main/generated/smoke-tests/apigateway/CMakeLists.txt#L4

sbera87: For reference, here is how the cmake is generated: https://github.com/aws/aws-sdk-cpp/blob/94627af2200b4ce3e791f46e95b6ace82931953a/tools/code-generation/smithy/codegen/cpp-smoke-tests-codegen/src/main/java/com/amazonaws/util/awsclientsmithygenerator/generators/SmokeTestsCMakeWriter.java#L27

sbera87 (commented):
Changes look fine overall. Just a minor comment and request to verify if name change affects smoke tests.

cmake/sdksCommon.cmake
```diff
@@ -27,39 +16,30 @@ function(get_multimap_element KEY VALUE_VAR)
...
-    get_multimap_element(${SERVICE_NAME} TEMP_VAR ${SDK_TEST_PROJECT_LIST})
+    get_flatmap_value(${SERVICE_NAME} TEMP_VAR ${SDK_TEST_PROJECT_LIST})
     set(${TEST_PROJECT_NAME_VAR} "${TEMP_VAR}" PARENT_SCOPE)
 endfunction()
 
 function(get_dependencies_for_sdk PROJECT_NAME DEPENDENCY_LIST_VAR)
-    get_map_element(${PROJECT_NAME} TEMP_VAR ${SDK_DEPENDENCY_LIST})
+    get_flatmap_value(${PROJECT_NAME} TEMP_VAR ${SDK_DEPENDENCY_LIST})
+    if (TEMP_VAR)
+        string(REPLACE "," ";" TEMP_LIST_VAR ${TEMP_VAR})
```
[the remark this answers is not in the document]
SergeyRyabinin: I saw this and even tried to get rid of this replacements as well, but
while it is indeed tries to return a semicolon-separated cmake list,
thanks to some cmake deep "magic", it is returned as a comma-separated list, and then it starts to fail in other places of the project in very strange ways (like when both comma and semicolon just disappear and the whole list is concatenated into a single word).

We can address it later (if it is even possible to fix)

cmake/sdksCommon.cmake
sbera87: Thanks for verifying.

cmake/sdksCommon.cmake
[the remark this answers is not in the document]
sbera87: Interesting. Thanks for letting me know

PR #3262, opened 2025-01-20
Smithy Identity auth refactor for Rest XML S3 client

sbera87:
*Issue #, if available:*

*Description of changes:*

*Check all that applies:*

- [ ] Linux
- [ ] Windows
- [ ] Android
- [ ] MacOS
- [ ] IOS
- [ ] Other Platforms


tools/code-generation/generator/src/main/java/com/amazonaws/util/awsclientgenerator/domainmodels/codegeneration/ServiceModel.java
```diff
@@ -90,4 +90,5 @@ public Operation getOperationForRequestShapeName(String requestShapeName) {
     ClientContextParams clientContextParams;
     boolean useSmithyClient;
     List<String> authSchemes;
+    String rawEndpointRules;
```
sbera87, line 93: This is needed to parse at the service model level since otherwise only shortened rules get stored

tools/code-generation/generator/src/main/java/com/amazonaws/util/awsclientgenerator/generators/cpp/CppClientGenerator.java
```diff
@@ -836,4 +842,105 @@ private void addRequestlessRequestObjectS(final ServiceModel serviceModel) {
                     requestlessOperations.add(operation.getName());
                 });
     }
+
+    //dfs
```
sbera87, line 846: Since this is the base class , added here so that it can be extended to any client code generator which needs multi auth

tools/code-generation/generator/src/main/java/com/amazonaws/util/awsclientgenerator/generators/cpp/CppClientGenerator.java
```diff
@@ -836,4 +842,105 @@ private void addRequestlessRequestObjectS(final ServiceModel serviceModel) {
...
+                if (!authSchemeSet.contains(authScheme)) {
+                    serviceModel.getAuthSchemes().add(authScheme);
+                    authSchemeSet.add(authScheme);
+                }
+            });
+        });
+        serviceModel.setAuthSchemes(authschemes);
+    }
+    //auth schemes can be named differently in endpoints/operations, this is a mapping
+    private static final Map<String, String> AuthSchemeNameMapping = ImmutableMap.of(
```
sbera87, line 941: mapping is needed to simplify auth scheme lookup

src/aws-cpp-sdk-core/source/smithy/client/AwsSmithyClientBase.cpp
```diff
@@ -552,3 +594,26 @@ void AwsSmithyClientBase::DisableRequestProcessing()
 {
     m_httpClient->DisableRequestProcessing();
 }
+
+void AwsSmithyClientBase::EnableRequestProcessing()
+{
+    m_httpClient->EnableRequestProcessing();
+}
+
+StreamOutcome AwsSmithyClientBase::MakeRequestWithUnparsedResponse(Aws::AmazonWebServiceRequest const * const request,
```
sbera87, line 603: Different outcome handling as legacy but reusing smithy functions

tests/aws-cpp-sdk-core-tests/smithy/client/SmithyClientTest.cpp
```diff
@@ -305,7 +305,7 @@ TEST_F(SmithyClientTest, testSigV4a) {
 
     EXPECT_EQ(res2.IsSuccess(), true);
 
-    EXPECT_TRUE(res2.GetResult()->GetSigningAccessKey().empty());
+    EXPECT_TRUE(!res2.GetResult()->GetSigningAccessKey().empty());
```
sbera87, line 308: previously, this was missing support which legacy had. Now added and tested

tests/aws-cpp-sdk-s3-integration-tests/BucketAndObjectOperationTest.cpp
```diff
@@ -150,13 +150,37 @@ namespace
         long CalculateDelayBeforeNextRetry(const AWSError<CoreErrors>&, long) const override { return 0; }
     };
 
+    class S3TestClient : public S3Client
```
sbera87, line 153: This wrapper client is used instead because of use of protected methods in smithy client. Else will need to be exposed.

src/aws-cpp-sdk-core/include/smithy/identity/signer/built-in/SignerProperties.h
```diff
@@ -0,0 +1,11 @@
+/**
+ * Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.
+ * SPDX-License-Identifier: Apache-2.0.
+ */
+#pragma once
+
+namespace smithy {
+    static const char *AUTH_SCHEME_PROPERTY = "signerName";
```
sbera87, line 8: known properties that are used as overrides for creds and signer resolution

src/aws-cpp-sdk-core/include/smithy/client/AwsSmithyClient.h
```diff
@@ -204,13 +221,61 @@ namespace client
...
             auto authSchemeOption = std::move(authSchemeOptionOutcome.GetResultWithOwnership());
-            if (AwsClientRequestSigning<AuthSchemesVariantT>::PreSignRequest(request, authSchemeOption, m_authSchemes, region, serviceName, expirationInSeconds).IsSuccess())
+            if (AwsClientRequestSigning<AuthSchemesVariantT>::PreSignRequest(request, authSchemeOption, m_authSchemes, region, serviceName, signerName, expirationInSeconds).IsSuccess())
             {
                 return request->GetURIString();
             }
             return {};
         }
 
+        Aws::String GeneratePresignedUrl(const Aws::Http::URI& uri,
```
sbera87, line 231: support both interfaces as legacy

src/aws-cpp-sdk-core/source/smithy/client/AwsSmithyClientBase.cpp
```diff
@@ -109,6 +113,8 @@ AwsSmithyClientBase::BuildHttpRequest(const std::shared_ptr<AwsSmithyClientAsync
         AddHeadersToRequest(httpRequest, pRequest->GetHeaders());
         AddHeadersToRequest(httpRequest, pRequest->GetAdditionalCustomHeaders());
 
+        httpRequest->SetHasEventStreamResponse(pRequest->HasEventStreamResponse());
```
sbera87, line 116: This was missing in new client without which wont flush the responses in streaming operations

src/aws-cpp-sdk-core/source/smithy/client/AwsSmithyClientBase.cpp
```diff
@@ -361,9 +401,11 @@ void AwsSmithyClientBase::HandleAsyncReply(std::shared_ptr<AwsSmithyClientAsyncR
 
     Aws::Client::HttpResponseOutcome outcome = [&]()
     {
-        bool hasEmbeddedError = pRequestCtx->m_pRequest &&
-                      pRequestCtx->m_pRequest->HasEmbeddedError(httpResponse->GetResponseBody(), httpResponse->GetHeaders());
-        if (Utils::DoesResponseGenerateError(httpResponse) || hasEmbeddedError)
+        auto hasEmbeddedError = [&]() {
```
sbera87, line 404: important to make it lambda to abide by legacy behavior, else it would evaluate wrong condition out of order which tampers with the stream pointers setting off fail bit.

src/aws-cpp-sdk-core/source/auth/signer/AWSAuthV4Signer.cpp
```diff
@@ -209,7 +210,8 @@ bool AWSAuthV4Signer::SignRequestWithCreds(Aws::Http::HttpRequest& request, cons
             0 /* expirationTimeInSeconds doesn't matter for HttpRequestViaHeaders */, Aws::Crt::Auth::SignatureType::HttpRequestViaHeaders);
     }
 
-    if (!credentials.GetSessionToken().empty())
+    //additional check is needed for s3 header since this is invoked by s3 signer
+    if (!credentials.GetSessionToken().empty() && !request.HasHeader(S3_EXPRESS_HEADER))
```
sbera87, line 214: necessary to avoid double header add for s3 express workflow

src/aws-cpp-sdk-core/include/smithy/client/AwsSmithyClient.h
```diff
@@ -98,9 +99,33 @@ namespace client
...
+          //move other base class members
+          m_endpointProvider = std::move(other.m_endpointProvider);
+          m_authSchemeResolver =  std::move(other.m_authSchemeResolver);
+          m_authSchemes = std::move(other.m_authSchemes);
+          m_serializer = std::move(other.m_serializer);
+          m_errorMarshaller = std::move(other.m_errorMarshaller);
+        }
 
-        AwsSmithyClientT& operator=(AwsSmithyClientT&&) = default;
+        AwsSmithyClientT& operator=(AwsSmithyClientT&& other) {
```
sbera87, line 114: This is exposed by a test case in s3control and is needed due to the provided inline comments.

src/aws-cpp-sdk-core/include/smithy/identity/signer/built-in/S3ExpressSigner.h
```diff
@@ -0,0 +1,113 @@
...
+            : BASECLASS(serviceName, region)
+        {
+        }
+
+        explicit S3ExpressSigner(const Aws::String& serviceName, const Aws::String& region, Aws::Client::AWSAuthV4Signer::PayloadSigningPolicy policy, bool escapeUrl)
+            : BASECLASS(serviceName, region, policy, escapeUrl)
+        {
+        }
+
+        SigningFutureOutcome sign(std::shared_ptr<Aws::Http::HttpRequest> httpRequest, const AwsCredentialIdentityBase& identity, SigningProperties properties) override
```
sbera87, line 45: This file no longer needed to be auto generated as all the params are provided for signer resolution in the new smithy way that an handle v4a as well as v4 flavor of instantiation

tools/code-generation/generator/src/main/java/com/amazonaws/util/awsclientgenerator/transform/C2jModelToGeneratorModelTransformer.java
```diff
@@ -196,6 +196,7 @@ public ServiceModel convert() {
             shortenedRules += "\0";
             serviceModel.setEndpointRules(shortenedRules);
         }
+        serviceModel.setRawEndpointRules(c2jServiceModel.getEndpointRules() );
```
sbera87, line 199: necessary because need raw endpoint rules to be parsed at service model level (which doesn't have hook to top level model)

src/aws-cpp-sdk-core/source/smithy/client/AwsSmithyClientBase.cpp
```diff
@@ -494,7 +550,7 @@ void AwsSmithyClientBase::HandleAsyncReply(std::shared_ptr<AwsSmithyClientAsyncR
         if (retryWithCorrectRegion)
         {
             Aws::String newEndpoint = m_errorMarshaller->ExtractEndpoint(outcome.GetError());
-            if (newEndpoint.empty()) {
+            if (!newEndpoint.empty()) {
```
sbera87, line 553: bug fix. Type error

src/aws-cpp-sdk-core/source/smithy/client/AwsSmithyClientBase.cpp
```diff
@@ -216,6 +237,30 @@ void AwsSmithyClientBase::MakeRequestAsync(Aws::AmazonWebServiceRequest const* c
     AttemptOneRequestAsync(std::move(pRequestCtx));
 }
 
+void AwsSmithyClientBase::UpdateAuthSchemeFromEndpoint(const Aws::Endpoint::AWSEndpoint& endpoint, AuthSchemeOption& authscheme) const
+{
+    //get signer Name from end point and pass this info
+    if (endpoint.GetAttributes()) {
+        auto authschemeName = endpoint.GetAttributes()->authScheme.GetName();
```
sbera87, line 244: Known properties needed  as per original implementation. This is used in signing and credential resolution appropriately. The properties are maintained in a separate file for clarity

src/aws-cpp-sdk-core/source/smithy/client/AwsSmithyClientBase.cpp
```diff
@@ -552,3 +608,32 @@ void AwsSmithyClientBase::DisableRequestProcessing()
...
+        return  StreamOutcome(Aws::AmazonWebServiceResult<Aws::Utils::Stream::ResponseStream>(
+            httpResponseOutcome.GetResult()->SwapResponseStreamOwnership(),
+            httpResponseOutcome.GetResult()->GetHeaders(), httpResponseOutcome.GetResult()->GetResponseCode()));
+    }
+
+    return StreamOutcome(std::move(httpResponseOutcome));
+}
+
+void AwsSmithyClientBase::AppendToUserAgent(const Aws::String& valueToAppend)
+{
```
sbera87, line 636: without this legacy test cases fail in s3 encryption

tools/code-generation/generator/src/main/java/com/amazonaws/util/awsclientgenerator/generators/cpp/CppClientGenerator.java
```diff
@@ -836,4 +842,105 @@ private void addRequestlessRequestObjectS(final ServiceModel serviceModel) {
...
+    }
+
+    protected void updateAuthSchemesFromEndpointRules(ServiceModel serviceModel, String rawjson)
+    {
+        if(rawjson == null || rawjson.isEmpty())
+        {
+            return;
+        }
+        
+        List<String> authschemes =  new ArrayList<>(serviceModel.getAuthSchemes());
```
sbera87, line 883: Needed to use a copy of list , otherwise it crashes for some services.

generated/src/aws-cpp-sdk-s3/include/aws/s3/S3ClientConfiguration.h
```diff
@@ -58,10 +57,6 @@ namespace Aws
             bool useArnRegion = false;
             Client::AWSAuthV4Signer::PayloadSigningPolicy payloadSigningPolicy = Client::AWSAuthV4Signer::PayloadSigningPolicy::RequestDependent;
             bool disableS3ExpressAuth = false;
-            using IdentityProviderSupplier = std::function<std::shared_ptr<S3ExpressIdentityProvider> (const S3Client &)>;
```
sbera87, line 61: No longer needed here as the provider is passed in the smithy constructor

src/aws-cpp-sdk-core/include/smithy/client/AwsSmithyClientBase.h
```diff
@@ -135,6 +136,13 @@ namespace client
                                             Aws::Http::HttpMethod method,
                                             EndpointUpdateCallback&& endpointCallback) const;
 
+        StreamOutcome MakeRequestWithUnparsedResponse(Aws::AmazonWebServiceRequest const * const request,
```
sbera87, line 139: needed for some operations which dont parse the response. Check s3 client for usage

src/aws-cpp-sdk-core/source/smithy/client/AwsSmithyClientBase.cpp
```diff
@@ -73,10 +84,11 @@ void AwsSmithyClientBase::baseCopyInit() {
   createFromFactoriesIfPresent(m_clientConfig->readRateLimiter, m_clientConfig->configFactories.readRateLimiterCreateFn);
   createFromFactoriesIfPresent(m_clientConfig->telemetryProvider, m_clientConfig->configFactories.telemetryProviderCreateFn);
   if (m_clientConfig && m_clientConfig->retryStrategy) {
-    m_interceptors.emplace_back(Aws::MakeShared<UserAgentInterceptor>(AWS_SMITHY_CLIENT_LOG,
+    m_userAgentInterceptor = Aws::MakeShared<UserAgentInterceptor>(AWS_SMITHY_CLIENT_LOG,
```
sbera87, line 87: non smithy client has this. Smithy client has vector of interceptors but we need the exact instance of derived class type useragent interceptor.  looping to find out which one is the first useragent interceptor will involve casting. One simple approach is to just using one allocated resource and 2 shared ptrs pointing to it in the same object. 
An alternative considered was just to store the index , but even that approach will need dynamic casting to child type. hence discarded.

src/aws-cpp-sdk-core/source/smithy/client/AwsSmithyClientBase.cpp
```diff
@@ -361,15 +409,23 @@ void AwsSmithyClientBase::HandleAsyncReply(std::shared_ptr<AwsSmithyClientAsyncR
...
+          pRequestCtx->m_pRequest->HasEmbeddedError(httpResponse->GetResponseBody(), httpResponse->GetHeaders());
+        };
+        if (Utils::DoesResponseGenerateError(httpResponse))
         {
             AWS_LOGSTREAM_DEBUG(AWS_SMITHY_CLIENT_LOG, "Request returned error. Attempting to generate appropriate error codes from response");
             assert(m_errorMarshaller);
             auto error = m_errorMarshaller->BuildAWSError(httpResponse);
             return HttpResponseOutcome(std::move(error));
         }
+        else if (hasEmbeddedError()) {
```
sbera87, line 423: Follows legacy design. Without this face error marshalling test errors. Embedded error is handled differently that other errors. This invokes client overriden implementation which changes the underlying iostream ptrs as a result of which one fetching the response body more than once sets the fail bit BuildAWSError causes incorrect parsing of the error

tools/code-generation/generator/src/main/resources/com/amazonaws/util/awsclientgenerator/velocity/cpp/smithy/SmithyClientSourceInit.vm
```diff
@@ -84,12 +92,97 @@ ${className}::${className}(const std::shared_ptr<AWSCredentialsProvider>& creden
...
+            {${entry}{Aws::MakeShared<DefaultS3ExpressIdentityResolver>(ALLOCATION_TAG, *this, credentialsProvider ), GetServiceName(), Aws::Region::ComputeSignerRegion(clientConfiguration.region), clientConfiguration.payloadSigningPolicy, false}},
+#else
             {${entry}{ Aws::MakeShared<smithy::AwsCredentialsProviderIdentityResolver>(ALLOCATION_TAG, credentialsProvider), GetServiceName(), clientConfiguration.region}}
+#end
 #end
         })
 {}
 
 /* Legacy constructors due deprecation */
+#if($serviceModel.metadata.serviceId == "S3")
```
sbera87, line 105: s3 has specific signature of legacy, hence  need this

generated/src/aws-cpp-sdk-s3/include/aws/s3/S3Client.h
```diff
@@ -4,14 +4,22 @@
  */
 
 #pragma once
+
```
sbera87, line 7: added for readability

src/aws-cpp-sdk-core/include/smithy/identity/auth/AuthSchemeOption.h
```diff
@@ -23,8 +23,24 @@ namespace smithy {
 
         const char* schemeId = nullptr;
 
-        PropertyBag virtual identityProperties() const { return PropertyBag{}; };
-        PropertyBag virtual signerProperties() const { return PropertyBag{}; };
+        PropertyBag virtual identityProperties() const { return m_identityProperties; };
+        PropertyBag virtual signerProperties() const { return m_signerProperties; };
+        void putIdentityProperty( const Aws::String& key,const Aws::Crt::Variant<Aws::String, bool>& value) {
```
sbera87, line 28: Need this now as the resolution interfaces accepts properties which will need all the parameters that was used previously but now simply respective property bag for identity or signer

sbiscigl (commented):
high level feedback:
* there are some backwards incompatible changes that we cant do
* we shouldnt care about signer in identity provider, we should remove all references to signer in identity provider.
* theres now s3-crt in here, we likely want to do that at the same time.
* s3 components should be scoped to the s3 library and shouldnt leak into core

generated/src/aws-cpp-sdk-s3/source/S3ExpressIdentityResolver.cpp
```diff
@@ -0,0 +1,228 @@
...
+    {
+       if (signerName->second != smithy::S3_EXPRESS_SIGNER_NAME)
+       {
+        if(!m_credsProvider)
+        {
+            AWS_LOGSTREAM_ERROR(S3_EXPRESS_IDENTITY_PROVIDER, "Failed to make S3Express Connect Call")
+            return ResolveIdentityFutureOutcome();
+        }
+        auto creds = m_credsProvider->GetAWSCredentials();
+        return ResolveIdentityFutureOutcome(Aws::MakeUnique<AwsCredentialIdentity>("DefaultAwsCredentialIdentityResolver",  AwsCredentialIdentity{creds.GetAWSAccessKeyId(), creds.GetAWSSecretKey(), creds.GetSessionToken(), creds.GetExpiration()}));
```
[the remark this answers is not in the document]
sbera87: COpy paste error. Will fix

src/aws-cpp-sdk-core/include/smithy/client/AwsSmithyClient.h
```diff
@@ -98,9 +99,9 @@ namespace client
             return *this;
         }
 
-        AwsSmithyClientT (AwsSmithyClientT&&) = default;
+        AwsSmithyClientT (AwsSmithyClientT&& other) = default;
```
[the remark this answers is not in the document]
sbera87: artifact of latest add and remove. Will restore

src/aws-cpp-sdk-core/include/smithy/client/AwsSmithyClient.h
```diff
@@ -162,7 +177,7 @@ namespace client
             if (authSchemeOptionIt != authSchemeOptions.end()) {
                 return SelectAuthSchemeOptionOutcome(*authSchemeOptionIt);
             }
-            return AWSError(Aws::Client::CoreErrors::CLIENT_SIGNING_FAILURE,
+            return Aws::Client::AWSError<CoreErrors>(Aws::Client::CoreErrors::CLIENT_SIGNING_FAILURE,
```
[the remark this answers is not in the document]
sbera87: pipeline warning  fix on one of the platforms

generated/src/aws-cpp-sdk-s3/include/aws/s3/S3ClientConfiguration.h
sbera87: ok, will restore this and reuse it internally

generated/src/aws-cpp-sdk-s3/source/S3ExpressIdentityResolver.cpp
```diff
@@ -0,0 +1,228 @@
...
+               if (value.holds_alternative<Aws::String>())
+               {
+                   return value.get<Aws::String>();
+               }
+               return {};
+           });
+    }
+
+    //if signer name is not s3 express as set in signer properties, get from credential provider
+    auto signerName = params->parameterMap.find(smithy::AUTH_SCHEME_PROPERTY);
```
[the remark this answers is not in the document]
sbera87: ok, made the change in the next commit

src/aws-cpp-sdk-core/include/smithy/client/common/AwsSmithyRequestSigning.h
```diff
@@ -145,7 +146,19 @@ namespace smithy
                     return;
                 }
 
-                auto identityResult = identityResolver->getIdentity(m_targetAuthSchemeOption.identityProperties(), m_targetAuthSchemeOption.identityProperties());
+                //relay service params in additional properties which will be relevant in credential resolution
```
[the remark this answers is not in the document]
sbera87: will remove this once auth scheme based identity resolver added.

src/aws-cpp-sdk-core/source/smithy/client/AwsSmithyClientBase.cpp
```diff
@@ -192,13 +206,20 @@ void AwsSmithyClientBase::MakeRequestAsync(Aws::AmazonWebServiceRequest const* c
     auto epResolutionOutcome = this->ResolveEndpoint(std::move(epParams), std::move(endpointCallback));
     if (!epResolutionOutcome.IsSuccess())
     {
-        pExecutor->Submit([epResolutionOutcome, responseHandler]() mutable
+        auto epOutcome = ResolveEndpointOutcome(Aws::Client::AWSError<Aws::Client::CoreErrors>{
```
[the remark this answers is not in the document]
sbera87: This is needed as it was missing from smithy implementation which our test cases catch . This restores this error validation on endpoint resolution failure

src/aws-cpp-sdk-core/include/smithy/identity/auth/built-in/SigV4MultiAuthResolver.h
```diff
@@ -0,0 +1,45 @@
...
+#include <smithy/identity/auth/AuthSchemeResolverBase.h>
+#include <smithy/identity/auth/built-in/SigV4AuthSchemeOption.h>
+#include <smithy/identity/auth/built-in/SigV4aAuthSchemeOption.h>
+#include <smithy/identity/auth/built-in/S3ExpressSigV4AuthSchemeOption.h>
+#include <aws/core/auth/signer/AWSAuthV4Signer.h>
+#include <smithy/identity/signer/built-in/S3ExpressSigner.h>
+
+namespace smithy {
+    template<typename ServiceAuthSchemeParametersT = DefaultAuthSchemeResolverParameters>
+    class SigV4MultiAuthSchemeResolver : public AuthSchemeResolverBase<ServiceAuthSchemeParametersT>
```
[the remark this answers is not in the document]
sbera87: This is actually the geenric multi auth resolver. What i can do i I can add a s3 class which inherits from this for the s3 additional part. But this is needed in the core as the other generic resolvers are there.

src/aws-cpp-sdk-core/include/smithy/identity/auth/built-in/S3ExpressSigV4AuthScheme.h
```diff
@@ -0,0 +1,88 @@
...
+#include <smithy/identity/auth/AuthSchemeResolverBase.h>
+#include <smithy/identity/identity/AwsCredentialIdentityBase.h>
+#include <smithy/identity/signer/built-in/S3ExpressSigner.h>
+
+
+
+namespace smithy {
+    constexpr char SIGV4_EXPRESS[] = "sigv4-s3express";
+
+    class S3ExpressSigV4AuthScheme : public AuthScheme<AwsCredentialIdentityBase>
```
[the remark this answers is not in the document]
sbera87: All auth schemes are in core already. Putting this in S3 is fine since s3 name in it

src/aws-cpp-sdk-core/include/smithy/identity/signer/built-in/S3ExpressSigner.h
```diff
@@ -0,0 +1,113 @@
...
+
+    template <>
+    struct IsValidS3ExpressSigner<AwsSigV4Signer> : std::true_type {};
+
+    template <>
+    struct IsValidS3ExpressSigner<AwsSigV4aSigner> : std::true_type {};
+
+    //Ensuring S3 Express Signer can use Sigv4 or Sigv4a signing algorithm
+    template <typename BASECLASS>
+    class S3ExpressSigner : public std::enable_if<IsValidS3ExpressSigner<BASECLASS>::value, BASECLASS>::type
```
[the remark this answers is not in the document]
sbera87: Sure, since s3 name in it , I will relocate this back to where it was

tools/scripts/codegen/legacy_c2j_cpp_gen.py
```diff
@@ -19,7 +19,8 @@
 from codegen.model_utils import ServiceModel
 
 SMITHY_SUPPORTED_CLIENTS = [
-    "dynamodb"
+    "dynamodb",
+    "s3"
```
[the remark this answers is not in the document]
sbera87: In a seaprate pr. This hosts the special handling services. Json services which have multi auth and bi-streaming are not enabled. Mult auth clients will be enabled in a separate pr and json bi-streaming support in smithy needs consensus on design

src/aws-cpp-sdk-core/include/smithy/identity/signer/built-in/SignerProperties.h
```diff
@@ -0,0 +1,11 @@
+/**
+ * Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.
+ * SPDX-License-Identifier: Apache-2.0.
```
sbera87, line 3: override properties used for signing. This is updated from endpoints

tests/aws-cpp-sdk-s3-integration-tests/S3ExpressTest.cpp
```diff
@@ -483,4 +506,57 @@ namespace {
...
+      ~S3TestClient(){}
+      private:
+      FRIEND_TEST(S3ExpressTest, ExpressSignerBackwardCompatibility);
+      void overrideS3ExpressSigner()
+      {
+        for(auto& auth : m_authSchemes)
+        {
+          if(auth.first == S3::S3ExpressSigV4AuthSchemeOption::s3ExpressSigV4AuthSchemeOption.schemeId)
+          {
+            auth.second.get<S3::S3ExpressSigV4AuthScheme>().signer() =
```
sbera87, line 532: legacy signer overriden here for demonstrating usage of backwards compatible interface

generated/src/aws-cpp-sdk-s3/source/S3Client.cpp
```diff
@@ -1548,32 +1413,28 @@ GetBucketAnalyticsConfigurationOutcome S3Client::GetBucketAnalyticsConfiguration
...
       request.SetServiceSpecificParameters(
-        [&]() -> std::shared_ptr<Http::ServiceSpecificParameters> {
-          Aws::Map<Aws::String, Aws::String> params;
-          params.emplace("bucketName", request.GetBucket());
-          ServiceSpecificParameters serviceSpecificParameters{params};
-          return Aws::MakeShared<ServiceSpecificParameters>(ALLOCATION_TAG, serviceSpecificParameters);
-        }());
-      return GetBucketAnalyticsConfigurationOutcome(MakeRequest(request, endpointResolutionOutcome.GetResult(), Aws::Http::HttpMethod::HTTP_GET));
+       [&]() -> std::shared_ptr<Http::ServiceSpecificParameters> {
+       Aws::Map<Aws::String, Aws::String> params;
```
sbiscigl, line 1431: we dont need service specific parameters anymore, please remove them, they are legacy and they cause us pain, now that we resolve endpoints in the smithy client we dont need them in generated code anymore

generated/src/aws-cpp-sdk-s3/source/S3Client.cpp
```diff
@@ -149,184 +157,171 @@ namespace Aws
...
+            }()
+        })
+{}
 
 S3Client& S3Client::operator=(const S3Client &rhs) {
     if (&rhs == this) {
       return *this;
     }
-    BASECLASS::operator=(rhs);
-    m_signerProvider = Aws::MakeShared<Aws::Auth::S3ExpressSignerProvider>(ALLOCATION_TAG,
```
sbiscigl, line 171: the reason we had copy constructors and move constructors one S3 was because the S3 express identity provider has a cache, and we dont want copied clients to share the cache with the parent. in this it would appear that they do. we cant do that.

generated/src/aws-cpp-sdk-s3/source/S3ExpressSigner.cpp
```diff
@@ -82,16 +65,78 @@ bool S3ExpressSigner::PresignRequest(Aws::Http::HttpRequest &request,
...
+    {
+        auto identity = m_S3ExpressIdentityProvider->GetS3ExpressIdentity(serviceSpecificParameters);
+        return {identity.getAccessKeyId(), identity.getSecretKeyId()};
+    }
+    return Aws::Auth::AWSCredentials();
+}
+
+//for backwards compatibility
+bool S3ExpressSigner::SignRequest(Aws::Http::HttpRequest& ) const {
+    return false;
```
sbiscigl, line 85: The signer should be signing the request? not returning false? wont this break anyone who was using the signer directly?

generated/src/aws-cpp-sdk-s3/include/aws/s3/S3ExpressSigner.h
```diff
@@ -12,63 +12,133 @@
...
+            std::lock_guard<std::mutex> lock(m_requestProcessing);
+            m_requestsProcessing.erase(requestId);
+        }
+
+        mutable std::set<Aws::String> m_requestsProcessing;
+        mutable std::mutex m_requestProcessing;
+    };
+
+
+    class AWS_S3_API S3ExpressSigner : public Aws::Client::AWSAuthSigner, public S3ExpressSignerBase<smithy::AwsSigV4Signer>{
```
sbiscigl, line 101: previously `S3ExpressSigner` was defined as

`class AWS_S3_API S3ExpressSigner : public Aws::Client::AWSAuthV4Signer`

which means S3ExpressSigner`is a` Aws::Client::AWSAuthV4Signer. is this public interface change not a breaking change?

generated/src/aws-cpp-sdk-s3/include/aws/s3/S3ExpressIdentityProvider.h
```diff
@@ -21,80 +21,81 @@ namespace Aws {
...
-            virtual ~S3ExpressIdentityProvider() {}
-
-        protected:
-            std::shared_ptr<std::mutex> GetMutexForBucketName(const Aws::String& bucketName);
-
-        private:
-            const S3Client &m_s3Client;
-            mutable std::mutex m_bucketNameMapMutex;
-            Aws::Map<Aws::String, std::shared_ptr<std::mutex>> m_bucketNameMutex;
+        class S3ExpressIdentityProvider : public smithy::IdentityResolverBase<smithy::AwsCredentialIdentityBase> {
```
sbiscigl, line 24: the breaks `is a` public inheritance no? i.e. before we were

```
class S3ExpressIdentityProvider: public smithy::IdentityResolverBase<S3ExpressIdentity> 
```

now it is

```
class S3ExpressIdentityProvider : public smithy::IdentityResolverBase<smithy::AwsCredentialIdentityBase>
```

dont we need to implement both public contracts here for backwards compatibility?

sbera87: This method ensures backwards compatibility
 `           GetS3ExpressIdentity(const std::shared_ptr<Aws::Http::ServiceSpecificParameters> &serviceSpecificParameters);`

generated/src/aws-cpp-sdk-s3/include/aws/s3/S3ExpressSigner.h
```diff
@@ -12,63 +12,133 @@
...
+    {
+      smithy::AwsCredentialIdentity operator()(const smithy::AwsCredentialIdentityBase& identity) const {
+        return smithy::AwsCredentialIdentity{identity.accessKeyId(), identity.secretAccessKey(), identity.sessionToken(),identity.expiration()};
+      }
+    };
+
+    template <>
+    struct ExtractIdentity<smithy::AwsSigV4Signer> {
+      auto operator()(const smithy::AwsCredentialIdentityBase& identity) const -> smithy::AwsCredentialIdentity {
+        return smithy::AwsCredentialIdentity{identity.accessKeyId(), identity.secretAccessKey(), {},{}};
```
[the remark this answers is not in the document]
sbera87: If we pass expiration and session token it adds other headers which breaks functionality. This is because the logic :
https://github.com/aws/aws-sdk-cpp/blob/4012ba9cbdd19aa849b0c682c58039e4af6ec58c/src/aws-cpp-sdk-core/source/auth/signer/AWSAuthV4Signer.cpp#L214 
This was needed to remove the check for presence of S3_EXPRESS_HEADER in previous commit.
The same exists in the legacy callflow where the session token is not passed to add session token header for S3 express callflow because session token is already set for s3 express in the signer

generated/src/aws-cpp-sdk-s3/include/aws/s3/S3ExpressIdentityProvider.h
sbiscigl: consider the following code, this would work before your refactor and fail after

```cpp
#include <aws/core/Aws.h>
#include <aws/s3/S3Client.h>
#include <aws/s3/S3ClientConfiguration.h>
#include <aws/s3/S3ExpressIdentity.h>
#include <aws/s3/S3ExpressIdentityProvider.h>

using namespace Aws;
using namespace Aws::Http;
using namespace Aws::Utils;
using namespace Aws::S3;

class MyIdentityProvider : public S3ExpressIdentityProvider {
 public:
  explicit MyIdentityProvider(const S3Client& client):S3ExpressIdentityProvider(client) {}
  ~MyIdentityProvider() override = default;

  S3ExpressIdentity GetS3ExpressIdentity(
    const std::shared_ptr<ServiceSpecificParameters>&
    serviceSpecificParameters) override
  {
   return S3ExpressIdentity{"access_key",
     "secret_key",
     "sessions_token",
     DateTime::Now()};
  }

  ResolveIdentityFutureOutcome getIdentity(
      const IdentityProperties& identityProperties,
      const AdditionalParameters& additionalParameters) override
  {
    return Aws::MakeUnique<S3ExpressIdentity>("log",
      "access_key",
      "secret_key",
      "sessions_token",
      DateTime::Now());
  }
};

auto main() -> int {
  SDKOptions options;
  InitAPI(options);
  {
    S3ClientConfiguration configuration{};
    configuration.identityProviderSupplier = [](const S3Client &client) -> std::shared_ptr<S3ExpressIdentityProvider> {
      AWS_UNREFERENCED_PARAM(client);
      return Aws::MakeShared<MyIdentityProvider>("log", client);
    };
  }
  ShutdownAPI(options);
}
```

sbera87: It won't because I already have implemented
`S3ExpressIdentity GetS3ExpressIdentity(
    const std::shared_ptr<ServiceSpecificParameters>&
    serviceSpecificParameters) `
    This internally already converts from smithy identity class to legacy s3expressidentity class. The internal implementation is different

src/aws-cpp-sdk-core/include/smithy/client/AwsSmithyClient.h
```diff
@@ -150,6 +151,20 @@ namespace client
                     }
                 }
             }
+
+            //resolve endpoint first time to fetch auth schemes
+            Aws::Endpoint::EndpointParameters epParams = ctx.m_pRequest ? ctx.m_pRequest->GetEndpointContextParams() : Aws::Endpoint::EndpointParameters();
+            auto epResolutionOutcome = this->ResolveEndpoint(epParams, [](Aws::Endpoint::AWSEndpoint&){});
```
SergeyRyabinin, line 157: This endpoint resolution call should be called within 
```
m_authSchemeResolver->resolveAuthScheme(identityParams);
```
which type is statically selected during code-generation / compilation time and will contain:
1. endpoint call on multi-auth services;
2. static auth scheme option return on single-auth services.

The (1) is to be generated for S3 with resolveAuthScheme method overwritten.

sbera87: ok, I have implemented the change in resolveAuthScheme  in subsequent commit  in such a way that helps s3 mult auth and any other mult auth client in later prs

generated/src/aws-cpp-sdk-s3/source/S3ExpressSignerProvider.cpp
```diff
@@ -21,12 +21,12 @@ Aws::Auth::S3ExpressSignerProvider::S3ExpressSignerProvider(
         region,
         signingPolicy,
         urlEscapePath) {
-    m_signers.emplace_back(Aws::MakeShared<Aws::S3::S3ExpressSigner>(CLASS_TAG,
+    m_signers.emplace_back(std::static_pointer_cast<Aws::Client::AWSAuthSigner>(Aws::MakeShared<Aws::S3::S3ExpressSigner>(CLASS_TAG,
```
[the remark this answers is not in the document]
sbera87: I will just maintain old and new interfaces in next commit. The above is because just wanted to make old interface use new implementation and inherit from abstract base class. This will be changed.

src/aws-cpp-sdk-core/source/smithy/client/AwsSmithyClientBase.cpp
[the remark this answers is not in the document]
sbera87: `auto epOutcome = ResolveEndpointOutcome(Aws::Client::AWSError<Aws::Client::CoreErrors>{
        Aws::Client::CoreErrors::ENDPOINT_RESOLUTION_FAILURE,  
        epResolutionOutcome.GetError().GetExceptionName(), 
        epResolutionOutcome.GetError().GetMessage(), 
        false});
`
This is the error code expected in resolve endpoint failure. Without this tests fail as expected behavior is broken

generated/src/aws-cpp-sdk-s3/source/S3Client.cpp
sbera87: as discussed, this will have to be a separate PR as this PR ensures same functionality as before but in smithy call flow

generated/src/aws-cpp-sdk-s3/include/aws/s3/S3ExpressSigner.h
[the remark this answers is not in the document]
sbera87: And when this session token is overriden we get 
Error details: HTTP response code: 403
Exception name: AccessDenied
Error message: Access Denied
causing all write operations to fail

src/aws-cpp-sdk-core/include/smithy/client/common/AwsSmithyRequestSigning.h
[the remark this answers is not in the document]
sbera87: as discussed, removing service request params will need to be in a separate PR. This code is needed as this currently relays the bucket name in service request params to identity resolution

generated/src/aws-cpp-sdk-s3/include/aws/s3/S3ClientConfiguration.h
sbiscigl: on the latest commit the above code still break for for backwards compatibility. reproduction in full:

```cpp
#include <aws/core/Aws.h>
#include <aws/s3/S3Client.h>

using namespace Aws;
using namespace Aws::S3;

auto main() -> int {
    SDKOptions options{};
    InitAPI(options);
    {
        S3ClientConfiguration configuration{};
        configuration.identityProviderSupplier =
          [](const S3Client &client) -> std::shared_ptr<S3ExpressIdentityProvider> {
              return Aws::MakeShared<DefaultS3ExpressIdentityProvider>("log_tag", client);
          };
        S3Client client{configuration};
    }
    ShutdownAPI(options);
    return 0;
}
```

generated/src/aws-cpp-sdk-s3/include/aws/s3/S3ExpressSigner.h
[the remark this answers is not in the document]
sbiscigl: OK debugged through this and i think the missing part is in the specification

> Note: S3 Express SessionToken is different from STS SessionToken and X-Amz-Security-Token MUST NOT be set for S3 Express requests when S3 Express Session Credentials are used.

so session token is NOT used, but expiry is, so please set the expiry

Also with the templates and `ExtractIdentity` I dont think there is a need for this level of template indirection in the sigv4 signer itself. would prefer you to just create an extra identity object in sign that drops these values, or a different signer class altogether.

generated/src/aws-cpp-sdk-s3/include/aws/s3/S3ClientConfiguration.h
sbera87: I have to support both the old and new one in this case too, even though this supplier can't be used in the auth schemes in smithy. Added in next commit and added this compilation test as well.

generated/src/aws-cpp-sdk-s3/include/aws/s3/S3ExpressSigner.h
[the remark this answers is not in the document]
sbera87: Thanks for finding the wording in the sep. resolved in next commit.

tools/code-generation/generator/src/main/java/com/amazonaws/util/awsclientgenerator/generators/cpp/CppClientGenerator.java
```diff
@@ -836,4 +847,105 @@ private void addRequestlessRequestObjectS(final ServiceModel serviceModel) {
...
+
+        // parse the JSON into a JsonElement tree
+        JsonElement jsonElement = JsonParser.parseString(rawjson);
+
+        // search for the "authSchemes" field in endpoint rules recursively to get all authschemes
+        List<JsonElement> authSchemes = new ArrayList<>();
+        findNestedField(jsonElement, "authSchemes", authSchemes);
+        // extract authschemes
+        authSchemes.stream()
+        .filter(entry -> entry.isJsonArray()) // check if the element is a JsonArray
```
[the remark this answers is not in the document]
sbera87: sure, this shd be done now

src/aws-cpp-sdk-core/include/smithy/client/AwsSmithyClient.h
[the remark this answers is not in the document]
sbera87: This should be done now

generated/src/aws-cpp-sdk-s3/source/S3ExpressSigner.cpp
sbera87: This is resolved now with old interface as is

src/aws-cpp-sdk-core/include/smithy/identity/auth/built-in/SigV4MultiAuthResolver.h
```diff
@@ -0,0 +1,73 @@
...
+#pragma once
+
+#include <aws/core/auth/signer/AWSAuthV4Signer.h>
+#include <aws/core/endpoint/EndpointParameter.h>
+#include <smithy/client/AwsSmithyClientBase.h>
+#include <smithy/identity/auth/AuthSchemeResolverBase.h>
+#include <smithy/identity/auth/built-in/SigV4AuthSchemeOption.h>
+#include <smithy/identity/auth/built-in/SigV4aAuthSchemeOption.h>
+
+#include "SigV4AuthScheme.h"
```
sbiscigl, line 14: no quote imports

sbiscigl (approved):
Fix nit and ship

src/aws-cpp-sdk-core/include/smithy/identity/auth/built-in/SigV4AuthScheme.h
```diff
@@ -36,6 +36,20 @@ namespace smithy {
             assert(m_signer);
         }
 
+        //delegate constructor
```
SergeyRyabinin, line 39: nit, no need to fix: this does not sound like a delegating constructor.

SergeyRyabinin (approved):

Issue #3264, opened 2025-01-22
RDS - DescribeDBClusterEndpointsRequest with Filters returns InvalidParameterCombination error


karenc-bq:
### Describe the bug

Sending a RDS [DescribeDBClusterEndpointsRequest](https://sdk.amazonaws.com/cpp/api/LATEST/aws-cpp-sdk-rds/html/class_aws_1_1_r_d_s_1_1_model_1_1_describe_d_b_cluster_endpoints_request.html) with a Filter returns `InvalidParameterCombination - The values list cannot be null for the filter db-cluster-endpoint-type.`

```cpp
Aws::RDS::Model::Filter filter;
filter.SetName("db-cluster-endpoint-type");
filter.SetValues({"CUSTOM"});

Aws::RDS::Model::DescribeDBClusterEndpointsRequest request;
request.SetFilters({filter});
```
When inspecting the `DescribeDBClusterEndpointsRequest` object in debug mode `m_values` has a size of 1 so it is not empty.
Please let me know if I am not using Filter correctly, thanks in advance.

### Regression Issue

- [ ] Select this option if this issue appears to be a regression.

### Expected Behavior

The custom endpoint matching the filter is returned.

### Current Behavior

`InvalidParameterCombination` error is returned.

### Reproduction Steps

1. Create a custom endpoint with 1 instance on RDS
2. Create a `DescribeDBClusterEndpointsRequest `
```cpp
Aws::RDS::Model::Filter filter;
filter.SetName("db-cluster-endpoint-type");
filter.SetValues({"CUSTOM"});

Aws::RDS::Model::DescribeDBClusterEndpointsRequest request;
request.SetFilters({filter});
const auto response = client->DescribeDBClusterEndpoints(request);
```

Got the same results with the following filter.
```
Aws::RDS::Model::Filter filter;
filter.SetName("db-cluster-endpoint-status");
filter.AddValues("available");
```

### AWS CPP SDK version used

1.11.488

### Compiler and Version used

9.2.0

### Operating System and version

Issue seem on both Windows and Ubuntu

sbiscigl:
replicated and confirmed its a bug in our serialization of header lists in xml

we send
```
Action=DescribeDBClusterEndpoints
  &Filters.member.1.Name=db-cluster-endpoint-type
  &Filters.member.1.Value.1=CUSTOM
  &Version=2014-10-31
```

should be sending
```
Action=DescribeDBClusterEndpoints
[...]

PR #3266, opened 2025-01-22
Fix queryxml serialization

sbiscigl:
*Issue #, if available:*

https://github.com/aws/aws-sdk-cpp/issues/3264

*Description of changes:*

There was a error with queryxml serialization that broke serialization for several APIs. name that when when using the xmlName trait

smithy docs: https://smithy.io/2.0/aws/protocols/aws-query-protocol.html

> Each value provided in the list is serialized as a separate key with a "." separator, the string "member", a "." separator, and a "1" indexed incrementing counter appended to the container's key. The xmlName trait can be used to serialize a property using a custom name instead of "member". The xmlFlattened trait can be used to unwrap the values into a containing structure or union, with the key not containing the initial "." separator and member segment.

specifically the line that says 

>The xmlName trait can be used to serialize a property using a custom name instead of "member"

before this fix we would always serialize with the name "member" instead of this location name. now we will use the xmlName if present.

*Check all that applies:*

- [x] Linux
- [x] Windows
- [x] Android
- [x] MacOS
- [ ] IOS
- [ ] Other Platforms


SergeyRyabinin (approved):

tests/aws-cpp-sdk-rds-integration-tests/RDSTest.cpp
```diff
@@ -302,4 +304,14 @@ namespace
         preSignedUrl = ExtractPreSignedUrlFromPayload(TestingMonitoringMetrics::s_lastPayload.c_str());
         ASSERT_STREQ(TESTING_PRESIGNED_URL, preSignedUrl.c_str());
     }
+
+    TEST_F(RDSTest, ShouldDescribeClusterEndpoots) {
+        DescribeDBClusterEndpointsRequest request{};
```
sbera87, line 309: I think we can also add check for the expected serialized filters since the format changed
For example, you can invoke
`request.SerializePayload()` and test against expected result

sbera87 (approved):

sbera87 (approved):

PR #3269, opened 2025-01-27
[ProtocolTests] #1 Protocol tests generation stub and cmake

SergeyRyabinin:
*Issue #, if available:*
Protocol tests
*Description of changes:*
Adds generation of stubs (empty dummy projects) for protocol tests.
Also, generates the cmake script and wires them to a corresponding test client.
Build of protocol tests can be enabled in cmake by adding `-DENABLE_PROTOCOL_TESTS=ON` (OFF by default).
However, if attempted to build tests, the build fail because of some issues that these tests uncover, I will cover actual tests generation and build issues fixes in the following PRs.

Tested by re-generating all and building.

The generated tests will look like:
```
generated/protocol-tests/tests
├── input
│   ├── ec2
│   │   ├── CMakeLists.txt
│   │   └── RunTests.cpp
│   ├── json
│   │   ├── CMakeLists.txt
│   │   └── RunTests.cpp
│   ├── json_1_0
│   │   ├── CMakeLists.txt
│   │   └── RunTests.cpp
│   ├── query
│   │   ├── CMakeLists.txt
│   │   └── RunTests.cpp
│   ├── rest-json
│   │   ├── CMakeLists.txt
│   │   └── RunTests.cpp
│   ├── rest-xml
│   │   ├── CMakeLists.txt
│   │   └── RunTests.cpp
│   └── rest-xml-1
│       ├── CMakeLists.txt
│       └── RunTests.cpp
└── output
    ├── ec2
    │   ├── CMakeLists.txt
    │   └── RunTests.cpp
    ├── json
    │   ├── CMakeLists.txt
    │   └── RunTests.cpp
    ├── json_1_0
    │   ├── CMakeLists.txt
    │   └── RunTests.cpp
    ├── query
    │   ├── CMakeLists.txt
    │   └── RunTests.cpp
    ├── rest-json
    │   ├── CMakeLists.txt
    │   └── RunTests.cpp
    ├── rest-xml
    │   ├── CMakeLists.txt
    │   └── RunTests.cpp
    └── rest-xml-1
        ├── CMakeLists.txt
        └── RunTests.cpp
```

*Check all that applies:*

- [x] Linux
- [x] Windows
- [ ] Android
- [ ] MacOS
- [ ] IOS
- [ ] Other Platforms


tools/code-generation/generator/src/main/java/com/amazonaws/util/awsclientgenerator/domainmodels/c2j_protocol_test/C2jGiven.java
```diff
@@ -0,0 +1,33 @@
...
+ */
+
+package com.amazonaws.util.awsclientgenerator.domainmodels.c2j_protocol_test;
+
+import lombok.Data;
+
+@Data
+public class C2jGiven {
+    @Data
+    public class C2jGivenHttp {
```
sbiscigl, line 13: nit: could be static class i believe

tools/code-generation/generator/src/main/java/com/amazonaws/util/awsclientgenerator/domainmodels/c2j_protocol_test/C2jOutputResponse.java
```diff
@@ -0,0 +1,17 @@
...
+
+package com.amazonaws.util.awsclientgenerator.domainmodels.c2j_protocol_test;
+
+import lombok.Data;
+
+import java.util.Map;
+
+@Data
+public class C2jOutputResponse {
+    Integer status_code;
```
sbiscigl, line 14: nit: use stack `int` instead of heap `Integer` since we're not making a collection

tools/code-generation/generator/src/main/java/com/amazonaws/util/awsclientgenerator/generators/DirectFromC2jGenerator.java
```diff
@@ -168,4 +176,39 @@ public ByteArrayOutputStream generateServiceTestSourceFromModels(String rawJson,
...
+     * A function to generate C++ source for service client tests
+     *
+     * @throws Exception
+     */
+    public ByteArrayOutputStream generateProtocolTestSourceFromModels(String c2jModelJson, String protocolTestsJson,
+                                                                      String protocolTestsType, String protocolTestsName,
+                                                                      String serviceName) throws Exception {
+        GsonBuilder gsonBuilder = new GsonBuilder();
+        gsonBuilder.registerTypeAdapter(C2jXmlNamespace.class, new C2jXmlNamespaceDeserializer());
+        gsonBuilder.registerTypeAdapter(JsonNode.class, new JsonNodeDeserializer());
```
sbiscigl, line 190: do we actually need `JsonNodeDeserializer` and a type adapter registered? this pattern is kind of weird. we use a jackson type adapter in gson. we should really only be using one or the other. is there any reason we actually need this type adapter? from what i can tell we dont do any special serialization we just call object mapper, so I'm pretty sure we dont actually need this adapter to exist.

tools/code-generation/generator/src/main/java/com/amazonaws/util/awsclientgenerator/generators/cpp/CppProtocolTestGenerator.java
```diff
@@ -0,0 +1,105 @@
...
+import com.amazonaws.util.awsclientgenerator.domainmodels.codegeneration.ServiceModel;
+
+import com.amazonaws.util.awsclientgenerator.generators.ClientGenerator;
+import com.amazonaws.util.awsclientgenerator.generators.exceptions.SourceGenerationFailedException;
+import org.apache.velocity.Template;
+import org.apache.velocity.VelocityContext;
+import org.apache.velocity.app.VelocityEngine;
+import org.apache.velocity.runtime.RuntimeConstants;
+import org.apache.velocity.runtime.resource.loader.ClasspathResourceLoader;
+import org.slf4j.helpers.NOPLoggerFactory;
```
sbiscigl, line 20: doesnt seem like we use this import

tools/code-generation/generator/src/main/java/com/amazonaws/util/awsclientgenerator/generators/cpp/CppProtocolTestGenerator.java
```diff
@@ -0,0 +1,105 @@
...
+        context.put("output.encoding", StandardCharsets.UTF_8.name());
+        return context;
+    }
+
+    protected final SdkFileEntry makeFile(Template template, VelocityContext context, String path, boolean needsBOM) throws IOException {
+        StringWriter sw = new StringWriter();
+        template.merge(context, sw);
+
+        try {
+            sw.close();
```
sbiscigl, line 91: you should be using the [try with resources pattern](https://docs.oracle.com/javase/tutorial/essential/exceptions/tryResourceClose.html) here i.e.

```
try(final StringWriter sw = new StringWriter();) {
    //...do stuff with sw
}

```

sbiscigl (approved):

tools/code-generation/generator/src/main/java/com/amazonaws/util/awsclientgenerator/generators/cpp/CppProtocolTestGenerator.java
SergeyRyabinin: we actually use
```
velocityEngine.addProperty(RuntimeConstants.RUNTIME_LOG_INSTANCE, new NOPLoggerFactory().getLogger(""));
```

tools/code-generation/generator/src/main/java/com/amazonaws/util/awsclientgenerator/generators/DirectFromC2jGenerator.java
SergeyRyabinin: in protocol tests, there is a test suite definition, with a pre-defined schema which is known and can be coded into a known object model tree.
However, the input/output parameters definition is not known and evaluated only during runtime:

https://github.com/aws/aws-sdk-cpp/blob/main/tools/code-generation/protocol-tests/output/json_1_0.json#L882-L888
The tests are design to hold an arbitrary json, such as:
```
                "result": {
                    "stringValue": "string",
                    "documentValue": [
                        true,
                        false
                    ]
                },
```
and it will be different per each test/operation.
My plan is to not parse this structure into an object tree, but evaluate during the runtime of test generation.

I've chosen JsonNode because I've already used it on a similar project.
Essentially, I need just some `duck-typed dict()` class here, and JsonNode would suite it.
Also, having gson to parse the known Json structure and Jackson JsonNode is a kind of mental distinguishing exercise here.

sbiscigl (approved):

sbera87 (approved):
lgtm

Issue #3282, opened 2025-01-31
Deadlock on calling Aws::ShutdownAPI via s_registryMutex


Markus87:
### Describe the bug

It seems the issue does not always occur.
<details><summary>Aws::ShutdownAPI holding s_registryMutex and joining thread who needs same mutex to finish</summary>

```
 	[Waiting on Thread 4440, double-click or press enter to switch to thread]	
 	ntdll.dll!00007ffccf5f0054()	Unknown
 	KERNELBASE.dll!00007ffccb930f33()	Unknown
 	msvcp140.dll!00007ffcafaf255f()	Unknown
 	[Inline Frame] aws-cpp-sdk-core.dll!std::thread::join() Line 133	C++
 	aws-cpp-sdk-core.dll!Aws::Utils::Threading::DefaultExecutor::~DefaultExecutor() Line 79	C++
 	aws-cpp-sdk-core.dll!Aws::Utils::Threading::DefaultExecutor::`vector deleting destructor'(unsigned int)	C++
 	[Inline Frame] aws-cpp-sdk-s3.dll!std::_Ref_count_base::_Decref() Line 1163	C++
 	[Inline Frame] aws-cpp-sdk-s3.dll!std::_Ptr_base<Aws::Utils::Threading::Executor>::_Decref() Line 1380	C++
 	[Inline Frame] aws-cpp-sdk-s3.dll!std::shared_ptr<Aws::Utils::Threading::Executor>::{dtor}() Line 1685	C++
 	[Inline Frame] aws-cpp-sdk-s3.dll!std::shared_ptr<Aws::Utils::Threading::Executor>::reset() Line 1732	C++
 	aws-cpp-sdk-s3.dll!Aws::Client::ClientWithAsyncTemplateMethods<Aws::S3::S3Client>::ShutdownSdkClient(void * pThis, __int64 timeoutMs) Line 121	C++
 	aws-cpp-sdk-core.dll!Aws::Utils::ComponentRegistry::TerminateAllComponents() Line 88	C++
>	aws-cpp-sdk-core.dll!Aws::ShutdownAPI(const Aws::SDKOptions & options) Line 207	C++
 	[Inline Frame] Bfx.Abstract.AWSEngine.dll!BitFactory::AWS::Initializer::{dtor}() Line 59	C++
[...]

PR #3288, opened 2025-02-05
Fix DefaultExecutor destruction deadlock if worker thread owns the Executor

SergeyRyabinin:
*Issue #, if available:*
https://github.com/aws/aws-sdk-cpp/issues/3282
*Description of changes:*
Built-in SDK thread executors to handle the case when one of their workers/tasks is the only/final owner of the executor class.
Avoids the deadlock on the cyclic destruction.

Also, additionally, force log flush after FATAL log event has been logged to avoid logs buffering.

*Check all that applies:*

- [x] Linux
- [x] Windows
- [ ] Android
- [ ] MacOS
- [ ] IOS
- [ ] Other Platforms

### DIFF
--- a/src/aws-cpp-sdk-core/include/aws/core/utils/logging/LogMacros.h
+++ b/src/aws-cpp-sdk-core/include/aws/core/utils/logging/LogMacros.h
@@ -77,8 +77,10 @@
         }
 
     #define AWS_LOGSTREAM_FATAL(tag, streamExpression) \
+    do { \
       AWS_LOGSTREAM(Aws::Utils::Logging::LogLevel::Fatal, tag, streamExpression) \
-      AWS_LOG_FLUSH()
+      AWS_LOG_FLUSH() \
+    } while(0)
     #define AWS_LOGSTREAM_ERROR(tag, streamExpression)  AWS_LOGSTREAM(Aws::Utils::Logging::LogLevel::Error, tag, streamExpression)
     #define AWS_LOGSTREAM_WARN(tag,  streamExpression)  AWS_LOGSTREAM(Aws::Utils::Logging::LogLevel::Warn,  tag, streamExpression)
     #define AWS_LOGSTREAM_INFO(tag,  streamExpression)  AWS_LOGSTREAM(Aws::Utils::Logging::LogLevel::Info,  tag, streamExpression)
--- a/src/aws-cpp-sdk-core/include/aws/core/utils/logging/LogMacros.h
+++ b/src/aws-cpp-sdk-core/include/aws/core/utils/logging/LogMacros.h
@@ -76,7 +76,9 @@
             } \
         }
 
-    #define AWS_LOGSTREAM_FATAL(tag, streamExpression)  AWS_LOGSTREAM(Aws::Utils::Logging::LogLevel::Fatal, tag, streamExpression)
+    #define AWS_LOGSTREAM_FATAL(tag, streamExpression) \
+      AWS_LOGSTREAM(Aws::Utils::Logging::LogLevel::Fatal, tag, streamExpression) \
+      AWS_LOG_FLUSH()
     #define AWS_LOGSTREAM_ERROR(tag, streamExpression)  AWS_LOGSTREAM(Aws::Utils::Logging::LogLevel::Error, tag, streamExpression)
     #define AWS_LOGSTREAM_WARN(tag,  streamExpression)  AWS_LOGSTREAM(Aws::Utils::Logging::LogLevel::Warn,  tag, streamExpression)
     #define AWS_LOGSTREAM_INFO(tag,  streamExpression)  AWS_LOGSTREAM(Aws::Utils::Logging::LogLevel::Info,  tag, streamExpression)
--- a/src/aws-cpp-sdk-core/include/aws/core/utils/threading/DefaultExecutor.h
+++ b/src/aws-cpp-sdk-core/include/aws/core/utils/threading/DefaultExecutor.h
@@ -31,6 +31,28 @@ namespace Aws
 
                 void WaitUntilStopped() override;
             protected:
+                class DefaultExecutorTask {
+                public:
+                  DefaultExecutorTask(std::function<void()>&& task, DefaultExecutor* executor);
+                  DefaultExecutorTask(const DefaultExecutorTask&) = delete;
+                  DefaultExecutorTask& operator=(const DefaultExecutorTask&) = delete;
+                  DefaultExecutorTask(DefaultExecutorTask&&) = default;
+
+                  /**
+                   * Detaches the task from the executor
+                   */
+                  void DoNotDetach();
+                  /**
+                   * Starts task execution on a new thread
+                   */
+                  static std::pair<std::thread, DefaultExecutorTask*> Launch(DefaultExecutorTask* task);
+                private:
+                  void Execute();
+
+                  std::function<void()> m_task;
+                  DefaultExecutor* m_executor = nullptr;
+                };
+
                 enum class State
                 {
                     Free, Locked, Shutdown
@@ -38,7 +60,9 @@ namespace Aws
                 bool SubmitToThread(std::function<void()>&&) override;
                 void Detach(std::thread::id id);
                 std::atomic<State> m_state;
-                Aws::UnorderedMap<std::thread::id, std::thread> m_threads;
+
+                using DefaultExecutorTaskPair = std::pair<std::thread, DefaultExecutorTask*>;
+                Aws::UnorderedMap<std::thread::id, DefaultExecutorTaskPair> m_tasks;
             };
         } // namespace Threading
     } // namespace Utils
--- a/src/aws-cpp-sdk-core/include/aws/core/utils/threading/DefaultExecutor.h
+++ b/src/aws-cpp-sdk-core/include/aws/core/utils/threading/DefaultExecutor.h
@@ -5,65 +5,40 @@
 
 #pragma once
 
+#include <aws/core/utils/memory/stl/AWSMap.h>
 #include <aws/core/utils/threading/Executor.h>
 
-#include <aws/core/utils/memory/stl/AWSMap.h>
-
-#include <atomic>
 #include <functional>
-#include <mutex>
-#include <thread>
 
-namespace Aws
-{
-    namespace Utils
-    {
-        namespace Threading
-        {
-            /**
-            * Default Executor implementation. Simply spawns a thread and detaches it.
-            */
-            class AWS_CORE_API DefaultExecutor : public Executor
-            {
-            public:
-                DefaultExecutor() : m_state(State::Free) {}
-                ~DefaultExecutor();
+namespace Aws {
+namespace Utils {
+namespace Threading {
+/**
+ * Default Executor implementation. Simply spawns a thread and detaches it.
+ */
+class AWS_CORE_API DefaultExecutor : public Executor {
+  // API contract
+ public:
+  DefaultExecutor();
+  DefaultExecutor(const DefaultExecutor& other);
+  DefaultExecutor& operator=(const DefaultExecutor&);
+  DefaultExecutor(DefaultExecutor&& other) = default;
+  DefaultExecutor& operator=(DefaultExecutor&&) = default;
 
-                void WaitUntilStopped() override;
-            protected:
-                class DefaultExecutorTask {
-                public:
-                  DefaultExecutorTask(std::function<void()>&& task, DefaultExecutor* executor);
-                  DefaultExecutorTask(const DefaultExecutorTask&) = delete;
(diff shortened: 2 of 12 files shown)


tests/aws-cpp-sdk-core-tests/utils/threading/DefaultExecutorTest.cpp
```diff
@@ -42,3 +42,72 @@ TEST_F(DefaultExecutorTest, ThreadsDetachIfNotShuttingDown)
...
+      cv.notify_one();
+    }
+  });
+
+  if (!taskStarted) {
+    std::unique_lock<std::mutex> lock(mtx);
+    cv.wait_for(lock, std::chrono::seconds(60), [&taskStarted]() { return taskStarted; });
+  }
+  ASSERT_TRUE(taskStarted) << "Async task has not started within 60 seconds!";
+  if (!taskStarted) {
```
sbera87, line 88: if task has not started the assertion will end the test in the previous line, so  I don't think  this is necessary

src/aws-cpp-sdk-core/include/aws/core/utils/threading/DefaultExecutor.h
```diff
@@ -31,14 +31,39 @@ namespace Aws
...
+                  DefaultExecutorTask& operator=(const DefaultExecutorTask&) = delete;
+                  DefaultExecutorTask(DefaultExecutorTask&&) = default;
+
+                  void Execute() {
+                    assert(m_task);
+                    m_task();
+                    if (m_executor) {
+                      m_executor->Detach(std::this_thread::get_id());
+                    }
+                    Aws::Delete(this);
```
sbiscigl, line 50: im not sure i love this because a segfault for calling execute twice seems like a bad failure mode for this. could we maybe place a level indirection above the task that prevents it from being called twice ever?

tests/aws-cpp-sdk-core-tests/utils/threading/DefaultExecutorTest.cpp
```diff
@@ -42,3 +42,72 @@ TEST_F(DefaultExecutorTest, ThreadsDetachIfNotShuttingDown)
...
+    }
+
+    if (!taskCanContinue) {
+      std::unique_lock<std::mutex> lock(mtx);
+      cv.wait_for(lock, std::chrono::seconds(60), [&taskCanContinue]() { return taskCanContinue; });
+    }
+
+    ASSERT_TRUE(taskCanContinue) << "Async task has not been allowed to continue withing 60 seconds!";
+    ASSERT_TRUE(pExec);
+    ASSERT_EQ(1, pExec.use_count());
```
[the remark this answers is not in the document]
sbera87: Ok I see it waits for another signal where one reference count will be decreased

tests/aws-cpp-sdk-core-tests/utils/threading/DefaultExecutorTest.cpp
SergeyRyabinin: But it may let the abandoned thread to be run, resulting in the test application continuing to run.
This is just a safeguard from stuck CI jobs.

src/aws-cpp-sdk-core/include/aws/core/utils/logging/LogMacros.h
```diff
@@ -76,7 +76,9 @@
             } \
         }
 
-    #define AWS_LOGSTREAM_FATAL(tag, streamExpression)  AWS_LOGSTREAM(Aws::Utils::Logging::LogLevel::Fatal, tag, streamExpression)
+    #define AWS_LOGSTREAM_FATAL(tag, streamExpression) \
```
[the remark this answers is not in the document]
SergeyRyabinin: The whole purpose of flushing is to flush buffered log lines.
The whole purpose of buffering is to increase the performance.

If we flush on every statement - then buffering is not needed and also logging performance will be low.

The idea here is that SDK may crash somewhere near FATAL log statements, and in this case we prefer to not lose buffered logs over performance.

src/aws-cpp-sdk-core/include/aws/core/utils/threading/DefaultExecutor.h
SergeyRyabinin: I has the same thought... if you think about it more, after the task has been destructed, you cannot trust anything under `this->`, as it is already destroyed.

The indirection here is 
```
using DefaultExecutorTaskPair = std::pair<std::thread, DefaultExecutorTask*>;
Aws::UnorderedMap<std::thread::id, DefaultExecutorTaskPair> m_tasks;
```

I.e. the indirection is thread_id used as a key.
We need this structure to *update* the task *after* it has been launched, so we need to have access to the task context from the Executor, but also we need to allow the Task to self-destroy.
There is another potential issue: Executor cannot access/update any DefaultExecutorTask (this will create a race condition), it can do it if being called from the same thread of the Task.

tests/aws-cpp-sdk-core-tests/utils/threading/PooledThreadExecutorTest.cpp
```diff
@@ -0,0 +1,83 @@
...
+
+class PooledThreadExecutorTest : public Aws::Testing::AwsCppSdkGTestSuite
+{
+};
+
+TEST_F(PooledThreadExecutorTest, WorkerThreadTheOnlyOwner)
+{
+  // If somehow the shared_ptr of the Executor gets owned by a worker thread of that Executor - Executor shutdown must not deadlock
+  auto pExec = Aws::MakeShared<PooledThreadExecutor>("WorkerThreadTheOnlyOwner", 4);
+
```
sbera87, line 20: Food for thought: Is it possible to simplify the test a little more for ease of maintenance and debug? Conceptually I get that you want to test the detached worker thread to solely own the ptr to executor and then shut  it down. Instead of signally back and forth in between the main and detached thread, can we not just signal from the main thread to the detached thread once .  The main thread creates and submit task to detached thread, resets its ownership and just wait for test case finish. The worker thread just waits till the signal , ensures sole ownership and shuts down.  So main thread has one notify, one wait. Worker thread has one wait and then one notify.

src/aws-cpp-sdk-core/include/aws/core/utils/threading/DefaultExecutor.h
```diff
@@ -31,14 +31,39 @@ namespace Aws
...
+                    : m_task(std::move(task)), m_executor(executor) {
+                    assert(m_task);
+                  }
+
+                  DefaultExecutorTask(const DefaultExecutorTask&) = delete;
+                  DefaultExecutorTask& operator=(const DefaultExecutorTask&) = delete;
+                  DefaultExecutorTask(DefaultExecutorTask&&) = default;
+
+                  void Execute() {
+                    assert(m_task);
```
sbera87, line 45: will m_task ever be null here? its already validated  in the constructor where it was moved to.

src/aws-cpp-sdk-core/include/aws/core/utils/threading/DefaultExecutor.h
```diff
@@ -31,14 +31,39 @@ namespace Aws
 
                 void WaitUntilStopped() override;
             protected:
+                struct DefaultExecutorTask {
+                  DefaultExecutorTask(std::function<void()>&& task, DefaultExecutor* executor)
```
sbera87, line 35: explicit is needed right? else DefaultExecutorTask() can cause issue in execute

src/aws-cpp-sdk-core/include/aws/core/utils/threading/DefaultExecutor.h
```diff
@@ -31,14 +31,39 @@ namespace Aws
...
+                    assert(m_task);
+                    m_task();
+                    if (m_executor) {
+                      m_executor->Detach(std::this_thread::get_id());
+                    }
+                    Aws::Delete(this);
+                  }
+
+                  std::function<void()> m_task;
+                  DefaultExecutor* m_executor = nullptr;
```
sbera87, line 54: should we not use reference to shared ptr and then use weak ptr to access it for more thread safety

src/aws-cpp-sdk-core/include/aws/core/utils/threading/DefaultExecutor.h
SergeyRyabinin: That's a very good comment!
The only issue is that...... the owning executor class is not aware if it is inside shared pointer or an independent object.

tests/aws-cpp-sdk-core-tests/utils/threading/PooledThreadExecutorTest.cpp
SergeyRyabinin: I wrapped synchronization blocks into `sendSignal` and `waitForSignal` functions, it should make the test more readable.

> can we not just signal from the main thread to the detached thread once . The main thread creates and submit task to detached thread, resets its ownership and just wait for test case finish

This creates a race condition in the test, potentially not testing the intended idea.
By not waiting in the task, there a chance that async task will finish before main thread releases the executor.
The current synchronization logic always forces the tested scenario.

src/aws-cpp-sdk-core/include/aws/core/utils/threading/DefaultExecutor.h
```diff
@@ -5,41 +5,40 @@
...
+
+ protected:
+  bool SubmitToThread(std::function<void()>&&) override;
+
+  // implementation details
+ public:
+  struct impl;
+
+ private:
+  std::shared_ptr<impl> pImpl;
```
sbiscigl, line 40: nit: why not unique ptr

tests/aws-cpp-sdk-core-tests/utils/threading/PooledThreadExecutorTest.cpp
sbera87: I would imagine that after the worker thread is started (in immediate dispatch mode) then the race condition u mentioned can be avoided but sure I see the point why its complex

sbera87 (approved):

tests/aws-cpp-sdk-core-tests/utils/threading/PooledThreadExecutorTest.cpp
SergeyRyabinin: Well, computers/implementations are different, and "immediate" is surprisingly in fact not always immediate.

src/aws-cpp-sdk-core/source/utils/threading/DefaultExecutor.cpp
```diff
@@ -3,80 +3,187 @@
...
 }
 
-DefaultExecutor::~DefaultExecutor()
-{
+DefaultExecutor& DefaultExecutor::operator=(const DefaultExecutor& other) {
+  if (this != &other) {
     WaitUntilStopped();
+    pImpl.reset();
+  }
+  return *this;
```
SergeyRyabinin, line 156: the executor is in invalid state after this method, need to re-create impl.

src/aws-cpp-sdk-core/source/utils/threading/DefaultExecutor.cpp
```diff
@@ -3,80 +3,187 @@
...
-        }
-    } 
-    while(expected != State::Shutdown);
+bool DefaultExecutor::impl::SubmitToThread(const std::shared_ptr<impl>& spThis, std::function<void()>&& fx) {
+  if (State::Shutdown == m_state.load()) {
+    AWS_LOGSTREAM_ERROR(DEFAULT_EXECUTOR_LOG_TAG, "Unable to submit async task: the executor is shut down!");
+    return false;
+  }
+
+  auto* task = Aws::New<DefaultExecutorTask>(DEFAULT_EXECUTOR_LOG_TAG, std::move(fx), spThis);
```
SergeyRyabinin, line 100: If Executor is in Shutdown after this line - it will be a memory leak.

src/aws-cpp-sdk-core/source/utils/threading/DefaultExecutor.cpp
```diff
@@ -3,80 +3,187 @@
...
+      return {};
+    }
+    return {std::thread(&DefaultExecutorTask::Execute, task), task};
+  }
+
+ private:
+  void Execute() {
+    assert(m_task);
+    m_task();
+    if (const auto spExecutor = m_executor.lock()) {
```
SergeyRyabinin, line 70: if m_executor pimpl is being destructed, the lock() will return null
and `Aws::Delete(this)` will result in Detach being skipped, then this task is deleted

then the destructor of pimpl may try to call because the task is still present on the task map.
```
taskItem.second.second->DoNotDetach();
```

sbera87 (approved):

PR #3291, opened 2025-02-10
Replace dashes with underscore for member var names

SergeyRyabinin:
*Issue #, if available:*
n/a
*Description of changes:*
Replace - with _
*Check all that applies:*

- [ ] Linux
- [ ] Windows
- [ ] Android
- [ ] MacOS
- [ ] IOS
- [ ] Other Platforms

### DIFF
--- a/tools/code-generation/generator/src/main/java/com/amazonaws/util/awsclientgenerator/domainmodels/codegeneration/cpp/CppViewHelper.java
+++ b/tools/code-generation/generator/src/main/java/com/amazonaws/util/awsclientgenerator/domainmodels/codegeneration/cpp/CppViewHelper.java
@@ -140,7 +140,7 @@ public class CppViewHelper {
     }
 
     public static String computeVariableName(String memberName) {
-        return memberName.substring(0, 1).toLowerCase() + memberName.substring(1);
+        return lowercasesFirstChar(memberName).replace("-", "_");
     }
 
     public static String convertToUpperCamel(String lowerCamel) {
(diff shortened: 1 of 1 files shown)


tools/code-generation/generator/src/main/java/com/amazonaws/util/awsclientgenerator/domainmodels/codegeneration/cpp/CppViewHelper.java
```diff
@@ -140,7 +140,7 @@ public static String computeDefaultValue(Shape shape) {
     }
 
     public static String computeVariableName(String memberName) {
-        return memberName.substring(0, 1).toLowerCase() + memberName.substring(1);
+        return lowercasesFirstChar(memberName).replace("-", "_");
```
sbiscigl, line 143: instead of underscores/string manipulation should we just copy what we do for [upper camel case](https://github.com/aws/aws-sdk-cpp/blob/main/tools/code-generation/generator/src/main/java/com/amazonaws/util/awsclientgenerator/domainmodels/codegeneration/cpp/CppViewHelper.java#L146-L148), but make it for lower camel case? since its _only variable_ names shouldn't be publicly impacting, but it could make a nasty delta.

sbiscigl (approved):

tools/code-generation/generator/src/main/java/com/amazonaws/util/awsclientgenerator/domainmodels/codegeneration/cpp/CppViewHelper.java
SergeyRyabinin: well, the issue with `CaseFormat` is that we need to know the source case convention, but we don't know it.

sbera87 (approved):

tools/code-generation/generator/src/main/java/com/amazonaws/util/awsclientgenerator/domainmodels/codegeneration/cpp/CppViewHelper.java
sbiscigl: wouldn't it always be lower camelcase?

SergeyRyabinin: we don't control it, it comes from the model file.

Issue #3008, opened 2024-06-18
WinHttp TLS is completely disabled if m_verifySSL is false


renjipanicker:
### Describe the bug

In the file;:
`aws-cpp-sdk-core/source/http/windows/WinHttpSyncHttpClient.cpp`

Line 541 (in function OpenRequest):
`DWORD requestFlags = request->GetUri().GetScheme() == Scheme::HTTPS && m_verifySSL ? WINHTTP_FLAG_SECURE : 0;`

This code passes 0 instead of WINHTTP_FLAG_SECURE if m_verifySSL is false. Therefore the connection is not TLS enabled.

This is incorrect behaviour. 

### Expected Behavior

If the scheme is HTTPS and m_verifySSL  is false, it should initiate a TLS connection, and ignore certificate errors.

### Current Behavior

If the scheme is HTTPS and m_verifySSL  is false, it initiates a non-TLS connection.

### Reproduction Steps

Create an outgoing HTTPS request to S3, and set m_verifySSL to false.
Use Wireshark or equivalent to observe that the connection is not TLS enabled.

### Possible Solution

Remove the `&& m_verifySSL`  from the condition.

### AWS CPP SDK version used

1.11.285

### Compiler and Version used

Visual Studio 2019

### Operating System and version

Windows Server 2019

jmklix:
Thanks for finding this bug and pointing it out to us. We are working on a fix.

sbiscigl:
Created https://github.com/aws/aws-sdk-cpp/pull/3313 which will revert to the old behavior, and creates a client configuraiton for users to use anonymous auth to satisfy the original issue.

sbiscigl:
should be fixed on main now, tagged this afternoon, give a shout if you have any questions

Issue #1445, opened 2020-08-25
Failed to disable SSL certificate verification


rupal-bq:
Confirm by changing [ ] to [x] below to ensure that it's a bug:
- [ ] I've gone though [Developer Guide](https://docs.aws.amazon.com/sdk-for-cpp/v1/developer-guide/welcome.html) and [API reference](http://sdk.amazonaws.com/cpp/api/LATEST/index.html)
- [ ] I've searched for [previous similar issues](https://github.com/aws/aws-sdk-cpp/issues) and didn't find any solution
  
**Describe the bug**
Setting verifySSL option to false in ClientConfiguration does not disable SSL certificate verification on windows. Request returns response code `REQUEST_NOT_MADE`.

**SDK version number**
1.8.34

**Platform/OS/Hardware/Device**
Windows 10

**To Reproduce (observed behavior)**
1. Run [Open Distro for Elasticsearch](https://opendistro.github.io/for-elasticsearch-docs/docs/install/windows/)
2. Use [this code](https://github.com/rupal-bq/sql/blob/feature/odbc/verifySSL/sql-odbc/src/UnitTests/UTAwsSdkCpp/test_aws_sdk_cpp.cpp#L74) to send requests to the server. 

**Expected behavior**
The test should return the response code as `OK`.

Request `curl -XGET https://localhost:9200/_cat/plugins?format=json -u admin:admin --insecure` returns response code as `OK`

**Logs/output**
[aws_sdk_2020-08-25-22.log](https://github.com/aws/aws-sdk-cpp/files/5126535/aws_sdk_2020-08-25-22.log)
```
[INFO] 2020-08-25 22:07:16.837 Aws_Init_Cleanup [14136] Initiate AWS SDK for C++ with Version:1.8.34
[TRACE] 2020-08-25 22:07:16.837 FileSystem [14136] Checking USERPROFILE for the home directory.
[DEBUG] 2020-08-25 22:07:16.837 FileSystem [14136] Environment value for variable USERPROFILE is C:\Users\Rupal
[INFO] 2020-08-25 22:07:16.837 Aws::Config::AWSConfigFileProfileConfigLoader [14136] Initializing config loader against fileName C:\Users\Rupal\.aws\credentials and using profilePrefix = 0
[TRACE] 2020-08-25 22:07:16.837 FileSystem [14136] Checking USERPROFILE for the home directory.
[...]

PR #3304, opened 2025-02-17
Fix tls disabled if verifySSL is not ON on Windows

phoebusm:
*Issue #, if available:*
https://github.com/aws/aws-sdk-cpp/issues/3008

*Description of changes:*
Basically reverting https://github.com/aws/aws-sdk-cpp/commit/6805ec77535cbd9b3fdf9175188e21b17af1a165
Maintainers agree with the fact that this is an issue: https://github.com/aws/aws-sdk-cpp/issues/3008#issuecomment-2179130108

TLS connection should not be coupled whether verifying certificate, regardless of platform.
The issue has been found affecting Windows only

*Check all that applies:*

- [ ] Linux
- [x] Windows
- [ ] Android
- [ ] MacOS
- [ ] IOS
- [ ] Other Platforms

### DIFF
--- a/src/aws-cpp-sdk-core/source/http/windows/WinHttpSyncHttpClient.cpp
+++ b/src/aws-cpp-sdk-core/source/http/windows/WinHttpSyncHttpClient.cpp
@@ -533,7 +533,7 @@ void* WinHttpSyncHttpClient::OpenRequest(const std::shared_ptr<HttpRequest>& req
 {
     LPCWSTR accept[2] = { nullptr, nullptr };
 
-    DWORD requestFlags = request->GetUri().GetScheme() == Scheme::HTTPS && m_verifySSL ? WINHTTP_FLAG_SECURE : 0;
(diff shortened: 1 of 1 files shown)


sbiscigl:
Hey @phoebusm took a look at the root issue https://github.com/aws/aws-sdk-cpp/issues/3008 and cross referenced with the original issue that we made this change for https://github.com/aws/aws-sdk-cpp/issues/1445. You are right this is wrong and we should not be doing that, and yes we can revert it.

we also have to address our customers that use it, and are relying on it. so for this i created https://github.com/aws/aws-sdk-cpp/pull/3313 which allows for the original use case of opting into anonymous auth which is what the original requester wanted.

sorry for creating _another pr_ but hopefully this will work for your use and will remove the need for this revert.

sbiscigl:
closing in favor of https://github.com/aws/aws-sdk-cpp/pull/3313 which has been merged

PR #3311, opened 2025-02-24
Copy bi-directional event stream request

SergeyRyabinin:
*Issue #, if available:*
Undefined behavior:
* on request copying;
* undefined behavior if the lifetime of the request is not maintained by user
 
*Description of changes:*
Create a request copy for async execution.
Avoid capturing objects by raw pointers into lambdas factories to avoid dangling pointers/reference.

*Check all that applies:*

- [x] Linux
- [ ] Windows
- [ ] Android
- [ ] MacOS
- [ ] IOS
- [ ] Other Platforms

### DIFF
--- a/src/aws-cpp-sdk-core/include/aws/core/client/AWSClientEventStreamingAsyncTask.h
+++ b/src/aws-cpp-sdk-core/include/aws/core/client/AWSClientEventStreamingAsyncTask.h
@@ -1,0 +1,130 @@
+/**
+ * Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.
+ * SPDX-License-Identifier: Apache-2.0.
+ */
+
+#pragma once
+
+#include <aws/core/Core_EXPORTS.h>
+#include <aws/core/NoResult.h>
+#include <aws/core/client/AWSClient.h>
+#include <aws/core/utils/event/EventDecoderStream.h>
+#include <aws/core/utils/event/EventEncoderStream.h>
+#include <aws/core/utils/event/EventStreamHandler.h>
+#include <aws/core/utils/logging/ErrorMacros.h>
+
+namespace Aws {
+namespace Client {
+
+class AsyncCallerContext;
+
+/**
+ * A template class to create bi-directional eventstreaming async task.
+ * This is essentially a C++14 capture-by-move lambda
+ */
+template <typename OutcomeT, typename ClientT, typename AWSEndpointT, typename RequestT, typename HandlerT>
+class AWS_CORE_LOCAL BidirectionalEventStreamingTask final {
+ public:
+  BidirectionalEventStreamingTask(const ClientT* pClientThis, AWSEndpointT&& endpoint, const std::shared_ptr<RequestT>& pRequest,
+                                  const HandlerT& handler, const std::shared_ptr<const Aws::Client::AsyncCallerContext>& handlerContext,
+                                  const std::shared_ptr<Utils::Event::EventEncoderStream>& stream, const Http::HttpMethod method,
+                                  char const* const signerName)
+      : m_clientThis(pClientThis),
+        m_endpoint(std::move(endpoint)),
+        m_pRequest(pRequest),
+        m_handler(handler),
+        m_handlerContext(handlerContext),
+        m_stream(stream),
+        m_method(method),
+        m_signerName(signerName),
+        m_sem(Aws::MakeShared<Aws::Utils::Threading::Semaphore>("BidirectionalEventStreamingTask", 0, 1)) {
+    assert(m_clientThis);
+    assert(m_pRequest);
+    assert(m_signerName);
+
+    // Reset EventStream handler as it still has a raw dangling pointer of the original user request
+    m_pRequest->SetEventStreamHandler(m_pRequest->GetEventStreamHandler());
+
+    // Perform synchronization logic extracted from generated code
+    // Setup a signal to notify the submitter thread that the task has actually started
+    std::shared_ptr<Aws::Utils::Threading::Semaphore> sem = m_sem;
+    m_pRequest->SetRequestSignedHandler([stream, sem](const Aws::Http::HttpRequest& httpRequest) {
+      stream->SetSignatureSeed(Aws::Client::GetAuthorizationHeader(httpRequest));
+      sem->ReleaseAll();
+    });
+
+    std::weak_ptr<RequestT> wRequest = m_pRequest;
+    // Setup InitialResponse handler to use the new actual request object
+    if (!m_pRequest->GetHeadersReceivedEventHandler()) {
+      m_pRequest->SetHeadersReceivedEventHandler([wRequest](const Http::HttpRequest*, Http::HttpResponse* response) {
+        auto request = wRequest.lock();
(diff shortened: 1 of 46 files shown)


src/aws-cpp-sdk-core/source/client/AWSClient.cpp
```diff
@@ -305,7 +305,8 @@ HttpResponseOutcome AWSClient::AttemptExhaustively(const Aws::Http::URI& uri,
         coreMetrics.httpClientMetrics = httpRequest->GetRequestMetrics();
         TracingUtils::EmitCoreHttpMetrics(httpRequest->GetRequestMetrics(),
             *m_telemetryProvider->getMeter(this->GetServiceClientName(), {}),
-            {{TracingUtils::SMITHY_METHOD_DIMENSION, request.GetServiceRequestName()},{TracingUtils::SMITHY_SERVICE_DIMENSION, this->GetServiceClientName()}});
+            {{TracingUtils::SMITHY_METHOD_DIMENSION, request.GetServiceRequestName()},
```
SergeyRyabinin, line 308: if it crashes again we will know if service client or request is in UB.

src/aws-cpp-sdk-core/source/utils/event/EventStreamDecoder.cpp
```diff
@@ -53,6 +53,7 @@ namespace Aws
 
             void EventStreamDecoder::ResetEventStreamHandler(EventStreamHandler* handler)
             {
+                m_eventStreamHandler = handler;
```
SergeyRyabinin, line 56: it was a big one.

tools/code-generation/generator/src/main/resources/com/amazonaws/util/awsclientgenerator/velocity/cpp/RequestHeader.vm
```diff
@@ -137,7 +137,7 @@ namespace Model
     /**
      * Underlying Event Stream Handler which is used to define callback functions.
      */
-    inline const ${operation.name}Handler& GetEventStreamHandler() const { return m_handler; }
+    inline ${operation.name}Handler& GetEventStreamHandler() { return m_handler; }
```
SergeyRyabinin, line 140: returning object with only non-const methods by const reference is quite meaningless otherwise.

src/aws-cpp-sdk-core/include/aws/core/client/AWSClientEventStreamingAsyncTask.h
```diff
@@ -0,0 +1,137 @@
...
+    }
+
+    // Setup ResponseStreamFactory to provide EventStream decoder based on the new actual request object, not the original one.
+    m_pRequest->SetResponseStreamFactory([wRequest]() -> Aws::IOStream* {
+      auto request = wRequest.lock();
+      if (!request) {
+        AWS_LOGSTREAM_FATAL(ClientT::GetAllocationTag(),
+                            "Unexpected nullptr bi-directional streaming request on response streaming factory call!");
+        assert(false);
+        return nullptr;
```
sbiscigl, line 85: instead of nullptr should we return a empty string stream? even though we have entered a invalid state, a empty stream might help someone more gracefully exit than a null ptr

src/aws-cpp-sdk-core/include/aws/core/client/AWSClientEventStreamingAsyncTask.h
```diff
@@ -0,0 +1,137 @@
...
+
+    std::weak_ptr<RequestT> wRequest = m_pRequest;
+    // Setup InitialResponse handler to use the new actual request object
+    if (!m_pRequest->GetHeadersReceivedEventHandler()) {
+      m_pRequest->SetHeadersReceivedEventHandler([wRequest](const Http::HttpRequest*, Http::HttpResponse* response) {
+        auto request = wRequest.lock();
+        if (!request) {
+          AWS_LOGSTREAM_FATAL(ClientT::GetAllocationTag(), "Unexpected nullptr bi-directional streaming request on initial response!");
+          assert(false);
+          return;
```
sbiscigl, line 64: nit i prefer no returns on a void function and to rely on if - else behavior i.e.
```
if (!request) {
  ....
} else if (!response) {
  ....
} else {
  .....
}
```

src/aws-cpp-sdk-core/include/aws/core/client/AWSClientEventStreamingAsyncTask.h
```diff
@@ -0,0 +1,137 @@
...
+                                  char const* const signerName)
+      : m_clientThis(pClientThis),
+        m_endpoint(std::move(endpoint)),
+        m_pRequest(pRequest),
+        m_handler(handler),
+        m_handlerContext(handlerContext),
+        m_stream(stream),
+        m_method(method),
+        m_signerName(signerName),
+        m_sem(Aws::MakeShared<Aws::Utils::Threading::Semaphore>("BidirectionalEventStreamingTask", 0, 1)) {
```
sbiscigl, line 40: why do we want to use a semaphore and not a mutex if max value of semaphore is 1?

tests/aws-cpp-sdk-transcribestreaming-integ-tests/TranscribeErrorCaseTests.cpp
```diff
@@ -154,4 +154,179 @@ TEST_F(TranscribeStreamingErrorTests, TranscribeAudioFile) {
...
+    if (outcome.IsSuccess()) {
+      TestTrace(Aws::String("OnResponseCallback, success"));
+    } else {
+      TestTrace(Aws::String("OnResponseCallback, error: ") + outcome.GetError().GetMessage());
+    }
+    semaphore.ReleaseAll();
+  };
+  for (size_t i = 0; i < 2; ++i) {
+    client->StartStreamTranscriptionAsync(request, OnStreamReady, OnResponseCallback, nullptr /*context*/);
+    semaphore.WaitOne();
```
sbiscigl, line 250: please add a reasonable "drop dead" timeout so that we can avoid locking unit tests if something goes wrong

src/aws-cpp-sdk-core/include/aws/core/client/AWSClientEventStreamingAsyncTask.h
SergeyRyabinin: ResponseStream is already handling nullptr condition in the c-tor.
https://github.com/aws/aws-sdk-cpp/blob/9c322a9857208116ee6e6f41b5483073c3fa4c73/src/aws-cpp-sdk-core/source/utils/stream/ResponseStream.cpp#L26-L30

src/aws-cpp-sdk-core/include/aws/core/client/AWSClientEventStreamingAsyncTask.h
SergeyRyabinin: That's the original generated design code.
We could switch to a singal, but it will take 2 more member variables (mutex and a conditional_variable) to pass around instead of this single one.

sbiscigl (approved):

sbiscigl (approved):

PR #3317, opened 2025-02-27
Prerequisites to Enable smithy client

sbera87:
*Issue #, if available:*

*Description of changes:*
This set of changes are needed to enable Smithy client for S3 with applicable backwards compatible changes.
- Relevant s3 smithy code generation from PR: https://github.com/aws/aws-sdk-cpp/pull/3298
- Added legacy APIs in smithy client for backwards compatibility
- Modify Generate Presigned URL to be smithy compliant workflow

*Check all that applies:*

- [x] Linux
- [x] Windows
- [ ] Android
- [x] MacOS
- [ ] IOS
- [ ] Other Platforms


src/aws-cpp-sdk-core/source/smithy/client/AwsSmithyClientBase.cpp
```diff
@@ -290,18 +460,21 @@ void AwsSmithyClientBase::AttemptOneRequestAsync(std::shared_ptr<AwsSmithyClient
         return;
     }
 
-    pRequestCtx->m_interceptorContext->SetTransmitRequest(pRequestCtx->m_httpRequest);
-    for (const auto& interceptor : m_interceptors)
+    if (pRequestCtx->m_interceptorContext)
```
sbera87, line 463: just added code block under null guard

src/aws-cpp-sdk-core/source/smithy/client/AwsSmithyClientBase.cpp
```diff
@@ -152,10 +152,129 @@ AwsSmithyClientBase::BuildHttpRequest(const std::shared_ptr<AwsSmithyClientAsync
...
+        requestName,
+        method,
+        std::move(resolveEndpointCb),
+        std::move(responseHandler),
+        pExecutor);
+}
+
+
+
+void AwsSmithyClientBase::MakeRequestAsyncHelper(std::shared_ptr<AwsSmithyClientAsyncRequestContext>& pRequestCtx,
```
sbera87, line 273: This helper abstracts endpoint resolution into a callback and is invoked contextually by different variations of APIs used for backwards compatibilty

generated/src/aws-cpp-sdk-s3/source/S3Client.cpp
```diff
@@ -4624,25 +4232,14 @@ Aws::String S3Client::GeneratePresignedUrl(const Aws::String& bucket,
...
-    }
-    Aws::Endpoint::AWSEndpoint& endpoint = computeEndpointOutcome.GetResult();
-    endpoint.AddPathSegments(key);
     Aws::Map<Aws::String, Aws::String> params;
     params.emplace("bucketName", bucket);
     ServiceSpecificParameters serviceSpecificParameters{params};
     auto serviceParams = Aws::MakeShared<ServiceSpecificParameters>(ALLOCATION_TAG, serviceSpecificParameters);
-    return AWSClient::GeneratePresignedUrl(endpoint, method, customizedHeaders, expirationInSeconds, Aws::Auth::SIGV4_SIGNER,
-        nullptr, nullptr, serviceParams);
+    EndpointUpdateCallback endpointCallback = [&](Aws::Endpoint::AWSEndpoint& endpoint){
```
sbera87, line 4239: Changes to resolve endpoint after select auth scheme

src/aws-cpp-sdk-core/include/smithy/client/AwsSmithyClient.h
```diff
@@ -153,7 +154,26 @@ namespace client
                 }
             }
 
-            Aws::Vector<AuthSchemeOption> authSchemeOptions = m_authSchemeResolver->resolveAuthScheme(identityParams);
+            Aws::Vector<AuthSchemeOption> authSchemeOptions;
+            /*for backwards compatibility, signer name override is used to filter auth schemes by id equivalent to legacy signer name*/
+            if(ctx.m_signerNameOverride.has_value())
```
sbera87, line 159: This is added to make the backwards compatible api functional. 
In legacy client, signer name passed was used to fetch the signer object via
`auto signer = GetSignerByName(signerName);`
In smithy design, this name passed is simply re-mapped and used to filter desired auth scheme (with requested signer) from list of supported auth schemes

src/aws-cpp-sdk-core/include/smithy/client/AwsSmithyClient.h
```diff
@@ -240,8 +314,141 @@ namespace client
             }
             return GeneratePresignedUrl(uri, method, signerRegionOverride, signerServiceNameOverride, expirationInSeconds, customizedHeaders, serviceSpecificParameters);
         }
+        
+        ResponseT MakeRequest(const Aws::Http::URI& uri,
```
sbera87, line 318: Legacy APIs

src/aws-cpp-sdk-core/include/smithy/client/AwsSmithyClientBase.h
```diff
@@ -74,16 +74,17 @@ namespace client
         using HttpRequest = Aws::Http::HttpRequest;
         using HttpResponse = Aws::Http::HttpResponse;
         using CoreErrors = Aws::Client::CoreErrors;
-        using AWSError = Aws::Client::AWSError<CoreErrors>;
-        using ClientError = AWSError;
-        using SigningError = AWSError;
+        using AWSCoreError = Aws::Client::AWSError<CoreErrors>;
+        using ClientError = AWSCoreError;
```
sbera87, line 78: Needed to ensure backwards compatibility . External clients use AwsError as a templated type

src/aws-cpp-sdk-core/include/smithy/client/AwsSmithyClientBase.h
```diff
@@ -144,6 +145,42 @@ namespace client
...
+                const char* signerName,
+                const char* signerRegionOverride,
+                const char* signerServiceNameOverride,
+                const char* requestName,
+                Aws::Http::HttpMethod method,
+                ResponseHandlerFunc&& responseHandler,
+                std::shared_ptr<Aws::Utils::Threading::Executor> pExecutor
+            ) const;
+
+        const std::shared_ptr<Aws::Client::AWSErrorMarshaller>& GetErrorMarshaller() const
```
sbera87, line 179: Needed as External Users use this

src/aws-cpp-sdk-core/include/smithy/client/AwsSmithyClientBase.h
```diff
@@ -200,7 +237,7 @@ namespace client
         Aws::Vector<std::shared_ptr<smithy::interceptor::Interceptor>> m_interceptors{};
         std::shared_ptr<smithy::client::UserAgentInterceptor> m_userAgentInterceptor;
     private:
-        void UpdateAuthSchemeFromEndpoint(const Aws::Endpoint::AWSEndpoint& endpoint, AuthSchemeOption& authscheme) const;
+        void UpdateAuthSchemeFromEndpoint(std::shared_ptr<AwsSmithyClientAsyncRequestContext>& pRequestCtx) const;
```
sbera87, line 240: Interface updated to facilitate legacy API usage where endpoint may  not  be available

src/aws-cpp-sdk-core/include/smithy/client/AwsSmithyClient.h
sbiscigl: > This is added to make the backwards compatible api functional.

This is really confusing me, you say we added this for backwards compatibility, but then we don't declare `GetSignerByName` so we break backwards incompatibility with that

i.e. this compiles before but is broken after. So why are we doing something for backwards compatibility, then we are actually breaking backwards compatibility?

```cpp
#include <aws/core/Aws.h>
#include <aws/s3/S3Client.h>

using namespace Aws;
using namespace Aws::S3;

class MyS3Client : public Aws::S3::S3Client {
public:
  MyS3Client() = default;
  ~MyS3Client() override = default;
  void Foo() {
    std::cout << this->GetSignerByName("some_name") << std::endl;
  }
};

auto main() -> int {
  SDKOptions options{};
  options.loggingOptions.logLevel = Aws::Utils::Logging::LogLevel::Debug;
  InitAPI(options);
  {
    MyS3Client client{};
    client.Foo();
  }
  ShutdownAPI(options);
  return 0;
}
```

src/aws-cpp-sdk-core/include/smithy/client/AwsSmithyClientAsyncRequestContext.h
```diff
@@ -70,6 +70,10 @@ namespace smithy
             ResponseHandlerFunc m_responseHandler;
             std::shared_ptr<Aws::Utils::Threading::Executor> m_pExecutor;
             std::shared_ptr<interceptor::InterceptorContext> m_interceptorContext;
+            //For Backwards Compatible APIs
```
sbiscigl, line 73: what backwards compatible APIs?

src/aws-cpp-sdk-core/include/smithy/client/AwsSmithyClient.h
sbiscigl: can we move this to a different class? its makes it hard to understand what is going on here if we add them in the smithy client, move it to a new class then have the smithy client just reference it. we do not want to copy and paste the old work flow into the new one.

src/aws-cpp-sdk-core/include/smithy/client/AwsSmithyClient.h
```diff
@@ -240,8 +314,141 @@ namespace client
...
+        }
+
+        ResponseT MakeRequest(const Aws::Http::URI& uri,
+            Aws::Http::HttpMethod method = Aws::Http::HttpMethod::HTTP_POST,
+            const char* signerName = Aws::Auth::SIGV4_SIGNER,
+            const char* requestName = "",
+            const char* signerRegionOverride = nullptr,
+            const char* signerServiceNameOverride = nullptr) const
+        {
+            std::shared_ptr<Aws::Utils::Threading::Executor> pExecutor = Aws::MakeShared<Aws::Utils::Threading::SameThreadExecutor>(ServiceNameT);
```
sbiscigl, line 359: you've more or less copy and pasted the old code into the new class, this is not what we want. this diverges the flow significantly. these functions _should_call_ `MakeRequestDeserialize` and should not diverge the call path. This is just creating both clients in a a single class. we should defer calling to the new methods and emit a warning log about calling these methods directly.

sbiscigl (commented):
I think we've got a maintainibilty issue here. You've more or less copy and pasted AWSClient into the SmithyClient, which I don't think is the way forward here. If we want to maintain backward compatible interfaces for newer clients we should maintain the interface, but not copy and paste the code directly.

src/aws-cpp-sdk-core/include/smithy/client/AwsSmithyClient.h
```diff
@@ -190,13 +210,68 @@ namespace client
             return m_serializer->Deserialize(std::move(httpResponseOutcome), GetServiceClientName(), requestName);
         }
 
+        Aws::String GeneratePresignedUrl(
+            const Aws::String& bucket,
```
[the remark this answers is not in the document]
sbera87: The above example is already handled in another PR (where other rest xml clients use the API you mentioned.) This PR deals with S3 related changes only

generated/src/aws-cpp-sdk-s3/include/aws/s3/S3ExpressAuthSchemeResolver.h
```diff
@@ -0,0 +1,64 @@
...
+                auto epResolutionOutcome = m_endpointProviderForAuth->ResolveEndpoint(epParams);
+                if (epResolutionOutcome.IsSuccess())
+                {
+                    auto endpoint = std::move(epResolutionOutcome.GetResultWithOwnership());
+                    if (endpoint.GetAttributes())
+                    {
+                        auto authSchemeName = endpoint.GetAttributes()->authScheme.GetName();
+                        if(strcmp(authSchemeName.c_str(),Aws::Auth::ASYMMETRIC_SIGV4_SIGNER) == 0)
+                        {
+                        return {smithy::SigV4aAuthSchemeOption::sigV4aAuthSchemeOption};
```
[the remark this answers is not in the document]
sbera87: So, its a vector of one for the appropriate scheme id. If you see smithy v4, v4a authscheme resolvers they return respective vector of one authscheme options. Since this is a child of multi auth scheme resolver it simply uses if else condition block for the different scheme ids.

generated/src/aws-cpp-sdk-s3/include/aws/s3/S3ClientConfiguration.h
```diff
@@ -62,6 +62,10 @@ namespace Aws
             IdentityProviderSupplier identityProviderSupplier = [](const S3Client &client) -> std::shared_ptr<S3ExpressIdentityProvider> {
                 return Aws::MakeShared<DefaultS3ExpressIdentityProvider>("S3ClientConfiguration", client);
             };
+            using SmithyIdentityProviderSupplier = std::function<std::shared_ptr<SmithyS3ExpressIdentityProvider> (const S3Client &)>;
```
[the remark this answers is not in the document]
sbera87: Though the functionality is achieved the same, the interfaces are different and hence added to support both old and new
`class AWS_S3_API S3ExpressIdentityProvider : public smithy::IdentityResolverBase<S3ExpressIdentity>`
vs
`class AWS_S3_API SmithyS3ExpressIdentityProvider : public smithy::IdentityResolverBase<smithy::AwsCredentialIdentityBase>` 

In smithy templated implementation, the latter interface is needed
as S3 express authscheme uses S3express signer which internally use smithy::AwsCredentialIdentityBase identity type and not S3ExpressIdentity.

`S3ExpressSigV4AuthScheme : public smithy::AuthScheme< smithy::AwsCredentialIdentityBase>`
`AwsSigV4Signer : public AwsSignerBase<AwsCredentialIdentityBase>`

generated/src/aws-cpp-sdk-s3/source/S3Client.cpp
```diff
@@ -697,29 +692,25 @@ DeleteBucketOutcome S3Client::DeleteBucket(const DeleteBucketRequest& request) c
...
+      return DeleteBucketOutcome(MakeRequestDeserialize(&request, request.GetServiceRequestName(), Aws::Http::HttpMethod::HTTP_DELETE, [&](Aws::Endpoint::AWSEndpoint& resolvedEndpoint) ->  void {
       request.SetServiceSpecificParameters(
-        [&]() -> std::shared_ptr<Http::ServiceSpecificParameters> {
-          Aws::Map<Aws::String, Aws::String> params;
-          params.emplace("bucketName", request.GetBucket());
-          ServiceSpecificParameters serviceSpecificParameters{params};
-          return Aws::MakeShared<ServiceSpecificParameters>(ALLOCATION_TAG, serviceSpecificParameters);
-        }());
-      return DeleteBucketOutcome(MakeRequest(request, endpointResolutionOutcome.GetResult(), Aws::Http::HttpMethod::HTTP_DELETE));
+       [&]() -> std::shared_ptr<Http::ServiceSpecificParameters> {
```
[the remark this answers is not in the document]
sbera87: I double checked. This honors the indentation used in the new smithy code gen template. Changing this will impact other smithy clients too.

src/aws-cpp-sdk-core/include/smithy/client/AwsSmithyClient.h
sbera87: Its actually new code in old interface. But, I have changed all the legacy APIs to call MakeRequestDeserialize using CRTP pattern in smithy client

src/aws-cpp-sdk-core/include/smithy/client/AwsSmithyClient.h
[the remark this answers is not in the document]
sbera87: I have made changes to this to eliminate bucket from the interface

generated/src/aws-cpp-sdk-s3/include/aws/s3/S3ExpressAuthSchemeResolver.h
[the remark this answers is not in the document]
sbiscigl: > So, its a vector of one for the appropriate scheme id.

no you should be returning a vector of supported auth schemes, and using the first supported one. i.e. if you are using v4a it should return `[v4a, v4]` then the caller gets to decide which one is supported, thats the way the interface is supposed to wok

generated/src/aws-cpp-sdk-s3/include/aws/s3/S3ExpressAuthSchemeResolver.h
```diff
@@ -0,0 +1,64 @@
...
+                        }
+                        else if(strcmp(authSchemeName.c_str(), S3::S3_EXPRESS_SIGNER_NAME) == 0)
+                        {
+                            return {S3ExpressSigV4AuthSchemeOption::s3ExpressSigV4AuthSchemeOption};
+                        }
+                    }
+                }
+            }
+
+            return {smithy::SigV4AuthSchemeOption::sigV4AuthSchemeOption};
```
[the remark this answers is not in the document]
sbiscigl: that doesnt make sense, this should return a empty list. doing something because "it makes a test pass" is not a good reason for a implementation detail

get object shuold be using a auth scheme that is recognized, why isnt it.

generated/src/aws-cpp-sdk-s3/source/S3Client.cpp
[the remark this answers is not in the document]
sbiscigl: well if it was a problem before we should fix it, it doesnt honor the correct indentation, please fix it

generated/src/aws-cpp-sdk-s3/source/S3ExpressSignerProvider.cpp
```diff
@@ -21,12 +21,12 @@ Aws::Auth::S3ExpressSignerProvider::S3ExpressSignerProvider(
         region,
         signingPolicy,
         urlEscapePath) {
-    m_signers.emplace_back(Aws::MakeShared<Aws::S3::S3ExpressSigner>(CLASS_TAG,
+    m_signers.emplace_back(std::static_pointer_cast<Aws::Client::AWSAuthSigner>(Aws::MakeShared<Aws::S3::S3ExpressSigner>(CLASS_TAG,
```
sbiscigl, line 24: remove the static pointer class it should not be needed

src/aws-cpp-sdk-core/include/smithy/client/AwsLegacyClient.h
```diff
@@ -0,0 +1,106 @@
...
+#pragma once
+
+#include <aws/core/AmazonWebServiceRequest.h>
+#include <aws/core/endpoint/AWSEndpoint.h>
+
+namespace smithy {
+namespace client
+{   
+    /*
+        This Class exists only to provide a way to access the legacy client APIs
```
sbiscigl, line 14: niiiiiiiiiiiice, i like this a lot

src/aws-cpp-sdk-core/include/smithy/client/AwsSmithyClient.h
```diff
@@ -240,13 +312,21 @@ namespace client
...
-
-    protected:
+    
         ServiceClientConfigurationT& m_clientConfiguration;
         std::shared_ptr<EndpointProviderT> m_endpointProvider{};
         std::shared_ptr<ServiceAuthSchemeResolverT> m_authSchemeResolver{};
         Aws::UnorderedMap<Aws::String, AuthSchemesVariantT> m_authSchemes{};
         std::shared_ptr<SerializerT> m_serializer{};
+    private:
+        friend class AwsLegacyClientT<ServiceNameT, ResponseT, AwsSmithyClientT<ServiceNameT,
```
sbiscigl, line 322: why friend class and not CRTP?

generated/src/aws-cpp-sdk-s3/source/S3ExpressSignerProvider.cpp
sbera87: For the return auth schemes list comment, we can't return all the schemes in a list  for multi auth case.
For multi auth support, we do a resolve endpoint twice, the first time happens in select auth scheme. 

Based on the auth scheme, we build the list as aforementioned. What you are referring to as pick the first supported auth scheme actually happens here inside the resolver. This is what based on previous PR comments to move this auth scheme vector selection inside the auth scheme resolver class from smithy client class (else it would end up doing double endpoint resolution for all clients and not just S3).

So, for the negative test case I mentioned, epResolutionOutcome is not a success and hence we have empty auth schemes list.

src/aws-cpp-sdk-core/include/smithy/client/AwsSmithyClient.h
sbera87: Without indicating friend class, it won't compile as legacy client calls MakeRequestDeserialize (which is protected) 
Note:
`class AwsSmithyClientT : public AwsSmithyClientBase, public AwsLegacyClientT<... >`
this keeps the AwsSmithyClientT as derived class inheriting from Instantiation of itself per CRTP pattern.

generated/src/aws-cpp-sdk-s3/source/S3ExpressSignerProvider.cpp
sbera87: Removed the cast in next commit

generated/src/aws-cpp-sdk-s3/source/S3Client.cpp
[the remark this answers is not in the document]
sbera87: As discussed, this is just the change smithy codegen brings with own consistent indentation that it is consistent with
.

sbiscigl (approved):

SergeyRyabinin (approved):

Issue #3315, opened 2025-02-26
mingw-w64 support


kou:
### Describe the feature

#1836 was raised and rejected in 2022. Is this policy still valid?

MSYS2 has patches for this:
* https://github.com/msys2/MINGW-packages/blob/master/mingw-w64-aws-sdk-cpp/0001-fix-dll-import-libary-folder.patch
* https://github.com/msys2/MINGW-packages/blob/master/mingw-w64-aws-sdk-cpp/aws-sdk-cpp-pr-1333.patch

### Use Case

I'm an Apache Arrow https://arrow.apache.org/ developer.

Apache Arrow uses aws-sdk-cpp for S3 support. Apache Arrow uses system aws-sdk-cpp or bundled aws-sdk-cpp by `ExternalProject_Add()`/`FetchContent`.

Apache Arrow supports mingw-w64 based environments. If aws-sdk-cpp supports mingw-w64, Apache Arrow don't need to apply patches with `ExternalProject_Add()`/`FetchContent`.

### Proposed Solution

How about merging the patches in MSYS2?

### Acknowledgements

- [x] I may be able to implement this feature request
- [ ] This feature might incur a breaking change

SergeyRyabinin:
Hi @kou,
Thank you for submitting this issue and thank you for providing this patches.
We are not against adding mingw support. The main reason the previous PRs were not merged is because we don't have this platform on our internal CI build system, so we were not able to ensure we would not break it, so we could not claim this target as supported.

Would it be possible for you to provide an example how you build the SDK on mingw-w64? I.e. the set of commands executed, and the prerequisites required so we could reproduce the build.

Best regards,
Sergey

kou:
Could you check #3319?

SergeyRyabinin:
Hi!
Thank you a lot for submitting this sample PR. I'll use it as an example while adding the mingw support.

Best regards,
Sergey

kou:
Thanks!

If you have any question, please let me know.

kou:
Is there any progress?

SergeyRyabinin:
Hi @kou ,

Please sorry for a delayed response.
I will try to work on this item next week.

Best regards,
S

kou:
Thanks!!!

kou:
Can we close this (and #3319) because #3385 was merged?
(Thanks!)

kai-ion:
[...]

PR #3319, opened 2025-02-28
Don't merge: Add a GitHub Actions workflow for mingw-w64 with MSYS2

kou:
*Issue #, if available:* #3315

*Description of changes:*

This just shows how to setup mingw-w64 build environment with MSYS2.

This uses GitHub Actions but we'll use internal CI build system instead of GitHub Actions.

The added workflow has 2 jobs. One job applies a patch ( https://github.com/msys2/MINGW-packages/blob/master/mingw-w64-aws-sdk-cpp/aws-sdk-cpp-pr-1333.patch ) in MSYS2 and another job builds aws-sdk-cpp as-is.

Both jobs do the followings:
 
1. Setup build environment with MSYS2
2. Run `cmake`
3. Build

I'll add review comments for details.

*Check all that applies:*

- [ ] Linux
- [x] Windows
- [ ] Android
- [ ] MacOS
- [ ] IOS
- [ ] Other Platforms

### DIFF
--- a/.github/workflows/mingw-w64.yml
+++ b/.github/workflows/mingw-w64.yml
@@ -1,0 +1,65 @@
+name: mingw-w64
+
+on:
+  - push
+  - pull_request
+
+jobs:
+  build:
+    name: Build
+    runs-on: windows-latest
+    strategy:
+      fail-fast: false
+      matrix:
+        with-patch:
+          - true
+          - false
+    env:
+      MINGW_PACKAGE_PREFIX: mingw-w64-ucrt-x86_64
+      MSYSTEM: UCRT64
+    steps:
+      - name: checkout
+        uses: actions/checkout@v4
+        with:
+          submodules: recursive
+      - name: dependencies
+        shell: c:\msys64\usr\bin\bash.exe -e -l {0}
+        run: |
+          pacman \
+            --needed \
+            --noconfirm \
+            --sync \
+            ${MINGW_PACKAGE_PREFIX}-cc \
+            ${MINGW_PACKAGE_PREFIX}-cmake \
+            ${MINGW_PACKAGE_PREFIX}-curl \
+            ${MINGW_PACKAGE_PREFIX}-ninja \
+            ${MINGW_PACKAGE_PREFIX}-openssl \
+            ${MINGW_PACKAGE_PREFIX}-zlib
+      - name: patch
+        if: matrix.with-patch
+        shell: c:\msys64\usr\bin\bash.exe -e -l {0}
+        run: |
+          pacman \
+            --needed \
+            --noconfirm \
+            --sync \
+            patch
+          cd ${GITHUB_WORKSPACE}
+          curl https://raw.githubusercontent.com/msys2/MINGW-packages/refs/heads/master/mingw-w64-aws-sdk-cpp/aws-sdk-cpp-pr-1333.patch | \
+            patch -p1
+      - name: configure
+        shell: c:\msys64\usr\bin\bash.exe -e -l {0}
+        run: |
+          set -x
+          cmake \
+            -B build \
+            -DAWS_SDK_WARNINGS_ARE_ERRORS=OFF \
+            -DBUILD_ONLY="s3;s3-crt;dynamodb;logs;kms;sqs;firehose;kinesis;sns;mediastore;mediastore-data;monitoring;secretsmanager;athena;kafka;cognito-idp;rds;ecs" \
+            -DBUILD_SHARED_LIBS=OFF \
+            -DENABLE_TESTING=OFF \
+            -GNinja \
+            -S ${GITHUB_WORKSPACE}
+      - name: build
+        shell: c:\msys64\usr\bin\bash.exe -e -l {0}
+        run: |
+          cmake --build build


.github/workflows/mingw-w64.yml
```diff
@@ -0,0 +1,65 @@
+name: mingw-w64
+
+on:
+  - push
+  - pull_request
+
+jobs:
+  build:
+    name: Build
+    runs-on: windows-latest
```
kou, line 10: It uses Windows 2022. GitHub Actions runner for Windows 2022 has MSYS2: https://github.com/actions/runner-images/blob/main/images/windows/Windows2022-Readme.md#msys2

So, this doesn't install MSYS2 explicitly.

We can install MSYS2 via command line: https://www.msys2.org/docs/installer/#cli-usage-examples

.github/workflows/mingw-w64.yml
```diff
@@ -0,0 +1,65 @@
...
+    strategy:
+      fail-fast: false
+      matrix:
+        with-patch:
+          - true
+          - false
+    env:
+      MINGW_PACKAGE_PREFIX: mingw-w64-ucrt-x86_64
```
kou, line 18: This environment variable isn't required. It's just for reusing the same value.

.github/workflows/mingw-w64.yml
```diff
@@ -0,0 +1,65 @@
...
+      MSYSTEM: UCRT64
```
kou, line 19: This environment variable is required. `bash -l` (using bash as a login shell) refers this environment variable and setup `PATH` and so on.

.github/workflows/mingw-w64.yml
```diff
@@ -0,0 +1,65 @@
...
+    steps:
+      - name: checkout
+        uses: actions/checkout@v4
+        with:
+          submodules: recursive
+      - name: dependencies
+        shell: c:\msys64\usr\bin\bash.exe -e -l {0}
```
kou, line 26: `-l` (login shell) is important here.

If we don't have `-l`, mingw-w64 build environment isn't prepared.

.github/workflows/mingw-w64.yml
```diff
@@ -0,0 +1,65 @@
...
+          pacman \
+            --needed \
+            --noconfirm \
+            --sync \
+            ${MINGW_PACKAGE_PREFIX}-cc \
+            ${MINGW_PACKAGE_PREFIX}-cmake \
+            ${MINGW_PACKAGE_PREFIX}-curl \
+            ${MINGW_PACKAGE_PREFIX}-ninja \
+            ${MINGW_PACKAGE_PREFIX}-openssl \
+            ${MINGW_PACKAGE_PREFIX}-zlib
```
kou, line 37: `${MINGW_PACKAGE_PREFIX}` is `mingw-w64-ucrt-x86_64`.

It means:

* `mingw-w64-ucrt`: We use mingw-w64 GCC with universal C runtime
* `x86_64`: We use Intel/AMD 64bit CPU (We can use `aarch64` for Arm 64bit CPU)

.github/workflows/mingw-w64.yml
```diff
@@ -0,0 +1,65 @@
...
+        shell: c:\msys64\usr\bin\bash.exe -e -l {0}
+        run: |
+          pacman \
+            --needed \
+            --noconfirm \
+            --sync \
+            ${MINGW_PACKAGE_PREFIX}-cc \
+            ${MINGW_PACKAGE_PREFIX}-cmake \
+            ${MINGW_PACKAGE_PREFIX}-curl \
+            ${MINGW_PACKAGE_PREFIX}-ninja \
```
kou, line 35: We can use `${MINGW_PACKAGE_PREFIX}-make` but Ninja is faster than Make.

.github/workflows/mingw-w64.yml
```diff
@@ -0,0 +1,65 @@
...
+            ${MINGW_PACKAGE_PREFIX}-openssl \
+            ${MINGW_PACKAGE_PREFIX}-zlib
+      - name: patch
+        if: matrix.with-patch
```
kou, line 39: This step is used by one job. For example: https://github.com/kou/aws-sdk-cpp/actions/runs/13583720209/job/37974163071

.github/workflows/mingw-w64.yml
```diff
@@ -0,0 +1,65 @@
...
+            --needed \
+            --noconfirm \
+            --sync \
+            ${MINGW_PACKAGE_PREFIX}-cc \
+            ${MINGW_PACKAGE_PREFIX}-cmake \
+            ${MINGW_PACKAGE_PREFIX}-curl \
+            ${MINGW_PACKAGE_PREFIX}-ninja \
+            ${MINGW_PACKAGE_PREFIX}-openssl \
+            ${MINGW_PACKAGE_PREFIX}-zlib
+      - name: patch
```
kou, line 38: This step applies only https://github.com/msys2/MINGW-packages/blob/master/mingw-w64-aws-sdk-cpp/aws-sdk-cpp-pr-1333.patch .
https://github.com/msys2/MINGW-packages/blob/master/mingw-w64-aws-sdk-cpp/0001-fix-dll-import-libary-folder.patch isn't applied.

Because we need only the former for passing build. The latter is for fixing install location for shared library.

For now, the former patch is for static build. We can't build aws-sdk-cpp for shared library with the former patch. So we don't need the latter patch yet.

.github/workflows/mingw-w64.yml
```diff
@@ -0,0 +1,65 @@
...
+          cd ${GITHUB_WORKSPACE}
+          curl https://raw.githubusercontent.com/msys2/MINGW-packages/refs/heads/master/mingw-w64-aws-sdk-cpp/aws-sdk-cpp-pr-1333.patch | \
+            patch -p1
+      - name: configure
+        shell: c:\msys64\usr\bin\bash.exe -e -l {0}
+        run: |
+          set -x
+          cmake \
+            -B build \
+            -DAWS_SDK_WARNINGS_ARE_ERRORS=OFF \
```
kou, line 56: There are some warnings with mingw-w64.

.github/workflows/mingw-w64.yml
```diff
@@ -0,0 +1,65 @@
...
+            -DBUILD_ONLY="s3;s3-crt;dynamodb;logs;kms;sqs;firehose;kinesis;sns;mediastore;mediastore-data;monitoring;secretsmanager;athena;kafka;cognito-idp;rds;ecs" \
```
kou, line 57: This is same as the one in simple-build.yml.

.github/workflows/mingw-w64.yml
```diff
@@ -0,0 +1,65 @@
...
+            -DBUILD_SHARED_LIBS=OFF \
```
kou, line 58: We can't build shared library even with the patch in MSYS2.

.github/workflows/mingw-w64.yml
```diff
@@ -0,0 +1,65 @@
...
+            -DENABLE_TESTING=OFF \
```
kou, line 59: We can't build tests even with the patch in MSYS2.

PR #3320, opened 2025-02-28
migrate code gen to use protocol list

sbiscigl:
*Description of changes:*

Migrate code generation to use protocol list instead of single protocol value. i.e. from s3

```
  "metadata":{
    "apiVersion":"2006-03-01",
    "checksumFormat":"md5",
    "endpointPrefix":"s3",
    "globalEndpoint":"s3.amazonaws.com",
    "protocol":"rest-xml",
    "protocols":["rest-xml"],
    "serviceAbbreviation":"Amazon S3",
    "serviceFullName":"Amazon Simple Storage Service",
    "serviceId":"S3",
    "signatureVersion":"s3",
    "uid":"s3-2006-03-01",
    "auth":["aws.auth#sigv4"]
  },
```

This will allow for services to support different protocols, for now we will just select the first protocol that is not CBOR as we do not support it yet.

*Check all that applies:*

- [x] Linux
- [x] Windows
- [x] Android
- [x] MacOS
- [ ] IOS
- [ ] Other Platforms


tools/code-generation/generator/src/main/java/com/amazonaws/util/awsclientgenerator/domainmodels/codegeneration/Metadata.java
```diff
@@ -22,7 +24,7 @@ public class Metadata {
     private String signatureVersion;
     private String signingName;
     private String targetPrefix;
-    private String protocol;
```
SergeyRyabinin, line 25: protocol field is still present on the C2J model files.
moreover, old models, that were not recently updated-or-touched, still have protocol field but not the list.

sbiscigl: ahhh thought all of them had them

SergeyRyabinin (approved):

PR #3321, opened 2025-02-28
Pre-requisite Changes Enable smithy query + rest-xml clients

sbera87:
*Issue #, if available:*

*Description of changes:*
This PR includes 
- changes needed to enable xml, query protocol clients and special customized client like RDS
- changes for multi auth support across all protocols (Refactor to minimize code/logic duplication in Java codegen classes)
- backwards compatible changes for generate presign url APIs
- Examples of changes on clients from different protocols.

*Check all that applies:*

- [ ] Linux
- [ ] Windows
- [ ] Android
- [ ] MacOS
- [ ] IOS
- [ ] Other Platforms

### DIFF
--- a/src/aws-cpp-sdk-core/include/smithy/client/AwsSmithyClient.h
+++ b/src/aws-cpp-sdk-core/include/smithy/client/AwsSmithyClient.h
@@ -228,7 +228,7 @@ namespace client
             auto httpResponseOutcome = MakeRequestSync(request, requestName, method, std::move(endpointCallback));
             return m_serializer->Deserialize(std::move(httpResponseOutcome), GetServiceClientName(), requestName);
         }
-
+        
         Aws::String GeneratePresignedUrl(
             EndpointUpdateCallback&& endpointCallback,
             Aws::Http::HttpMethod method,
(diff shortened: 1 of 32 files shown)


tools/code-generation/generator/src/main/java/com/amazonaws/util/awsclientgenerator/generators/cpp/JsonCppClientGenerator.java
```diff
@@ -176,7 +177,7 @@ protected List<SdkFileEntry> generateClientSourceFile( List<ServiceModel> servic
         {
             if(serviceModels.get(index).isUseSmithyClient() && !serviceModels.get(index).hasEventStreamingRequestShapes())
             {
-                return GenerateSmithyClientSourceFile(serviceModels.get(index), index);
+                return GenerateSmithyClientSourceFile(serviceModels.get(index), index, Optional.empty());
```
sbera87, line 180: Reuse base class function

tools/code-generation/generator/src/main/java/com/amazonaws/util/awsclientgenerator/generators/cpp/RestXmlCppClientGenerator.java
```diff
@@ -42,11 +46,39 @@ protected SdkFileEntry generateErrorMarshallerHeaderFile(ServiceModel serviceMod
...
     protected SdkFileEntry generateClientHeaderFile(final ServiceModel serviceModel) throws Exception {
-        Template template = velocityEngine.getTemplate("/com/amazonaws/util/awsclientgenerator/velocity/cpp/xml/XmlServiceClientHeader.vm", StandardCharsets.UTF_8.name());
 
         VelocityContext context = createContext(serviceModel);
         context.put("CppViewHelper", CppViewHelper.class);
         context.put("RequestlessOperations", requestlessOperations);
+        Template template;
+        if (serviceModel.isUseSmithyClient())
+        {
+            template = velocityEngine.getTemplate("/com/amazonaws/util/awsclientgenerator/velocity/cpp/smithy/SmithyClientHeader.vm", StandardCharsets.UTF_8.name());
```
sbera87, line 56: Refactor to use base class function to avoid duplicate code and logic

tools/code-generation/generator/src/main/java/com/amazonaws/util/awsclientgenerator/generators/cpp/dynamodb/DynamoDBJsonCppClientGenerator.java
```diff
@@ -62,7 +63,7 @@ protected List<SdkFileEntry> generateClientSourceFile(List<ServiceModel> service
         serviceModelsIndices.stream().forEach(index -> {
             if(serviceModels.get(index).isUseSmithyClient())
             {
-                smithyClients.add(GenerateSmithyClientSourceFile(serviceModels.get(index), index));
+                smithyClients.add(GenerateSmithyClientSourceFile(serviceModels.get(index), index,Optional.empty()));
```
sbera87, line 66: To make API reusable, accommodate interface change

tools/scripts/codegen/model_utils.py
```diff
@@ -40,6 +40,9 @@
     , "lexv2-runtime"
     , "qbusiness"
     , "transcribestreaming"
+    , "s3-crt"
```
sbera87, line 43: In exclusion list because s3 was addressed in separate PR, and S3-crt,s3control pending

src/aws-cpp-sdk-core/include/smithy/client/AwsSmithyClient.h
```diff
@@ -284,6 +284,49 @@ namespace client
             return {};
         }
 
+        Aws::String GeneratePresignedUrl(
```
sbiscigl, line 287: having four generate presigned URL methods seems excessive and that we are creating bloat in this class. can we do something similar to how we did `AwsLegacyClientT` and move this into a `PresignMixinT` then invoke a single smithy client method?

[In fact we already do this](https://github.com/aws/aws-sdk-cpp/blob/main/src/aws-cpp-sdk-core/include/aws/core/client/AWSUrlPresigner.h) on the old client but have it as a signer to look up. lets just add it as a mix in.

tools/code-generation/generator/src/main/java/com/amazonaws/util/awsclientgenerator/generators/cpp/CppClientGenerator.java
```diff
@@ -718,15 +718,25 @@ protected SdkFileEntry generateClientSmithyHeaderFile(final ServiceModel service
...
+            Optional<String> firstAuthScheme = serviceModel.getAuthSchemes().stream().filter(entry->ResolverMapping.containsKey(entry)).findFirst();
+
+            if(firstAuthScheme.isPresent())
+            {
+                context.put("AuthSchemeResolver", ResolverMapping.get(firstAuthScheme.get()));
+            }
+            else
+            {
+                throw new RuntimeException(String.format("authSchemes '%s'",serviceModel.getAuthSchemes().stream().collect(Collectors.toList())
+                ));
```
sbiscigl, line 738: fix these trailing params, put them on the same line as everything else, it reads weird

src/aws-cpp-sdk-core/include/smithy/client/AwsSmithyClient.h
SergeyRyabinin, line 287: is there a to duplicate this method body?

tools/code-generation/generator/src/main/java/com/amazonaws/util/awsclientgenerator/generators/cpp/CppClientGenerator.java
```diff
@@ -736,23 +746,34 @@ protected SdkFileEntry generateClientSmithyHeaderFile(final ServiceModel service
...
 
         VelocityContext context = createContext(serviceModel);
         context.put("CppViewHelper", CppViewHelper.class);
-        Optional<String> firstAuthScheme = serviceModel.getAuthSchemes().stream().filter(entry->ResolverMapping.containsKey(entry)).findFirst();
-        if(firstAuthScheme.isPresent())
+            
+        if(serviceModel.getAuthSchemes().size() > 1)
         {
-            context.put("AuthSchemeResolver", ResolverMapping.get(firstAuthScheme.get()));
+            context.put("AuthSchemeResolver", "SigV4MultiAuthSchemeResolver");
```
sbiscigl, line 759: why are we hard coding it to sigv4 and no the the first auth scheme? we should be using the first in the list, hard coding it seems wrong

tools/code-generation/generator/src/main/java/com/amazonaws/util/awsclientgenerator/generators/cpp/CppClientGenerator.java
```diff
@@ -736,23 +746,34 @@ protected SdkFileEntry generateClientSmithyHeaderFile(final ServiceModel service
...
         }
         else
         {
-            throw new RuntimeException(String.format("authSchemes '%s'",serviceModel.getAuthSchemes().stream().collect(Collectors.toList())
-            ));
+            Optional<String> firstAuthScheme = serviceModel.getAuthSchemes().stream().filter(entry->ResolverMapping.containsKey(entry)).findFirst();
```
sbiscigl, line 763: this is the third time this has been copied so likely it should be broken off. into its own function

tools/code-generation/generator/src/main/java/com/amazonaws/util/awsclientgenerator/generators/cpp/RestXmlCppClientGenerator.java
```diff
@@ -42,11 +46,39 @@ protected SdkFileEntry generateErrorMarshallerHeaderFile(ServiceModel serviceMod
...
+            if(serviceModel.getAuthSchemes().size() > 1)
```
sbiscigl, line 57: this same block appears all of the place, we need to move it to a single place to avoid code duplicaiton

tools/code-generation/generator/src/main/java/com/amazonaws/util/awsclientgenerator/generators/cpp/CppClientGenerator.java
sbera87: If you think about it, the existing ResolverMapping maps scheme id to an Auth scheme resolver, but in case there are multiple auth schemes detected for a service across different operations and endpoint rules, there is not a single scheme id to map against any resolver, hence it is handled separately to map explicitly to a MultiAuthSchemeResolver class.  Using a first in the list wont work for this case.

tools/code-generation/generator/src/main/java/com/amazonaws/util/awsclientgenerator/generators/cpp/CppClientGenerator.java
sbera87: One dilemma I faced was that different protocol classes may have different template contexts that are also loaded in the same code block. But, let me try to make this a common function.

tools/code-generation/generator/src/main/java/com/amazonaws/util/awsclientgenerator/generators/cpp/RestXmlCppClientGenerator.java
sbera87: Sure, let me address this in next commit

src/aws-cpp-sdk-core/include/smithy/client/AwsSmithyClient.h
sbera87: There are overlaps in the message body for the auth scheme selection bit, but differs in processing uri with request.  I will push a commit with an attempt to reuse the overlapping block.

sbera87: Made the changes to have all legacy presigners in legacy client and smithy client have only one implementation. Under the hood, they all reuse a private implementation .

src/aws-cpp-sdk-core/include/smithy/client/AwsSmithyClient.h
```diff
@@ -222,7 +222,7 @@ namespace client
             auto httpResponseOutcome = MakeRequestSync(request, requestName, method, std::move(endpointCallback));
             return m_serializer->Deserialize(std::move(httpResponseOutcome), GetServiceClientName(), requestName);
         }
-
+        
         Aws::String GeneratePresignedUrl(
```
sbiscigl, line 226: so fan of having on presigner now, only question i really have is why is this in the the Smithy client and the others legcy. I think something that we could talk about is "what the public interface of a pre signer should be". but since this is already here, im open to moving forward without having a answer for that for now

sbiscigl (approved):

SergeyRyabinin (approved):

SergeyRyabinin (approved):

PR #3331, opened 2025-03-06
[ProtocolTests] query sub object serialization refactor

SergeyRyabinin:
*Issue #, if available:*
Protocol tests implementation
*Description of changes:*
Refactor query sub-object serialization to avoid template duplication

It generates an ugly diff on base64 blob serialization, but it is expected:
- base64 may contain + and / characters;
- these characters are reserved per rfc3986 and must be %-encoded. This was done for few cases originally of base64-encodind, but not all, and since this PR unifies how elements are serialized - this effect occurs.

Also it generates an ugly diff on enum serialization, the issue is the same: sometimes we were rfc-encoding enums, sometimes not. While service model owners might be responsible for having their enums rfc-compliant, it is a safer bet for SDK to be compliant and consistent on enum encoding.

*Check all that applies:*

- [x] Linux
- [x] Windows
- [ ] Android
- [ ] MacOS
- [ ] IOS
- [ ] Other Platforms


generated/protocol-tests/test-clients/aws-cpp-sdk-json-protocol/include/aws/json-protocol/JsonProtocolClient.h
```diff
@@ -6,33 +6,21 @@
...
-  class AWS_JSONPROTOCOL_API JsonProtocolClient : smithy::client::AwsSmithyClientT<Aws::JsonProtocol::SERVICE_NAME,
-      Aws::JsonProtocol::JsonProtocolClientConfiguration,
-      smithy::SigV4AuthSchemeResolver<>,
-      Aws::Crt::Variant<smithy::SigV4AuthScheme>,
-      JsonProtocolEndpointProviderBase,
-      smithy::client::JsonOutcomeSerializer,
-      smithy::client::JsonOutcome,
-      Aws::Client::JsonProtocolErrorMarshaller>,
-    Aws::Client::ClientWithAsyncTemplateMethods<JsonProtocolClient>
+  class AWS_JSONPROTOCOL_API JsonProtocolClient : public Aws::Client::AWSJsonClient, public Aws::Client::ClientWithAsyncTemplateMethods<JsonProtocolClient>
```
sbiscigl, line 18: looks like the smithy refactor rollback had unintended consequences here. how should the "protocol clients" handle this migrations/should they?

sbiscigl (approved):

generated/protocol-tests/test-clients/aws-cpp-sdk-json-protocol/include/aws/json-protocol/JsonProtocolClient.h
SergeyRyabinin: I'd say it was a miss when we did the revert.
Once tests are stable - they will be added to a regular regeneration and build-execution list.

sbera87 (approved):
One generic comment I have is to consider as food for thought if it's better to generate different classes following appropriate design pattern to handle 'ec2' vs rest cases. The different classes would come from smaller template files for ec2 vs rest cases. This can help maintainability (due to less branching in code) and readability.

Issue #2949, opened 2024-05-09
DefaultRateLimiter get stuck indefinitely


qqwangxiaow:
### Describe the bug

We have set a 100MB upload rate limit for the AWS S3 SDK, but sometimes the upload threads get stuck 
indefinitely.

`config.writeRateLimiter = Aws::MakeShared<Aws::Utils::RateLimits::DefaultRateLimiter<>>("s3", 100 * 1024 * 1024);
`

stack

#0  0x00007fca25e097fd in nanosleep () at ../sysdeps/unix/syscall-template.S:84
#1  0x00000000005db796 in std::this_thread::sleep_for<long, std::ratio<1l, 1000l> > (__rtime=<synthetic pointer>...) at /usr/include/c++/9/thread:378
#2  Aws::Utils::RateLimits::DefaultRateLimiter<std::chrono::_V2::system_clock, std::chrono::duration<long, std::ratio<1l, 1l> >, true>::ApplyAndPayForCost (this=this@entry=0x3dc4510, cost=614)
    at external/_main~data_deps_ext~aws/aws-cpp-sdk-core/include/aws/core/utils/ratelimiter/DefaultRateLimiter.h:115
#3  0x000000000084e9d1 in Aws::Http::CurlHttpClient::MakeRequestInternal(Aws::Http::HttpRequest&, std::shared_ptr<Aws::Http::Standard::StandardHttpResponse>&, Aws::Utils::RateLimits::RateLimiterInterface*, Aws::Utils::RateLimits::RateLimiterInterface*) const () at external/_main~data_deps_ext~aws/aws-cpp-sdk-core/source/http/curl/CurlHttpClient.cpp:461
#4  0x0000000000851acb in Aws::Http::CurlHttpClient::MakeRequest(std::shared_ptr<Aws::Http::HttpRequest> const&, Aws::Utils::RateLimits::RateLimiterInterface*, Aws::Utils::RateLimits::RateLimiterInterface*) const () at external/_main~data_deps_ext~aws/aws-cpp-sdk-core/source/http/curl/CurlHttpClient.cpp:719
#5  0x000000000099a26c in Aws::Client::AWSClient::AttemptOneRequest(std::shared_ptr<Aws::Http::HttpRequest> const&, Aws::AmazonWebServiceRequest const&, char const*, char const*) const ()
    at /usr/include/c++/9/bits/char_traits.h:372
[...]

Issue #806, opened 2018-02-21
Stuck forever in List Object call with rate limiter


newpoo:
The stack trace is as below. It's sleeping in DefaultRateLimiter for an insanely large number of seconds. The aws-sdk-cpp version is 1.2.7.

The issue goes away if we disable rate limiter.

### What platform/OS are you using?
Ubuntu 14.04

### What compiler are you using? what version?
gcc 5.4

### What's your CMake arguments?
cmake -DCMAKE_BUILD_TYPE=Release -DCUSTOM_MEMORY_MANAGEMENT=0 -DSTATIC_LINKING=1 -DBUILD_ONLY="s3"

### Can you provide a TRACE level log? (sanitize any sensitive information)

#0  0x00007fbc30d44b9d in nanosleep () from /mnt/rockstore/bin/libpthread.so.0
#1  0x0000000000705462 in sleep_for<long, std::ratio<1l, 1000l> > (__rtime=<synthetic pointer>) at /usr/include/c++/5/thread:292
#2  Aws::Utils::RateLimits::DefaultRateLimiter<std::chrono::_V2::system_clock, std::chrono::duration<long, std::ratio<1l, 1l> >, true>::ApplyAndPayForCost (this=0x7fbc18e680d0, cost=5850) at /usr/local/include/aws/core/utils/ratelimiter/DefaultRateLimiter.h:115
#3  0x00007fbc32a43344 in Aws::Http::CurlHttpClient::WriteData(char*, unsigned long, unsigned long, void*) [clone .part.32] () from /mnt/rockstore/bin/libaws-cpp-sdk-core.so
#4  0x00007fbc2d3ebe10 in ?? () from /mnt/rockstore/bin/libcurl.so.4
#5  0x00007fbc2d405700 in ?? () from /mnt/rockstore/bin/libcurl.so.4
#6  0x00007fbc2d3fff37 in ?? () from /mnt/rockstore/bin/libcurl.so.4
#7  0x00007fbc2d409c2c in ?? () from /mnt/rockstore/bin/libcurl.so.4
#8  0x00007fbc2d40a3d1 in curl_multi_perform () from /mnt/rockstore/bin/libcurl.so.4
#9  0x00007fbc2d401a13 in curl_easy_perform () from /mnt/rockstore/bin/libcurl.so.4
[...]

PR #3338, opened 2025-03-10
fix default limiter overflow

qqwangxiaow:
*Issue #, if available:*
#806 
#2949 

*Description of changes:*
The rate limiter refills the token bucket based on the time difference, and the unit is in nanoseconds or smaller. This could potentially cause an int64 overflow, resulting in a negative value, which would lead to an excessively long sleep that cannot recover. Here, we handle the overflow case separately, refilling the token bucket completely when an overflow occurs.

After our service is initialized, not every instance will upload to S3 in real-time. However, a master-slave switch might occur, causing an overflow when calculating the token bucket refill, because time interval exceeds 6 days. it could result in an indefinite sleep.

here is the coredump data
image

*Check all that applies:*

- [x] Linux
- [ ] Windows
- [ ] Android
- [ ] MacOS
- [ ] IOS
- [ ] Other Platforms


sbera87 (commented):
The logic is correct. However, I think we can make the code concise
```
//check for overflow case
if (m_replenishNumerator != 0 &&
    (elapsedTime > std::numeric_limits<int64_t>::max() / m_replenishNumerator ||
     m_accumulatorFraction > std::numeric_limits<int64_t>::max() - (elapsedTime * m_replenishNumerator))) {
    m_accumulator = m_maxRate;
    m_accumulatorFraction = 0;
} else {
    // replenish the accumulator based on how much time has passed
    auto temp = elapsedTime * m_replenishNumerator + m_accumulatorFraction;
    m_accumulator += temp / m_replenishDenominator;
    m_accumulatorFraction = temp % m_replenishDenominator;

    // the accumulator is capped based on the maximum rate
    m_accumulator = std::min(m_accumulator, m_maxRate);
    if (m_accumulator == m_maxRate) {
        m_accumulatorFraction = 0;
    }
}
```

qqwangxiaow:
@sbera87  I fixed the comments, but clang-format seems to have made quite a few changes

sbera87 (approved):

sbiscigl (approved):

SergeyRyabinin (approved):

PR #3341, opened 2025-03-11
fix code gen for retryable errors

sbiscigl:
*Description of changes:*

We were incorrectly generating some internal 500 errors as non-retry-able when in fact they can and should be retried.

*Check all that applies:*

- [x] Linux
- [x] Windows
- [x] Android
- [x] MacOS
- [ ] IOS
- [ ] Other Platforms


tools/code-generation/generator/src/main/java/com/amazonaws/util/awsclientgenerator/transform/C2jModelToGeneratorModelTransformer.java
```diff
@@ -108,6 +108,8 @@ public class C2jModelToGeneratorModelTransformer {
         LEGACY_SERVICE_IDS.add("dynamodbstreams");
     }
 
+    private static Set<Integer> RESPONSE_CODES_TO_RETRY = ImmutableSet.of(500, 502, 503, 504);
```
DmitriyMusatkin, line 111: nit: add a comment on those magical numbers, even if it seems really obvious

sbiscigl: added

SergeyRyabinin (approved):

DmitriyMusatkin (approved):

tools/code-generation/generator/src/main/java/com/amazonaws/util/awsclientgenerator/transform/C2jModelToGeneratorModelTransformer.java
```diff
@@ -936,12 +941,14 @@ Error convertError(C2jError c2jError, String operationName) {
...
-        if (c2jError.getError() != null && c2jError.getError().getCode() != null) {
-            if(c2jError.getError().getHttpStatusCode() >= 500 || !c2jError.getError().isSenderFault()) {
+        if (c2jError.getError() != null) {
+            if(RESPONSE_CODES_TO_RETRY.contains(c2jError.getError().getHttpStatusCode()) && !c2jError.getError().isSenderFault()) {
                 error.setRetryable(true);
             }
 
-            error.setText(c2jError.getError().getCode());
+            if (c2jError.getError().getCode() != null) {
+                error.setText(c2jError.getError().getCode());
```
sbera87, line 950: Previously, for query protocol, we would set text for ((errors >= 500) || (errors not sender fault))
After the change, I understand that error is marked retryable for all protocols for select 5XX errors and if not sender fault. But, it also changes the logic for setText, which I am not sure is desired. 
i.e. now we will set text for any error for query protocol. Is this expected?

sbiscigl: > i.e. now we will set text for any error for query protocol. Is this expected?

yes we want to use the code from the query xml error code if it is present i.e. the comment above it
`//query xml loads this inner structure to do this work.`

additionally you can see query xml service remain unchanged/error text as a whole remains unchanged in this change

sbiscigl: actuallty even better from [smithy docs](https://smithy.io/2.0/aws/protocols/aws-query-protocol.html#aws-protocols-awsqueryerror-trait)

> aws.protocols#awsQueryError trait
Summary
Provides a custom "Code" value for awsQuery errors and an HTTP response code. The "Code" of an awsQuery error is used by clients to determine which type of error was encountered.

sbera87 (approved):

PR #3186, opened 2024-11-11
fix trailing checksum

sbiscigl:
*Issue #, if available:*
[issues/3132](https://github.com/aws/aws-sdk-cpp/issues/3132)

*Description of changes:*
There was a nasty bit of undefined behavior in the aws-chunked logic for curl. The original implementation assumed that [curl would always provide us with a chunk](https://github.com/aws/aws-sdk-cpp/blob/main/src/aws-cpp-sdk-core/source/http/curl/CurlHttpClient.cpp#L333) larger than 8kb to per ReadFunc callback. That is simply not true, and it will cause requests where curl calls readfunc with less than 8kb to fail consistently.This buffers the input stream and outputs to curl valid sized chunks, specifically 64kb chunks as [S3 specifies as their preference](https://docs.aws.amazon.com/AmazonS3/latest/API/sigv4-streaming.html)

>  The chunk size must be at least 8 KB. We recommend a chunk size of a least 64 KB for better performance. This chunk size applies to all chunks except the last one. The last chunk you send can be smaller than 8 KB. If your payload is small and can fit into one chunk, then it can be smaller than the 8 KB.

*Check all that applies:*
- [x] Did a review by yourself.
- [x] Added proper tests to cover this PR. (If tests are not applicable, explain.)
- [x] Checked if this PR is a breaking (APIs have been changed) change.
- [x] Checked if this PR will _not_ introduce cross-platform inconsistent behavior.
- [x] Checked if this PR would require a ReadMe/Wiki update.

Check which platforms you have built SDK on to verify the correctness of this PR.
- [x] Linux
- [x] Windows
- [x] Android
- [x] MacOS
- [ ] IOS
- [ ] Other Platforms

By submitting this pull request, I confirm that you can use, modify, copy, and redistribute this contribution, under the terms of your choice.

PR #3348, opened 2025-03-18
Update crt fix chunked

sbiscigl:
*Description of changes:*

* updates the CRT to version `v0.31.1`
* Fixes aws-chunked encoding in the CRT http cient, it did not work before.

A simple put object/get object with the CRT HTTP client and vanillia s3 client does not work before this patch as aws-chunked encoding was broken

i.e.

```c++
#include <aws/core/Aws.h>
#include <aws/s3/S3Client.h>
#include <aws/s3/model/GetObjectRequest.h>
#include <aws/s3/model/PutObjectRequest.h>

using namespace Aws;
using namespace Aws::Utils;
using namespace Aws::S3;

namespace {
const char* LOG_TAG = "test-application";
}

auto main() -> int {
  SDKOptions options{};
  InitAPI(options);
  {
    S3Client client{};
    auto request = PutObjectRequest();
    request.SetBucket("sbiscigl");
    request.SetKey("test-crt-http-client");
    auto body = Aws::MakeShared<StringStream>(LOG_TAG, "some test string body");
    request.SetBody(body);
    const auto response = client.PutObject(request);
    assert(response.IsSuccess());
    auto get_object_request = GetObjectRequest();
    get_object_request.SetBucket(BUCKET_NAME);
    get_object_request.SetKey(KEY_NAME);
    const auto get_object_result = client.GetObject(get_object_request);
    assert(get_object_result.IsSuccess());
    std::cout << get_object_result.GetResult().GetBody().rdbuf() << std::endl;
  }
  ShutdownAPI(options);
  return 0;
}
```

This fixes the CRT HTTP client to use the same patch that we used for the original curl issuse(https://github.com/aws/aws-sdk-cpp/pull/3186).

*Check all that applies:*

- [x] Linux
- [x] Windows
- [x] Android
- [x] MacOS
- [ ] IOS
- [ ] Other Platforms

### DIFF
--- a/src/aws-cpp-sdk-core/include/aws/core/utils/stream/AwsChunkedStream.h
+++ b/src/aws-cpp-sdk-core/include/aws/core/utils/stream/AwsChunkedStream.h
@@ -6,6 +6,7 @@
 #include <aws/core/utils/HashingUtils.h>
 #include <aws/core/utils/Outcome.h>
 #include <aws/core/utils/crypto/Hash.h>
+#include <aws/crt/io/Stream.h>
 
 namespace Aws {
 namespace Utils {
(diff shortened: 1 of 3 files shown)


SergeyRyabinin (approved):

src/aws-cpp-sdk-core/include/aws/core/utils/stream/AwsChunkedStream.h
```diff
@@ -56,6 +60,14 @@ class AwsChunkedStream {
     return static_cast<size_t>(m_chunkingStream->gcount());
   }
 
+protected:
+  bool ReadImpl(Crt::ByteBuf &buffer) noexcept override {
+    size_t amountToRead = buffer.capacity - buffer.len;
+    const auto bytesRead = BufferedRead(reinterpret_cast<char *>(buffer.buffer), amountToRead);
+    buffer.len += bytesRead;;
+    return bytesRead > 0;
```
graebm, line 68: https://github.com/awslabs/aws-crt-cpp/blob/743ab3891c4fb015a74713b7b631c720ce0ffe8e/include/aws/crt/io/Stream.h#L108-L113
```
                 * @return true if nothing went wrong.
                 * Return true even if you read 0 bytes because the end-of-file has been reached.
                 * Return true even if you read 0 bytes because data is not currently available.
                 *
                 * Return false if an actual failure condition occurs,
                 * you SHOULD also raise an error via aws_raise_error().
```

SergeyRyabinin (approved):

src/aws-cpp-sdk-core/include/aws/core/utils/stream/AwsChunkedStream.h
sbiscigl: updated

Issue #2908, opened 2024-03-27
curl-originated error messages not informative enough


pitrou:
### Describe the bug

While trying to debug a TLS certificate issue, I found out that the AWS SDK returned poorly informative error messages, such as:
```
OSError: When listing buckets: AWS Error NETWORK_CONNECTION during ListBuckets operation: curlCode: 60, SSL peer certificate or SSH remote key was not OK
```

I then tried using the `curl` command line directly and, to my surprise, the error message was much more informative:
```
curl: (60) SSL: no alternative certificate subject name matches target host name '127.0.0.1'
More details here: https://curl.se/docs/sslcerts.html

curl failed to verify the legitimacy of the server and therefore could not
establish a secure connection to it. To learn more about this situation and
how to fix it, please visit the web page mentioned above.
```

I inquired on the [curl issue tracker](https://github.com/curl/curl/issues/13199) and it turns out that the AWS SDK should use the [`CURLOPT_ERRORBUFFER`](https://curl.se/libcurl/c/CURLOPT_ERRORBUFFER.html) option so as to get the detailed error message.

### Expected Behavior

The AWS SDK should give out the detailed error made available by libcurl, such as "no alternative certificate subject name matches target host name: ...".

### Current Behavior

The AWS SDK gives out an unhelpful error message, as returned by libcurl when `CURLOPT_ERRORBUFFER` is not set.

### Reproduction Steps

Sorry, I have nothing self-contained unless you think of installing PyArrow as "self-contained". However, this is straightforward enough that it probably doesn't need reproduction?

In any case, see https://github.com/apache/arrow/issues/37001#issuecomment-2022451857 for an example.

### Possible Solution

The AWS SDK could reserve a dedicated error buffer for each libcurl connection and set the CURLOPT_ERRORBUFFER with it.

### AWS CPP SDK version used

1.11.267, but current git main seems affected

### Compiler and Version used

gcc 12.3.0

### Operating System and version

Ubuntu 22.04

jmklix:
This PR should help with this: https://github.com/aws/aws-sdk-cpp/pull/2676

pitrou:
Thank you @SergeyRyabinin !

PR #3349, opened 2025-03-19
use curl error buffer for more informative error message

SergeyRyabinin:
*Issue #, if available:*
https://github.com/aws/aws-sdk-cpp/issues/2908
https://github.com/aws/aws-sdk-cpp/pull/2912
*Description of changes:*

Uses [CURLOPT_ERRORBUFFER](https://curl.se/libcurl/c/CURLOPT_ERRORBUFFER.html) as a more informative source of error messages for network errors when encountered. if error buffer is not available falls back to old way, which is also the recommended curl method via the example in the error buffer documentation.

example error message before:
`"curlCode: 6, Couldn't resolve host name"`

after
`"curlCode: 6, Couldn't resolve host name; Details: Could not resolve host: 0xxxabcdefg123456789.com"`

*Check all that applies:*

- [x] Linux
- [ ] Windows
- [ ] Android
- [ ] MacOS
- [ ] IOS
- [ ] Other Platforms

### DIFF
--- a/src/aws-cpp-sdk-core/source/http/curl/CurlHttpClient.cpp
+++ b/src/aws-cpp-sdk-core/source/http/curl/CurlHttpClient.cpp
@@ -849,6 +849,14 @@ std::shared_ptr<HttpResponse> CurlHttpClient::MakeRequest(const std::shared_ptr<
             }
         }
 
+        ByteBuffer errorBuffer(CURL_ERROR_SIZE);
+        if (errorBuffer.GetUnderlyingData() && errorBuffer.GetSize() >= CURL_ERROR_SIZE) {
+          errorBuffer[0] = '\0';
+          curl_easy_setopt(connectionHandle, CURLOPT_ERRORBUFFER, errorBuffer.GetUnderlyingData());
+        } else {
+          AWS_LOGSTREAM_ERROR(CURL_HTTP_CLIENT_TAG, "Failed to allocate CURLOPT_ERRORBUFFER");
+        }
+
         OverrideOptionsOnConnectionHandle(connectionHandle);
         Aws::Utils::DateTime startTransmissionTime = Aws::Utils::DateTime::Now();
         CURLcode curlResponseCode = curl_easy_perform(connectionHandle);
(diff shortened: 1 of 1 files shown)


src/aws-cpp-sdk-core/source/http/curl/CurlHttpClient.cpp
```diff
@@ -858,9 +866,17 @@ std::shared_ptr<HttpResponse> CurlHttpClient::MakeRequest(const std::shared_ptr<
             response->SetClientErrorType(CoreErrors::NETWORK_CONNECTION);
             Aws::StringStream ss;
             ss << "curlCode: " << curlResponseCode << ", " << curl_easy_strerror(curlResponseCode);
+            if (errorBuffer.GetUnderlyingData() && errorBuffer.GetSize() >= CURL_ERROR_SIZE) {
+              ss << "; Details: ";
+              for(size_t i = 0; i < CURL_ERROR_SIZE; ++i) {
```
sbiscigl, line 871: any reason we have to loop over the error buffer instead of just streaming? the following works as expected

```c++
#include <aws/core/Aws.h>

using namespace Aws;
using namespace Aws::Utils;

auto main() -> int {
  SDKOptions options{};
  InitAPI(options);
  {
    ByteBuffer buffer{"some error string"};
    std::cout << buffer.GetUnderlyingData() << "\n";
  }
  ShutdownAPI(options);
  return 0;
}
```

src/aws-cpp-sdk-core/source/http/curl/CurlHttpClient.cpp
```diff
@@ -976,6 +992,8 @@ std::shared_ptr<HttpResponse> CurlHttpClient::MakeRequest(const std::shared_ptr<
             AWS_LOGSTREAM_ERROR(CURL_HTTP_CLIENT_TAG, ss.str());
         }
         request->AddRequestMetric(GetHttpClientMetricNameByType(HttpClientMetricsType::RequestLatency), (DateTime::Now() - startTransmissionTime).count());
+
+        curl_easy_setopt(connectionHandle, CURLOPT_ERRORBUFFER, nullptr);
```
sbiscigl, line 996: would be kind of neat to put this in a RAII container so we dont forget about setting to null ptr i.e.
```cpp
class CurlErrorBuffRAII {
 public:
  explicit CurlErrorBuffRAII(CURL* connectionHandle) : m_errorHandle{connectionHandle} {
    if (m_errorBuffer.GetUnderlyingData() && m_errorBuffer.GetSize() >= CURL_ERROR_SIZE) {
      m_errorBuffer[0] = '\0';
      curl_easy_setopt(m_errorHandle, CURLOPT_ERRORBUFFER, m_errorBuffer.GetUnderlyingData());
    } else {
      AWS_LOGSTREAM_ERROR(CURL_HTTP_CLIENT_TAG, "Failed to allocate CURLOPT_ERRORBUFFER");
    }
  }

  ByteBuffer& AccessBuffer() { return m_errorBuffer; }

  ~CurlErrorBuffRAII() { curl_easy_setopt(m_errorHandle, CURLOPT_ERRORBUFFER, nullptr); }

 private:
  ByteBuffer m_errorBuffer{CURL_ERROR_SIZE};
  CURL* m_errorHandle{nullptr};
};
```

sbiscigl (approved):

src/aws-cpp-sdk-core/source/http/curl/CurlHttpClient.cpp
SergeyRyabinin: https://curl.se/libcurl/c/CURLOPT_ERRORBUFFER.html

It is just a defensive programming against libcurl not nulling the buffer.
Well I'm already setting the null terminator at the beginning, probably it could replace the loop.

PR #3374, opened 2025-04-04
only check response checksum for returned header

sbiscigl:
*Description of changes:*

Only updates response validation hashes for the first checksum in the response

*Check all that applies:*

- [x] Linux
- [x] Windows
- [x] Android
- [x] MacOS
- [ ] IOS
- [ ] Other Platforms

### DIFF
--- a/src/aws-cpp-sdk-core/source/http/crt/CRTHttpClient.cpp
+++ b/src/aws-cpp-sdk-core/source/http/crt/CRTHttpClient.cpp
@@ -246,13 +246,19 @@ namespace Aws
             }
 
             //TODO: handle the read rate limiter here, once back pressure is setup.
+            assert(response);
             for (const auto& hashIterator : request->GetResponseValidationHashes())
             {
+              std::stringstream headerStr;
+              headerStr<<"x-amz-checksum-"<<hashIterator.first;
+              if(response->HasHeader(headerStr.str().c_str()))
+              {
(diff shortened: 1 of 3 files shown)


sbera87 (approved):

SergeyRyabinin (approved):

src/aws-cpp-sdk-core/source/http/crt/CRTHttpClient.cpp
```diff
@@ -246,13 +246,19 @@ namespace Aws
             }
 
             //TODO: handle the read rate limiter here, once back pressure is setup.
+            assert(response);
             for (const auto& hashIterator : request->GetResponseValidationHashes())
             {
+              std::stringstream headerStr;
+              headerStr<<"x-amz-checksum-"<<hashIterator.first;
```
jeking3, line 253: An improvement here would be to embed the c string representation of the header into the hash class, avoiding a stringstream to string to c string for every buffer that arrives.

src/aws-cpp-sdk-core/source/http/crt/CRTHttpClient.cpp
```diff
@@ -246,13 +246,19 @@ namespace Aws
             }
 
             //TODO: handle the read rate limiter here, once back pressure is setup.
+            assert(response);
             for (const auto& hashIterator : request->GetResponseValidationHashes())
```
jeking3, line 250: An improvement here would be to delay hash initialization until it's first used.

PR #3377, opened 2025-04-10
Remove maps built on each call of some conversion functions.

svart-riddare:
The said map should have been static as a 'TODO' in the code explains. Since the static keyword was removed when the `std::map` was changed to an `Aws::Map`, the map is rebuilt at each calls, which is a lot of work, and calls to `malloc`/`free`. Although the search is now linear instead of logarithmic, the number of entries is quite small and it's always better than rebuilding the map at each call.

This saves about 300 mallocs per PutObject request for example.

*Issue #, if available:*

*Description of changes:*

*Check all that applies:*

- [x] Linux
- [x] Windows
- [ ] Android
- [ ] MacOS
- [ ] IOS
- [ ] Other Platforms

### DIFF
--- a/src/aws-cpp-sdk-core/source/monitoring/HttpClientMetrics.cpp
+++ b/src/aws-cpp-sdk-core/source/monitoring/HttpClientMetrics.cpp
@@ -3,7 +3,7 @@
  * SPDX-License-Identifier: Apache-2.0.
  */
 
-#include <aws/core/utils/HashingUtils.h>
+#include <aws/core/utils/memory/stl/AWSArray.h>
 #include <aws/core/monitoring/HttpClientMetrics.h>
 
 namespace Aws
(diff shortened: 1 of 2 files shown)


src/aws-cpp-sdk-core/source/monitoring/HttpClientMetrics.cpp
```diff
@@ -18,64 +18,41 @@ namespace Aws
...
         static const char HTTP_CLIENT_METRICS_SSL_LATENCY[] = "SslLatency";
-        static const char HTTP_CLIENT_METRICS_THROUGHPUT[] = "Throughput";
         static const char HTTP_CLIENT_METRICS_DOWNLOAD_SPEED[] = "DownloadSpeed";
+        static const char HTTP_CLIENT_METRICS_THROUGHPUT[] = "Throughput";
         static const char HTTP_CLIENT_METRICS_UPLOAD_SPEED[] = "UploadSpeed";
         static const char HTTP_CLIENT_METRICS_UNKNOWN[] = "Unknown";
 
-        using namespace Aws::Utils;
-        HttpClientMetricsType GetHttpClientMetricTypeByName(const Aws::String& name)
+        static const Aws::Array<std::pair<HttpClientMetricsType, const char*>, 12> httpClientMetricsNames =
```
svart-riddare, line 26: Not sure which casing I should use.

SergeyRyabinin (approved):

src/aws-cpp-sdk-core/source/monitoring/HttpClientMetrics.cpp
```diff
@@ -18,64 +18,41 @@ namespace Aws
...
-            };
-
-            auto it = metricsTypeToName.find(static_cast<int>(type));
-            if (it == metricsTypeToName.end())
-            {
-                return HTTP_CLIENT_METRICS_UNKNOWN;
-            }
-            return Aws::String(it->second.c_str());
+            assert(httpClientMetricsNames[static_cast<int>(type)].first == type);
+            return Aws::String(httpClientMetricsNames[static_cast<int>(type)].second);
```
sbiscigl, line 55: would prefer you to use the `std::find` lookup i.e.

```cpp
const auto metricName = std::find_if(httpClientMetricsNames.begin(),
  httpClientMetricsNames.end(),
  [&type](const std::pair<HttpClientMetricsType, const char *>& nameEntry) -> bool { return nameEntry.first == type; });
if(metricName == httpClientMetricsNames.end()) {
  return HTTP_CLIENT_METRICS_UNKNOWN;
}
return metricName->second;
```

as this preserves the behavior from before. i recognize that the map logically doesnt have any illegal access, but this makes sure that there is a defined behavior for when none are found when assertions are disabled.

sbiscigl (commented):
One comment to use `std::find_if` to preserve old behavior, but outside of that, really appriciate you taking the time to address a TODO that we left. sorry that we left it, and thanks for taking the time to do it.

src/aws-cpp-sdk-core/source/monitoring/HttpClientMetrics.cpp
svart-riddare: Hi, I was unsure whether you wanted to prevent an array out of bound index, or to prevent someone adding some entries in the constant table not in the correct order.

I have added a check for this first case; I can of course put an `std::find` if you want to prevent against the second case.

sbiscigl: > or to prevent someone adding some entries in the constant table not in the correct order.

yeah my main concern is that someone edits this out of order, creates a binary incompatibility and it shows up in downstream package managers. I think we can change it ourselves to std::find if we find that is the case down the road, and leave the PR as is for now

PR #3381, opened 2025-04-16
add user agent business metric for account id endpoints

sbiscigl:
*Description of changes:*

Adds business metrics for the account id endpoint feature. Moves account id configuraiton from client specific paramters to general parameters as we should have initially.

*Check all that applies:*

- [x] Linux
- [x] Windows
- [x] Android
- [x] MacOS
- [ ] IOS
- [ ] Other Platforms


src/aws-cpp-sdk-core/source/client/ClientConfiguration.cpp
```diff
@@ -278,12 +280,21 @@ void setConfigFromEnvOrProfile(ClientConfiguration &config)
...
+        AWS_ACCOUNT_ID_ENDPOINT_MODE_CONFIG_FILE_OPTION,
+        {"required", "disabled", "preferred"}, /* allowed values */
+        "preferred" /* default value */);
 }
 
 ClientConfiguration::ClientConfiguration()
 {
     this->disableIMDS = false;
     setLegacyClientConfigurationParameters(*this);
+    setConfigFromEnvOrProfile(*this);;
```
sbiscigl, line 297: interesting fact -- this meant functionally that the `disableIMDSv1` configuration did not work if a user set a region in configuration

sbera87 (approved):

src/aws-cpp-sdk-core/include/aws/core/client/ClientConfiguration.h
```diff
@@ -465,6 +465,16 @@ namespace Aws
...
               bool useAnonymousAuth = false;
             } winHTTPOptions;
+
+          /**
+            * The AWS AccountId used for the request.
+            */
+          Aws::String accountId;
+
+          /**
+           * The AccountId Endpoint Mode.
```
SergeyRyabinin, line 475: the comparison is done by
```
strcmp(mapping.first, accountIdMode.c_str()) == 0; });
```

i.e. not case insensitive.

In an ideal world, I'd suggest to use an enum here.
On practice, please just add a comment with possible allowed values and the requirements of the lowercase.

SergeyRyabinin (approved):

src/aws-cpp-sdk-core/include/aws/core/client/ClientConfiguration.h
sbiscigl: I agree this is a better user experience, however this would be backwards breaking because on main because of how codegen is currently generating it.

```cpp
#include <aws/core/Aws.h>
#include <aws/dynamodb/DynamoDBClient.h>

using namespace Aws;
using namespace Aws::DynamoDB;

auto main() -> int {
  SDKOptions options;
  InitAPI(options);
  {
    DynamoDBClientConfiguration config;
    config.accountIdEndpointMode = "preferred";
    DynamoDBClient client{config};
  }
  ShutdownAPI(options);
  return 0;
}
```

> In an ideal world/On practice

yep yep, will add comment regarding

SergeyRyabinin (approved):

Issue #3389, opened 2025-04-23
fatal error: 'aws/bedrock-runtime/model/InternalServerException.h' file not found


Chen188:
### Describe the bug

Failed to compile BedrockRuntime ConverseStream sample code. Looks like the CPP SDK lakes of `InternalServerException.h` file, which is included in `aws/bedrock-runtime/model/ConverseStreamOutput.h`.

### Regression Issue

- [ ] Select this option if this issue appears to be a regression.

### Expected Behavior

Fix the error: `'aws/bedrock-runtime/model/InternalServerException.h' file not found`

### Current Behavior

```
% cmake .
-- Found AWS SDK for C++, Version: 1.11.552, Install Root:/usr/local, Platform Prefix:, Platform Dependent Libraries: pthread;curl
-- Components specified for AWSSDK: bedrock-runtime;core, application will be depending on libs: aws-cpp-sdk-bedrock-runtime;aws-cpp-sdk-core
-- Try finding aws-cpp-sdk-core
-- Performing Test CMAKE_HAVE_LIBC_PTHREAD
-- Performing Test CMAKE_HAVE_LIBC_PTHREAD - Success
-- Found Threads: TRUE
-- Found aws-cpp-sdk-core
-- Try finding aws-cpp-sdk-bedrock-runtime
-- Found aws-cpp-sdk-bedrock-runtime
-- Configuring done (0.9s)
-- Generating done (0.0s)
-- Build files have been written to: /Users/xx/test-cpp-sdk-bedrock/build
```

```
% make

[ 50%] Building CXX object CMakeFiles/converse_stream_sample.dir/converse_stream_sample.cpp.o
In file included from /Users/xx/test-cpp-sdk-bedrock/converse_stream_sample.cpp:4:
/usr/local/include/aws/bedrock-runtime/model/ConverseStreamOutput.h:14:10: fatal error: 'aws/bedrock-runtime/model/InternalServerException.h' file not found
   14 | #include <aws/bedrock-runtime/model/InternalServerException.h>
      |          ^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
1 error generated.
make[2]: *** [CMakeFiles/converse_stream_sample.dir/converse_stream_sample.cpp.o] Error 1
make[1]: *** [CMakeFiles/converse_stream_sample.dir/all] Error 2
make: *** [all] Error 2
```

### Reproduction Steps

[...]

PR #3391, opened 2025-04-23
Add generator to generate tests that service client headers are buildable

SergeyRyabinin:
*Issue #, if available:*
https://github.com/aws/aws-sdk-cpp/issues/3389
*Description of changes:*
Prevent such issues by a include all headers test.
*Check all that applies:*

- [ ] Linux
- [ ] Windows
- [ ] Android
- [ ] MacOS
- [ ] IOS
- [ ] Other Platforms


tools/scripts/codegen/include_tests_util.py
```diff
@@ -0,0 +1,125 @@
...
+        return IncludeTestsUtil._generate_test(headers, namespace, cpp_name, service_name, test_output_dir)
+
+    @staticmethod
+    def _collect_headers(service_client_dir: str) -> list:
+        service_client_include_dir = f"{service_client_dir}/include/"
+        headers = []
+
+        for root, dirs, files in os.walk(service_client_include_dir):
+            for filename in files:
+                if filename.endswith(".h") or filename.endswith(".hpp"):
```
sbiscigl, line 87: do any of our files actually end with .hpp?

tools/scripts/codegen/include_tests_util.py
```diff
@@ -0,0 +1,125 @@
...
+        rendered_test = jinja2_template.render(t_headers=headers,
+                                               t_test_name=test_name,
+                                               t_namespace=namespace,
+                                               t_cpp_name=cpp_name)
+        test_file_path = f"{test_output_dir}/{test_name}s.cpp"
+
+        with open(f"{test_file_path}", mode="w", encoding="utf-8") as rendered_test_f:
+            rendered_test_f.write(rendered_test)
+
+        return 0
```
sbiscigl, line 106: nit: why return 0 if we dont return any other value? maybe prefer void

sbiscigl (approved):

tools/scripts/codegen/include_tests_util.py
SergeyRyabinin: nope, but if ever anyone is going to add an .hpp - this generator will pick them up.

tools/scripts/codegen/include_tests_util.py
SergeyRyabinin: just following the pattern of "generator return 0" on success, and throws an exception on error (and something such as return -1 in the middle).

PR #3398, opened 2025-04-28
handle brackets in ipv6 authority string

sbiscigl:
*Description of changes:*

Right now a ipv6 override with brackets does not function correctly, i.e.

```cpp
#include <aws/core/Aws.h>
#include <aws/s3/S3Client.h>

using namespace Aws;
using namespace Aws::S3;
using namespace Aws::Utils;

namespace {
const char* LOG_TAG = "test-app";
}

auto main() -> int {
  SDKOptions options{};
  InitAPI(options);
  {
    S3ClientConfiguration client_configuration{};
    client_configuration.endpointOverride = "http://[::]:9000/";
    S3Client client{client_configuration};
    const auto list_buckets_response = client.ListBuckets();
    if (list_buckets_response.IsSuccess()) {
      std::cout << "Found buckets:\n";
      for (const auto& bucket : list_buckets_response.GetResult().GetBuckets()) {
        std::cout << bucket.GetName() << "\n";
      }
    } else {
      std::cout << "Call failed\n";
      std::cout << list_buckets_response.GetError().GetExceptionName() << "\n";
      std::cout << list_buckets_response.GetError().GetMessage() << "\n";
    }
  }
  ShutdownAPI(options);
  return 0;
}
```

This is because we pass the result of `GetAuthority` from our [URI class](https://github.com/aws/aws-sdk-cpp/blob/main/src/aws-cpp-sdk-core/include/aws/core/http/URI.h#L73-L76) to check if the host is valid. This authority included the brackets.

In this change we now strip out the brackets that we pass to CRT when we check for host validity.

*Check all that applies:*

- [x] Linux
- [x] Windows
- [x] Android
- [x] MacOS
- [ ] IOS
- [ ] Other Platforms

### DIFF
--- a/src/aws-cpp-sdk-core/include/aws/core/http/URI.h
+++ b/src/aws-cpp-sdk-core/include/aws/core/http/URI.h
@@ -71,12 +71,14 @@ namespace Aws
             void SetScheme(Scheme value);
 
             /**
-            * Gets the domain portion of the uri
+            * Gets the authority portion of the URI. Differs from the authority definition in RFC 3986. user information
+            * and the port information are not included. It is functionally the host as defined by  RFC 3986 with encoding included.
             */
             inline const Aws::String& GetAuthority() const { return m_authority; }
 
             /**
-            * Sets the domain portion of the uri
+            * Sets the authority portion of the URI. Differs from the authority definition in RFC 3986. user information
+            * and the port information are not included. It is functionally the host as defined by  RFC 3986 with encoding included.
             */
             inline void SetAuthority(const Aws::String& value) { m_authority = value; }
 
@@ -212,6 +214,17 @@ namespace Aws
              * URLEncodes the path portion of the URI according to RFC3986
              */
             static Aws::String URLEncodePathRFC3986(const Aws::String& path, bool rfcCompliantEncoding = false);
+
+            /**
+             * The host portion of the URI as described in rfc3986.
+             *
+             * The host subcomponent of authority is identified by an IPv6 literal
+             * encapsulated within square brackets, an IPv4 address in dotted-
+             * decimal form, or a registered name.
+             *
+             * @return The host portion of the URI
+             */
+            Aws::String GetHost() const;
 
         private:
             void ParseURIParts(const Aws::String& uri);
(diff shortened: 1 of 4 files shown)


src/aws-cpp-sdk-core/source/utils/DNS.cpp
```diff
@@ -7,6 +7,21 @@
...
 #include <aws/crt/DnsUtils.h>
 
+namespace {
+  /**
+   * Extracts the IPv6 authority by removing brackets. returns the authority
+   * unchanged if brackets are invalid or not present.
+   */
+  Aws::String ExtractIpv6Authority(const Aws::String& authority) {
+    const auto begin = authority.find('[');
+    const auto end = authority.find(']', begin);
```
sbera87, line 17: nit: This is more efficient
`const auto end = authority.rfind(']');`

src/aws-cpp-sdk-core/source/utils/DNS.cpp
```diff
@@ -7,6 +7,21 @@
...
+    if (begin == Aws::String::npos || end == Aws::String::npos) {
+      return authority;
+    }
+    return authority.substr(begin + 1, end - 1);
```
sbera87, line 21: This can be a bug when` []` leads to empty string

sbiscigl: thats also not a valid IPv6 string so im not that concerned

SergeyRyabinin (approved):

SergeyRyabinin (approved):

src/aws-cpp-sdk-core/include/aws/core/http/URI.h
```diff
@@ -213,6 +215,17 @@ namespace Aws
              */
             static Aws::String URLEncodePathRFC3986(const Aws::String& path, bool rfcCompliantEncoding = false);
 
+            /**
+             * The host portion of the URI as described in rfc3986.
+             *
+             * The host subcomponent of authority is identified by an IP literal
```
DmitriyMusatkin, line 221: nit: ipv6 in square brackets

src/aws-cpp-sdk-core/include/aws/core/http/URI.h
```diff
@@ -71,12 +71,14 @@ namespace Aws
             void SetScheme(Scheme value);
 
             /**
-            * Gets the domain portion of the uri
+            * Gets the authority portion of the URI. Differs from the authority definition in RFC 3986. user information
```
DmitriyMusatkin, line 74: nit: maybe add that its essentially host part of authority with encoding still present

DmitriyMusatkin (approved):

PR #3414, opened 2025-05-08
Protocol tests parsed response validation

SergeyRyabinin:
*Issue #, if available:*

*Description of changes:*

*Check all that applies:*

- [ ] Linux
- [ ] Windows
- [ ] Android
- [ ] MacOS
- [ ] IOS
- [ ] Other Platforms


tools/scripts/run_protocol_tests.py
```diff
@@ -0,0 +1,136 @@
...
+    def __del__(self):
+        """Destructor that sends Ctrl-C/SIGINT to the subprocess."""
+        if self.process and self.process.poll() is None:
+            print(f"Sending SIGINT to subprocess with PID {self.process.pid}")
+            if os.name == 'nt':  # Windows
+                self.process.send_signal(signal.CTRL_C_EVENT)
+            else:  # Unix-like
+                self.process.send_signal(signal.SIGINT)
+
+            time.sleep(0.5)
```
sbiscigl, line 55: do we need static sleep? what are we waiting for?

tools/scripts/run_protocol_tests.py
```diff
@@ -0,0 +1,136 @@
...
+            print(f"Mock http logs:\n{self.process.communicate()[0]}\n")
+
+    def is_running(self):
+        """Check if the subprocess is still running."""
+        return self.process.poll() is None if self.process else False
+
+    def wait(self):
+        """Wait for the subprocess to complete."""
+        if self.process:
+            return self.process.wait()
```
sbiscigl, line 72: [waits without timeouts](https://docs.python.org/3/library/subprocess.html#subprocess.Popen.wait) have heavily traumatized my up to this point, lets add a reasonable timeout to prevent zombie hosts

tools/code-generation/generator/src/main/java/com/amazonaws/util/awsclientgenerator/generators/cpp/CppProtocolTestGenerator.java
```diff
@@ -157,15 +159,48 @@ private List<String> computeTestSourceIncludes(ProtocolTestSuite testSuite) {
...
 
                     }
-                    // include the request shape itself
+                    return includes;
+                })
+                .flatMap(entrySet -> entrySet.stream())
+                .collect(Collectors.toSet());
+        // include the request shapes
+        Set<String> requestShapes = testSuite.getCases().stream()
+                .map(entry -> {
```
sbiscigl, line 168: looks like we only ever add one item to the set ever, simplified this could just be
```java
testSuite.getCases().stream()
  .map(entry -> { return String.format("<aws/%s/model/%s.h>", 
      serviceModel.getMetadata().getProjectName(),
      entry.getGiven().getName() + "Request")})
  .collect(Collectors.toSet());
````
to avoid creating a extra set per test case

File: tools/code-generation/generator/src/main/resources/com/amazonaws/util/awsclientgenerator/velocity/cpp/protocoltests/result/ValidateOutcome.vm [hunk 4]
```diff
@@ -0,0 +1,19 @@
+#parse("com/amazonaws/util/awsclientgenerator/velocity/cpp/protocoltests/result/ValidateParamsMacro.vm")
+#if($case.input || $case.output.get().successResult)
+  AWS_ASSERT_SUCCESS(outcome) << outcome.GetError();
+#if($case.output && $case.output.get().successResult)
+#set($expectedResult = $case.output.get().successResult.get().result.get())
+#set($responseShape = $serviceModel.operations.get("${case.given.name}").result.shape)
+#if(!$expectedResult.isEmpty())
+  const ${responseShape.name}& result = outcome.GetResult();
+#end
+  /* expectedResult = R"( ${expectedResult} )" */
```
sbiscigl, line 10: love the debug comment

sbiscigl (approved):

tools/code-generation/generator/src/main/java/com/amazonaws/util/awsclientgenerator/generators/cpp/CppProtocolTestGenerator.java
```diff
@@ -157,15 +159,48 @@ private List<String> computeTestSourceIncludes(ProtocolTestSuite testSuite) {
...
+
+        set.addAll(requestShapes);
         set.removeAll(HEADERS_TO_NOT_INCLUDE);
         return set.stream().sorted().collect(Collectors.toList());
     }
+
+    /**
+     * Perform legacy patching of the ec2 model present from the very beginning.
+     */
+    private void legacyPatchEc2Model() {
```
sbera87, line 186: I think this can be optimized to use one loop

```
private void legacyPatchEc2Model() {
    if (!"ec2".equals(serviceModel.getMetadata().getProtocol())) {
        return;
    }

    Map<String, Shape> shapes = serviceModel.getShapes();
    Map<String, Shape> updatedShapes = new HashMap<>();

    for (Map.Entry<String, Shape> entry : shapes.entrySet()) {
        String key = entry.getKey();
        Shape shape = entry.getValue();

        String newKey = key.replaceAll("Result$", "Response");
        shape.setName(shape.getName().replaceAll("Result$", "Response"));
        shape.setType(shape.getType().replaceAll("Result$", "Response"));

        updatedShapes.put(newKey, shape);
    }

    shapes.clear();
    shapes.putAll(updatedShapes);
}
```

tools/scripts/run_protocol_tests.py
```diff
@@ -0,0 +1,136 @@
...
+        self.args = args
+        self.fail = False
+
+    def run(self):
+        tests = self._collect_tests()
+        if not tests or not len(tests):
+            print(f"No protocol tests found in {self.build_dir}")
+            self.fail = True
+        for test in tests:
+            mock_server = MockHttpServerHandle()
```
sbera87, line 92: should be used with 'with' for cleanup

tools/scripts/run_protocol_tests.py
```diff
@@ -0,0 +1,136 @@
...
+    """A RAII-style helper wrapper around mini mock http server.
+    Launches a server on construction and kills it on destruction
+    """
+
+    def __init__(self):
+        python_cmd = [shutil.which("python3"), PROTO_TEST_MOCK_HANDLER]
+        self.process = subprocess.Popen(python_cmd, stdout=subprocess.PIPE, stderr=subprocess.PIPE)
+        print(f"Started mock server with PID {self.process.pid}")
+
+    def __del__(self):
```
sbera87, line 46: This may not be deterministically called by the interpreter per https://docs.python.org/3.4/reference/datamodel.html 
Better to use https://book.pythontips.com/en/latest/context_managers.html

sbera87, line 46: This may not be deterministically called by the interpreter per https://docs.python.org/3.4/reference/datamodel.html 
Better to use https://book.pythontips.com/en/latest/context_managers.html

tools/scripts/run_protocol_tests.py
```diff
@@ -0,0 +1,136 @@
...
+    return arg_map
+
+
+class MockHttpServerHandle:
+    """A RAII-style helper wrapper around mini mock http server.
+    Launches a server on construction and kills it on destruction
+    """
+
+    def __init__(self):
+        python_cmd = [shutil.which("python3"), PROTO_TEST_MOCK_HANDLER]
```
sbera87, line 42: No exception handling found

tools/scripts/run_protocol_tests.py
```diff
@@ -0,0 +1,136 @@
...
+    def run(self):
+        tests = self._collect_tests()
+        if not tests or not len(tests):
+            print(f"No protocol tests found in {self.build_dir}")
+            self.fail = True
+        for test in tests:
+            mock_server = MockHttpServerHandle()
+            try:
+                process = subprocess.run([test], timeout=6 * 60, check=True)
+                process.check_returncode()
```
sbera87, line 95: this is redundant as check is True 
per python documentation 
`If check is true, and the process exits with a non-zero exit code, a [CalledProcessError](https://docs.python.org/3/library/subprocess.html#subprocess.CalledProcessError) exception will be raised. Attributes of that exception hold the arguments, the exit code, and stdout and stderr if they were captured.
`

tools/code-generation/generator/src/main/java/com/amazonaws/util/awsclientgenerator/generators/cpp/CppProtocolTestGenerator.java
SergeyRyabinin: copy-paste mistake, thank you!

tools/code-generation/generator/src/main/java/com/amazonaws/util/awsclientgenerator/generators/cpp/CppProtocolTestGenerator.java
SergeyRyabinin: The complexity is the same, in addition, this was a copy-pasted method from the actual EC2 client generator.
I refactored it a bit and re-using the existing one.

Additionally, 
```
Map<String, Shape> updatedShapes = new HashMap<>();
```
would re-shuffle shapes in a map as far as I remember, Json parsers use Linked structures.

tools/scripts/run_protocol_tests.py
SergeyRyabinin: I've added a bit of exception handling here, but can't do much more than that.
Catch-trace-crash/return -1 is the current approach.

tools/scripts/run_protocol_tests.py
SergeyRyabinin: Thank you, I've refactored the code to explicitly implicitly call new method `stop()` with the `with` statement.

tools/scripts/run_protocol_tests.py
SergeyRyabinin: we need to wait a bit to let the process and OS to terminate the process.

tools/scripts/run_protocol_tests.py
SergeyRyabinin: this is actually a left-over from the AI generated slop I used as an initial draft.
I just removed it.

tools/scripts/run_protocol_tests.py
```diff
@@ -0,0 +1,136 @@
...
+            return self.process.wait()
+        return None
+
+
+class ProtocolTestRunner:
+    """Actual test runner: collects available tests and runs them
+    """
+
+    def __init__(self, args):
+        self.debug = args.get("debug", False)
```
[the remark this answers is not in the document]
SergeyRyabinin: yes, unused.
Thank you for noticing!

tools/scripts/run_protocol_tests.py
SergeyRyabinin: thank you, I've refactored this piece with enter/exit methods and `with`.

tools/scripts/run_protocol_tests.py
SergeyRyabinin: Thank you, I removed the check.

tools/scripts/run_protocol_tests.py
```diff
@@ -0,0 +1,149 @@
...
+PROTO_TEST_MOCK_HANDLER = f"{SCRIPTS_DIR}/protocol_tests_mock_server/protocol_tests_mock_server.py"
+
+
+def parse_arguments() -> dict:
+    """Parse arguments and/or print help
+
+    :return: dict with parsed arguments
+    """
+    parser = argparse.ArgumentParser(description="Protocol tests driver for AWS-SDK-CPP")
+    parser.add_argument("--debug", action="store_true", help="Print additional information")
```
DmitriyMusatkin, line 28: tiny suggestion: parser.add_argument("--debug", action="store_true", default = False, help="Print additional information")

and remove logic later

tools/scripts/run_protocol_tests.py
```diff
@@ -0,0 +1,149 @@
...
+
+
+class MockHttpServerHandle:
+    """A RAII-style helper wrapper around mini mock http server.
+    Launches a server on construction and kills it on destruction
+    """
+
+    def __init__(self):
+        try:
+            python_cmd = [shutil.which("python3"), PROTO_TEST_MOCK_HANDLER]
```
DmitriyMusatkin, line 43: sys.executable if you want to make sure its the same python?

DmitriyMusatkin (approved):

tools/scripts/run_protocol_tests.py
SergeyRyabinin: No need to be sure it is the same interpreter
- test driver
- mock server
- test app
All can be seen as an independent apps.

sbera87 (approved):

Issue #3412, opened 2025-05-08
Unable to access SSO credentials due to msimatch in hashing behavior compared to the AWS CLI


justinpombrio:
### Describe the bug

If the `sso_start_url` ends with `/#`, then there's a difference in the hashes computed by `aws sso login` (which includes the `#`) and this CPP client (which excludes the `#`). These hashes are used as the filenames in `~/.aws/sso/cache/*.json`, so the CPP client can't find the sso cache file, so it fails to connect and prints a 403 error.

### Regression Issue

- [ ] Select this option if this issue appears to be a regression.

### Expected Behavior

After I run `aws sso login` and `aws s3 ls` (both successful), I can also access S3 resources using this CPP client.

### Current Behavior

Despite `aws sso login` and `aws s3 ls` both succeeding, the CPP client reports the error:

```
[ERROR] 2025-05-07 16:24:48.374 AWSXmlClient [131720638338688] HTTP response code: 403
Resolved remote host IP address: 52.219.93.58
Request ID: TPGTKY5PXV0B0BRG
Exception name: AccessDenied
Error message: AccessDenied
```

### Reproduction Steps

Use a `sso_start_url` that ends with `#` in `~/.aws/config`.

(If you already have cached credentials for that start url without the `#`, also delete the caches.)

### Possible Solution

From the behavior I've seen, I suspect that `GetCachedConfigProfile` is  normalizing the url (removing the trailing `#`) before its hashed by `SSOCredentialsProvider`. It should be normalizing after hashing, instead.

### AWS CPP SDK version used

1.11.555

### Compiler and Version used

clang v16.0.6

### Operating System and version

Linux Mint 22.1 (essentially equivalent to Ubuntu 24.04)

sbiscigl:
Dug into this a little bit root cause is INI format, and whether or not a `#` can live in a INI file.

[according to some docs on INI comments](https://en.wikipedia.org/wiki/INI_file#Comments)

> Some INI dialects furthermore allow use of the [number sign](https://en.wikipedia.org/wiki/Number_sign) (#, ASCII 0x23) to denote a comment, mirroring Unix [shell](https://en.wikipedia.org/wiki/Bourne_shell) comments.

So in a INI file that looks like

```
[profile MyTestProfile]
sso_start_url = https://some-start-url.com#this is technically a comment
```
[...]

PR #3425, opened 2025-05-19
allow for the hash character in configuration values

sbiscigl:
*Issue #, if available:*

https://github.com/aws/aws-sdk-cpp/issues/3412

*Description of changes:*

There was a issue users encountered that prevented them from using the `#` character in configuration value, specially sso start urls. This now makes sure that a comment inline has to include white space before the comment character.

[go saw something similar as well and made the same change to align with other SDKs](https://github.com/aws/aws-sdk-go-v2/pull/2371)

*Check all that applies:*

- [x] Linux
- [x] Windows
- [x] Android
- [x] MacOS
- [ ] IOS
- [ ] Other Platforms

### DIFF
--- a/src/aws-cpp-sdk-core/source/config/AWSConfigFileProfileConfigLoader.cpp
+++ b/src/aws-cpp-sdk-core/source/config/AWSConfigFileProfileConfigLoader.cpp
@@ -4,11 +4,17 @@
  */
 
 #include <aws/core/config/AWSProfileConfigLoader.h>
+#include <aws/core/utils/memory/stl/AWSArray.h>
 #include <aws/core/utils/memory/stl/AWSSet.h>
 #include <aws/core/utils/memory/stl/AWSStreamFwd.h>
 #include <aws/core/utils/StringUtils.h>
 #include <aws/core/utils/logging/LogMacros.h>
 #include <fstream>
+
+namespace {
+Aws::Array<const char*, 4> COMMENT_START_SEQ{" #", " ;", "\t#", "\t;"};
+Aws::Array<const char, 2> COMMENT_CHARS{'#', ';'};
+}
 
 namespace Aws
 {
@@ -46,8 +52,6 @@ namespace Aws
         static const size_t IDENTIFIER_ALLOWED_CHARACTERS_SZ = sizeof(IDENTIFIER_ALLOWED_CHARACTERS) - 1;
         const char WHITESPACE_CHARACTERS[]         = "\t ";
         static const size_t WHITESPACE_CHARACTERS_SZ = sizeof(WHITESPACE_CHARACTERS) - 1;
-        const char COMMENT_START[]                 = "#;";
-        static const size_t COMMENT_START_SZ = sizeof(COMMENT_START) - 1;
 
         struct ProfilePropertyAccessFunctions
         {
@@ -125,7 +129,7 @@ namespace Aws
                         rawLine.pop_back(); // Remove carriage return character ('\r')
                     }
 
-                    Aws::String line = rawLine.substr(0, rawLine.find_first_of(COMMENT_START)); // ignore comments
+                    const Aws::String line{StripCommentFromLine(rawLine)};
                     if (line.empty() || line.length() < ASSUME_EMPTY_LEN || line.find_first_not_of(WHITESPACE_CHARACTERS) == Aws::String::npos)
                     {
                         continue;
@@ -384,8 +388,7 @@ namespace Aws
                     }
                     pos++;
                     pos = line.find_first_not_of(WHITESPACE_CHARACTERS, pos);
-                    if(pos != Aws::String::npos &&
-                       std::find(COMMENT_START, COMMENT_START + COMMENT_START_SZ, line[pos]) == COMMENT_START + COMMENT_START_SZ)
+                    if(pos != Aws::String::npos && std::find(COMMENT_CHARS.begin(), COMMENT_CHARS.end(), line[pos]) == COMMENT_CHARS.end())
                     {
                         AWS_LOGSTREAM_ERROR(PARSER_TAG, "Found unexpected characters after closing bracket of Section Identifier " << line);
                         break;
@@ -530,6 +533,26 @@ namespace Aws
                 {
                     AWS_LOGSTREAM_FATAL(PARSER_TAG, "Unknown parser error: unexpected state " << currentState);
                 }
+            }
+
+            /**
+             * Returns a configuration file line with inline comments removed.
+             *
+             * @param line a raw line read in from configuration.
+             * @return the raw line with any commented out sections removed.
+             */
+            static Aws::String StripCommentFromLine(const Aws::String& line) {
+              // entire line is commented if the first char is comment, otherwise we want whitespace + comment delimitation
+              if (line.empty() || std::any_of(COMMENT_CHARS.begin(), COMMENT_CHARS.end(),
+                                              [&line](const char firstChar) -> bool { return firstChar == line.front(); })) {
+                return {};
+              }
+
+              size_t commentLocation = Aws::String::npos;
+              for (const auto* const commentChar : COMMENT_START_SEQ) {
+                commentLocation = std::min(commentLocation, line.find(commentChar));
+              }
+              return {line.substr(0, commentLocation)};
             }
 
             Aws::Map<String, Profile> m_foundProfiles;
--- a/tests/aws-cpp-sdk-core-tests/aws/config/AWSConfigFileProfileConfigLoaderTest.cpp
+++ b/tests/aws-cpp-sdk-core-tests/aws/config/AWSConfigFileProfileConfigLoaderTest.cpp
@@ -348,13 +348,13 @@ R"(
      ;another comment
 [  default ]
 source_profile   =   base)" "\t" R"(
-credential_process = echo '{ "Version": 1, "AccessKeyId": "ASIARTESTID", "SecretAccessKey": "TESTSECRETKEY", "SessionToken": "TESTSESSIONTOKEN", "Expiration": "2022-05-02T18:36:00+00:00" }'#COMMENT
+credential_process = echo '{ "Version": 1, "AccessKeyId": "ASIARTESTID", "SecretAccessKey": "TESTSECRETKEY", "SessionToken": "TESTSESSIONTOKEN", "Expiration": "2022-05-02T18:36:00+00:00" }' #COMMENT
  ; Comment to be ignored
 )"
 R"(                  )" // to avoid blank space removals by an IDE.
 R"(
 [  profile   )" "\t\t\t" R"(    base  ]
-    region  =   us-east-1#sa-east-3
+    region  =   us-east-1 #sa-east-3
     #region = commented-out region
 )";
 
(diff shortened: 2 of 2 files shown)


tests/aws-cpp-sdk-core-tests/aws/config/AWSConfigFileProfileConfigLoaderTest.cpp
```diff
@@ -348,13 +348,13 @@ R"(
      ;another comment
 [  default ]
 source_profile   =   base)" "\t" R"(
-credential_process = echo '{ "Version": 1, "AccessKeyId": "ASIARTESTID", "SecretAccessKey": "TESTSECRETKEY", "SessionToken": "TESTSESSIONTOKEN", "Expiration": "2022-05-02T18:36:00+00:00" }'#COMMENT
+credential_process = echo '{ "Version": 1, "AccessKeyId": "ASIARTESTID", "SecretAccessKey": "TESTSECRETKEY", "SessionToken": "TESTSESSIONTOKEN", "Expiration": "2022-05-02T18:36:00+00:00" }' #COMMENT
```
sbiscigl, line 351: these test cases had "invalid comments", they require white space before the actual comment

src/aws-cpp-sdk-core/source/config/AWSConfigFileProfileConfigLoader.cpp
```diff
@@ -532,6 +535,26 @@ namespace Aws
                 }
             }
 
+            /**
+             * Returns a configuration file line with inline comments removed.
+             *
+             * @param line a raw line read in from configuration.
+             * @return the raw line with any commented out sections removed.
+             */
+            static Aws::String StripCommentFromLine(const Aws::String& line) {
```
sbera87, line 544: can be simplified to usage of just one delimiter list. Also need to strip out trailing whitespaces.
think of example: <code line>           #comment    

```
static Aws::String StripCommentFromLine(const Aws::String& line) {
    // return empty if line starts with comment char
    if (line.empty() || std::any_of(COMMENT_CHARS.begin(), COMMENT_CHARS.end(),
        [&line](char c) { return c == line[0]; })) {
        return {};
    }

    // find first comment char that has a space before it, then strip out code line sans trailing spaces
    size_t commentPos = Aws::String::npos;
    for (char c : COMMENT_CHARS) {
        size_t pos = line.find(' ' + Aws::String(1, c));
        if (pos != Aws::String::npos) {
            commentPos = std::min(commentPos, pos);
        }
    }

    if (commentPos == Aws::String::npos) {
        return line;
    }

    // find last non-whitespace character before the comment
    while (commentPos > 0 && std::isspace(line[commentPos - 1])) {
        --commentPos;
    }

    return line.substr(0, commentPos);
}
```

tests/aws-cpp-sdk-core-tests/aws/config/AWSConfigFileProfileConfigLoaderTest.cpp
sbera87, line 351: what about test case for multiple spaces after code end or  `<code line>   # # `

src/aws-cpp-sdk-core/source/config/AWSConfigFileProfileConfigLoader.cpp
```diff
@@ -4,12 +4,18 @@
...
+#include <aws/core/utils/memory/stl/AWSArray.h>
 #include <aws/core/utils/memory/stl/AWSSet.h>
 #include <aws/core/utils/memory/stl/AWSStreamFwd.h>
 #include <aws/core/utils/StringUtils.h>
 #include <aws/core/utils/logging/LogMacros.h>
 #include <fstream>
 
+namespace {
+Aws::Array<const char*, 2> COMMENT_START_SEQ{" #", " ;"};
+Aws::Array<const char, 2> COMMENT_CHARS{'#', ';'};
```
sbera87, line 16: why not  '//' ?

sbiscigl: thats not a comment in the [INI spec](https://en.wikipedia.org/wiki/INI_file#Comments), which is what this file is parsing, which is how a [aws configuration file is defiend](https://docs.aws.amazon.com/cli/v1/userguide/cli-configure-files.html).

src/aws-cpp-sdk-core/source/config/AWSConfigFileProfileConfigLoader.cpp
sbiscigl: > Also need to strip out trailing whitespaces.

why does this funciton _need to_ strip out white space? that happens in the the [`ParseSectionDeclaration`](https://github.com/aws/aws-sdk-cpp/blob/main/src/aws-cpp-sdk-core/source/config/AWSConfigFileProfileConfigLoader.cpp#L274-L317) function. this function removes a inline comment. or returns a empty line for a commented line, simple as that, no need to add more functionality to it that is covered other places


> can be simplified to usage of just one delimiter list.
```
' ' + Aws::String(1, c)
```

how would be creating a temporary value eachtime be beneficial vs having a statically defined list? its just a few more bytes to have a static value instead of a runtime value.

src/aws-cpp-sdk-core/source/config/AWSConfigFileProfileConfigLoader.cpp
```diff
@@ -4,12 +4,18 @@
...
 #include <aws/core/config/AWSProfileConfigLoader.h>
+#include <aws/core/utils/memory/stl/AWSArray.h>
 #include <aws/core/utils/memory/stl/AWSSet.h>
 #include <aws/core/utils/memory/stl/AWSStreamFwd.h>
 #include <aws/core/utils/StringUtils.h>
 #include <aws/core/utils/logging/LogMacros.h>
 #include <fstream>
 
+namespace {
+Aws::Array<const char*, 2> COMMENT_START_SEQ{" #", " ;"};
```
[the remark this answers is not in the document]
sbiscigl: is there a benefit to having the runtime temp value instead of the few more bytes static? personally i dont really care either way. but I err on the side of less runtime evaluation when possible.

src/aws-cpp-sdk-core/source/config/AWSConfigFileProfileConfigLoader.cpp
sbera87: Its maintaining two hard coded lists  vs one. The algorithm is far from optimal using std find, so question of static vs runtime optimization is moot imo. For pattern matching there are better  algos (KMP) but thats an overkill here.

src/aws-cpp-sdk-core/source/config/AWSConfigFileProfileConfigLoader.cpp
sbera87: Just had to ask. This can be a generic helper that we can use here. which can use delimiter strings vs delimiter characters. Just food for thought

src/aws-cpp-sdk-core/source/config/AWSConfigFileProfileConfigLoader.cpp
sbera87: if there is a piece of code
`<code>       ;#`
current logic extracts `<code>     ` including trailing space which we dont need . It may be relvant when say someone has
an commented line
 `     # comment`
 in that case we don't  really need to strip the leading whitespace as a line. Or if we do, caller can decide what to do with it , but the point is this helper can indicate its all spaces and no other content to the caller to avoid further parsing

sbiscigl: > Its maintaining two hard coded lists vs one. The algorithm is far from optimal using std find, so question of static vs runtime optimization is moot imo.

I personally dont care, i just avoid runtime eval when i can, if you feel as if its a blocker though I can change it.

> if there is a piece of code

this is a private helper function, that is not invoked outside of this class, this is encapsulating one line of code and moving it to a private functoin, no need to overthink here. the old [operation didnt remove white space](https://github.com/aws/aws-sdk-cpp/blob/main/src/aws-cpp-sdk-core/source/config/AWSConfigFileProfileConfigLoader.cpp#L128), so i dont see why this should now either.

sbera87: the old operation  shouldn't be the standard, since we have a shot at bettering it. I think we should have additional tests with lines with indented comments only (which will yield empty lines or no line depending on the final logic you choose) as well as setting a string variable with "# or ;" in it akin to .
I don't think the current logic handles the latter case.

```
  configFile << R"([profile maelle])" << "\n"
             << R"(sso_account_id = 333333333333)" << "\n"
             << R"(sso_role_name = PictoAccess)" << "\n"
             << R"(sso_start_url = https://lumiere.com/for-those-who-come-after # stendahl pre nerf)" << "\n"
             << R"(                 ;# dummy comment here)" << "\n"     //this should be ok with/without empty line  
             << R"(StringWithSpaces="This is # a comment with comment chars #")" << "\n"   // this should also be handled 
             << R"(#sso_start_url = https://lumiere.com/verso)" << "\n"
             << R"(sso_region = eu-west-3)" << "\n"
             << R"(sso_registration_scopes = sso:account:access)" << "\n"
             << R"(region = eu-west-3)" << "\n"
             << R"(output = json)" << "\n";
```

sbera87: ini files support quoted strings. If the code is meant to only parse specific subset of case and is not supposed to be foolproof line stripper, I can approve with the aforementioned comments. To your point, don't we then need to validate file doesn't have quotes? because if we do it will be incorrect then and we should at least flag that

sbera87 (approved):
Approved with comments: Pointed out logic/implementation caveats and more test cases and negative test for validation of spec

src/aws-cpp-sdk-core/source/config/AWSConfigFileProfileConfigLoader.cpp
sbiscigl: > ini files support quoted strings.

[from the spec](https://en.wikipedia.org/wiki/INI_file#Quoted_values)

> Some implementations allow values to be quoted, typically using [double quotes](https://en.wikipedia.org/wiki/Quotation_mark) and/or [apostrophes](https://en.wikipedia.org/wiki/Apostrophe). This allows for explicit declaration of whitespace, and/or for quoting of special characters (equals, semicolon, etc.). The standard Windows function [GetPrivateProfileString](https://docs.microsoft.com/en-us/windows/win32/api/winbase/nf-winbase-getprivateprofilestringw) supports this, and will remove quotation marks that surround the values.

spcifically in this context "Some implementations allow". Currently we do not nor have we in the past.

>  is not supposed to be foolproof line stripper

this is _not_ supposed to be a foolproof INI parser, nor this change to catpure all nuances of INI parsing. this change is _just to fix a customer issue_. in the future we have ezpressed interest into moving to the [CRT configuraiton parser](https://github.com/awslabs/aws-c-sdkutils/blob/main/include/aws/sdkutils/aws_profile.h), aligning with how other SDKs parse INI files in a central dependency. that migrattion is not happening as part of this change.

PR #3437, opened 2025-05-29
Add CRT version check at cmake and runtime

SergeyRyabinin:
*Issue #, if available:*
By default, git does not pull with recursive submodules pull, resulting in builds with outdated CRT submodule dependency, which in turn results in issues reported such as
https://github.com/aws/aws-sdk-cpp/discussions/3433
*Description of changes:*
Add a check validating that the CRT submodule version is one set and validated by the AWS SDK team at the cmake configuration step.
Additionally, add a non-blocking logging error if CRT built-with and runtime-found versions are different.
*Check all that applies:*

- [ ] Linux
- [ ] Windows
- [ ] Android
- [ ] MacOS
- [ ] IOS
- [ ] Other Platforms


sbiscigl (approved):
('-')7

cmake/check_crt_version.cmake
```diff
@@ -0,0 +1,26 @@
+# Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.
+# SPDX-License-Identifier: Apache-2.0.
+
+function(check_crt_version)
+    include(expected_crt_version)
+
+    file(READ "${aws-crt-cpp_SOURCE_DIR}/VERSION" CRT_SIMPLE_VERSION)
```
SergeyRyabinin, line 7: this variable probably won't be set if you are importing the CRT library instead of building it as a part of the SDK build, so in case of import, you want to disable the check
```
-DENABLE_CRT_VERSION_CHECK=OFF
```

PR #3465, opened 2025-06-24
Implement the monitoring interface for performance tests

Alan4506:
*Description of changes:*

This PR adds performance metrics reporting for SDK service operations. It implements a monitoring interface that collects latency data and outputs it in JSON format.

*Check all that applies:*

- [X] Linux
- [X] Windows
- [X] Android
- [X] MacOS
- [ ] IOS
- [ ] Other Platforms

### DIFF
--- a/tests/performance-tests/include/performance_tests/reporting/JsonReportingMetrics.h
+++ b/tests/performance-tests/include/performance_tests/reporting/JsonReportingMetrics.h
@@ -1,0 +1,165 @@
+/**
+ * Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.
+ * SPDX-License-Identifier: Apache-2.0.
+ */
+
+#pragma once
+
+#include <aws/core/client/AWSClient.h>
+#include <aws/core/http/HttpRequest.h>
+#include <aws/core/monitoring/CoreMetrics.h>
+#include <aws/core/monitoring/MonitoringFactory.h>
+#include <aws/core/monitoring/MonitoringInterface.h>
+#include <aws/core/utils/memory/AWSMemory.h>
+#include <aws/core/utils/memory/stl/AWSSet.h>
+#include <aws/core/utils/memory/stl/AWSString.h>
+#include <aws/core/utils/memory/stl/AWSVector.h>
+
+#include <cstdint>
+#include <memory>
+#include <utility>
+
+namespace PerformanceTest {
+namespace Reporting {
+/**
+ * Container for a single performance metric record that stores measurement data and associated metadata.
+ */
+struct PerformanceMetricRecord {
+  Aws::String name;
+  Aws::String description;
+  Aws::String unit;
+  int64_t date;
+  Aws::Vector<int64_t> measurements;
+  Aws::Vector<std::pair<Aws::String, Aws::String>> dimensions;
+};
+
+/**
+ * An implementation of the MonitoringInterface that collects performance metrics
+ * and reports them in a JSON format.
+ */
+class JsonReportingMetrics : public Aws::Monitoring::MonitoringInterface {
+ public:
+  ~JsonReportingMetrics() override;
+
+  /**
+   * Called when an AWS request is started. Returns context for tracking.
+   * @param serviceName Name of the AWS service
+   * @param requestName Name of the operation
+   * @param request HTTP request object
+   * @return Context pointer (always returns nullptr)
+   */
+  void* OnRequestStarted(const Aws::String& serviceName, const Aws::String& requestName,
+                         const std::shared_ptr<const Aws::Http::HttpRequest>& request) const override;
+
+  /**
+   * Called when an AWS request succeeds. Records performance metrics.
+   * @param serviceName Name of the AWS service
+   * @param requestName Name of the operation
+   * @param request HTTP request object
+   * @param outcome HTTP response outcome
+   * @param metrics Core metrics collection containing latency data
+   * @param context Request context (unused)
+   */
+  void OnRequestSucceeded(const Aws::String& serviceName, const Aws::String& requestName,
+                          const std::shared_ptr<const Aws::Http::HttpRequest>& request, const Aws::Client::HttpResponseOutcome& outcome,
+                          const Aws::Monitoring::CoreMetricsCollection& metrics, void* context) const override;
+
+  /**
+   * Called when an AWS request fails. Records performance metrics.
+   * @param serviceName Name of the AWS service
+   * @param requestName Name of the operation
+   * @param request HTTP request object
+   * @param outcome HTTP response outcome
+   * @param metrics Core metrics collection containing latency data
+   * @param context Request context (unused)
+   */
+  void OnRequestFailed(const Aws::String& serviceName, const Aws::String& requestName,
+                       const std::shared_ptr<const Aws::Http::HttpRequest>& request, const Aws::Client::HttpResponseOutcome& outcome,
+                       const Aws::Monitoring::CoreMetricsCollection& metrics, void* context) const override;
+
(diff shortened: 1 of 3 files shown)


tests/performance-tests/include/performance_tests/reporting/JsonReportingMetrics.h
```diff
@@ -0,0 +1,165 @@
...
+/**
+ * Container for a single performance metric record that stores measurement data and associated metadata.
+ */
+struct PerformanceMetricRecord {
+  Aws::String name;
+  Aws::String description;
+  Aws::String unit;
+  int64_t date;
+  Aws::Vector<int64_t> measurements;
+  Aws::Vector<std::pair<Aws::String, Aws::String>> dimensions;
```
sbiscigl: should this be a `Aws::Map<Aws::String, Aws::String>`? would we ever have duplicate keys or want to handle duplicate keys?

tests/performance-tests/include/performance_tests/reporting/JsonReportingMetrics.h
```diff
@@ -0,0 +1,165 @@
...
+namespace PerformanceTest {
+namespace Reporting {
+/**
+ * Container for a single performance metric record that stores measurement data and associated metadata.
+ */
+struct PerformanceMetricRecord {
+  Aws::String name;
+  Aws::String description;
+  Aws::String unit;
+  int64_t date;
```
sbiscigl: why `int64_t` instead of [`Aws::Utils::Datetime`](https://github.com/aws/aws-sdk-cpp/blob/main/src/aws-cpp-sdk-core/include/aws/core/utils/DateTime.h#L55)?  using the datetime data type is likely what we want to do, then extract the time we want from it.

tests/performance-tests/include/performance_tests/reporting/JsonReportingMetrics.h
```diff
@@ -0,0 +1,165 @@
...
+  Aws::Vector<int64_t> measurements;
```
sbiscigl: instead of using `int64_t` for measurements, you may want to make a `Measurement` type that can have different implementations, so we can support floating point measurements in the future.

tests/performance-tests/include/performance_tests/reporting/JsonReportingMetrics.h
```diff
@@ -0,0 +1,165 @@
...
+                            const Aws::Monitoring::CoreMetricsCollection& metricsFromCore) const;
+
+  /**
+   * Outputs aggregated performance metrics to JSON file.
+   * Groups records by name and dimensions, then writes to configured output file.
+   */
+  void DumpJson() const;
+
+  mutable Aws::Vector<PerformanceMetricRecord> m_performanceRecords;
+  static Aws::Vector<std::pair<Aws::String, Aws::String>> TestDimensions;
```
sbiscigl: these should not be static variables to the class, they should be member variables that are assigned in the constructor.

tests/performance-tests/include/performance_tests/reporting/JsonReportingMetrics.h
```diff
@@ -0,0 +1,165 @@
...
+   * @param context Request context (unused)
+   */
+  void OnFinish(const Aws::String& serviceName, const Aws::String& requestName,
+                const std::shared_ptr<const Aws::Http::HttpRequest>& request, void* context) const override;
+
+  /**
+   * Sets test dimensions that will be included with all performance records.
+   * @param dimensions Vector of key-value pairs representing test dimensions (e.g., size, bucket type)
+   */
+  static void SetTestContext(const Aws::Vector<std::pair<Aws::String, Aws::String>>& dimensions);
```
sbiscigl: i comment on this further, but we should avoid creating static state that is shared between instances of the metrics reporter, instead using member variables that are assigned in a constructor.

tests/performance-tests/src/reporting/JsonReportingMetrics.cpp
```diff
@@ -0,0 +1,179 @@
...
+
+void JsonReportingMetrics::SetProductInfo(const Aws::String& productId, const Aws::String& sdkVersion, const Aws::String& commitId) {
+  ProductId = productId;
+  SdkVersion = sdkVersion;
+  CommitId = commitId;
+}
+
+void JsonReportingMetrics::SetOutputFilename(const Aws::String& filename) { OutputFilename = filename; }
+
+JsonReportingMetrics::~JsonReportingMetrics() { DumpJson(); }
```
sbiscigl: nice RAII!

tests/performance-tests/src/reporting/JsonReportingMetrics.cpp
```diff
@@ -0,0 +1,179 @@
...
+    return;
+  }
+
+  int64_t durationMs = 0;
+  Aws::String const latencyKey = Aws::Monitoring::GetHttpClientMetricNameByType(Aws::Monitoring::HttpClientMetricsType::RequestLatency);
+
+  auto iterator = metricsFromCore.httpClientMetrics.find(latencyKey);
+  if (iterator != metricsFromCore.httpClientMetrics.end()) {
+    durationMs = iterator->second;
+  }
```
sbiscigl: if we don't find the record we likely want to fail instead of reporting the metric. is there anyway we can report back to the runner that we have failed to collect metrics instead of reporting 0?

tests/performance-tests/src/reporting/JsonReportingMetrics.cpp
```diff
@@ -0,0 +1,179 @@
...
+  record.date = Aws::Utils::DateTime::Now().Seconds();
+  record.measurements.push_back(durationMs);
+  record.dimensions = TestDimensions;
+
+  m_performanceRecords.push_back(record);
+}
+
+void* JsonReportingMetrics::OnRequestStarted(const Aws::String&, const Aws::String&,
+                                             const std::shared_ptr<const Aws::Http::HttpRequest>&) const {
+  return nullptr;
```
sbiscigl: I think we might have a little problem here specifically about how you create `durationMs` and are tracking it here. and i suggest you look at `DefaultMonitoring` about [how we do it in the default monitor](https://github.com/aws/aws-sdk-cpp/blob/main/src/aws-cpp-sdk-core/source/monitoring/DefaultMonitoring.cpp#L199-L209) . we need to start the duration count here, and finish it when we succeed. You actually try following the pattern in that file.

tests/performance-tests/include/performance_tests/reporting/JsonReportingMetrics.h
Alan4506: We will not have duplicate keys here. In this case, Aws::Map<Aws::String, Aws::String> can be more efficient.

tests/performance-tests/src/reporting/JsonReportingMetrics.cpp
```diff
@@ -0,0 +1,220 @@
...
+
+  Aws::String const latencyKey = Aws::Monitoring::GetHttpClientMetricNameByType(Aws::Monitoring::HttpClientMetricsType::RequestLatency);
+  auto iterator = metricsFromCore.httpClientMetrics.find(latencyKey);
+  if (iterator != metricsFromCore.httpClientMetrics.end() && iterator->second > 0) {
+    requestContext->serviceName = serviceName;
+    requestContext->requestName = requestName;
+    requestContext->request = request;
+    requestContext->durationMs = iterator->second;
+  } else {
+    m_hasInvalidLatency = true;
```
sbiscigl: this could likely be on the request context, and oculd be called `failed`, the you just dont add it to top level list of metrics

PR #3470, opened 2025-07-01
Add s3 service performance tests

Alan4506:
*Description of changes:*

This PR adds performance tests for the s3 service. It tests operations including PutObject and GetObject with multiple test dimensions: bucket type and object size. 

*Check all that applies:*

- [X] Linux
- [X] Windows
- [X] Android
- [X] MacOS
- [ ] IOS
- [ ] Other Platforms

### DIFF
--- a/tests/performance-tests/include/performance-tests/Utils.h
+++ b/tests/performance-tests/include/performance-tests/Utils.h
@@ -1,0 +1,71 @@
+/**
+ * Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.
+ * SPDX-License-Identifier: Apache-2.0.
+ */
+
+#pragma once
+
+#include <aws/core/Aws.h>
+#include <aws/core/utils/StringUtils.h>
+#include <aws/core/utils/UUID.h>
+#include <aws/core/utils/memory/stl/AWSString.h>
+
+#include <algorithm>
+#include <cstddef>
+#include <cstdlib>
+#include <iostream>
+#include <random>
+#include <string>
+
+namespace PerformanceTest {
+namespace Utils {
+/**
+ * Generate a random string of specified length.
+ * @param length The desired length of the string
+ * @return A random string containing lowercase letters and digits
+ */
+static inline Aws::String RandomString(size_t length) {
+  auto randchar = []() -> char {
+    const char charset[] =
+        "0123456789"
+        "abcdefghijklmnopqrstuvwxyz";
+    const size_t max_index = (sizeof(charset) - 1);
+    return charset[rand() % max_index];
+  };
+  Aws::String str(length, 0);
+  std::generate_n(str.begin(), length, randchar);
+  return str;
+}
+
+/**
+ * Log error messages to stderr.
+ * @param service The name of the AWS service
+ * @param operation The name of the operation that failed
+ * @param message The error message to log
+ */
+static inline void LogError(const Aws::String& service, const Aws::String& operation, const Aws::String& message) {
+  std::cerr << "[ERROR] " << service << ":" << operation << " failed: " << message << '\n';
+}
+
+/**
+ * Generate a unique identifier using UUID.
+ * @return A 10-character lowercase UUID substring
+ */
+static inline Aws::String GenerateUniqueId() {
+  Aws::String const rawUUID = Aws::Utils::UUID::RandomUUID();
+  return Aws::Utils::StringUtils::ToLower(rawUUID.c_str()).substr(0, 10);
+}
+
+/**
+ * Get the current SDK version as a formatted string.
+ * @return SDK version in the format "major.minor.patch" (e.g., "1.11.0")
+ */
+static inline Aws::String GetSDKVersionString() {
+  Aws::SDKOptions::SDKVersion const version;
+  return Aws::Utils::StringUtils::to_string(static_cast<int>(version.major)) + "." +
+         Aws::Utils::StringUtils::to_string(static_cast<int>(version.minor)) + "." +
+         Aws::Utils::StringUtils::to_string(static_cast<int>(version.patch));
+}
+
+}  // namespace Utils
+}  // namespace PerformanceTest
--- a/tests/performance-tests/include/performance-tests/services/s3/S3PerformanceTest.h
+++ b/tests/performance-tests/include/performance-tests/services/s3/S3PerformanceTest.h
@@ -1,0 +1,62 @@
+/**
+ * Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.
+ * SPDX-License-Identifier: Apache-2.0.
+ */
+
+#pragma once
+
+#include <aws/core/utils/memory/stl/AWSString.h>
+#include <aws/s3/S3Client.h>
+
+#include <cstddef>
+
+namespace PerformanceTest {
+namespace Services {
+namespace S3 {
+/**
+ * Configuration for S3 performance test cases.
+ */
+struct TestCase {
+  Aws::String sizeLabel;
+  size_t sizeBytes;
+  Aws::String bucketTypeLabel;
+};
+
+/**
+ * Runs S3 performance test by creating a bucket, performing PutObject and GetObject operations,
+ * then cleaning up resources.
+ * @param s3 S3 client instance to use for operations
+ * @param config Test configuration containing object size and bucket type parameters
+ * @param availabilityZoneId Availability zone ID for S3 Express buckets
+ * @param iterations Number of put/get operations to perform
+ */
+void RunTest(Aws::S3::S3Client& s3, const TestCase& config, const Aws::String& availabilityZoneId, int iterations = 3);
+
+/**
+ * Create S3 bucket for testing.
+ * @param s3 S3 client instance
+ * @param config Test configuration
+ * @param availabilityZoneId Availability zone ID for S3 Express buckets
+ * @return Bucket name if successful, empty string if failed
+ */
+Aws::String SetupBucket(Aws::S3::S3Client& s3, const TestCase& config, const Aws::String& availabilityZoneId);
+
+/**
+ * Run PutObject and GetObject operations.
+ * @param s3 S3 client instance
+ * @param bucketName Name of the bucket to use
+ * @param config Test configuration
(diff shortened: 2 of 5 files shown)


tests/performance-tests/include/performance-tests/services/s3/S3PerformanceTest.h
```diff
@@ -0,0 +1,62 @@
...
+
+/**
+ * Runs S3 performance test by creating a bucket, performing PutObject and GetObject operations,
+ * then cleaning up resources.
+ * @param s3 S3 client instance to use for operations
+ * @param config Test configuration containing object size and bucket type parameters
+ * @param availabilityZoneId Availability zone ID for S3 Express buckets
+ * @param iterations Number of put/get operations to perform
+ */
+void RunTest(Aws::S3::S3Client& s3, const TestCase& config, const Aws::String& availabilityZoneId, int iterations = 3);
```
sbiscigl: we should represent a performance test as a class not as unstructured functions in a namespace, also we should give some structure to them. i.e.

```cpp
#include <memory>

class PerformanceTest {
public:
  virtual ~PerformanceTest() = default;
  virtual void Setup() = 0;
  virtual void TearDown() = 0;
  virtual void Run() = 0;
};

class S3PerformanceTest : public PerformanceTest {
 public:
  ~S3PerformanceTest() override;
  void Setup() override {
    //Do setup
  };
  void TearDown() override {
    //Do teardown
  }
  void Run() override {
    //Run the test
  }
};

int main() {
  auto performanceTest = std::make_unique<PerformanceTest>();
  performanceTest->Setup();
  performanceTest->Run();
  performanceTest->TearDown();
}
```

This is beneficial as when we add more services, we have more subclasses, and the "runner" is aware of the interface and not the actual implementation.

tests/performance-tests/src/services/s3/main.cpp
```diff
@@ -0,0 +1,61 @@
...
+
+#include <string>
+
+int main(int argc, char** argv) {
+  Aws::String region = "us-east-1";
+  Aws::String availabilityZoneId = "use1-az4";
+  Aws::String commitId = "unknown";
+  int iterations = 10;
+
+  for (int i = 1; i < argc; ++i) {
```
sbiscigl: since this is a test and not SDK code we should feel free to use a __real__  arg parser instead of hand coding one. i personally like [cxxopts](https://github.com/jarro2783/cxxopts).

tests/performance-tests/include/performance-tests/Utils.h
```diff
@@ -0,0 +1,71 @@
...
+static inline Aws::String GenerateUniqueId() {
+  Aws::String const rawUUID = Aws::Utils::UUID::RandomUUID();
+  return Aws::Utils::StringUtils::ToLower(rawUUID.c_str()).substr(0, 10);
+}
+
+/**
+ * Get the current SDK version as a formatted string.
+ * @return SDK version in the format "major.minor.patch" (e.g., "1.11.0")
+ */
+static inline Aws::String GetSDKVersionString() {
```
sbiscigl: why cant you just use [`GetVersionString`](https://github.com/aws/aws-sdk-cpp/blob/main/src/aws-cpp-sdk-core/include/aws/core/Version.h#L13) from `aws/core/Version.h`

tests/performance-tests/src/services/s3/main.cpp
```diff
@@ -0,0 +1,61 @@
...
+        PerformanceTest::Services::S3::TestConfig::OutputFilename);
+  }};
+
+  Aws::InitAPI(options);
+
+  {
+    Aws::Client::ClientConfiguration cfg;
+    cfg.region = region;
+
+    Aws::S3::S3Client s3(cfg);
```
sbiscigl: initializing the client should likely be scoped to the instance of the `PerformanceTest` object, `main` should just make a decision which performance test object to create and run it

tests/performance-tests/include/performance-tests/Utils.h
```diff
@@ -0,0 +1,71 @@
...
+  return str;
+}
+
+/**
+ * Log error messages to stderr.
+ * @param service The name of the AWS service
+ * @param operation The name of the operation that failed
+ * @param message The error message to log
+ */
+static inline void LogError(const Aws::String& service, const Aws::String& operation, const Aws::String& message) {
```
sbiscigl: instead of writing to std::err, lets use the macro [`AWS_LOG_ERROR`](https://github.com/aws/aws-sdk-cpp/blob/main/src/aws-cpp-sdk-core/include/aws/core/utils/logging/LogMacros.h#L62) from `aws/core/utils/logging/LogMacros.h`

tests/performance-tests/include/performance-tests/services/s3/S3TestConfig.h
```diff
@@ -0,0 +1,26 @@
...
+
+#include <aws/core/utils/memory/stl/AWSSet.h>
+#include <aws/core/utils/memory/stl/AWSString.h>
+#include <aws/core/utils/memory/stl/AWSVector.h>
+
+namespace PerformanceTest {
+namespace Services {
+namespace S3 {
+namespace TestConfig {
+const Aws::Set<Aws::String> TestOperations = {"PutObject", "GetObject"};
```
sbiscigl: `Aws::Set`/`Aws::Vector`/`Aws::String` should not be used in this case because the static, and known at compilation time, they will not be added to at runtime and therefor do not need to be allocator aware.

something like
```cpp
struct Widget {
  const char* name;
  const int value;
};

namespace {
  const std::array<const char*, 4> my_array{"foo", "bar"};
  const std::array<const Widget, 2> m_widgets{{"name", 64}};
}
```
is more preferable as they are statically sized and defined

tests/performance-tests/src/services/s3/main.cpp
```diff
@@ -0,0 +1,61 @@
...
+#include <aws/core/monitoring/MonitoringFactory.h>
+#include <aws/core/utils/memory/AWSMemory.h>
+#include <aws/core/utils/memory/stl/AWSString.h>
+#include <aws/s3/S3Client.h>
+#include <performance-tests/Utils.h>
+#include <performance-tests/reporting/JsonReportingMetrics.h>
+#include <performance-tests/services/s3/S3PerformanceTest.h>
+#include <performance-tests/services/s3/S3TestConfig.h>
+
+#include <string>
```
sbiscigl, line 18: why `std::string`? maybe a unused import

tests/performance-tests/CMakeLists.txt
```diff
@@ -0,0 +1,34 @@
...
+    add_executable(${SERVICE}-performance-test
+            src/services/${SERVICE}/main.cpp
+            src/reporting/JsonReportingMetrics.cpp
+            src/services/${SERVICE}/${PERF_TEST_FILE}
+    )
+    set_compiler_flags(${SERVICE}-performance-test)
+    set_compiler_warnings(${SERVICE}-performance-test)
+    target_include_directories(${SERVICE}-performance-test PRIVATE include)
+    target_link_libraries(${SERVICE}-performance-test PRIVATE aws-cpp-sdk-core ${SDK_LIB} cxxopts::cxxopts)
+    target_compile_options(${SERVICE}-performance-test PRIVATE -std=c++17 -fexceptions)
```
sbiscigl: dont use ` target_compile_options` instead use `set_property` and `CXX_STANDARD` per the [cmake documenation](https://cmake.org/cmake/help/latest/prop_tgt/CXX_STANDARD.html)

`set_property(TARGET ${SERVICE}-performance-test PROPERTY CXX_STANDARD 20)`

also since its testing that is optional that we only run, feel free to bump to cpp 20

tests/performance-tests/include/performance-tests/services/s3/S3TestConfig.h
```diff
@@ -0,0 +1,30 @@
...
+#include <performance-tests/services/s3/S3PerformanceTest.h>
+
+#include <array>
+#include <set>
+
+namespace PerformanceTest {
+namespace Services {
+namespace S3 {
+namespace TestConfig {
+const std::set<const char*> TestOperations = {"PutObject", "GetObject"};
```
sbiscigl: use `std::array`, `std::set` can [allocate memory](https://en.cppreference.com/w/cpp/container/set.html), and since this is constant space we want to use std::array

tests/performance-tests/src/services/s3/S3PerformanceTest.cpp
```diff
@@ -0,0 +1,119 @@
...
+    AWS_LOG_ERROR("PerformanceTest", "S3:Run - Bucket setup failed, skipping test");
+    return;
+  }
+
+  const auto randomPayload = PerformanceTest::Utils::RandomString(m_config.sizeBytes);
+
+  // Run PutObject multiple times
+  for (int i = 0; i < m_iterations; i++) {
+    auto stream = Aws::MakeShared<Aws::StringStream>("PerfStream");
+    *stream << randomPayload;
```
sbiscigl: you can just pass the string to the string stream constructor i.e.

```c++
const auto auto stream = Aws::MakeShared<Aws::StringStream>("PerfStream", randomPayload);
```

tests/performance-tests/src/services/s3/S3PerformanceTest.cpp
```diff
@@ -0,0 +1,119 @@
...
+
+    Aws::S3::Model::PutObjectRequest por;
+    por.WithBucket(m_bucketName).WithKey("test-object-" + Aws::Utils::StringUtils::to_string(i)).SetBody(stream);
+    por.SetAdditionalCustomHeaderValue("test-dimension-size", m_config.sizeLabel);
+    por.SetAdditionalCustomHeaderValue("test-dimension-bucket-type", m_config.bucketTypeLabel);
+    auto putOutcome = m_s3->PutObject(por);
```
sbiscigl, line 86: at this point `m_s3` is not guaranteed to not be a nullptr, it is dependent that `SetUp` is called exactly once. You should put handling to ensure that `SetUp` is called once, or that the pointer here is not nullpr

sbiscigl (approved):

PR #3472, opened 2025-07-08
Add DynamoDB service performance test

Alan4506:
*Description of changes:*

This PR adds performance tests for the DynamoDB service. It tests operations including PutItem and GetItem with the object size as the test dimension: 

*Check all that applies:*

- [X] Linux
- [X] Windows
- [X] Android
- [X] MacOS
- [ ] IOS
- [ ] Other Platforms

### DIFF
--- a/tests/performance-tests/include/performance-tests/services/dynamodb/DynamoDBPerformanceTest.h
+++ b/tests/performance-tests/include/performance-tests/services/dynamodb/DynamoDBPerformanceTest.h
@@ -5,23 +5,17 @@
 
 #pragma once
 
+#include <aws/core/client/ClientConfiguration.h>
+#include <aws/core/utils/Outcome.h>
 #include <aws/core/utils/memory/AWSMemory.h>
 #include <aws/core/utils/memory/stl/AWSString.h>
 #include <aws/dynamodb/DynamoDBClient.h>
 #include <performance-tests/PerformanceTestBase.h>
-
-#include <cstddef>
+#include <performance-tests/services/dynamodb/DynamoDBTestConfig.h>
 
 namespace PerformanceTest {
 namespace Services {
 namespace DynamoDB {
-/**
- * Configuration for DynamoDB performance test cases.
- */
-struct TestCase {
-  const char* sizeLabel;
-  size_t sizeBytes;
-};
 
 /**
  * DynamoDB performance test implementation.
@@ -29,18 +23,18 @@ struct TestCase {
  */
 class DynamoDBPerformanceTest : public PerformanceTestBase {
  public:
-  DynamoDBPerformanceTest(const Aws::String& region, const TestCase& config, int iterations = 10);
+  DynamoDBPerformanceTest(const Aws::Client::ClientConfiguration& clientConfig, const TestConfig::TestCase& testConfig,
+                          int iterations = 10);
 
-  void Setup() override;
-  void TearDown() override;
+  Aws::Utils::Outcome<bool, Aws::String> Setup() override;
   void Run() override;
+  void TearDown() override;
 
  private:
-  const TestCase& m_config;
-  const Aws::String m_region;
+  const Aws::UniquePtr<Aws::DynamoDB::DynamoDBClient> m_dynamodb;
+  const TestConfig::TestCase m_testConfig;
   const int m_iterations;
-  Aws::UniquePtr<Aws::DynamoDB::DynamoDBClient> m_dynamodb;
-  Aws::String m_tableName;
+  const Aws::String m_tableName;
 };
 
 }  // namespace DynamoDB
--- a/tests/performance-tests/include/performance-tests/services/dynamodb/DynamoDBPerformanceTest.h
+++ b/tests/performance-tests/include/performance-tests/services/dynamodb/DynamoDBPerformanceTest.h
@@ -1,0 +1,48 @@
+/**
+ * Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.
+ * SPDX-License-Identifier: Apache-2.0.
+ */
+
+#pragma once
+
+#include <aws/core/utils/memory/AWSMemory.h>
+#include <aws/core/utils/memory/stl/AWSString.h>
+#include <aws/dynamodb/DynamoDBClient.h>
+#include <performance-tests/PerformanceTestBase.h>
+
+#include <cstddef>
+
+namespace PerformanceTest {
+namespace Services {
+namespace DynamoDB {
+/**
+ * Configuration for DynamoDB performance test cases.
+ */
+struct TestCase {
+  const char* sizeLabel;
+  size_t sizeBytes;
+};
+
+/**
+ * DynamoDB performance test implementation.
+ * Tests PutItem and GetItem operations with different payload sizes.
+ */
+class DynamoDBPerformanceTest : public PerformanceTestBase {
+ public:
+  DynamoDBPerformanceTest(const Aws::String& region, const TestCase& config, int iterations = 10);
+
+  void Setup() override;
+  void TearDown() override;
+  void Run() override;
+
+ private:
+  const TestCase& m_config;
+  const Aws::String m_region;
+  const int m_iterations;
+  Aws::UniquePtr<Aws::DynamoDB::DynamoDBClient> m_dynamodb;
+  Aws::String m_tableName;
+};
+
+}  // namespace DynamoDB
+}  // namespace Services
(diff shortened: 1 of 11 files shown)


tests/performance-tests/src/services/dynamodb/DynamoDBPerformanceTest.cpp
```diff
@@ -0,0 +1,149 @@
...
+#include <thread>
+
+PerformanceTest::Services::DynamoDB::DynamoDBPerformanceTest::DynamoDBPerformanceTest(const Aws::String& region, const TestCase& config,
+                                                                                      int iterations)
+    : m_config(config), m_region(region), m_iterations(iterations) {}
+
+void PerformanceTest::Services::DynamoDB::DynamoDBPerformanceTest::Setup() {
+  Aws::Client::ClientConfiguration cfg;
+  cfg.region = m_region;
+  m_dynamodb = Aws::MakeUnique<Aws::DynamoDB::DynamoDBClient>("DynamoDBPerformanceTest", cfg);
```
sbiscigl: why not move client creation to the constructor?

tests/performance-tests/src/services/dynamodb/DynamoDBPerformanceTest.cpp
```diff
@@ -0,0 +1,149 @@
...
+      std::this_thread::sleep_for(std::chrono::seconds(1));
+    } else {
+      AWS_LOG_ERROR("PerformanceTest", ("DynamoDB:DescribeTable failed: " + describeOutcome.GetError().GetMessage()).c_str());
+      m_tableName.clear();
+      return;
+    }
+    count++;
+  }
+
+  if (count >= MAX_QUERIES) {
```
sbiscigl, line 86: should we change SetUp to return a Outcome of whether or not set up was successful? if set up failed, the test will fail so we should fail it quickly.

tests/performance-tests/src/services/dynamodb/main.cpp
```diff
@@ -0,0 +1,59 @@
...
+    return Aws::MakeUnique<PerformanceTest::Reporting::JsonReportingMetricsFactory>(
+        "JsonReportingMetricsFactory", operations, "cpp1", versionStr, commitId,
+        PerformanceTest::Services::DynamoDB::TestConfig::OutputFilename);
+  }};
+
+  Aws::InitAPI(sdkOptions);
+
+  {
+    for (const auto& config : PerformanceTest::Services::DynamoDB::TestConfig::TestMatrix) {
+      auto performanceTest = Aws::MakeUnique<PerformanceTest::Services::DynamoDB::DynamoDBPerformanceTest>("DynamoDBPerformanceTest",
```
sbiscigl: this can be a stack allocation we dont need it as a unique ptr

tests/performance-tests/include/performance-tests/services/dynamodb/DynamoDBPerformanceTest.h
```diff
@@ -0,0 +1,48 @@
...
+class DynamoDBPerformanceTest : public PerformanceTestBase {
+ public:
+  DynamoDBPerformanceTest(const Aws::String& region, const TestCase& config, int iterations = 10);
+
+  void Setup() override;
+  void TearDown() override;
+  void Run() override;
+
+ private:
+  const TestCase& m_config;
```
sbiscigl: why is this a ref? that could create lifetime issues

tests/performance-tests/include/performance-tests/services/dynamodb/DynamoDBPerformanceTest.h
```diff
@@ -0,0 +1,48 @@
...
+#include <cstddef>
+
+namespace PerformanceTest {
+namespace Services {
+namespace DynamoDB {
+/**
+ * Configuration for DynamoDB performance test cases.
+ */
+struct TestCase {
+  const char* sizeLabel;
```
sbiscigl: why a size label? since we are only supporting a handful of sizes, should this be a enum?

tests/performance-tests/src/services/dynamodb/DynamoDBPerformanceTest.cpp
```diff
@@ -0,0 +1,149 @@
...
+    AWS_LOG_ERROR("PerformanceTest", "DynamoDB:Table did not become active within timeout");
+    m_tableName.clear();
+    return;
+  }
+}
+
+void PerformanceTest::Services::DynamoDB::DynamoDBPerformanceTest::Run() {
+  if (!m_dynamodb) {
```
sbiscigl: these should be `assert` statements, we should not be handling these isntances gracefully we should be crashing as we are in a invalid state

sbiscigl:
also please include the cmake file for the exectuable

tests/performance-tests/include/performance-tests/services/dynamodb/DynamoDBPerformanceTest.h
Alan4506: The reason why I put the sizeLable here is that later I pass this sizeLabel to the JsonReportingMetrics using the header: `getItemRequest.SetAdditionalCustomHeaderValue("test-dimension-size", m_config.sizeLabel);`. It is a text description of the actual size of the object, like "64KB" or "1MB", just for convenience. Since there is a conversion formula between the text description and the actual size (e.g., from 64KB to 64 * 1024), it is also acceptable to simply keep `const char* sizeLabel` and remove `size_t sizeBytes`. However, I am not sure if it is good to convert the TestCase itself to an enum class, since other services like s3 may have other attributes such as bucket type in TestCase:
```
struct TestCase {
  const char* sizeLabel;
  const char* bucketTypeLabel;
};
```
In this case, should we use enum class for each attribute, i.e, create enum class SizeLable and BucketTypeLabel, and then include them in the TestCase like the following: 
```
enum class SizeLabel {
  K8,
  K64,
  M1
};

const char* ToString(SizeLabel label) {
  switch (label) {
    case SizeLabel::K8: return "8KB";
    case SizeLabel::K64: return "64KB";
    case SizeLabel::M1:  return "1MB";
    default: return "UnknownSize";
  }
}

enum class BucketTypeLabel {
  Standard,
  Express
};

const char* ToString(BucketTypeLabel label) {
  switch (label) {
    case BucketTypeLabel::Standard: return "s3-standard";
    case BucketTypeLabel::Express: return "s3-express";
    default: return "unknownBucket";
  }
}

struct TestCase {
  SizeLabel size;
  BucketTypeLabel bucketType;
};
```
Is it a good pattern, or is it overcomplicated and we should stick to the original one:
```
struct TestCase {
  const char* sizeLabel;
  const char* bucketTypeLabel;
};
```

tests/performance-tests/src/services/dynamodb/DynamoDBPerformanceTest.cpp
Alan4506: It makes sense to me that we should just crash for Run() when we are in a invalid state, for example: 
```
assert(m_dynamodb && "DynamoDB client not initialized - Setup() must succeed before Run()");
assert(!m_tableName.empty() && "DynamoDB table name empty - Setup() must succeed before Run()");
```
But in this case we are not using `AWS_LOG_ERROR` anymore?

tests/performance-tests/include/performance-tests/services/dynamodb/DynamoDBPerformanceTest.h
```diff
@@ -0,0 +1,42 @@
...
+/**
+ * DynamoDB performance test implementation.
+ * Tests PutItem and GetItem operations with different payload sizes.
+ */
+class DynamoDBPerformanceTest : public PerformanceTestBase {
+ public:
+  DynamoDBPerformanceTest(const Aws::Client::ClientConfiguration& clientConfig, const TestConfig::TestCase& testConfig,
+                          int iterations = 10);
+
+  Aws::Utils::Outcome<bool, Aws::String> Setup() override;
```
sbiscigl: you should use
```
Aws::Utils::Outcome<Aws::NoResult, Aws::String> Setup() override;
```
instead of bool to denote that that there is actually no result, and not actually a boolean outcome. we do that [a lot of places in the SDK](https://github.com/aws/aws-sdk-cpp/blob/ec3ed43d6ccc5308626f28b44c0e215e18dc1095/generated/src/aws-cpp-sdk-mq/include/aws/mq/MQServiceClientModel.h#L121C15-L121C58) already.

sbiscigl (approved):

PR #3473, opened 2025-07-10
add option to ignore content length in S3Crt

sbiscigl:
*Description of changes:*

the `S3CrtClient` has the ability to alias a `PutObject` call to a underlying multipart upload call if content-length is not present optimizing uploads, and more importantly, functioning as a "transfer manager" to allow the easy upload of streams where seeking for content lengths is unreasonable.

*Check all that applies:*

- [x] Linux
- [x] Windows
- [x] Android
- [x] MacOS
- [ ] IOS
- [ ] Other Platforms


generated/src/aws-cpp-sdk-s3-crt/include/aws/s3-crt/S3CrtClientConfiguration.h
```diff
@@ -160,6 +160,16 @@ namespace Aws
...
             Aws::Vector<Aws::String> networkInterfaceNames;
+
+            /**
+             * Configuration on how to handle missing content length. By default the SDK attempts to seek the stream
+             * to find the total length of the stream before streaming. CRT will fallback to multipart upload if there
+             * is no content length, creating multipart uploads of partSize before knowing the stream has ended.
+             */
+            enum class CONTENT_LENGTH_CONFIGURATION {
+              SEEK_STREAM,
+              MULTIPART_UPLOAD,
```
DmitriyMusatkin: lets not lock in crt to implementation here - yeah right now it does mpu always, i can imagine in the future us doing buffering and then determining whether to mpu

DmitriyMusatkin (approved):

generated/src/aws-cpp-sdk-s3-crt/include/aws/s3-crt/S3CrtClientConfiguration.h
sbiscigl: i can rename to `SKIP_CONTENT_LENGTH`

PR #3478, opened 2025-07-15
ignore s3 select tests, new aws acconts dont support s3 select

sbiscigl:
*Description of changes:*

[S3 Select is not accepting new customers](https://docs.aws.amazon.com/AmazonS3/latest/userguide/selecting-content-from-objects.html)

> Amazon S3 Select is no longer available to new customers. Existing customers of Amazon S3 Select can continue to use the feature as usual. [Learn more](https://aws.amazon.com/blogs/storage/how-to-optimize-querying-your-data-in-amazon-s3/)

as such we need to disable our CI tests dependent on them as we cannot run them in new aws accounts.

*Check all that applies:*

- [x] Linux
- [x] Windows
- [x] Android
- [x] MacOS
- [ ] IOS
- [ ] Other Platforms

### DIFF
--- a/tests/aws-cpp-sdk-s3-integration-tests/BucketAndObjectOperationTest.cpp
+++ b/tests/aws-cpp-sdk-s3-integration-tests/BucketAndObjectOperationTest.cpp
@@ -56,6 +56,11 @@
 #endif
 #endif
 
+#if defined(_WIN32)
+// disable "warning C4702: unreachable code" from GTEST_SKIP on newer MSVS
+#pragma warning(disable: 4702)
+#endif
+
 #include <aws/core/http/standard/StandardHttpRequest.h>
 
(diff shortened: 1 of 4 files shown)


sbaluja (approved):

tests/aws-cpp-sdk-s3-integration-tests/BucketAndObjectOperationTest.cpp
sbaluja, line 1: See line 1898, would that test need to be skipped as well? It's using s3 select there as well. Do correct me if I'm wrong

sbiscigl: yep, fixed

PR #3483, opened 2025-07-18
Fix incorrect signature with `SetPreservePathSeparators(true)` caused by stale `m_pathHasTrailingSlash`

shurarama:
This PR fixes an issue where pathless requests (e.g., STS or Athena) fail signature validation when `Aws::Http::SetPreservePathSeparators(true)` is enabled.

The root cause lies in the handling of `URI::SetPath(...)`:

- When an empty path is parsed, the SDK assigns `/` via `SetPath("/")`, which sets `m_pathHasTrailingSlash = true`.
- Later, the SDK internally re-applies the path via `uri.SetPath(uri.GetURLEncodedPathRFC3986());`, assuming it is idempotent.
- But in `AddPathSegments(...)`, this flag interferes with leading empty segment removal logic:

```cpp
if (s_preservePathSeparators && m_pathSegments.empty()
    && !split.empty() && split.front().empty()
    && !m_pathHasTrailingSlash)
{
    split.erase(split.begin());
}
```

We reset m_pathHasTrailingSlash to false in URI::SetPath() before calling AddPathSegments(...), ensuring consistent re-parsing.

*Check all that applies:*

- [x] Linux
- [x] Windows
- [ ] Android
- [ ] MacOS
- [ ] IOS
- [ ] Other Platforms

### DIFF
--- a/src/aws-cpp-sdk-core/source/http/URI.cpp
+++ b/src/aws-cpp-sdk-core/source/http/URI.cpp
@@ -254,6 +254,7 @@ Aws::String URI::GetURLEncodedPathRFC3986() const
 void URI::SetPath(const Aws::String& value)
 {
     m_pathSegments.clear();
+    m_pathHasTrailingSlash = false;
     AddPathSegments(value);
 }
 


sbiscigl:
fixed with https://github.com/aws/aws-sdk-cpp/pull/3499, closing this PR in favor of that one

Issue #3496, opened 2025-07-28
Checksum verification fails for composite checkum types


marack:
### Describe the bug

S3 GetObject fails with a "Response checksums mismatch" error when the target object has a "Composite" checksum type.  The only way to successfully retrieve such an object seems to be by setting checksum validation to "WHEN_REQUIRED".

There is a TODO comment in the source where the error is generated that suggests support for these checksums may not be implemented at all: https://github.com/aws/aws-sdk-cpp/blame/5849c282350f3ed1667b8dcd0c34c1259a026928/src/aws-cpp-sdk-core/include/smithy/client/features/ChecksumInterceptor.h#L174

Note that composite checksums end in a dash and number as mentioned in the comment.  See also: https://docs.aws.amazon.com/AmazonS3/latest/userguide/checking-object-integrity.html?icmpid=docs_amazons3_console#ChecksumTypes)

The boto3 API defaults to using mutli-part uploads for files > 8MB, resulting in a Composite checksum by default.  However the C++ API does not seems to support these, but defaults to attempting to verify them anyway.

This seems like a problematic situation where PutObject issued via boto3 will succeed, but then a matching GetObject via C++ will fail even though both calls are issued using the default request parameters.

### Regression Issue

- [x] Select this option if this issue appears to be a regression.

### Expected Behavior

S3 GetObject on an object with a Composite checksum type should succeed

### Current Behavior

S3 GetObject on an object with a Composite checksum fails with an error "Response checksums mismatch".

### Reproduction Steps

Upload a file to S3 using boto3 python API.  File must be > 8MB to trigger use of Composite checksum using default settings.

```python
import boto3
s3 = boto3.client('s3')
s3.upload_file('test.data', 'my-bucket', 'test.data')
```

Now try to dowload the same file with C++ API
```C++
#include <aws/core/Aws.h>
#include <aws/s3/S3Client.h>
#include <aws/s3/model/GetObjectRequest.h>
#include <aws/s3/model/GetObjectResult.h>
#include <iostream>

int main(int argc, char const* argv[])
{
  Aws::InitAPI({});

  Aws::Client::ClientConfiguration config;
  // uncomment the below line to make test pass (but should not be needed and is not recommended)
  //config.checksumConfig.responseChecksumValidation = Aws::Client::ResponseChecksumValidation::WHEN_REQUIRED;

  auto client = Aws::S3::S3Client{config};

  auto req = Aws::S3::Model::GetObjectRequest{};
[...]

PR #3497, opened 2025-07-28
skip response validation when a composite checksum is used

sbiscigl:
*Issue #, if available:*

https://github.com/aws/aws-sdk-cpp/issues/3496

*Description of changes:*

There was a `TODO` left in the [original flex checksum](https://github.com/aws/aws-sdk-cpp/commit/01e61b3137f90509e71765ee7e3b4a39e2e8de91#diff-0d6ccd061cd510e9343ec544ade9358c104f8b004ac66bb6d441c18cd3b73d2aR506) implementation that will cause get requests to fail for [composite checksums](https://docs.aws.amazon.com/AmazonS3/latest/userguide/checking-object-integrity.html?icmpid=docs_amazons3_console#ChecksumTypes). This implements that `TODO` by checking if a `-` exists in the checksum as composite checksums have a dash, and that is not a [valid base64 character encoding](https://en.wikipedia.org/wiki/Base64#Design)

We align with other SDKs in that we will not validate composite checksum for get object

*Check all that applies:*

- [x] Linux
- [x] Windows
- [x] Android
- [x] MacOS
- [ ] IOS
- [ ] Other Platforms


src/aws-cpp-sdk-core/include/smithy/client/features/ChecksumInterceptor.h
```diff
@@ -171,9 +171,12 @@ class ChecksumInterceptor : public smithy::interceptor::Interceptor {
     }
     for (const auto& hashIterator : httpRequest->GetResponseValidationHashes()) {
       Aws::String checksumHeaderKey = Aws::String("x-amz-checksum-") + hashIterator.first;
-      // TODO: If checksum ends with -#, then skip
       if (httpResponse->HasHeader(checksumHeaderKey.c_str())) {
         const Aws::String& checksumHeaderValue = httpResponse->GetHeader(checksumHeaderKey);
+        if (checksumHeaderValue.find('-') != Aws::String::npos) {
```
DmitriyMusatkin: rfind and make sure that number after parses as valid unint

sbiscigl: or actually should we just use the `x-amz-checksum-type` header and skip based on that?

src/aws-cpp-sdk-core/include/smithy/client/features/ChecksumInterceptor.h
```diff
@@ -171,9 +171,17 @@ class ChecksumInterceptor : public smithy::interceptor::Interceptor {
...
       Aws::String checksumHeaderKey = Aws::String("x-amz-checksum-") + hashIterator.first;
-      // TODO: If checksum ends with -#, then skip
       if (httpResponse->HasHeader(checksumHeaderKey.c_str())) {
         const Aws::String& checksumHeaderValue = httpResponse->GetHeader(checksumHeaderKey);
+        // Handle composite checksum case where checksum has a trailing value indicating part size
+        // i.e. c8Sk6w==-2 which denotes a 2 part composite checksum.
+        const auto compositeChecksumIter = checksumHeaderValue.find('-');
+        if (compositeChecksumIter != Aws::String::npos &&
+            compositeChecksumIter + 1 < checksumHeaderValue.length() &&
+            Aws::Utils::StringUtils::ConvertToInt32(checksumHeaderValue.substr(compositeChecksumIter + 1).c_str()) != 0) {
```
sbiscigl, line 181: fwiw `ConvertToInt32` [uses std::atol](https://github.com/aws/aws-sdk-cpp/blob/main/src/aws-cpp-sdk-core/source/utils/StringUtils.cpp#L358) under the hood which [according to docs](https://en.cppreference.com/w/cpp/string/byte/atoi.html)

> Return value
> Integer value corresponding to the contents of str on success.
> If no conversion can be performed, ​0​ is returned.

in most circumstances i would be concerned about the `0` case but from [s3 documentation](https://docs.aws.amazon.com/AmazonS3/latest/userguide/checking-object-integrity.html)

> If you use a multipart upload with Checksums for composite (or part-level) checksums, the multipart upload part numbers must be consecutive and begin with 1. If you try to complete a multipart upload request with nonconsecutive part numbers, Amazon S3 generates an HTTP 500 Internal Server error.

so using `0` as `invalid, error` in this case works

src/aws-cpp-sdk-core/include/smithy/client/features/ChecksumInterceptor.h
sbiscigl: will do `-N`

src/aws-cpp-sdk-core/include/smithy/client/features/ChecksumInterceptor.h
```diff
@@ -171,9 +171,17 @@ class ChecksumInterceptor : public smithy::interceptor::Interceptor {
     }
     for (const auto& hashIterator : httpRequest->GetResponseValidationHashes()) {
       Aws::String checksumHeaderKey = Aws::String("x-amz-checksum-") + hashIterator.first;
-      // TODO: If checksum ends with -#, then skip
       if (httpResponse->HasHeader(checksumHeaderKey.c_str())) {
         const Aws::String& checksumHeaderValue = httpResponse->GetHeader(checksumHeaderKey);
+        // Handle composite checksum case where checksum has a trailing value indicating part size
+        // i.e. c8Sk6w==-2 which denotes a 2 part composite checksum.
+        const auto compositeChecksumIter = checksumHeaderValue.find('-');
```
DmitriyMusatkin: nit: rfind?

src/aws-cpp-sdk-core/include/smithy/client/features/ChecksumInterceptor.h
```diff
@@ -171,9 +171,17 @@ class ChecksumInterceptor : public smithy::interceptor::Interceptor {
...
+        if (compositeChecksumIter != Aws::String::npos &&
+            compositeChecksumIter + 1 < checksumHeaderValue.length() &&
+            Aws::Utils::StringUtils::ConvertToInt32(checksumHeaderValue.substr(compositeChecksumIter + 1).c_str()) != 0) {
+          AWS_LOGSTREAM_DEBUG(AWS_SMITHY_CLIENT_CHECKSUM, "Skipping checksum validation for composite checksum: " << checksumHeaderValue);
```
DmitriyMusatkin, line 182: do we log similar elsewhere? this could cause excessive logging

DmitriyMusatkin (approved):

src/aws-cpp-sdk-core/include/smithy/client/features/ChecksumInterceptor.h
sbiscigl: oops

src/aws-cpp-sdk-core/include/smithy/client/features/ChecksumInterceptor.h
sbiscigl: [we do log successful checksum comparison at debug](https://github.com/aws/aws-sdk-cpp/blob/main/src/aws-cpp-sdk-core/include/smithy/client/features/ChecksumInterceptor.h#L188-L190), so this would be as noisy as that

Issue #3486, opened 2025-07-20
preservePathSeparators breaks sns requests


eyal-di:
### Describe the bug

when using `options.httpOptions.preservePathSeparators = true;` for enabling '/' dirs in s3, sns requests' signatures become invalid.

```cpp
Aws::SDKOptions options;
options.httpOptions.preservePathSeparators = true;
Aws::InitAPI(options);
...
sns = Aws::SNS::SNSClient();
Aws::SNS::Model::ListTopicsRequest request;
auto outcome = sns.ListTopics(request); // outcome is no successful
...
s3 = Aws::S3:S3Client();
Aws::S3::Model::PutObjectRequest request;
request.SetBucket(...);
request.SetKey("/" + filename); // for example
request.SetBody(...);
Aws::S3::Model::PutObjectOutcome outcome = s3.PutObject(request); //outcome is successfull
```

### Expected Behavior

sns client should sign requests correctly even preservePathSeparators is turned on.

### Current Behavior

when sending an sns request the outcome object holds an error:
> The request signature we calculated does not match the signature you provided. Check your AWS Secret Access Key and signing method. Consult the service documentation for details.

### Reproduction Steps

```cpp
int main(){
Aws::SDKOptions options;
options.httpOptions.preservePathSeparators = true;
Aws::InitAPI(options);
sns = Aws::SNS::SNSClient();
Aws::SNS::Model::ListTopicsRequest request;
auto outcome = sns.ListTopics(request); // outcome is not successful
}
```

### Additional Information/Context

log from the faulty run:
[DEBUG] 2025-07-17 09:18:37.875 AWSAuthV4Signer [140324431722304] Canonical Request String: POST
//

amz-sdk-invocation-id:48927949-CB62-47C5-9874-69E5F944D030
amz-sdk-request:attempt=1
content-length:36
content-type:application/x-www-form-urlencoded
host:sns.us-east-2.amazonaws.com
x-amz-api-version:2010-03-31
x-amz-content-sha256:55e772bd3d832d8507965af84ab07fd17d9be3cbb788151ceb0d7fc2be442a54
x-amz-date:20250717T091837Z

amz-sdk-invocation-id;amz-sdk-request;content-length;content-type;host;x-amz-api-version;x-amz-content-sha256;x-amz-date
55e772bd3d832d8507965af84ab07fd17d9be3cbb788151ceb0d7fc2be442a54
[...]

PR #3499, opened 2025-07-30
fix(core): prevent extra trailing slash when s_preservePathSeparators is true

pulimsr:
… is enabled

*Issue #, [3486](https://github.com/aws/aws-sdk-cpp/issues/3486)*

*Description of changes:* Fixed URI methods adding extra trailing slash when s_preservePathSeparators 
is true. Trailing slashes are already preserved in path segments, so no 
additional slash should be added.

Affects: GetPath(), GetURLEncodedPath(), GetURLEncodedPathRFC3986()

*Check all that applies:*

- [x] Linux
- [x] Windows
- [x] Android
- [x] MacOS
- [ ] IOS
- [ ] Other Platforms


src/aws-cpp-sdk-core/source/http/URI.cpp
```diff
@@ -206,7 +206,7 @@ Aws::String URI::GetPath() const
         path.append(segment);
     }
 
-    if (m_pathSegments.empty() || m_pathHasTrailingSlash)
+    if (m_pathSegments.empty() || (m_pathHasTrailingSlash && !s_preservePathSeparators))
```
sbiscigl: I think we're verrrrrrrry close to the root of the problem, but might not be there yet. so in the original code
```c++
if (m_pathSegments.empty() || m_pathHasTrailingSlash) {
   ss << '/';
}
```

we only add the trailing slash if there are no segments _or_ `m_pathHasTrailingSlash` is set to true. we only set that variable to true when [we add a path](https://github.com/aws/aws-sdk-cpp/blame/main/src/aws-cpp-sdk-core/include/aws/core/http/URI.h#L156). so instead of repeating that logic here, lets move that up to where we actually set `m_pathHasTrailingSlash`. This is likely because with preserve path we keep the slash, as before when we normalized it we removed it.

so something like
```
 m_pathHasTrailingSlash = !s_preservePathSeparators && (!segments.empty() && segments.back() == '/');
```
would likely do the trick

sbiscigl (approved):

PR #3500, opened 2025-07-31
Reuse checksum on streaming retries + refactor ChecksumInterceptor

sbaluja:
*Issue #, if available:*

*Description of changes:*
- On streaming operations, if the request payload was fully consumed and request checksum calculation was complete we will save the request checksum algorithm and value in memory to be reused for retry.
- Refactored the ChecksumInterceptor logic for setting checksum headers/values to be split across separate functions to follow single responsibility and allow for easier addition of checksum algorithms in the future.

*Check all that applies:*
  - Tests already exist to validate these changes.

- [x] Linux
- [x] Windows
- [x] Android
- [x] MacOS
- [ ] IOS
- [ ] Other Platforms

### DIFF
--- a/src/aws-cpp-sdk-core/include/aws/core/AmazonWebServiceRequest.h
+++ b/src/aws-cpp-sdk-core/include/aws/core/AmazonWebServiceRequest.h
@@ -35,6 +35,10 @@ namespace Aws
     typedef std::function<void(const AmazonWebServiceRequest&)> RequestRetryHandler;
     typedef std::function<void(const Aws::Http::HttpRequest&)> RequestSignedHandler;
 
+    struct RetryContext {
+      std::pair<Aws::String, std::shared_ptr<Aws::Utils::Crypto::Hash>> m_requestHash;
+    };
+
     /**
      * Base level abstraction for all modeled AWS requests
      */
@@ -222,7 +226,9 @@ namespace Aws
          */
         Aws::Set<Aws::Client::UserAgentFeature> GetUserAgentFeatures() const { return m_userAgentFeatures; }
 
-      inline virtual bool RequestChecksumRequired() const { return false; }
+        inline virtual bool RequestChecksumRequired() const { return false; }
+
+        mutable Aws::RetryContext m_retryContext;
     protected:
         /**
          * Default does nothing. Override this to convert what would otherwise be the payload of the
--- a/src/aws-cpp-sdk-core/include/smithy/client/features/ChecksumInterceptor.h
+++ b/src/aws-cpp-sdk-core/include/smithy/client/features/ChecksumInterceptor.h
@@ -82,56 +82,10 @@ class ChecksumInterceptor : public smithy::interceptor::Interceptor {
         // For non-streaming payload, the resolved checksum location is always header.
         // For streaming payload, the resolved checksum location depends on whether it is an unsigned payload, we let
         // AwsAuthSigner decide it.
-        if (request.IsStreaming() && checksumValueAndAlgorithmProvided) {
-          addChecksumFeatureForChecksumName(checksumAlgorithmName, request);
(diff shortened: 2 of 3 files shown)


src/aws-cpp-sdk-core/include/aws/core/AmazonWebServiceRequest.h
```diff
@@ -222,7 +226,9 @@ namespace Aws
          */
         Aws::Set<Aws::Client::UserAgentFeature> GetUserAgentFeatures() const { return m_userAgentFeatures; }
 
-      inline virtual bool RequestChecksumRequired() const { return false; }
+        inline virtual bool RequestChecksumRequired() const { return false; }
+
+        mutable Aws::RetryContext m_retryContext;
```
sbiscigl: please make private with getters/setters

src/aws-cpp-sdk-core/include/smithy/client/features/ChecksumInterceptor.h
```diff
@@ -233,6 +172,94 @@ class ChecksumInterceptor : public smithy::interceptor::Interceptor {
...
+        httpRequest->SetRequestHash(algorithm, hash);
+      }
+    } else {
+      httpRequest->SetHeaderValue(checksumType, checksumValue);
+    }
+  }
+
+  void calculateAndSetChecksum(const Aws::AmazonWebServiceRequest& request, std::shared_ptr<Aws::Http::HttpRequest> httpRequest,
+                               const Aws::String& algorithm, const Aws::String& checksumType) {
+    static const Aws::Array<std::pair<const char*, ChecksumHandler>, 5> algorithmMap = {{
```
sbiscigl, line 190: should declare it as static in the compilation unit i.e. just in the `.cpp`. no reason to have it in the function.

src/aws-cpp-sdk-core/include/aws/core/AmazonWebServiceRequest.h
```diff
@@ -35,6 +35,10 @@ namespace Aws
     typedef std::function<void(const AmazonWebServiceRequest&)> RequestRetryHandler;
     typedef std::function<void(const Aws::Http::HttpRequest&)> RequestSignedHandler;
 
+    struct RetryContext {
+      std::pair<Aws::String, std::shared_ptr<Aws::Utils::Crypto::Hash>> m_requestHash;
```
sbiscigl: this will initialize to 
`("", nullptr)`,
 which is why you do 
`if (request.m_retryContext.m_requestHash.second == nullptr) {` 
later on to check for "absence". 
would be better to have this as
`std::shared_ptr<std::pair<Aws::String, std::shared_ptr<Aws::Utils::Crypto::Hash>>> m_requestHash`
or
`Aws::Crt::Optional<std::pair<Aws::String, std::shared_ptr<Aws::Utils::Crypto::Hash>>> m_requestHash`

to better call out "absent" or "present"

src/aws-cpp-sdk-core/include/smithy/client/features/ChecksumInterceptor.h
sbaluja: There is no `.cpp` associated with the ChecksumInterceptor, all functionality is within the header file, unless you're referring to something else.

src/aws-cpp-sdk-core/include/aws/core/AmazonWebServiceRequest.h
sbaluja: Fair enough, was keeping it consistent with how we check for absence on the httpRequest itself  - `httpRequest->GetRequestHash().second == nullptr`, but this sounds like a good change, will implement

sbiscigl (approved):

PR #3505, opened 2025-08-01
swap to use CRT implementation of STS Web Identity Provider

sbiscigl:
*Description of changes:*

We currently have a handwritten client for calling `AssumeRoleWithWebIdentity` in our credentials provider chain. We are moving this refer the the [common runtime implementation](https://github.com/awslabs/aws-c-auth).

*Check all that applies:*

- [x] Linux
- [x] Windows
- [x] Android
- [x] MacOS
- [ ] IOS
- [ ] Other Platforms


src/aws-cpp-sdk-core/source/auth/STSCredentialsProvider.cpp
```diff
@@ -2,166 +2,108 @@
...
-        }
-    }
-
-    if (m_tokenFile.empty())
-    {
-        AWS_LOGSTREAM_WARN(STS_ASSUME_ROLE_WEB_IDENTITY_LOG_TAG, "Token file must be specified to use STS AssumeRole web identity creds provider.");
-        return; // No need to do further constructing
+  Aws::Crt::Auth::CredentialsProviderSTSWebIdentityConfig stsConfig{};
+  stsConfig.Bootstrap = GetDefaultClientBootstrap();
+  Aws::Crt::Io::TlsContextOptions tlsCtxOptions = Aws::Crt::Io::TlsContextOptions::InitDefaultClient();
```
DmitriyMusatkin, line 24: doesnt cpp sdk have some option to override tls context? might want to pass it here

src/aws-cpp-sdk-core/source/auth/STSCredentialsProvider.cpp
```diff
@@ -2,166 +2,108 @@
...
-    config.retryStrategy = Aws::MakeShared<SpecifiedRetryableErrorsRetryStrategy>(STS_ASSUME_ROLE_WEB_IDENTITY_LOG_TAG, retryableErrors, 3/*maxRetries*/);
-
-    m_client = Aws::MakeUnique<Aws::Internal::STSCredentialsClient>(STS_ASSUME_ROLE_WEB_IDENTITY_LOG_TAG, config);
-    m_initialized = true;
-    AWS_LOGSTREAM_INFO(STS_ASSUME_ROLE_WEB_IDENTITY_LOG_TAG, "Creating STS AssumeRole with web identity creds provider.");
+    return UUID::RandomUUID();
+  }().c_str();
+  m_credentialsProvider = Aws::Crt::Auth::CredentialsProvider::CreateCredentialsProviderSTSWebIdentity(stsConfig);
+  if (m_credentialsProvider && m_credentialsProvider->IsValid()) {
+    m_state = STATE::INITIALIZED;
```
DmitriyMusatkin, line 41: does it need to be thread safe?

DmitriyMusatkin (approved):

src/aws-cpp-sdk-core/source/auth/STSCredentialsProvider.cpp
sbiscigl: i dont think it needs to be right now since `m_state` is only set in the constructor and is a private member. If it mutated outside the constructor then yes absolutely, but dont think its a concern right now.

src/aws-cpp-sdk-core/source/auth/STSCredentialsProvider.cpp
sbiscigl: [We allow someone to override the TLS connection options](https://github.com/aws/aws-sdk-cpp/blob/main/src/aws-cpp-sdk-core/source/Aws.cpp#L93-L104), which does contain a [underlying handle](https://github.com/awslabs/aws-crt-cpp/blob/main/include/aws/crt/io/TlsOptions.h#L204) that has access to the tls context -- let me see if there is a way to expose this, it should re-use that context.

sbiscigl: youre right i think we need to make a change in CRT to pass options, made the PR, let me know what you think https://github.com/awslabs/aws-crt-cpp/pull/763

sbiscigl: updated

PR #3515, opened 2025-08-13
Initial implementation of AWSRpcV2CborClient

sbaluja:
*Issue #, if available:*

*Description of changes:*
Initial implementation of the core CBOR Client. Purpose of this PR is to bring in changes incrementally. 
*Check all that applies:*

- [x] Linux
- [x] Windows
- [x] Android
- [x] MacOS
- [ ] IOS
- [ ] Other Platforms

### DIFF
--- a/src/aws-cpp-sdk-core/include/aws/core/client/AWSRpcV2CborClient.h
+++ b/src/aws-cpp-sdk-core/include/aws/core/client/AWSRpcV2CborClient.h
@@ -1,0 +1,112 @@
+/**
+ * Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.
+ * SPDX-License-Identifier: Apache-2.0.
+ */
+
+#pragma once
+#if !defined(AWS_CBOR_CLIENT_H)
+#define AWS_CBOR_CLIENT_H
+
+#include <aws/core/Core_EXPORTS.h>
+#include <aws/core/client/AWSClient.h>
+#include <aws/crt/cbor/Cbor.h>
+
+namespace Aws
+{
+    namespace Client
+    {
+        using RpcV2CborOutcome = Utils::Outcome<AmazonWebServiceResult<Aws::UniquePtr<Crt::Cbor::CborDecoder>>, AWSError<CoreErrors>>;
+
+        template <typename OutcomeT, typename ClientT, typename AWSEndpointT, typename RequestT, typename HandlerT>
+        class BidirectionalEventStreamingTask;
+
+        /**
+         *  AWSClient that handles marshalling cbor response bodies. You would inherit from this class
+         *  to create a client that uses Cbor as its payload format.
+         */
+        class AWS_CORE_API AWSRpcV2CborClient : public AWSClient
+        {
+        public:
+            typedef AWSClient BASECLASS;
+
+            /**
+             * Simply calls AWSClient constructor.
+             */
+            AWSRpcV2CborClient(const Aws::Client::ClientConfiguration& configuration,
+                const std::shared_ptr<Aws::Client::AWSAuthSigner>& signer,
+                const std::shared_ptr<AWSErrorMarshaller>& errorMarshaller);
+
+            /**
+             * Simply calls AWSClient constructor.
+             */
+            AWSRpcV2CborClient(const Aws::Client::ClientConfiguration& configuration,
+                    const std::shared_ptr<Aws::Auth::AWSAuthSignerProvider>& signerProvider,
+                    const std::shared_ptr<AWSErrorMarshaller>& errorMarshaller);
+
+            virtual ~AWSRpcV2CborClient() = default;
+
+        protected:
+            template <typename OutcomeT, typename ClientT, typename AWSEndpointT, typename RequestT, typename HandlerT>
+            friend class BidirectionalEventStreamingTask; // allow BidirectionalEventStreamingTask to MakeRequests
+            /**
+             * Converts/Parses an http response into a meaningful AWSError object using the cbor message structure.
+             */
+            virtual AWSError<CoreErrors> BuildAWSError(const std::shared_ptr<Aws::Http::HttpResponse>& response) const override;
+
+            /**
+             * Returns a pointer to a Cbor decoder or an error from the request. Does some marshalling cbor and raw streams,
+             * then just calls AttemptExhaustively.
+             *
+             * method defaults to POST
+             */
+            RpcV2CborOutcome MakeRequest(const Aws::AmazonWebServiceRequest& request,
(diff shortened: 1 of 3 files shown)


src/aws-cpp-sdk-core/source/client/AWSRpcV2CborClient.cpp
```diff
@@ -0,0 +1,259 @@
...
+#include <aws/core/utils/UUID.h>
+#include <aws/core/utils/event/EventStream.h>
+#include <aws/core/utils/logging/LogMacros.h>
+#include <aws/core/utils/memory/stl/AWSStringStream.h>
+#include <aws/core/utils/xml/XmlSerializer.h>
+#include <smithy/tracing/TracingUtils.h>
+
+#include <cassert>
+
+#include "aws/crt/cbor/Cbor.h"
```
sbiscigl: nit: use `<> imports`

src/aws-cpp-sdk-core/source/client/AWSRpcV2CborClient.cpp
```diff
@@ -0,0 +1,259 @@
...
+#include <aws/core/http/HttpClient.h>
+#include <aws/core/http/HttpResponse.h>
+#include <aws/core/http/URI.h>
+#include <aws/core/monitoring/MonitoringManager.h>
+#include <aws/core/utils/Outcome.h>
+#include <aws/core/utils/UUID.h>
+#include <aws/core/utils/event/EventStream.h>
+#include <aws/core/utils/logging/LogMacros.h>
+#include <aws/core/utils/memory/stl/AWSStringStream.h>
+#include <aws/core/utils/xml/XmlSerializer.h>
```
sbiscigl: dont think we need a xml serializer in here right?

src/aws-cpp-sdk-core/source/client/AWSRpcV2CborClient.cpp
```diff
@@ -0,0 +1,259 @@
...
+    return error;
+}
+
+Aws::UniquePtr<Aws::Crt::Cbor::CborDecoder> AWSRpcV2CborClient::CreateCborDecoder(
+    const HttpResponseOutcome& httpOutcome) const
+{
+    Aws::StringStream ss;
+    ss << httpOutcome.GetResult()->GetResponseBody().rdbuf();
+    Aws::Crt::ByteCursor cborDecoder= Aws::Crt::ByteCursorFromCString(ss.str().c_str());
+    return Aws::MakeUnique<Crt::Cbor::CborDecoder>("TODO:", cborDecoder);
```
sbiscigl: nit: `TODO` should be `AWS_CBOR_CLIENT_LOG_TAG`

src/aws-cpp-sdk-core/include/aws/core/client/AWSRpcV2CborClient.h
```diff
@@ -0,0 +1,112 @@
...
+        using RpcV2CborOutcome = Utils::Outcome<AmazonWebServiceResult<Aws::UniquePtr<Crt::Cbor::CborDecoder>>, AWSError<CoreErrors>>;
+
+        template <typename OutcomeT, typename ClientT, typename AWSEndpointT, typename RequestT, typename HandlerT>
+        class BidirectionalEventStreamingTask;
+
+        /**
+         *  AWSClient that handles marshalling cbor response bodies. You would inherit from this class
+         *  to create a client that uses Cbor as its payload format.
+         */
+        class AWS_CORE_API AWSRpcV2CborClient : public AWSClient
```
sbiscigl: one high level comment i have is that this class and `JsonClient` is about 99 percent the same. is there a opportunity to re-use some code between the two? it feels like the only thing that is different is the outcome type. can we template the outcome type and keep the logic the same? i.e.

```cpp
#include <type_traits>

class AwsOutcome {};

class AwsClient {
 public:
  AwsOutcome MakeRequest() { return AwsOutcome{}; }
};

template <typename OutcomeT,
  typename std::enable_if<std::is_constructible<OutcomeT, AwsOutcome>::value, int>::type = 0>
class BaseClient : public AwsClient {
 public:
  OutcomeT MakeRequest() { return OutcomeT{AwsClient::MakeRequest()}; }
};

class JsonOutcome {
 public:
  explicit JsonOutcome(AwsOutcome outcome) {}
};

class CborOutcome {
 public:
  explicit CborOutcome(AwsOutcome outcome) {}
};

class JsonClient : public BaseClient<JsonOutcome> {
 public:
  JsonOutcome MakeRequest() { return BaseClient::MakeRequest(); }
};

class CborClient : public BaseClient<CborOutcome> {
 public:
  CborOutcome MakeRequest() { return BaseClient::MakeRequest(); }
};

int main() {
  CborClient cbor_client{};
  cbor_client.MakeRequest();
  JsonClient json_client{};
  json_client.MakeRequest();
  return 0;
}
```

using sfinae to ensure that `OutcomeT` is constructible from `AwsOutcome` which in this case is `HttpResponseOutcome` but that general idea.

sbaluja: I agree that much of the code is the same, and this would be a perfect use case for a template-based approach. However, it would also increase the scope significantly, since that would be a change that could affect current client functionality. I think if we did want to do a refactor like this, it'd be best to do it in a separate PR after these changes are merged

sbiscigl: > However, it would also increase the scope significantly, since that would be a change that could affect current client functionality.

I don't think thats true, if anything this reduces scope as it re-uses a ton of functionality we already have implemented for json. copy-ing and pasting here would create duplicates of the same code path creating maintainability concerns. we're here editing this now, so lets do it the right way the first time

src/aws-cpp-sdk-core/source/client/AWSRpcV2CborClient.cpp
```diff
@@ -0,0 +1,107 @@
...
+
+        Aws::StringStream ss;
+        ss << "No response body.";
+        error = AWSError<CoreErrors>(errorCode, "", ss.str(),
+            IsRetryableHttpResponseCode(responseCode));
+    }
+    else
+    {
+        assert(httpResponse->GetResponseCode() != HttpResponseCode::OK);
+        error = GetErrorMarshaller()->Marshall(*httpResponse);
```
sbiscigl: since `GetErrorMarshaller` is a function on `AwsClient` and `Marshall` is a pure virtual interface, cant function be moved to the base `AWSProtocolClient` class? also looking at the diff it looks like theres actually no difference between the impls right now.

sbiscigl (approved):

sbiscigl (approved):

PR #3526, opened 2025-08-29
Smoke tests environment improvements

pulimsr:
*Description of changes:* Added proper exit codes to the Python script and removed unnecessary parts from the shell script

*Check all that applies:*

- [x] Linux
- [x] Windows
- [x] Android
- [x] MacOS
- [ ] IOS
- [ ] Other Platforms


tools/scripts/run_smoke_tests.py
```diff
@@ -37,14 +29,19 @@ def main():
     total_tests = 0
     all_results = defaultdict(list)
 
-    services=should_run_service(services)
+    service_ids = os.environ.get('AWS_SMOKE_TEST_SERVICE_IDS', '')
+    if not service_ids:
+        return 0
```
sbiscigl: i dont think this is what we want to do, this effectively creates a path where if `AWS_SMOKE_TEST_SERVICE_IDS` is removed for any reason we start ignoring smoke tests. this logic in the most abstract form should become

````
if (has service id AND service id is a known service in the SDK) {
  run _only_ the smoke test for service id
} else {
 run all tests
}
````

the makes it so we never skip smoke tests based on service id logic, we _must_ default to running all tests if we can make sense of _service_id_

File: tools/scripts/run_smoke_tests.py [hunk 2]
```diff
@@ -36,30 +28,66 @@ def main():
...
+                    all_results[service].append({
+                        'service': service,
+                        'name': test['name'],
+                        'status': 'SKIPPED',
+                        'failure': None,
                     })
+        except (FileNotFoundError, json.JSONDecodeError):
+            pass
+        finally:
+            os.unlink(json_output_path)
```
pulimsr, line 91: Following [Reporting skipped tests on the SEP](https://code.amazon.com/packages/AwsDrSeps/blobs/804fbd82f183eb49eb5bace479a1666e3afcac99/--/seps/accepted/shared/shared_smoke_test_runners.md#:~:text=Reporting%20skipped%20tests) and listing all the services that have smoke-tests but are not on AWS_SMOKE_TEST_SERVICE_IDS

sbiscigl (approved):

PR #3530, opened 2025-09-02
[C++] SDKs to track feature Ids for Credentials (ENV) and Custom UA

kai-ion:
*Issue #, if available:*

*Description of changes:*

*Check all that applies:*

- [x] Linux
- [x] Windows
- [x] Android
- [x] MacOS
- [ ] IOS
- [ ] Other Platforms

### DIFF
--- a/src/aws-cpp-sdk-core/source/auth/signer/AWSAuthV4Signer.cpp
+++ b/src/aws-cpp-sdk-core/source/auth/signer/AWSAuthV4Signer.cpp
@@ -8,6 +8,7 @@
 #include <aws/core/auth/signer/AWSAuthSignerHelper.h>
 
 #include <aws/core/auth/AWSCredentialsProvider.h>
+#include <aws/core/client/UserAgent.h>
 #include <aws/core/http/HttpRequest.h>
 #include <aws/core/http/URI.h>
 #include <aws/core/utils/DateTime.h>
@@ -25,6 +26,7 @@
 
 #include <iomanip>
 #include <cstring>
+#include <numeric>
 
 using namespace Aws;
 using namespace Aws::Client;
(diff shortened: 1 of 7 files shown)


src/aws-cpp-sdk-core/source/auth/signer/AWSAuthV4Signer.cpp
```diff
@@ -595,3 +605,56 @@ Aws::Auth::AWSCredentials AWSAuthV4Signer::GetCredentials(const std::shared_ptr<
...
 }
+
+void AWSAuthV4Signer::UpdateUserAgentWithCredentialFeatures(Aws::Http::HttpRequest& request, const Aws::Auth::CredentialsResolutionContext& context) const {
+    if (!request.HasHeader(USER_AGENT)) {
+        AWS_LOGSTREAM_DEBUG(v4LogTag, "Request does not have User-Agent header, skipping credential feature update");
+        return;
+    }
+
+    const auto userAgent = request.GetHeaderValue(USER_AGENT);
+    if (userAgent.find("m/") == Aws::String::npos) {
```
sbiscigl: if we are adding this check here why do we still have logic around it not existing further down? i.e. we are still doing
```cpp
  auto userAgentParsed = Aws::Utils::StringUtils::Split(userAgent, ' ');
  auto metricsSegment = std::find_if(userAgentParsed.begin(), userAgentParsed.end(),
      [](const Aws::String& value) { return value.find("m/") != Aws::String::npos; });

  if (metricsSegment != userAgentParsed.end()) {
      // Add new metrics to existing metrics section
      *metricsSegment = Aws::String{*metricsSegment + "," + credentialFeatures};
  } else {
      // No metrics section exists, add new one
      userAgentParsed.push_back("m/" + credentialFeatures);
  }
```

we should consolidate logic to avoid doing that search multiple times/handle checks that we know will be true

PR #3536, opened 2025-09-05
[C++] SDKs to track feature Ids for Credentials (SSO/LegacySSO Provider)

kai-ion:
*Issue #, if available:*

*Description of changes:*
Implement User-Agent 2.1 business metrics for credential provider tracking in AWS SDK for C++.

*Check all that applies:*

- [x] Linux
- [x] Windows
- [x] Android
- [x] MacOS
- [ ] IOS
- [ ] Other Platforms


src/aws-cpp-sdk-core/source/auth/SSOCredentialsProvider.cpp
```diff
@@ -101,6 +101,13 @@ void SSOCredentialsProvider::Reload()
     AWS_LOGSTREAM_TRACE(SSO_CREDENTIALS_PROVIDER_LOG_TAG, "Successfully retrieved credentials with AWS_ACCESS_KEY: " << result.creds.GetAWSAccessKeyId());
 
     m_credentials = result.creds;
+    if (!m_credentials.IsEmpty()) {
```
sbiscigl, line 104: will this actually be called if the credentials provider is called twice? given that the credentials did not expire between the calls?

tests/aws-cpp-sdk-core-tests/aws/auth/SSOCredentialTrackingTest.cpp
```diff
@@ -0,0 +1,275 @@
...
+// Minimal AWSClient wrapper so we can make a signed call and inspect User-Agent
+class CredentialTestingClient : public AWSClient {
+public:
+  explicit CredentialTestingClient(const ClientConfiguration& config,
+                                   const std::shared_ptr<AWSCredentialsProvider>& provider)
+      : AWSClient(config,
+                  Aws::MakeShared<Aws::Client::AWSAuthV4Signer>(TEST_LOG_TAG, provider, "service", config.region),
+                  Aws::MakeShared<MockAWSErrorMarshaller>(TEST_LOG_TAG)) {}
+
+  HttpResponseOutcome MakeRequest(const Aws::AmazonWebServiceRequest& request) {
```
sbiscigl: this is sort of weird, we shouldnt be subclassing `AwsClient` and then creating a function as a entry point, we should make a _real_ aws client and be testing that object. make a mock aws client like we do in the [RDS unit test](https://github.com/aws/aws-sdk-cpp/blob/60fdb8b114819dde8b6be7101c10ac6c43e95a90/tests/aws-cpp-sdk-rds-integration-tests/RDSTest.cpp#L59-L72). but not RDS specific and invoke the public functions on AWS client.

tests/aws-cpp-sdk-core-tests/aws/auth/SSOCredentialTrackingTest.cpp
```diff
@@ -0,0 +1,275 @@
...
+  }
+};
+
+class SSOCredentialsProviderTrackingTest : public Aws::Testing::AwsCppSdkGTestSuite {
+protected:
+  void SetUp() override {
+    AwsCppSdkGTestSuite::SetUp();
+
+    // Build paths the same way the SDK does
+    const Aws::String profileDir = ProfileConfigFileAWSCredentialsProvider::GetProfileDirectory();
```
sbiscigl, line 69: if this ran in on a users machine would this overrite their SSO configuraiton if it existed? If so we need to have this existing in a temp file/temp directory as we do in other tests. we cant mess with use credentials if its building on a users machine.

PR #3538, opened 2025-09-08
Cbor Serialization/Deserialization & Protocol Tests 

sbaluja:
*Issue #, if available:*

*Description of changes:*
- Support for code generation of Cbor clients and serialization/deserialization logic of request/results. 
- Updated Protocol Tests

*Check all that applies:*

- [x] Linux
- [x] Windows
- [x] Android
- [x] MacOS
- [ ] IOS
- [ ] Other Platforms


tools/code-generation/generator/src/main/java/com/amazonaws/util/awsclientgenerator/domainmodels/codegeneration/Metadata.java
```diff
@@ -17,6 +17,7 @@
 @Data
 public class Metadata {
     private static List<String> supportedProtocols = ImmutableList.of(
+            "smithy-rpc-v2-cbor",
```
sbiscigl: shouldnt we put this at the end or it will select cbor for existing services when running code gen right now?

tools/code-generation/generator/src/main/java/com/amazonaws/util/awsclientgenerator/domainmodels/codegeneration/Shape.java
```diff
@@ -181,6 +188,10 @@ public boolean hasMember(String member) {
         return members != null && members.keySet().stream().anyMatch(key -> key.equals(member));
     }
 
+    public boolean hasMembers() {
+        return members != null && !members.isEmpty();
```
sbiscigl, line 192: nit: prefer using [apache common lang](https://commons.apache.org/proper/commons-collections/apidocs/org/apache/commons/collections4/CollectionUtils.html#isEmpty(java.util.Collection)) for this 
```java
return CollectionUtils.isEmpty(members);
```

tools/code-generation/generator/src/main/java/com/amazonaws/util/awsclientgenerator/domainmodels/codegeneration/cpp/CppViewHelper.java
```diff
@@ -188,9 +189,9 @@ public static String computeJsonizeString(Shape shape) {
         return computeJsonizeString(shape, false);
     }
 
-    public static String computeCppType(Shape shape) {
+    static String computeCppTypeInternal(Shape shape, Map<String, String> typeMapping) {
```
sbiscigl, line 192: does this file need to change anymore sice we moved to a hide-parent-function approach in the the CBOR cpp shape helper?

tools/code-generation/generator/src/main/java/com/amazonaws/util/awsclientgenerator/generators/cpp/CborCppClientGenerator.java
```diff
@@ -0,0 +1,190 @@
...
+
+import java.nio.charset.StandardCharsets;
+import java.util.HashMap;
+import java.util.List;
+import java.util.Map;
+import java.util.Optional;
+import java.util.stream.Collectors;
+import java.util.stream.IntStream;
+
+public class CborCppClientGenerator extends CppClientGenerator {
```
sbiscigl, line 27: so `CborCppClientGenerator` and `JsonCppClientGenerator` are extremely similar just like how the clients are very similar, i do believe the majority of the difference is the template files and the helper class. would it make sense to extract a `ProtocolClientGenerator` that looks something like
```java
class ProtocolClientGenerator {
  public ProtocolClientGenerator(finalString templatePath1,
    String template path2.
     ...,
     Class helperClass) {
    ...   
  }

  ...common methods
}
```

the we could do each generator like

```java
class CborCppClientGenerator extends ProtocolClientGenerator {
  super( ...)
}
```

this would help reduce a lot of duplication.

tools/code-generation/generator/src/main/java/com/amazonaws/util/awsclientgenerator/generators/cpp/CborCppClientGenerator.java
```diff
@@ -0,0 +1,190 @@
...
+            {
+                return GenerateSmithyClientSourceFile(serviceModels.get(index), index, Optional.empty());
+            }
+            else
+            {
+                return GenerateLegacyClientSourceFile(serviceModels.get(index), index);
+            }
+        }).collect(Collectors.toList());
+    }
+}
```
sbiscigl, line 195: so this class is notably missing a overload for `generateEventStreamHandlerSourceFile`. according to [smithy docs](https://smithy.io/2.0/aws/amazon-eventstream.html)

> The value of this header is the [media type](https://datatracker.ietf.org/doc/html/rfc6838.html) of the payload. This value follows the rules of the particular service protocol being used. For example, a JSON payload MAY use a :content-type of application/json.

which does mean the event stream payload could be cbor in addition in our internal spec (sorry no link for outside readers) it says

> Support all major Smithy features. This includes documents and event streaming.

Which I think means that we need to support event streams where the payload type is event stream.

That said this PR has grown so we can actually implement that later, for now i think it would be good enough to override it and have a runtime exception mentioning that we dont support it yet so that we fail in a good known way

generated/protocol-tests/test-clients/aws-cpp-sdk-rpcv2protocol/source/model/ComplexError.cpp
```diff
@@ -0,0 +1,206 @@
...
+}
+
+return *this;
+}
+
+void ComplexError::CborEncode(Aws::Crt::Cbor::CborEncoder& encoder) const
+{
+
+    // Calculate map size
+    size_t mapSize = 0;
```
sbiscigl: yeah we gotta do something about clang-format and auto generated files there are two points we should discuss as a team:
* i think we can clang-tidy protocol tests today, i dont think anyone would disagree, we should do it today before it gets worse. should be as easy as [running clang format at the end of the protocol test generator](https://github.com/aws/aws-sdk-cpp/blob/main/tools/scripts/codegen/protocol_tests_gen.py#L82-L91). 
* we should be able to clang-tidy generated code. we will lose some git blame/history functionality, however since it is generated i think model history is more important than formatting history and then we can leave this whole formatting thing behind us. before we actually start having real cbor clients we should do this too.

generated/protocol-tests/test-clients/aws-cpp-sdk-rpcv2protocol/source/model/ComplexNestedErrorData.cpp
```diff
@@ -0,0 +1,180 @@
...
+                }
+                break;
+            }
+            auto val = decoder->PopNextTextVal();
+            if (val.has_value()) {
+                ss << Aws::String(reinterpret_cast<const char*>(val.value().ptr), val.value().len);
+            }
+        }
+        m_foo = ss.str();
+        ss.clear();
```
sbiscigl: since the string stream is scoped to the else statement is there a reason we need to clear it since it will be deleted after this scope?

tests/testing-resources/include/aws/testing/AwsProtocolTestHelpers.h
```diff
@@ -110,7 +110,7 @@ class AwsProtocolTestSuite : public Aws::Testing::AwsCppSdkGTestSuite {
     AWS_ASSERT_SUCCESS(setNextResponse);
   }
 
-  void ValidateRequestSent(const ExpectedRequest& expected = ExpectedRequest()) const {
+  void ValidateRequestSent(const ExpectedRequest& expected = ExpectedRequest(), const bool ignoreBody = false) const {
```
sbiscigl: boolean argmuents are a anti pattern that i would avoid, i would suggest creating two methods on that calls the other i.e.

```cpp
void ValidateRequestSent(const ExpectedRequest& expected = ExpectedRequest()) const {
  ValidateBody()
  ValidateHeaders()
}
```

making it composable, so in code gen you could call each method on its own if need be.

tests/testing-resources/include/aws/testing/AwsProtocolTestHelpers.h
```diff
@@ -119,7 +119,7 @@ class AwsProtocolTestSuite : public Aws::Testing::AwsCppSdkGTestSuite {
     if (!expected.method.empty()) {
       EXPECT_STREQ(expected.method.c_str(), receivedRequest.GetMethod().c_str());
     }
-    if (!expected.body.empty()) {
+    if (!expected.body.empty() && !ignoreBody) {
```
sbiscigl: OK so i see what is going on here you ignore body and code generate logic for checking in the tests. this isnt a bad approach but using boolean arguments is a anitpattern and something we should avoid. I think we extend this just a little bit more to make it so all "comparison" happens in the `ValidateRequestSent` function. we could extract the body comparison to a virtual function that the tests can override

i.e. 

```cpp
class ExpectedRequest {};

class TestHelper {
 public:
  void ValidateRequestSent(const ExpectedRequest& expected = ExpectedRequest()) {
    ExpectedRequest seen;
    BodyCompare(expected, seen);
  }
protected:
  virtual void BodyCompare(const ExpectedRequest& expected, const ExpectedRequest& seen) {
    //.. default logic
  }
};

class TestWithCustomBodyLogic: public TestHelper {
 public:
  void TestMethod() {
    ValidateRequestSent(ExpectedRequest());
  }
protected:
  void BodyCompare(const ExpectedRequest& expected, const ExpectedRequest& seen) override {
    //.. custom logic
  } 
};
```

src/aws-cpp-sdk-core/include/aws/core/utils/DateTime.h
```diff
@@ -77,6 +77,11 @@ namespace Aws
              */
             DateTime(double secondsSinceEpoch);
 
+           /**
+            * Initializes time point to epoch time in seconds
+            */
+            DateTime(uint64_t secondsSinceEpoch);
```
sbiscigl, line 83: im not sure we can get away with this, lets take the example program

```c++
#include <iostream>

void foo(double num) {
  std::cout << num << std::endl;
}

auto main() -> int {
  foo(1);
}
```

then we add a function that accepts a `uint64_t`

```cpp
#include <iostream>

void foo(double num) {
  std::cout << num << std::endl;
}

void foo(uint64_t num) {
  std::cout << num << std::endl;
}

auto main() -> int {
  foo(1);
}
```

this [no longer compiles](https://godbolt.org/z/zTdeMsnvP), so we have to be clever getting around this. we could add a static function 

```
static DateTime FromUnsignedSeconds(uint64_t secondsSinceEpoch)
```

that achieves the same result but without ambigious overloads

tools/code-generation/generator/src/main/java/com/amazonaws/util/awsclientgenerator/domainmodels/codegeneration/Metadata.java
sbaluja: Will change this before a merge, just wanted to see dry runs pass with cbor client generation

tools/code-generation/generator/src/main/java/com/amazonaws/util/awsclientgenerator/domainmodels/codegeneration/cpp/CppViewHelper.java
sbaluja: Its a refactor so the logic for computing the type remains in this class, and from the CborViewHelper we can just pass in a different typeMapping. Avoids duplication of the logic across both classes since we only want to change the type mappings for int -> int64_t

tools/code-generation/generator/src/main/java/com/amazonaws/util/awsclientgenerator/generators/cpp/CborCppClientGenerator.java
sbaluja: Hm, I think this makes more sense if we did this for _all_ protocol generators then, rather than just cbor/json, as they're all very similar, but that should probably be a separate thing on it's own. I think it's worth reconsidering the benefit of this refactor anyway, since we do intend to move to smithy code gen

generated/protocol-tests/test-clients/aws-cpp-sdk-rpcv2protocol/source/model/ComplexNestedErrorData.cpp
sbaluja: nah you're right, this should only be there for nested shapes, I'll change that

tests/testing-resources/include/aws/testing/AwsProtocolTestHelpers.h
sbaluja: Right now for protocol input tests we are comparing raw bodies, and only for CBOR are we converting to an object and comparing the properties, but I know we've discussed doing this for all protocols. I'd say this should be apart of that task, since the code snippet you supplied could also then be extended to output tests. For now, this behavior is specific to CBOR - for a short period of time since we intend to get all protocol tests green anyway, and that would mean all protocols tests would have to follow this approach

tools/code-generation/generator/src/main/java/com/amazonaws/util/awsclientgenerator/domainmodels/codegeneration/Shape.java
sbaluja: I agree, but we don't currently have that library as a dependency in our code gen project

src/aws-cpp-sdk-core/include/aws/core/utils/DateTime.h
sbaluja: This problem already exists since we have a constructor with int64_t, so ambiguous calls are already a problem with DateTimes. My changes would just co-exist with that issue

tests/testing-resources/include/aws/testing/AwsProtocolTestHelpers.h
sbaluja: Update: Updated all output protocol-tests and only cbor input protocol tests to use this approach

generated/protocol-tests/tests/output/smithy-rpc-v2-cbor/tests/RecursiveShapesTest.cpp
```diff
@@ -0,0 +1,86 @@
...
+  SetMockResponse(mockRs);
+
+  RecursiveShapesRequest request;
+
+  auto outcome = client.RecursiveShapes(request);
+  AWS_ASSERT_SUCCESS(outcome) << outcome.GetError();
+  const RecursiveShapesResult& result = outcome.GetResult();
+  ValidateRequestSent([&result](const ExpectedRequest&, const Aws::ProtocolMock::Model::Request&) -> void {
+    /* expectedResult = R"( {"nested":{"foo":"Foo1","nested":{"bar":"Bar1","recursiveMember":{"foo":"Foo2","nested":{"bar":"Bar2"}}}}} )" */
+    {
```
sbiscigl, line 32: nit: `{{` unneeded but it also sounds like it might be a codegen nightmare

generated/protocol-tests/test-clients/aws-cpp-sdk-rpcv2protocol/source/model/Float16Result.cpp
```diff
@@ -0,0 +1,161 @@
...
+using namespace Aws::Utils::Cbor;
+using namespace Aws;
+
+Float16Result::Float16Result(const Aws::AmazonWebServiceResult<Aws::Utils::Cbor::CborValue>& result) { *this = result; }
+
+Float16Result& Float16Result::operator=(const Aws::AmazonWebServiceResult<Aws::Utils::Cbor::CborValue>& result) {
+  const auto& cborValue = result.GetPayload();
+  const auto decoder = cborValue.GetDecoder();
+
+  if (decoder != nullptr) {
```
sbiscigl, line 29: So thought a lot about this factoring of it and tried to come up with a better was to do this to avoid this nesting logic, and right now I think its _OK_ given the circumstances and that its a two way door given that this will be in compiled code and we can change the implementation anytime we want.

At a high level though we should try to decompose this code into logical parts that can be re-used in code gen instead of code genning the entire implementation i.e. logically this is:
```
if (known map size) {
 deserialize_float
 deserialize_string
} else {
 deserialize_float
 deserialize_string
}
```

this could roughly become in templated code

```cpp
#pragma once
#include <aws/core/AmazonWebServiceResult.h>
#include <aws/core/utils/memory/stl/AWSString.h>
#include <aws/core/utils/cbor/CborValue.h>
#include <type_traits>


namespace Aws {
namespace Utils {
namespace Cbor {

class CborDecoderUtil final {
public:
  CborDecoderUtil() = delete;

  template <typename T, typename std::enable_if<std::is_same<T, Aws::String>::value, int>::type = 0>
  static T Decode(const AmazonWebServiceResult<Aws::Utils::Cbor::CborValue>& result) {
    auto decoder = result.GetPayload().GetDecoder();
    auto peekType = decoder->PeekType();
    if (peekType.has_value()) {
      if (peekType.value() == Aws::Crt::Cbor::CborType::Text) {
        auto val = decoder->PopNextTextVal();
        if (val.has_value()) {
          return Aws::String(reinterpret_cast<const char*>(val.value().ptr), val.value().len);
        }
      } else {
        decoder->ConsumeNextSingleElement();
        Aws::StringStream ss;
        while (decoder->LastError() == AWS_ERROR_UNKNOWN) {
          auto nextType = decoder->PeekType();
          if (!nextType.has_value() || nextType.value() == CborType::Break) {
            if (nextType.has_value()) {
              decoder->ConsumeNextSingleElement();  // consume the Break
            }
            break;
          }
          auto val = decoder->PopNextTextVal();
          if (val.has_value()) {
            ss << Aws::String(reinterpret_cast<const char*>(val.value().ptr), val.value().len);
          }
        }
        return ss.str();
      }
    }
    // likely log an error, we're in a weird state.
    return T{};
  }

  template <typename T, typename std::enable_if<std::is_same<T, double>::value, int>::type = 0>
  static T Decode(const AmazonWebServiceResult<Aws::Utils::Cbor::CborValue>& result) {
    auto decoder = result.GetPayload().GetDecoder();
    auto val = decoder->PopNextFloatVal();
    if (val.has_value()) {
      return val.value();
    }
    // likely log an error, we're in a weird state.
    return T{};
  }

  // TODO: do this for every type
};
}
}
}
```

where the trouble starts then is
> TODO: do this for every type

specifically where this gets into trouble is `Map` -> `Object` types. This is because C++ does not have reflection. for primative types this is tivial, but with complex types we need to actually map between a cbor "map" which is actually closer to a object and a c++ "object".

The way other C++/C parsers accomplish this is by using a intermediate object i.e.

```c++
#include <unordered_map>
#include <vector>

class cbor_float {};

class cbor_string {};

//other types

class cbor_value {
public:
  cbor_value(char* bytes) {
    //...create underlying type 
  }
private:
  std::unordered_map<
    std::string,
    std::variant<cbor_float, cbor_string, cbor_value>> data;
};

class cbor_constructible {
public:
  cbor_constructible(cbor_value value) {
    //...construct from free from objec
  }
};
```

the issue there though is then you have to construct a intermediate object, which is also not ideal.

so given the open choices:

* use reflection
pros:
It is the correct choice
cons:
[doesnt exist yet](https://isocpp.org/files/papers/P2996R4.html)

* create a temporary object
pros:
cleaner code
cons:
temporary object

* use code gen to create cbor parsing
pros:
no temp object
no implementaion specific doorway
cons:
messy code

i think we can move forward with codegen with the idea to revisit it if we want to use temporary objects or a better solutions presents itself

sbiscigl (approved):

PR #3545, opened 2025-09-16
[C++] SDKs to track feature Ids for Credentials (PROFILE_SOURCE_PROFILE/PROFILE_NAMED_PROVIDER)

kai-ion:
*Issue #, if available:*

*Description of changes:*
Added user agent feature tracking for STS profile-based credential provider

*Check all that applies:*

- [x] Linux
- [x] Windows
- [x] Android
- [x] MacOS
- [ ] IOS
- [ ] Other Platforms


src/aws-cpp-sdk-identity-management/source/auth/STSProfileCredentialsProvider.cpp
```diff
@@ -210,6 +210,12 @@ void STSProfileCredentialsProvider::Reload()
         profileIt->second.SetCredentials({});
     }
 
+    // Check for credential_source usage
```
sbiscigl: [credentials_source is not supported in the SDK so why are we tracking it](https://docs.aws.amazon.com/sdkref/latest/guide/feature-assume-role-credentials.html)?

src/aws-cpp-sdk-identity-management/source/auth/STSProfileCredentialsProvider.cpp
```diff
@@ -306,6 +312,11 @@ void STSProfileCredentialsProvider::Reload()
     }
 
     m_credentials = profileIt->second.GetCredentials();
+    
+    if (!m_credentials.IsEmpty()) {
```
sbiscigl: instead of updating `m_credentials` after the fact, why not update the credentials before we set the member variable when we knows its resolved from the source profile? i.e.

```cpp
const auto arn = sourceProfiles.back()->second.GetRoleArn();
const auto& assumedCreds = GetCredentialsFromSTS(stsCreds, arn);

if (!assumedCreds.IsEmpty()) {
  assumedCreds.AddUserAgentFeature(Aws::Client::UserAgentFeature::CREDENTIALS_PROFILE_SOURCE_PROFILE);
}

sourceProfiles.back()->second.SetCredentials(assumedCreds);
```

src/aws-cpp-sdk-identity-management/source/auth/STSProfileCredentialsProvider.cpp
kai-ion: Removed tracking because credentials_source is not supported for CPP

src/aws-cpp-sdk-identity-management/source/auth/STSProfileCredentialsProvider.cpp
kai-ion: refactored to track after creds are resolved from source profile

PR #3556, opened 2025-09-22
Install pdb files for windows builds 

sbaluja:
*Issue #, if available:*

*Description of changes:*
Install .pdb files for windows builds

*Check all that applies:*

- [x] Linux
- [x] Windows
- [x] Android
- [x] MacOS
- [ ] IOS
- [ ] Other Platforms


generated/src/aws-cpp-sdk-AWSMigrationHub/CMakeLists.txt
```diff
@@ -70,7 +70,9 @@ setup_install()
 
 install (FILES ${AWS_AWSMIGRATIONHUB_HEADERS} DESTINATION ${INCLUDE_DIRECTORY}/aws/AWSMigrationHub)
 install (FILES ${AWS_AWSMIGRATIONHUB_MODEL_HEADERS} DESTINATION ${INCLUDE_DIRECTORY}/aws/AWSMigrationHub/model)
-
+if (MSVC)
+    install (FILES $<TARGET_PDB_FILE:${PROJECT_NAME}> DESTINATION ${CMAKE_INSTALL_BINDIR} OPTIONAL)
+endif()
 do_packaging()
```
sbiscigl, line 76: so instead of each different cmake file doing the PDB operation why not move it to the [`do_packaging()` macro](https://github.com/aws/aws-sdk-cpp/blob/main/cmake/utilities.cmake#L110-L150)? so that were are only editing one place and anything using `do_packaging` gets it for free.

sbaluja: I'll make a separate PR, that's a better idea

Issue #3558, opened 2025-09-23
Regression: STSAssumeRoleWebIdentityCredentialsProvider not caching credentials: STS congestion and crashes


jukka-aro-navvis:
## Bug Description

The `STSAssumeRoleWebIdentityCredentialsProvider` in AWS SDK C++ bypasses the credential caching layer, causing severe performance issues and connection crashes in multi-threaded applications using Kubernetes IAM Roles for Service Accounts (IRSA).

**This is a regression** - credential caching was working correctly in AWS SDK C++ version 1.11.621 and was broken by commit https://github.com/aws/aws-sdk-cpp/commit/c6ed3e3c6c965a088ab7fb09e429e045134ffaa6

### Observed Symptoms

1. **STS Congestion**: Occasional very high latency through STS congestion due to a large number of concurrent calls from different threads when calling GetObject for S3 data access
2. **Connection Crashes**: Recurring crashes in aws-c-http connection handling at https://github.com/awslabs/aws-c-http/blob/main/source/connection.c#L429, suspected to be a double release bug in reference counting

### Environment

- **AWS SDK C++ Version**: 1.11.653 (regression introduced in 1.11.622+)
- **Last Working Version**: 1.11.621
- **Operating System**: Ubuntu 22.04
- **Authentication Method**: AssumeRoleWithWebIdentity (Kubernetes IRSA)
- **Cloud Setup**: IAM role-based authentication with Kubernetes service accounts annotated with AWS IAM roles
- **Application**: Multi-threaded application using `Aws::S3Crt::S3CrtClient` making concurrent AWS S3 operations

**Important**: This issue is particularly relevant when using the **S3 CRT Client** (`Aws::S3Crt::S3CrtClient`) as it may have different credential provider behavior compared to the standard S3 client.

### Primary Issue: Missing Credential Caching

**File**: `src/aws-cpp-sdk-core/source/auth/STSCredentialsProvider.cpp` (line 39)

```cpp
// Current problematic implementation - creates uncached provider
m_credentialsProvider = Aws::Crt::Auth::CredentialsProvider::CreateCredentialsProviderSTSWebIdentity(stsConfig);
```

[...]

PR #3560, opened 2025-09-23
fix cache and threading concerns with STS Webidentity provider

sbiscigl:
*Issue #, if available:*

https://github.com/aws/aws-sdk-cpp/issues/3558

*Description of changes:*

There was a fundamental misunderstanding when we swapped to [use the CRT implementation of the STSWebidentityProvider](https://github.com/aws/aws-sdk-cpp/commit/c6ed3e3c6c965a088ab7fb09e429e045134ffaa6) and that was that CRT would cache the credentials from STS using their expiration. This is incorrect, it is up to us to cache and expire the credentials.

This change moves the implementation back to using a reader/writer lock to cache the credentials between invocations, making a network call only when expiration is close.

*Check all that applies:*

- [x] Linux
- [x] Windows
- [x] Android
- [x] MacOS
- [ ] IOS
- [ ] Other Platforms

### DIFF
--- a/src/aws-cpp-sdk-core/source/auth/STSCredentialsProvider.cpp
+++ b/src/aws-cpp-sdk-core/source/auth/STSCredentialsProvider.cpp
@@ -37,11 +37,26 @@ STSAssumeRoleWebIdentityCredentialsProvider::STSAssumeRoleWebIdentityCredentials
     }
     return UUID::RandomUUID();
   }().c_str();
-  m_credentialsProvider = Aws::Crt::Auth::CredentialsProvider::CreateCredentialsProviderSTSWebIdentity(stsConfig);
+
+  // Create underlying STS provider
+  auto stsProvider = Aws::Crt::Auth::CredentialsProvider::CreateCredentialsProviderSTSWebIdentity(stsConfig);
+  if (!stsProvider || !stsProvider->IsValid()) {
+    AWS_LOGSTREAM_WARN(STS_LOG_TAG, "Failed to create underlying STS credentials provider");
+    return;
+  }
+
+  // Wrap with caching provider
+  Aws::Crt::Auth::CredentialsProviderCachedConfig cachedConfig;
+  cachedConfig.Provider = stsProvider;
+  cachedConfig.CachedCredentialTTL = credentialsConfig.stsCredentialsProviderConfig.credentialCacheCacheTTL;
+
+  m_credentialsProvider = Aws::Crt::Auth::CredentialsProvider::CreateCredentialsProviderCached(cachedConfig);
   if (m_credentialsProvider && m_credentialsProvider->IsValid()) {
     m_state = STATE::INITIALIZED;
+    AWS_LOGSTREAM_INFO(STS_LOG_TAG,
+                       "STS credentials provider initialized with cache TTL " << cachedConfig.CachedCredentialTTL.count() << " ms");
   } else {
-    AWS_LOGSTREAM_WARN(STS_LOG_TAG, "Failed to create STS credentials provider");
+    AWS_LOGSTREAM_WARN(STS_LOG_TAG, "Failed to create cached STS credentials provider");
   }
 }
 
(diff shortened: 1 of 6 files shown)


src/aws-cpp-sdk-core/source/auth/STSCredentialsProvider.cpp
```diff
@@ -37,11 +37,26 @@ STSAssumeRoleWebIdentityCredentialsProvider::STSAssumeRoleWebIdentityCredentials
...
+  auto stsProvider = Aws::Crt::Auth::CredentialsProvider::CreateCredentialsProviderSTSWebIdentity(stsConfig);
+  if (!stsProvider || !stsProvider->IsValid()) {
+    AWS_LOGSTREAM_WARN(STS_LOG_TAG, "Failed to create underlying STS credentials provider");
+    return;
+  }
+
+  // Wrap with caching provider
+  Aws::Crt::Auth::CredentialsProviderCachedConfig cachedConfig;
+  cachedConfig.Provider = stsProvider;
+  cachedConfig.CachedCredentialTTL = credentialsConfig.stsCredentialsProviderConfig.credentialCacheCacheTTL;
```
jukka-aro-navvis, line 51: @sbiscigl Sorry, I failed to think this properly through yesterday, but I think there is a flaw in providing the TTL value here.
- If I read this correctly, we always take the default (50 min) for credentialCacheCacheTTL now.
- However, AssumeRoleWithWebIdentity supports configurable DurationSeconds, see https://docs.aws.amazon.com/cli/latest/reference/sts/assume-role-with-web-identity.html.
 
What happens when DurationSeconds != 3600 (60min) while we use the default 50min here in the provider:
  - Short credentials (15min): Serves expired credentials → Authentication failures
  - Long credentials (120min): Unnecessary refreshes → Performance loss

Here's a quickly sketched proposal how to address the possibly varying credentials duration (not tested code -> no warranty):

```
  std::chrono::milliseconds calculateDynamicTTL(const AWSCredentials& credentials) {
    // 1. Get ACTUAL expiry from credential (whatever AWS issued)
    auto expiryTime = credentials.GetExpiration();
    auto currentTime = DateTime::Now();
    auto remainingTime = expiryTime.Millis() - currentTime.Millis();

    // 2. Calculate TTL with grace period (adapts to ANY credential lifetime)
    auto ttl = remainingTime - std::chrono::minutes(5); // 5-min grace period

    // 3. Ensure reasonable bounds
    return std::max(std::chrono::minutes(1),
                    std::min(ttl, remainingTime * 0.95));
  }
```
Possibly make the grace period configurable in the provider if needed?

src/aws-cpp-sdk-core/source/auth/STSCredentialsProvider.cpp
```diff
@@ -116,3 +115,65 @@ AWSCredentials STSAssumeRoleWebIdentityCredentialsProvider::GetAWSCredentials()
...
+  const auto sessionTokenCursor = crtCredentials.GetSessionToken();
+  credentials.SetSessionToken({reinterpret_cast<char*>(sessionTokenCursor.ptr), sessionTokenCursor.len});
+  return credentials;
+}
+
+AWSCredentials STSAssumeRoleWebIdentityCredentialsProvider::fetchCredentialsAsync() {
+  AWS_LOGSTREAM_DEBUG(STS_LOG_TAG, "Initiating credential fetch from STS/cache");
+
+  AWSCredentials credentials{};
+  std::atomic<bool> refreshDone{false};
```
jukka-aro-navvis, line 150: Sorry, did not think this through either yesterday.
- `std::atomic<bool>` is not working correctly in the general case since each thread will allocate its own flag on its own stack.
- in our case, however, this flag is redundant since we are already ensuring that only one thread ever enters fetchCredentialsAsync at any time - see usage of m_refreshInProgress in STSAssumeRoleWebIdentityCredentialsProvider::GetAWSCredentials().

Suggest to either remove this flag (if we agree to rely on synchronization only in the outer function) or implement it properly if we want to go 100% in each individual function.

jukka-aro-navvis: E. g. something like this should work, if we want to keep each individual function thread safe:
```
  // In STSCredentialsProvider.h - add member variable:
  private:
    mutable std::atomic<bool> m_refreshDone{false};

  // In STSCredentialsProvider.cpp - use member variable:
  AWSCredentials fetchCredentialsAsync() {
    m_refreshDone.store(false);  // ← Reset member variable

    m_credentialsProvider->GetCredentials([this, &credentials](auto crtCredentials, int errorCode) {
      // ... process credentials ...
      m_refreshDone.store(true);     // ← Set member variable
      m_refreshInProgress.store(false);
      m_refreshSignal.notify_all();
    });

    std::unique_lock<std::mutex> lock{m_refreshMutex};
    m_refreshSignal.wait_for(lock, timeout, [this]() { return m_refreshDone.load(); });

    return credentials;
  }
``

File: src/aws-cpp-sdk-core/source/auth/STSCredentialsProvider.cpp (hunk 1 above)
Comment (sbiscigl) [reply src/aws-cpp-sdk-core/source/auth/STSCredentialsProvider.cpp:51]: I actually thought of this too, and almost changed the entire implementation around it. but this actually incorrect and your original impl was actually correct.

in the cached [credentials provider we always use `expiration` if it is returned in the credentials](https://github.com/awslabs/aws-c-auth/blob/main/source/credentials_provider_cached.c#L99-L127)
```c
uint64_t credentials_expiration_timepoint_seconds = UINT64_MAX;
if (credentials != NULL) {
    credentials_expiration_timepoint_seconds = aws_credentials_get_expiration_timepoint_seconds(credentials);
}

/*
 * If the sourced credentials have an explicit expiration time, we should always use that time
 * rather than the much cruder, mechanical refresh setting on the caching wrapper.
 */
if (credentials_expiration_timepoint_seconds < UINT64_MAX) {
    uint64_t system_now = 0;
    if (!impl->system_clock_fn(&system_now)) {

        uint64_t system_now_seconds =
            aws_timestamp_convert(system_now, AWS_TIMESTAMP_NANOS, AWS_TIMESTAMP_SECS, NULL);
        if (credentials_expiration_timepoint_seconds >=
            system_now_seconds + REFRESH_CREDENTIALS_EARLY_DURATION_SECONDS) {
            next_refresh_time_in_ns = high_res_now;
            next_refresh_time_in_ns += aws_timestamp_convert(
                credentials_expiration_timepoint_seconds - system_now_seconds -
                    REFRESH_CREDENTIALS_EARLY_DURATION_SECONDS,
                AWS_TIMESTAMP_SECS,
                AWS_TIMESTAMP_NANOS,
                NULL);
        } else {
            next_refresh_time_in_ns = high_res_now;
        }
    }
}
```

This will use the adjustible `Expiration` field returned from STS, so think of this as. cached provider that defeaults to 50 minutes _unless_ we get the expiration back, in which case it is used.

src/aws-cpp-sdk-core/source/auth/STSCredentialsProvider.cpp
sbiscigl: will opt to move `m_refreshDone` to the class level to keep each indiviual function thread safe

src/aws-cpp-sdk-core/source/auth/STSCredentialsProvider.cpp
jukka-aro-navvis: Thanks for the explanation - it looks like this is indeed handled on the cached provider level.

PR #3562, opened 2025-09-24
track refresh done at class level

sbiscigl:
*Issue #, if available:*

https://github.com/aws/aws-sdk-cpp/issues/3558

*Description of changes:*

As brought up in the issue

> *  std::atomic<bool> is not working correctly in the general case since each thread will allocate its own flag on its own stack.
> * in our case, however, this flag is redundant since we are already ensuring that only one thread ever enters 

> fetchCredentialsAsync at any time - see usage of m_refreshInProgress in STSAssumeRoleWebIdentityCredentialsProvider::GetAWSCredentials().
Suggest to either remove this flag (if we agree to rely on synchronization only in the outer function) or implement it properly if we want to go 100% in each individual function.

We can either remove the redundant flag, or implement it so that the function is thread safe. Currently since the caller is calling it thread safe, there is not a compounded issues, however we should make the function threadsafe.

*Check all that applies:*

- [x] Linux
- [x] Windows
- [x] Android
- [x] MacOS
- [x] IOS
- [ ] Other Platforms


pulimsr (approved):

jukka-aro-navvis:
I did quick testing at our end (K8s, Ubuntu 22.04 LTS) and the fix seems to work correctly.

kai-ion (approved):

PR #3586, opened 2025-10-15
Update FSM config parser

kai-ion:
Add support for parsing [services] sections with service-specific endpoint_url configuration following AWS SDKs & Tools specification.

*Issue #, if available:*

*Description of changes:*
- Extend Profile class with endpoint_url and servicesEndpointUrl map
- Add FSM states for [services] and [services serviceId] sections
- Add GetServiceEndpointUrl() and GetGlobalEndpointUrl() API methods
- Support case-insensitive service ID lookup
- Add comprehensive test coverage

Supports config format:
[profile default]
endpoint_url = https://global.example.com

[services s3]
endpoint_url = http://localhost:9000

*Check all that applies:*

- [x] Linux
- [x] Windows
- [x] Android
- [x] MacOS
- [ ] IOS
- [ ] Other Platforms

### DIFF
--- a/src/aws-cpp-sdk-core/include/aws/core/config/AWSProfileConfig.h
+++ b/src/aws-cpp-sdk-core/include/aws/core/config/AWSProfileConfig.h
@@ -8,6 +8,7 @@
 #include <aws/core/utils/memory/stl/AWSString.h>
 #include <aws/core/utils/memory/stl/AWSMap.h>
 #include <aws/core/auth/AWSCredentials.h>
+#include <aws/crt/Optional.h>
 
 namespace Aws
 {
@@ -19,6 +20,24 @@ namespace Aws
         class Profile
         {
         public:
+            /*
+             * Data container for service endpoints.
+             */
+            class Services
+            {
+            public:
+                Services() = default;
+                Services(Aws::Map<Aws::String, Aws::String>&& endpoints, Aws::String name)
+                    : m_endpoints(std::move(endpoints)), m_name(std::move(name)) {}
+                
+                inline Aws::Map<Aws::String, Aws::String> GetEndpoints() const { return m_endpoints; }
+                inline Aws::String GetServiceBlockName() const { return m_name; }
+                inline bool IsSet() const { return !m_name.empty(); }
+            private:
+                Aws::Map<Aws::String, Aws::String> m_endpoints;
+                Aws::String m_name;
+            };
+
             /*
              * Data container for a sso-session config entry.
              * This is independent of the general profile configuration and used by a bearer auth token provider.
@@ -84,6 +103,10 @@ namespace Aws
             inline void SetSourceProfile(const Aws::String& value ) { m_sourceProfile = value; }
             inline const Aws::String& GetCredentialProcess() const { return m_credentialProcess; }
             inline void SetCredentialProcess(const Aws::String& value ) { m_credentialProcess = value; }
+            inline const Aws::String& GetGlobalEndpointUrl() const { return m_endpointUrl; }
+            inline void SetGlobalEndpointUrl(const Aws::String& value) { m_endpointUrl = value; }
+            inline Services GetServices() const { return m_services; }
+            inline void SetServices(Services&& services) { m_services = std::move(services); }
             inline void SetAllKeyValPairs(const Aws::Map<Aws::String, Aws::String>& map) { m_allKeyValPairs = map; }
             inline void SetAllKeyValPairs(Aws::Map<Aws::String, Aws::String>&& map) { m_allKeyValPairs = std::move(map); }
             inline const Aws::String GetValue(const Aws::String& key) const
@@ -111,7 +134,9 @@ namespace Aws
             Aws::String m_ssoAccountId;
             Aws::String m_ssoRoleName;
             Aws::String m_defaultsMode;
+            Aws::String m_endpointUrl;
             Aws::Map<Aws::String, Aws::String> m_allKeyValPairs;
+            Services m_services;
 
             bool m_ssoSessionSet = false;
             SsoSession m_ssoSession;
--- a/src/aws-cpp-sdk-core/source/config/AWSConfigFileProfileConfigLoader.cpp
+++ b/src/aws-cpp-sdk-core/source/config/AWSConfigFileProfileConfigLoader.cpp
@@ -40,6 +40,9 @@ namespace Aws
         static const char PROFILE_SECTION[]                  = "profile";
         static const char DEFAULT[]                          = "default";
         static const char SSO_SESSION_SECTION[]              = "sso-session";
+        static const char SERVICES_SECTION[]                 = "services";
+        static const char ENDPOINT_URL_KEY[]                 = "endpoint_url";
+        static const char IGNORE_CONFIGURED_ENDPOINT_URLS_KEY[] = "ignore_configured_endpoint_urls";
         static const char DEFAULTS_MODE_KEY[]                = "defaults_mode";
         static const char EQ                                 = '=';
         static const char LEFT_BRACKET                       = '[';
@@ -74,7 +77,8 @@ namespace Aws
                  {EXTERNAL_ID_KEY, &Profile::SetExternalId, &Profile::GetExternalId},
                  {CREDENTIAL_PROCESS_COMMAND, &Profile::SetCredentialProcess, &Profile::GetCredentialProcess},
                  {SOURCE_PROFILE_KEY, &Profile::SetSourceProfile, &Profile::GetSourceProfile},
-                 {DEFAULTS_MODE_KEY, &Profile::SetDefaultsMode, &Profile::GetDefaultsMode}};
+                 {DEFAULTS_MODE_KEY, &Profile::SetDefaultsMode, &Profile::GetDefaultsMode},
+                 {ENDPOINT_URL_KEY, &Profile::SetGlobalEndpointUrl, &Profile::GetGlobalEndpointUrl}};
 
         template<typename EntryT, size_t N>
         const EntryT* FindInStaticArray(const EntryT (&array)[N], const Aws::String& searchKey)
@@ -119,6 +123,7 @@ namespace Aws
                 static const size_t ASSUME_EMPTY_LEN = 3;
                 State currentState = START;
                 Aws::String currentSectionName;
+                Aws::String activeServiceId;
                 Aws::Map<Aws::String, Aws::String> currentKeyValues;
 
                 Aws::String rawLine;
@@ -142,6 +147,7 @@ namespace Aws
                     {
                         FlushSection(currentState, currentSectionName, currentKeyValues);
                         currentKeyValues.clear();
+                        activeServiceId.clear();
                         ParseSectionDeclaration(line, currentSectionName, currentState);
                         continue;
                     }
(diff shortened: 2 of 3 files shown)


src/aws-cpp-sdk-core/include/aws/core/config/AWSConfigFileProfileConfigLoader.h
```diff
@@ -39,13 +40,24 @@ namespace Aws
              */
             void SetFileName(const Aws::String& fileName) { m_fileName = fileName; }
 
+            /**
+             * Get service-specific endpoint URL for a given profile and service.
+             */
+            const Aws::String* GetServiceEndpointUrl(const Aws::String& profileName, const Aws::String& serviceId) const;
```
sbiscigl: why are we returning a pointer to a string? we should likely be returning by value and rely on RVO or NVRO, or returning bt ref if we actually want to edit the owned string, but we dont want to be returning by pointer.

src/aws-cpp-sdk-core/include/aws/core/config/AWSProfileConfig.h
```diff
@@ -93,6 +93,18 @@ namespace Aws
                 return iter->second;
             }
 
+            inline const Aws::String& GetServicesName() const {
```
sbiscigl: why return are we returing by ref? we definitely should not be editing the underlying value.

src/aws-cpp-sdk-core/include/aws/core/config/AWSProfileConfig.h
```diff
@@ -93,6 +93,18 @@ namespace Aws
...
+                auto iter = m_allKeyValPairs.find("services");
+                static const Aws::String empty;
+                return (iter != m_allKeyValPairs.end()) ? iter->second : empty;
```
sbiscigl: This likely not how we should be returning the idea of "lack of service endpoint". likely [Crt::Optional](https://github.com/awslabs/aws-crt-cpp/blob/main/include/aws/crt/Optional.h) is a good candidate for representing "lacking of value" instead of empty string.

src/aws-cpp-sdk-core/source/config/AWSConfigFileProfileConfigLoader.cpp
```diff
@@ -158,6 +164,36 @@ namespace Aws
...
+                        if (equalsPos == std::string::npos) {
+                            continue; // ignore garbage/blank in services section
+                        }
+
+                        auto left = StringUtils::Trim(line.substr(0, equalsPos).c_str());
+                        auto right = StringUtils::Trim(line.substr(equalsPos + 1).c_str());
+
+                        // New service block: "s3 =" (right hand side empty)
+                        if (!left.empty() && right.empty()) {
+                            activeServiceId = StringUtils::ToLower(left.c_str());
```
sbiscigl: from the spec

> This transformation MUST replace any spaces in the service ID with underscores and uppercase all letters

so `ToLower` here I think should be `ToUpper` no?

src/aws-cpp-sdk-core/source/config/AWSConfigFileProfileConfigLoader.cpp
```diff
@@ -665,5 +737,40 @@ namespace Aws
 
             return false;
         }
+
+        const Aws::String* AWSConfigFileProfileConfigLoader::GetServiceEndpointUrl(const Aws::String& profileName, const Aws::String& serviceId) const
+        {
+            auto profileIter = m_profiles.find(profileName);
+            if (profileIter == m_profiles.end()) 
+            {
+                return nullptr;
```
sbiscigl: yeah dont use "nullptr" to denote absence of a value, or a failed outcome, use a optional, or result type for that

src/aws-cpp-sdk-core/include/aws/core/config/AWSConfigFileProfileConfigLoader.h
```diff
@@ -39,13 +40,24 @@ namespace Aws
...
+
+            /**
+             * Get global endpoint URL for a given profile.
+             */
+            const Aws::String* GetGlobalEndpointUrl(const Aws::String& profileName) const;
```
sbiscigl: Alright i think we have a slightly larger design problem here that we should deal with lets take a look at the two new added APIS

```c++
Aws::String GetServiceEndpointUrl(const Aws::String& profileName, 
   const Aws::String& serviceId) const;
Aws::String* GetGlobalEndpointUrl(const Aws::String& profileName) const;
```

both of these functions have something in common, they take a `profileName`, and even in the comments you mention there is a mapping between profile and endpoint urls. This says to me instead of maintaining a foreign key mapping via this API, this is something that should exist in the profile object, i.e. [`Profile `](https://github.com/aws/aws-sdk-cpp/blob/main/src/aws-cpp-sdk-core/include/aws/core/config/AWSProfileConfig.h#L19C15-L19C22). so instead of adding this api, lets add this data to the `Profile` object that we store.

src/aws-cpp-sdk-core/include/aws/core/config/AWSConfigFileProfileConfigLoader.h
kai-ion: Updated by using `Aws::Crt::Optional<Aws::String>` instead of pointer returns.

src/aws-cpp-sdk-core/include/aws/core/config/AWSProfileConfig.h
kai-ion: Same as above. Updated to use `Aws::Crt::Optional<Aws::String>`

src/aws-cpp-sdk-core/include/aws/core/config/AWSProfileConfig.h
kai-ion: Same as above. Updated to use `Aws::Crt::Optional<Aws::String>`

src/aws-cpp-sdk-core/source/config/AWSConfigFileProfileConfigLoader.cpp
kai-ion: Fixed normalization to use ToUpper

src/aws-cpp-sdk-core/source/config/AWSConfigFileProfileConfigLoader.cpp
kai-ion: Same as above. Updated to use `Aws::Crt::Optional<Aws::String>`

src/aws-cpp-sdk-core/include/aws/core/config/AWSConfigFileProfileConfigLoader.h
kai-ion: Moved GetEndpointUrl() and GetServicesName() as computed getters in the profile object, also added a static helper that does the lookup for the service.

kai-ion: NVM on the static helper, removed it!
I was only using it as a temporary 'resolver' for the new unit tests anyways

src/aws-cpp-sdk-core/include/aws/core/config/AWSConfigFileProfileConfigLoader.h
```diff
@@ -46,6 +47,7 @@ namespace Aws
         private:
             Aws::String m_fileName;
             bool m_useProfilePrefix;
+            Aws::Map<Aws::String, Aws::Map<Aws::String, Aws::String>> m_services;
```
sbiscigl: i dont think this is being used anymore

kai-ion: removed this map in new commit

src/aws-cpp-sdk-core/include/aws/core/config/AWSProfileConfig.h
```diff
@@ -93,6 +94,16 @@ namespace Aws
                 return iter->second;
             }
 
+            inline Aws::Crt::Optional<Aws::String> GetServicesName() const {
```
sbiscigl: so i still think this is the wrong API, the profile object _will have_ a services section, once loaded, we wont be doing any further look ups or joining of the services section and the profile sections, from the outside a API caller will see a profile object, just not it will have a services section/global endpoint member

kai-ion: Updated to have the profile to include a service object that will contain the map of endpoints and service block name

src/aws-cpp-sdk-core/source/config/AWSConfigFileProfileConfigLoader.cpp
```diff
@@ -158,6 +163,36 @@ namespace Aws
                         }
```
sbaluja, line 164: Are we not handling global endpoint URL's for all services? 
Ex. 
```
[profile dev-global]
endpoint_url = https://play.min.io:9000
```

tests/aws-cpp-sdk-core-tests/aws/config/ServiceEndpointsConfigFileLoaderTest.cpp
```diff
@@ -0,0 +1,307 @@
...
+    ASSERT_NE(profiles.end(), profileIt);
+    const auto& profile = profileIt->second;
+
+    // Test services endpoints are parsed correctly
+    const auto& services = profile.GetServices();
+    ASSERT_TRUE(services.IsSet());
+    const auto& endpoints = services.GetEndpoints();
+    ASSERT_EQ(1u, endpoints.size());
+    ASSERT_EQ("http://foo.com", endpoints.at("S3"));
+}
```
sbaluja, line 398: Curious what the behavior would be if we have multiple `endpoint_url` key/values in a profile or service configuration. I think we should add tests for scenarios like
```
[profile dev-global]
endpoint_url = https://play.min.io:9000
endpoint_url = https://play2.min.io:9000
```
```
[services s3test]
s3
endpoint_url = https://play.min.io:9000
s3
endpoint_url = https://play2.min.io:9000
```

From my understanding of the implementation, I think we'd just override the the value, so we'd end up with `endpoint_url = https://play2.min.io:9000`, but it's probably worth discussing if we want to error out here. 

Either way this behavior is likely specific to OUR implementation since the SEP doesn't explicitly define this behavior, and this will be important to catch any regressions if we move to the CRT parser later on.

src/aws-cpp-sdk-core/source/config/AWSConfigFileProfileConfigLoader.cpp
kai-ion: global endpoint URL is handled by the [FlushSection](https://github.com/aws/aws-sdk-cpp/blob/c01a1cbb216c5a8d05ee03f125f279a0c50908bc/src/aws-cpp-sdk-core/source/config/AWSConfigFileProfileConfigLoader.cpp#L428), and added directly to the profile

tests/aws-cpp-sdk-core-tests/aws/config/ServiceEndpointsConfigFileLoaderTest.cpp
kai-ion: In the current parser it will just overwrite the url with the latest endpoint_url, I will add two tests for this behavior

src/aws-cpp-sdk-core/include/aws/core/config/AWSProfileConfig.h
```diff
@@ -19,6 +20,24 @@ namespace Aws
         class Profile
         {
         public:
+            /*
+             * Data container for service endpoints.
+             */
+            class Services
+            {
+            public:
+                Services() = default;
```
sbiscigl, line 29: why do we need a default constructor? seems like we should only have the constructor with endpoints and name

src/aws-cpp-sdk-core/include/aws/core/config/AWSProfileConfig.h
```diff
@@ -93,6 +112,14 @@ namespace Aws
                 return iter->second;
             }
 
+            inline const Services& GetServices() const { return m_services; }
+            inline void SetServices(Services&& services) { m_services = std::move(services); }
+
+            inline Aws::Crt::Optional<Aws::String> GetGlobalEndpointUrl() const {
+                const Aws::String& endpoint = GetValue("endpoint_url");
```
sbiscigl: this is the only `Get` function that uses `GetValue`, why not create a memeber of profile `endpointUrl` to be like all the other data members of `Profile`?

sbiscigl (approved):
Mentioned some minor fixes before your ship, but API wise looks good

PR #3591, opened 2025-10-22
Integration tests

pulimsr:
*Description of changes:* Adding service ID filtering for integration tests

*Check all that applies:*

- [x] Linux
- [x] Windows
- [ ] Android
- [x] MacOS
- [ ] IOS
- [ ] Other Platforms


tools/scripts/build-tests/run-al2-integ-tests.sh
```diff
@@ -21,6 +21,8 @@ AWS_ROLE_SESSION_NAME="$3"
 
 echo "Setting the run environment"
 export TEST_ASSUME_ROLE_ARN=arn:aws:iam::${AWS_ACCOUNT}:role/IntegrationTest
+cd ${PREFIX_DIR}/aws-sdk-cpp
```
sbiscigl: nit: for this type of operation we really should be using `pushd` and `popd`, i.e.

```
pushd ${PREFIX_DIR}/aws-sdk-cpp
# ...do whatever you want to do there
popd
```

this was the directory remains the same before and after the operatoin you ran. right now you cd into a dir and dont return, that may not have any complications now, but could in the future.

tools/scripts/build-tests/run-al2-integ-tests.sh
```diff
@@ -21,6 +21,8 @@ AWS_ROLE_SESSION_NAME="$3"
...
+SERVICE_ID=$(git status generated/src/aws-cpp-sdk-* --porcelain | grep "generated/src/" | sed -n 's|.*generated/src/aws-cpp-sdk-\([^/]*\).*|\1|p' | sort -u | tr "\n" "," | sed "s/,$//")
```
sbiscigl: on our daily releases i think we want to run _all_ tests not just the ones in `SERVICE_ID`. so while this does do what we want to it to i think we might want to add another argument called `RUN_ALL` or something to script to override running all of the tests.

also when this is run in the step does `git status generated/src/aws-cpp-sdk-* --porcelain` actually produce a output? this is because at this point things are actually committed when this run

sbiscigl (approved):

PR #3597, opened 2025-10-27
endpoint resolver/provider

kai-ion:
*Issue #, if available:*

*Description of changes:*
Adding an endpoint resolver template to be called in client init to resolve endpoint in order chain

*Check all that applies:*

- [x] Linux
- [x] Windows
- [x] Android
- [x] MacOS
- [ ] IOS
- [ ] Other Platforms

### DIFF
--- a/src/aws-cpp-sdk-core/include/aws/core/config/EndpointResolver.h
+++ b/src/aws-cpp-sdk-core/include/aws/core/config/EndpointResolver.h
@@ -1,0 +1,45 @@
+/**
+ * Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.
+ * SPDX-License-Identifier: Apache-2.0.
+ */
+
+#pragma once
+
+#include <aws/core/Core_EXPORTS.h>
+#include <aws/core/utils/memory/stl/AWSString.h>
+#include <aws/core/config/AWSProfileConfig.h>
+#include <aws/core/config/ConfigAndCredentialsCacheManager.h>
+#include <aws/core/platform/Environment.h>
+#include <aws/core/utils/StringUtils.h>
+#include <aws/core/utils/logging/LogMacros.h>
+#include <aws/core/http/Scheme.h>
+
+namespace Aws
+{
+    namespace Config
+    {
+        static const char ENDPOINT_RESOLVER_TAG[] = "EndpointResolver";
+
+        /**
+         * Resolver that sources endpoints from configuration.
+         */
+        namespace EndpointResolver
+        {
+            /**
+             * Convert service ID to environment variable suffix format
+             */
+            AWS_CORE_API Aws::String ToEnvSuffix(const Aws::String& serviceId);
+
+            /**
+             * Sources endpoint URL from configuration.
+             * 
+             * @param serviceId Service identifier (e.g., "s3", "Elastic Beanstalk")
+             * @param profileName Profile name for shared config lookup
+             * @return Optional endpoint URL if found, empty if not configured
+             */
+            AWS_CORE_API Aws::String EndpointSource(const Aws::String& serviceId,
+                                                   const Aws::String& profileName);
+        }
+
+    } // namespace Config
+} // namespace Aws
--- a/src/aws-cpp-sdk-core/include/aws/core/endpoint/BuiltInParameters.h
+++ b/src/aws-cpp-sdk-core/include/aws/core/endpoint/BuiltInParameters.h
@@ -26,6 +26,8 @@ namespace Aws
 
             virtual void SetFromClientConfiguration(const Client::ClientConfiguration& config);
             virtual void SetFromClientConfiguration(const Client::GenericClientConfiguration& config);
+            virtual void SetFromClientConfiguration(const Client::ClientConfiguration& config, const Aws::String& serviceName);
+            virtual void SetFromClientConfiguration(const Client::GenericClientConfiguration& config, const Aws::String& serviceName);
 
             virtual void OverrideEndpoint(const Aws::String& endpoint, const Aws::Http::Scheme& scheme = Aws::Http::Scheme::HTTPS);
 
@@ -35,6 +37,9 @@ namespace Aws
             void SetBooleanParameter(Aws::String name, bool value);
             void SetStringArrayParameter(Aws::String name, const Aws::Vector<Aws::String>&& value);
             const Aws::Vector<EndpointParameter>& GetAllParameters() const;
+
+        private:
+            void SetFromClientConfigurationImpl(const Client::ClientConfiguration& config, const Aws::String& serviceName);
 
         protected:
             Aws::Vector<EndpointParameter> m_params;
--- a/src/aws-cpp-sdk-core/include/aws/core/endpoint/EndpointProviderBase.h
+++ b/src/aws-cpp-sdk-core/include/aws/core/endpoint/EndpointProviderBase.h
@@ -50,6 +50,15 @@ namespace Aws
             virtual void InitBuiltInParameters(const ClientConfigurationT& config) = 0;
 
             /**
+             * Initialize client context parameters from a ClientConfiguration with service name
+             */
+            virtual void InitBuiltInParameters(const ClientConfigurationT& config, const Aws::String& serviceName)
+            {
+                AWS_UNREFERENCED_PARAM(serviceName);
+                InitBuiltInParameters(config);
+            }
+
+            /**
              * Function to override endpoint, i.e. to set built-in parameter "AWS::Endpoint"
              */
             virtual void OverrideEndpoint(const Aws::String& endpoint) = 0;
--- a/src/aws-cpp-sdk-core/source/config/EndpointResolver.cpp
+++ b/src/aws-cpp-sdk-core/source/config/EndpointResolver.cpp
@@ -1,0 +1,98 @@
+/**
+ * Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.
+ * SPDX-License-Identifier: Apache-2.0.
+ */
+
+#include <aws/core/config/EndpointResolver.h>
+#include <aws/core/utils/StringUtils.h>
+#include <aws/core/config/ConfigAndCredentialsCacheManager.h>
+#include <aws/core/platform/Environment.h>
+#include <aws/core/utils/logging/LogMacros.h>
+
+static const char* SERVICE_ENDPOINT_ENV_PREFIX = "AWS_ENDPOINT_URL_";
+static const char* GLOBAL_ENDPOINT_ENV_VAR = "AWS_ENDPOINT_URL";
+
+namespace Aws
+{
+    namespace Config
+    {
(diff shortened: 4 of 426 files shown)


src/aws-cpp-sdk-core/source/endpoint/BuiltInParameters.cpp
```diff
@@ -78,6 +79,51 @@ namespace Endpoint
         return SetFromClientConfiguration(static_cast<const Client::ClientConfiguration&>(config));
     }
 
+    void BuiltInParameters::SetFromClientConfiguration(const Client::ClientConfiguration& config, const Aws::String& serviceName)
+    {
+        bool forceFIPS = false;
```
sbiscigl: this logic is copy and pasted from the other function `SetFromClientConfiguration` with the exception that we handle the setting of override differently. lets not repeat ourselves and refactor this function so that we dont have copy and pasted code.

src/aws-cpp-sdk-core/source/endpoint/BuiltInParameters.cpp
```diff
@@ -78,6 +79,51 @@ namespace Endpoint
...
+            OverrideEndpoint(config.endpointOverride, config.scheme);
+            if (config.region.empty()) {
+                AWS_LOGSTREAM_WARN(ENDPOINT_BUILTIN_LOG_TAG,
+                                   "Endpoint is overridden but region is not set. "
+                                   "Region is required by many endpoint rule sets to resolve the endpoint. "
+                                   "And it is required to compute an aws signature.");
+                SetStringParameter(AWS_REGION, "region-not-set");
+            }
+        } else if (!serviceName.empty()) {
+            Aws::Config::EndpointResolver::EndpointSource(serviceName, config.profileName, config.scheme, *this);
```
sbiscigl: we shouldnt be passing `*this`, the objective here is figure out what the endpoint override it, we should not also be mutating self in this call. something along the lines of
```c++
OverrideEndpoint(
  Aws::Config::EndpointResolver::EndpointSource(serviceName, config), 
  config.scheme);
```

would be what we want to be doing

src/aws-cpp-sdk-core/include/aws/core/endpoint/BuiltInParameters.h
```diff
@@ -26,6 +26,8 @@ namespace Aws
 
             virtual void SetFromClientConfiguration(const Client::ClientConfiguration& config);
             virtual void SetFromClientConfiguration(const Client::GenericClientConfiguration& config);
+            virtual void SetFromClientConfiguration(const Client::ClientConfiguration& config, const Aws::String& serviceName);
```
sbiscigl, line 29: should the non - service name ones have a implementation? why would they not simply call the service name function but with a empty service name?

would this be breaking if someone subclasses BuiltInParameters and they were already overriding `SetFromClientConfiguration`?

src/aws-cpp-sdk-core/include/aws/core/config/EndpointResolver.h
```diff
@@ -0,0 +1,124 @@
...
+namespace Aws
+{
+    namespace Config
+    {
+        static const char ENDPOINT_RESOLVER_TAG[] = "EndpointResolver";
+
+        /**
+         * Resolver that sources endpoints and sets them on endpoint providers.
+         */
+        class AWS_CORE_API EndpointResolver
```
sbiscigl: why do we have a class if all the methods are static and we have no members? currently this class is default constructible.

src/aws-cpp-sdk-core/include/aws/core/endpoint/EndpointProviderBase.h
```diff
@@ -49,6 +49,11 @@ namespace Aws
              */
             virtual void InitBuiltInParameters(const ClientConfigurationT& config) = 0;
 
+            /**
+             * Initialize client context parameters from a ClientConfiguration with service name
+             */
+            virtual void InitBuiltInParameters(const ClientConfigurationT& config, const Aws::String& serviceName) = 0;
```
sbiscigl: adding a new pure virtual method means that anything else implementing this will break. why not have a default implementation that calls the previous function to preserve backwards compat?

src/aws-cpp-sdk-core/source/endpoint/BuiltInParameters.cpp
kai-ion: Refactor the logic to use a shared `SetFromClientConfigurationImpl `function

src/aws-cpp-sdk-core/source/endpoint/BuiltInParameters.cpp
kai-ion: Refactor endpointResolver to return an endpoint instead of mutating the config directly

src/aws-cpp-sdk-core/include/aws/core/endpoint/BuiltInParameters.h
kai-ion: Refactor to use a fallback call instead of implementing new virtual functions

src/aws-cpp-sdk-core/include/aws/core/config/EndpointResolver.h
kai-ion: Change this class into a name space instead

src/aws-cpp-sdk-core/include/aws/core/endpoint/EndpointProviderBase.h
kai-ion: Refactor to use a fallback call with empty service name

src/aws-cpp-sdk-core/include/aws/core/endpoint/BuiltInParameters.h
```diff
@@ -26,6 +26,8 @@ namespace Aws
 
             virtual void SetFromClientConfiguration(const Client::ClientConfiguration& config);
             virtual void SetFromClientConfiguration(const Client::GenericClientConfiguration& config);
+            void SetFromClientConfiguration(const Client::ClientConfiguration& config, const Aws::String& serviceName);
```
sbiscigl: why are we not allowing this to be virtual as it was before?

src/aws-cpp-sdk-core/source/config/EndpointResolver.cpp
```diff
@@ -0,0 +1,99 @@
...
+                    const Aws::String ignoreVal = profile.GetValue("ignore_configured_endpoint_urls");
+                    if (!ignoreVal.empty() && Utils::StringUtils::ConvertToBool(ignoreVal.c_str())) {
+                        AWS_LOGSTREAM_DEBUG(ENDPOINT_RESOLVER_TAG, "Configured endpoints ignored due to ignore_configured_endpoint_urls=true in profile: " << profileName);
+                        return "";
+                    }
+                }
+
+                // 2) Service-specific environment variable
+                {
+                    Aws::String service = "AWS_ENDPOINT_URL_";
```
sbiscigl: why do we create a new variable eachtime? shouldnt this be const and static

src/aws-cpp-sdk-core/source/config/EndpointResolver.cpp
```diff
@@ -0,0 +1,99 @@
...
+                    service += serviceKey;
+                    Aws::String fromEnv = Aws::Environment::GetEnv(service.c_str());
+                    if (!fromEnv.empty()) {
+                        AWS_LOGSTREAM_DEBUG(ENDPOINT_RESOLVER_TAG, "Resolved configured endpoint from service-specific environment variable: " << service);
```
sbiscigl: why are we using DEBUG and TRACE in the same if statement, that means that no matter which log level you choose one or both will be displayed. additionally when you want to know one, but not the other?

tests/aws-cpp-sdk-s3-integration-tests/EndpointResolverIntegrationTest.cpp
```diff
@@ -0,0 +1,246 @@
...
+using namespace Aws::Auth;
+using namespace Aws::Client;
+using namespace Aws::Http;
+using namespace Aws::Utils;
+using namespace Aws::Config;
+using namespace Aws::Environment;
+
+const char* ALLOCATION_TAG = "EndpointResolverIntegrationTest";
+
+class S3TestClient : public S3Client
```
sbiscigl: why does this class exist? it looks like it just calls the base class and nothing more

tests/aws-cpp-sdk-s3-integration-tests/EndpointResolverIntegrationTest.cpp
```diff
@@ -0,0 +1,246 @@
...
+    S3ClientConfiguration s3Config;
+    s3Config.region = "us-east-1";
+    s3Config.retryStrategy = MakeShared<NoRetry>(ALLOCATION_TAG);
+
+    const auto& seenRequest = ExecuteHeadBucketRequest(s3Config);
+
+    EXPECT_EQ("https://test-bucket.custom-s3.example.com", seenRequest.GetUri().GetURIString());
+}
+
+TEST_F(EndpointResolverIntegrationTest, GlobalEnvironmentVariable)
```
sbiscigl, line 149: I think we would want a test for `GlobalEnvironmentVariableUsedOverProfile` so that we validate that global endpoints are chosen over a specific profile endpoint.

kai-ion: Added a `GlobalEnvironmentVariableUsedOverProfile` test, works as expected with global environment > profile

tests/aws-cpp-sdk-s3-integration-tests/EndpointResolverIntegrationTest.cpp
kai-ion: Ops, copied over from s3unit test. Yes there is nothing in the test client being used in this integration test. (Headbucket request is already public).
Removed this class.

src/aws-cpp-sdk-core/source/config/EndpointResolver.cpp
kai-ion: Added `SERVICE_ENDPOINT_ENV_PREFIX` as static const char*, also did the same for `GLOBAL_ENDPOINT_ENV_VAR`

src/aws-cpp-sdk-core/source/config/EndpointResolver.cpp
kai-ion: Replaced the DEBUG + TRACE with single DEBUG messages that include both the source information and the resolved URL

src/aws-cpp-sdk-core/include/aws/core/endpoint/BuiltInParameters.h
kai-ion: This is me being stupid. Reverted back to virtual with the new impl private function to reduce code duplication.

PR #3604, opened 2025-10-31
Verifying ContentRange on response from GetObject

pulimsr:
Verifying ContentRange on response from GetObject

*Issue #, if available:*

*Description of changes:*

*Check all that applies:*

- [ ] Linux
- [ ] Windows
- [ ] Android
- [ ] MacOS
- [ ] IOS
- [ ] Other Platforms


tests/aws-cpp-sdk-transfer-tests/TransferTests.cpp
```diff
@@ -2328,6 +2328,40 @@ TEST_P(TransferTests, TransferManager_TestRelativePrefix)
...
+    uint64_t offset = 1024;
+    uint64_t partSize = 2048;
+    std::shared_ptr<TransferHandle> downloadPtr = transferManager->DownloadFile(GetTestBucketName(), RandomFileName, offset, partSize, createStreamFn);
+    
+    downloadPtr->WaitUntilFinished();
+    ASSERT_EQ(TransferStatus::COMPLETED, downloadPtr->GetStatus());
+    ASSERT_EQ(partSize, downloadPtr->GetBytesTotalSize());
+    ASSERT_EQ(partSize, downloadPtr->GetBytesTransferred());
+}
+
```
kai-ion, line 2364: Should there also be a test with a false content range to test for failures?

pulimsr: I'll look into creating a mock S3 response with an incorrect ContentRange to test the failure path

src/aws-cpp-sdk-transfer/source/transfer/TransferManager.cpp
```diff
@@ -1110,6 +1136,37 @@ namespace Aws
             }
             else
             {
+                if (request.RangeHasBeenSet())
+                {
+                    const auto& requestedRange = request.GetRange();
+                    const auto& responseContentRange = outcome.GetResult().GetContentRange();
+                    
+                    if (!responseContentRange.empty())
```
sbiscigl: is there any instance where we set range on the request where s3 does not return a range? would that be a error?

tests/aws-cpp-sdk-transfer-tests/TransferTests.cpp
sbiscigl: +1 i think even as going as far to have a "transfer manager unit test" might be good. you can largely copy the [S3 unit tests](https://github.com/aws/aws-sdk-cpp/blob/main/tests/aws-cpp-sdk-s3-unit-tests/S3UnitTests.cpp) to create the skeleton of it.

src/aws-cpp-sdk-transfer/source/transfer/TransferManager.cpp
```diff
@@ -1110,6 +1136,33 @@ namespace Aws
             }
             else
             {
+                if (request.RangeHasBeenSet())
```
sbiscigl: nit: `else if(request.RangeHasBeenSet()) { ... }` is preferred to `else { if { ... }}`

tests/aws-cpp-sdk-transfer-unit-tests/TransferUnitTests.cpp
```diff
@@ -0,0 +1,113 @@
+#include <gtest/gtest.h>
+#include <aws/core/Aws.h>
+#include <aws/testing/AwsTestHelpers.h>
+#include <aws/testing/MemoryTesting.h>
+
+using namespace Aws;
+
+const char* ALLOCATION_TAG = "TransferUnitTest";
+
+// Copy the VerifyContentRange function for testing
```
sbiscigl: I don't think this is what we meant by saying write unit tests for this. we dont want to copy and paste code to test it, we want to mock the interfaces that make actual calls such that we can test the code we added. I think the idea here is that you would create a transfer manager with a mocked out s3 client, so that when you run the transfer manager it uses this mocked out s3 client instead of a real one, so you can verify behavior without using a real network connection. something like this

```c++
#include <aws/core/Aws.h>
#include <aws/core/utils/threading/PooledThreadExecutor.h>
#include <aws/s3/S3Client.h>
#include <aws/transfer/TransferManager.h>

using namespace Aws;
using namespace Aws::Utils;
using namespace Aws::Utils::Logging;
using namespace Aws::Utils::Threading;
using namespace Aws::Transfer;
using namespace Aws::S3;

namespace {
const char* LOGTAG = "TestApplication";

class MockS3Client : public S3Client {
 public:
  Model::CompleteMultipartUploadOutcome
  CompleteMultipartUpload(const Model::CompleteMultipartUploadRequest& request) const override {
   //TODO: implement mock
  }

  Model::CreateMultipartUploadOutcome
  CreateMultipartUpload(const Model::CreateMultipartUploadRequest& request) const override {
    //TODO: implement mock
  }

  Model::GetObjectOutcome
  GetObject(const Model::GetObjectRequest& request) const override {
    //TODO: implement mock
  }

  Model::HeadObjectOutcome
  HeadObject(const Model::HeadObjectRequest& request) const override {
    //TODO: implement mock
  }

  Model::PutObjectOutcome
  PutObject(const Model::PutObjectRequest& request) const override {
    //TODO: implement mock
  }

  Model::UploadPartOutcome
  UploadPart(const Model::UploadPartRequest& request) const override {
    //TODO: implement mock
  }
};

}

class SdkContext {
 public:
  explicit SdkContext(SDKOptions&& options) : options_(std::move(options)) { InitAPI(options_); }
  ~SdkContext() { ShutdownAPI(options_); }

 private:
  SDKOptions options_;
};

auto main() -> int {
  SDKOptions options;
  options.loggingOptions.logLevel = LogLevel::Trace;
  SdkContext context(std::move(options));

  const auto executor = Aws::MakeUnique<PooledThreadExecutor>(LOGTAG, std::thread::hardware_concurrency());
  TransferManagerConfiguration configuration{executor.get()};
  configuration.s3Client = Aws::MakeShared<MockS3Client>("MockS3Client");
  // do testing operations
  return 0;
}
```

but wrapped by gtest

src/aws-cpp-sdk-transfer/source/transfer/TransferManager.cpp
```diff
@@ -1108,33 +1134,57 @@ namespace Aws
...
                     handle->ChangePartToFailed(partState);
+                    handle->SetError(error);
+                    TriggerErrorCallback(handle, error);
                 }
             }
+            else
+            {
+                handle->ChangePartToFailed(partState);
+            }
+            }
```
sbiscigl: nit:
```
}
}
```
feels like a formatting mismatch, lets format this block so its easier to read

src/aws-cpp-sdk-transfer/source/transfer/TransferManager.cpp
```diff
@@ -825,6 +825,33 @@ namespace Aws
...
+            if (requestedRange.empty() || responseContentRange.empty())
+            {
+                return false;
+            }
+
+            if (requestedRange.find("bytes=") != 0)
+            {
+                return false;
+            }
+            Aws::String requestRange = requestedRange.substr(6);
```
sbiscigl: `substr(6)` seems like a "magic number", can we make this based on a search? a hardcoded index seems like it could break if anything changes

src/aws-cpp-sdk-transfer/source/transfer/TransferManager.cpp
```diff
@@ -825,6 +825,33 @@ namespace Aws
...
+
+            if (responseContentRange.find("bytes ") != 0)
```
sbiscigl: `!= 0` is weird but valid, for searching string in cpp you should prefer the [`npos` value](http://en.cppreference.com/w/cpp/string/basic_string/npos.html) i.e.
```
if (responseContentRange.find("bytes ") != Aws::String::npos) {/*...*/}
```

src/aws-cpp-sdk-transfer/source/transfer/TransferManager.cpp
pulimsr: Alright

src/aws-cpp-sdk-transfer/source/transfer/TransferManager.cpp
pulimsr: okay

src/aws-cpp-sdk-transfer/source/transfer/TransferManager.cpp
pulimsr: okay

pulimsr: Yes, that makes sense - but since I'm specifically checking if the string starts with 'bytes', I'm using .substr(0, strlen(requestPrefix)) for the prefix comparison.

src/aws-cpp-sdk-transfer/source/transfer/TransferManager.cpp
pulimsr: I've updated the code to use strlen(requestPrefix) instead of the hardcoded value

src/aws-cpp-sdk-transfer/source/transfer/TransferManager.cpp
pulimsr: formatted this

PR #3608, opened 2025-11-04
Multi auth selection (Client level)

sbaluja:
*Issue #, if available:*

*Description of changes:*
This PR adds support for client-level auth resolution. It includes replacement of current authResolvers to be more generic, and allows for customers to set preferences for authentication methods. 

*Check all that applies:*

- [x] Linux
- [x] Windows
- [x] Android
- [x] MacOS
- [ ] IOS
- [ ] Other Platforms

### DIFF
--- a/src/aws-cpp-sdk-core/include/smithy/identity/auth/built-in/BearerTokenAuthSchemeResolver.h
+++ b/src/aws-cpp-sdk-core/include/smithy/identity/auth/built-in/BearerTokenAuthSchemeResolver.h
@@ -1,0 +1,28 @@
+/**
+ * Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.
+ * SPDX-License-Identifier: Apache-2.0.
+ */
+#pragma once
+
+#include <smithy/identity/auth/AuthSchemeResolverBase.h>
+#include <smithy/identity/auth/built-in/BearerTokenAuthSchemeOption.h>
+
+namespace smithy
+{
+template <typename ServiceAuthSchemeParametersT =
+              DefaultAuthSchemeResolverParameters>
+class BearerTokenAuthSchemeResolver
+    : public AuthSchemeResolverBase<ServiceAuthSchemeParametersT>
+{
+  public:
+    using ServiceAuthSchemeParameters = ServiceAuthSchemeParametersT;
+    virtual ~BearerTokenAuthSchemeResolver() = default;
+
+    Aws::Vector<AuthSchemeOption> resolveAuthScheme(
+        const ServiceAuthSchemeParameters &identityProperties) override
+    {
+        AWS_UNREFERENCED_PARAM(identityProperties);
+        return {BearerTokenAuthSchemeOption::bearerTokenAuthSchemeOption};
+    }
+};
+} // namespace smithy
--- a/src/aws-cpp-sdk-core/include/smithy/identity/auth/built-in/SigV4AuthSchemeResolver.h
+++ b/src/aws-cpp-sdk-core/include/smithy/identity/auth/built-in/SigV4AuthSchemeResolver.h
@@ -1,0 +1,25 @@
+/**
+* Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.
+ * SPDX-License-Identifier: Apache-2.0.
+ */
+#pragma once
+
+#include <smithy/identity/auth/AuthSchemeResolverBase.h>
+#include <smithy/identity/auth/built-in/SigV4AuthSchemeOption.h>
+
+
+namespace smithy {
+    template<typename ServiceAuthSchemeParametersT = DefaultAuthSchemeResolverParameters>
+    class SigV4AuthSchemeResolver : public AuthSchemeResolverBase<ServiceAuthSchemeParametersT>
+    {
+    public:
+        using ServiceAuthSchemeParameters = ServiceAuthSchemeParametersT;
+        virtual ~SigV4AuthSchemeResolver() = default;
+
+        Aws::Vector<AuthSchemeOption> resolveAuthScheme(const ServiceAuthSchemeParameters& identityProperties) override
+        {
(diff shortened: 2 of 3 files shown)


generated/src/aws-cpp-sdk-dynamodb/include/aws/dynamodb/DynamoDBClient.h
```diff
@@ -11,8 +11,9 @@
 #include <aws/dynamodb/DynamoDB_EXPORTS.h>
 #include <smithy/client/AwsSmithyClient.h>
 #include <smithy/client/serializer/JsonOutcomeSerializer.h>
+#include <smithy/identity/auth/built-in/GenericAuthSchemeResolver.h>
 #include <smithy/identity/auth/built-in/SigV4AuthScheme.h>
-#include <smithy/identity/auth/built-in/SigV4AuthSchemeResolver.h>
+#include <smithy/identity/auth/built-in/SigV4aAuthScheme.h>
```
sbiscigl: why are we adding the header for `V4a` that scares me a little bit because dynamo does not use `V4a`

src/aws-cpp-sdk-core/include/aws/core/client/ClientConfiguration.h
```diff
@@ -563,6 +563,12 @@ namespace Aws
               std::chrono::milliseconds credentialCacheCacheTTL = std::chrono::minutes(50);
             } stsCredentialsProviderConfig;
           } credentialProviderConfig;
+
+          /**
+           * Authentication scheme preferences in order of preference.
+           * First available auth scheme will be used for each operation.
+           */
+          Aws::Vector<Aws::String> authPreferences;
```
sbiscigl, line 588: not sure if we should put this on the top level `ClientConfiguration` object, should it be code-genned on the service specific client configuration? that gives us two advantages, only smithy clients witll have it, and additionally we can make it a `Aws::Array` of constant space.

src/aws-cpp-sdk-core/include/smithy/identity/auth/AuthSchemeResolverBase.h
```diff
@@ -9,9 +9,23 @@
...
 
 namespace smithy {
+
+static char SIGV4_PREFERENCE[] = "sigv4";
+static char SIGV4A_PREFERENCE[] = "sigv4a";
+static char BEARER_PREFERENCE[] = "bearer";
+static char NO_AUTH_PREFERENCE[] = "noauth";
+
+// Global map from auth scheme name (trimmed ID) to full ID for case insensitive lookup
+static const Aws::UnorderedMap<Aws::String, Aws::String> AUTH_SCHEME_NAME_TO_ID = {
```
sbiscigl: we cant do static const `Aws::UnorderedMap` because of customer allocators, should make it a `Aws::Array` so that we avoid the whole allocator buisness

src/aws-cpp-sdk-core/include/smithy/identity/auth/built-in/BearerTokenAuthSchemeResolver.h
```diff
@@ -1,28 +0,0 @@
-/**
- * Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.
- * SPDX-License-Identifier: Apache-2.0.
- */
-#pragma once
```
sbiscigl, line 5: cant remove installed headers, you can make the header point to something new, but removing installed headers can and will break customers if any of their code includes this header.

src/aws-cpp-sdk-core/include/smithy/identity/auth/AuthSchemeResolverBase.h
```diff
@@ -43,7 +58,31 @@ class AuthSchemeResolverBase
...
+protected:
+  virtual Aws::Vector<AuthSchemeOption> resolveAuthSchemeImpl(const ServiceAuthSchemeParameters& identityProperties) = 0;
+  
+  virtual Aws::Vector<AuthSchemeOption> filterByPreferences(const Aws::Vector<AuthSchemeOption>& options, 
+                                                           const Aws::Vector<Aws::String>& preferences) {
+    if (preferences.empty()) return options;
+    
+    Aws::Vector<AuthSchemeOption> filtered;
+    for (const auto& pref : preferences) {
+      auto prefSchemeIt = AUTH_SCHEME_NAME_TO_ID.find(Aws::Utils::StringUtils::ToLower(pref.c_str()));
```
sbiscigl: is there a way to code generate `AUTH_SCHEME_NAME_TO_ID` so that we dont have all the mappings, just the ones we know are relevant? and would that simplfy any assumptions?

generated/src/aws-cpp-sdk-dynamodb/include/aws/dynamodb/DynamoDBClient.h
sbaluja: Ya this shouldn't be there, its coming from the hasSigV4aAuth incorrectly returning an truthy value during code gen. Will fix

src/aws-cpp-sdk-core/include/smithy/identity/auth/AuthSchemeResolverBase.h
sbaluja: we _could_ probably do this, but it's gonna add complexity with little benefit imo. Since we construct `GenericAuthResolver` as part of the code gen, we could also construct with the mappings, and pass it down to the base class where the mapping is needed. Maybe missing something, but given that this is a 4 key/pair value lookup I don't think there's any gains here

src/aws-cpp-sdk-core/include/smithy/identity/auth/AuthSchemeResolverBase.h
sbaluja: ya good callout. While writing this something felt wrong about it and now I'm remembering why

PR #3610, opened 2025-11-05
validating url for sso and sts

pulimsr:
*Description of changes:* validating credential urls for sso and sts

*Check all that applies:*

- [ ] Linux
- [ ] Windows
- [ ] Android
- [ ] MacOS
- [ ] IOS
- [ ] Other Platforms


src/aws-cpp-sdk-core/source/internal/AWSHttpResourceClient.cpp
```diff
@@ -685,6 +696,16 @@ namespace Aws
             {
                 ss << ".cn";
             }
+            ss.str();
+
+            Aws::Http::URI uri(ss.str());
```
sbiscigl: instead of validating the the URI in the constructor, validate it when we make the call like how we currently do it in [AwsClient](https://github.com/aws/aws-sdk-cpp/blob/main/src/aws-cpp-sdk-core/source/client/AWSClient.cpp#L262), i.e. in [`AWSHttpResourceClient::GetResourceWithAWSWebServiceResult`](https://github.com/aws/aws-sdk-cpp/blob/main/src/aws-cpp-sdk-core/source/internal/AWSHttpResourceClient.cpp#L132C45-L132C102) we should validate there before making any calls

tests/aws-cpp-sdk-core-tests/aws/auth/SSRFValidationTest.cpp
```diff
@@ -0,0 +1,63 @@
+#include <aws/testing/AwsCppSdkGTestSuite.h>
+#include <aws/core/auth/SSOCredentialsProvider.h>
+#include <aws/core/internal/AWSHttpResourceClient.h>
+#include <aws/core/client/ClientConfiguration.h>
+
+using namespace Aws;
+using namespace Aws::Auth;
+using namespace Aws::Client;
+
+class SSRFValidationTest : public Aws::Testing::AwsCppSdkGTestSuite {
```
sbiscigl: lets not make this its own test, [`AWSCredentialsProviderTest`](https://github.com/aws/aws-sdk-cpp/blob/main/tests/aws-cpp-sdk-core-tests/aws/auth/AWSCredentialsProviderTest.cpp) has tests for both SSO and STS, lets place tests there that validate this behavor

sbiscigl (approved):

PR #3614, opened 2025-11-06
Remove unwanted c++ std injection into downstream projects via pkgconfig

WillAyd:
*Issue #, if available:*
3613

*Description of changes:*
Removed c++ std and fno-exceptions flag in pkgconfig

*Check all that applies:*
Not applicable to CI testing

- [ ] Linux
- [ ] Windows
- [ ] Android
- [ ] MacOS
- [ ] IOS
- [ ] Other Platforms

### DIFF
--- a/cmake/compiler_settings.cmake
+++ b/cmake/compiler_settings.cmake
@@ -36,7 +36,6 @@ endfunction()
 
 
 macro(set_gcc_flags)
-    list(APPEND AWS_COMPILER_FLAGS "-fno-exceptions" "-std=c++${CPP_STANDARD}")
     if(COMPILER_IS_MINGW)
         list(APPEND AWS_COMPILER_FLAGS -D__USE_MINGW_ANSI_STDIO=1)
     endif()


cmake/compiler_settings.cmake
```diff
@@ -36,7 +36,6 @@ endfunction()
 
 
 macro(set_gcc_flags)
-    list(APPEND AWS_COMPILER_FLAGS "-fno-exceptions" "-std=c++${CPP_STANDARD}")
```
WillAyd, line 39: There are other flags in this file that maybe we should remove? As is, flags that are required for internal compilation are leaking out via pkgconfig to third parties. Perhaps there is some historical intent, but I think this is undesirable (?)

sbiscigl:
closing in favor of https://github.com/aws/aws-sdk-cpp/pull/3792 which adds the opt-in cmake arg `NO_CPPSTD_IN_PKG_CONFIG` that you can set in your distribution.

Issue #3618, opened 2025-11-10
Bug when s_preservePathSeparators enabled


dongxiao1198:
### Describe the bug

src/aws-cpp-sdk-core/include/aws/core/http/URI.h

If we enable s_preservePathSeparators at initOption, we can try writing files like:
s3://bucket/test/test_dir/
it seems this path will ignore the tailing slash...and actually writing file
s3://bucket/test/test_dir

### Regression Issue

- [ ] Select this option if this issue appears to be a regression.

### Expected Behavior

can handle the tailing slash

### Current Behavior

trimmed

### Reproduction Steps

Modify this case can find this issue:
tests/aws-cpp-sdk-s3-unit-tests/S3UnitTests.cpp

TEST_F(S3UnitTest, S3UriPathPreservationOn) {

```
diff --git a/tests/aws-cpp-sdk-s3-unit-tests/S3UnitTests.cpp b/tests/aws-cpp-sdk-s3-unit-tests/S3UnitTests.cpp
index 80bc663d3ed..ea543c3b471 100644
--- a/tests/aws-cpp-sdk-s3-unit-tests/S3UnitTests.cpp
+++ b/tests/aws-cpp-sdk-s3-unit-tests/S3UnitTests.cpp
@@ -226,7 +226,7 @@ TEST_F(S3UnitTest, S3UriPathPreservationOn) {

   auto putObjectRequest = PutObjectRequest()
       .WithBucket("velvetunderground")
-      .WithKey("////stephanie////says////////////that////////she//wants///////to/know.txt");
+      .WithKey("////stephanie////says////////////that////////she//wants///////to/know.txt/");
```

And can test uri like:
stephanie////says////////////that////////she//wants///////to/know.txt
stephanie////says////////////that////////she//wants///////to/know.txt/
////stephanie////says////////////that////////she//wants///////to/know.txt
////stephanie////says////////////that////////she//wants///////to/know.txt/
and so on

### AWS CPP SDK version used

master

### Compiler and Version used

llvm18

### Operating System and version

macos

dongxiao1198:
@pulimsr this may related to your patch https://github.com/aws/aws-sdk-cpp/pull/3499

pulimsr:
Thanks for reporting, fixed this issue via PR [#3619](https://github.com/aws/aws-sdk-cpp/pull/3619). Closing this bug report.

PR #3619, opened 2025-11-10
fix/when s_preservePathSeparators enabled

pulimsr:
*Issue #3618*

*Description of changes:* Fixed trailing slash preservation logic in URI parsing to correctly handle both S3 object keys with trailing slashes and SNS signature compatibility when s_
preservePathSeparators is enabled.

*Check all that applies:*

- [ ] Linux
- [ ] Windows
- [ ] Android
- [ ] MacOS
- [ ] IOS
- [ ] Other Platforms

### DIFF
--- a/tools/scripts/run_integration_tests.py
+++ b/tools/scripts/run_integration_tests.py
@@ -44,6 +44,7 @@ def main():
         "aws-cpp-sdk-dynamodb-integration-tests",
         "aws-cpp-sdk-sqs-integration-tests",
         "aws-cpp-sdk-sqs-unit-tests",
+        "aws-cpp-sdk-sns-integration-tests",
         "aws-cpp-sdk-s3-integration-tests",
         "aws-cpp-sdk-s3-unit-tests",
         "aws-cpp-sdk-s3-crt-integration-tests",


tests/aws-cpp-sdk-s3-unit-tests/S3UnitTests.cpp
```diff
@@ -226,7 +226,7 @@ TEST_F(S3UnitTest, S3UriPathPreservationOn) {
 
   auto putObjectRequest = PutObjectRequest()
       .WithBucket("velvetunderground")
-      .WithKey("////stephanie////says////////////that////////she//wants///////to/know.txt");
```
sbiscigl, line 229: i dont think we want to _remove_ this test case. Its still a valuable test case, i think we just want to add more test cases. additionally can you add a test case for `.txt//` to make sure that it works for many slashes and not just one

src/aws-cpp-sdk-core/include/aws/core/http/URI.h
```diff
@@ -153,7 +153,7 @@ namespace Aws
                 {
                     m_pathSegments.push_back(segment);
                 }
-                m_pathHasTrailingSlash = (m_pathSegments.empty() || !s_preservePathSeparators) && (!segments.empty() && segments.back() == '/');
+                m_pathHasTrailingSlash = s_preservePathSeparators ? (!segments.empty() && segments.back() == '/') : m_pathSegments.empty() && (!segments.empty() && segments.back() == '/');
```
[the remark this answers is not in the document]
dongxiao1198: I debugged your case PreservePathSeparatorsTest, I think the root cause was URI::SetPath does not reset the m_pathHasTrailingSlash to false.

In AWSAuthSignerHelper.cpp line 42:
uriCpy.SetPath(uriCpy.GetURLEncodedPathRFC3986());
reset the path by '/', but uriCpy.m_pathHasTrailingSlash is already true, here doubled the '/'

I tried revert your previous patch and fixed like this, I think it is working, all case passed:

```
diff --git a/src/aws-cpp-sdk-core/source/http/URI.cpp b/src/aws-cpp-sdk-core/source/http/URI.cpp
index daf5554b7a4..b7cfb33f7af 100644
--- a/src/aws-cpp-sdk-core/source/http/URI.cpp
+++ b/src/aws-cpp-sdk-core/source/http/URI.cpp
@@ -254,6 +254,7 @@ Aws::String URI::GetURLEncodedPathRFC3986() const
 void URI::SetPath(const Aws::String& value)
 {
     m_pathSegments.clear();
+    m_pathHasTrailingSlash = false;
     AddPathSegments(value);
 }
```

How do you say?

src/aws-cpp-sdk-core/source/http/URI.cpp
```diff
@@ -206,7 +206,11 @@ Aws::String URI::GetPath() const
         path.append(segment);
     }
 
-    if (m_pathSegments.empty() || m_pathHasTrailingSlash)
+    if (m_pathSegments.empty())
```
sbiscigl: why did you change this?
```c++
if (m_pathSegments.empty() || m_pathHasTrailingSlash)
{
  path.push_back('/');
}
```

it is functionally the same as
```c++
if (m_pathSegments.empty())
{
  path.push_back('/');
}
else if (m_pathHasTrailingSlash)
{
  path.push_back('/');
}
```

pulimsr: You're right - this was a mistake while trying to change the code back. Updating now.

dongxiao1198:
If you revert code like this. you break https://github.com/aws/aws-sdk-cpp/pull/3499 again
look what I'm saying before

pulimsr:
Hi @dongxiao1198, - I've actually reverted that change since it wasn't necessary. I've added a test specifically addressing the scenario [here](https://github.com/aws/aws-sdk-cpp/blob/main/tests/aws-cpp-sdk-sns-integration-tests/PreservePathSeparatorsTest.cpp)

dongxiao1198:
> Hi @dongxiao1198, - I've actually reverted that change since it wasn't necessary. I've added a test specifically addressing the scenario [here](https://github.com/aws/aws-sdk-cpp/blob/main/tests/aws-cpp-sdk-sns-integration-tests/PreservePathSeparatorsTest.cpp)

this case has some issue：
options.httpOptions.preservePathSeparators = true; 
is not working since this set after the initialization in the main function in tests/aws-cpp-sdk-sns-integration-tests/RunTests.cpp 
so it actually run in options.httpOptions.preservePathSeparators = false mode.

PR #3626, opened 2025-11-19
Add Conan usage instructions in README

AbrilRBS:
*Description of changes:*
Hi! This PR adds instructions for using the sdk with Conan, which we thought would be a nice addition in the README file.

Please let me know if this is okay, or if you would prefer it placed elsewhere.

The latest available version is 1.11.692, and we usually update the available version every few months/on user requests.

*Check all that applies:*

- [ ] Linux
- [ ] Windows
- [ ] Android
- [ ] MacOS
- [ ] IOS
- [ ] Other Platforms

### DIFF
--- a/README.md
+++ b/README.md
@@ -120,11 +120,11 @@ The aws-sdk-cpp port in vcpkg is kept up to date by Microsoft team members and c
 
 ### Building aws-sdk-cpp - Using Conan
 
-You can install pre-built binaries for aws-sdk-cpp or build it from source using [Conan](https://conan.io/). Use the following command:
+You can download and install aws-sdk-cpp using [Conan](https://conan.io/). Use the following command:
 
-    conan install --requires="aws-sdk-cpp/[*]" --build=missing
+    conan install --requires="aws-sdk-cpp/[*]" --build="aws-*"
 
-The aws-sdk-cpp Conan recipe is kept up to date by Conan maintainers and community contributors.
+The aws-sdk-cpp Conan recipe is kept up to date by JFrog's Conan team members and community contributors.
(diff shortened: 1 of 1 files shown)


sbiscigl:
> You can install pre-built binaries

My main problem with this wording is that we are _not_ ABI stable and using pre-built binaries assuming that we are customers may run into issues. Also we are not providing these binaries. We may in the future and would like to but for now its not us providing these binaries.

[Also i think i would rather have this in our dev guide](https://docs.aws.amazon.com/sdk-for-cpp/v1/developer-guide/sdk-from-pm.html) i.e. we have "Get the SDK using vcpkg". adding a section "Get the SDK using conan" I would be a fan of. If its alright with you I can port what you have in your PR to that page and add it there.

AbrilRBS:
Hi @sbiscigl thanks for your insight!

> My main problem with this wording is that we are not ABI stable

Note that the command only downloads the necessary recipe to locally build the package, and Conan does not assume ABI compatibility by default, it only considers existing binaries if and only if they are compatible.

I do understand the sentiment and I've updated the command example to force local recompilation in all cases, just like it's done fin the vcpkg case. Would the current wording work better?

sbiscigl:
Sorry for taking a while to get back to you, yeah this is basically the instructions for vcpkg mirrored, so i'm good with this. I cant merge your PR for logistical reasons, so i created a PR with your commit on it, i will merge it once CI runs.

https://github.com/aws/aws-sdk-cpp/pull/3647/

sbiscigl:
https://github.com/aws/aws-sdk-cpp/pull/3647 merged resolving

PR #3629, opened 2025-11-24
Transfer manager checksums download

pulimsr:
*Description of changes:* Adding checksum validation on s3 download

*Check all that applies:*

- [x] Linux
- [x] Windows
- [x] Android
- [x] MacOS
- [ ] IOS
- [ ] Other Platforms


src/aws-cpp-sdk-transfer/source/transfer/TransferManager.cpp
```diff
@@ -874,6 +876,71 @@ namespace Aws
...
+                        actualChecksum = getResult.GetChecksumCRC32C();
+                    } else if (!getResult.GetChecksumSHA256().empty()) {
+                        actualChecksum = getResult.GetChecksumSHA256();
+                    } else if (!getResult.GetChecksumSHA1().empty()) {
+                        actualChecksum = getResult.GetChecksumSHA1();
+                    } else if (!getResult.GetChecksumCRC64NVME().empty()) {
+                        actualChecksum = getResult.GetChecksumCRC64NVME();
+                    }
+                    
+                    if (!actualChecksum.empty() && actualChecksum != handle->GetChecksum()) {
```
sbaluja: Do we need to check for `!actualChecksum.empty()`? Does it have to be non-empty or a simple mismatch is enough for us to error out?

src/aws-cpp-sdk-transfer/source/transfer/TransferManager.cpp
```diff
@@ -1117,6 +1199,20 @@ namespace Aws
...
+                        } else if (!getResult.GetChecksumCRC32C().empty()) {
+                            partState->SetChecksum(getResult.GetChecksumCRC32C());
+                        } else if (!getResult.GetChecksumSHA256().empty()) {
+                            partState->SetChecksum(getResult.GetChecksumSHA256());
+                        } else if (!getResult.GetChecksumSHA1().empty()) {
+                            partState->SetChecksum(getResult.GetChecksumSHA1());
+                        } else if (!getResult.GetChecksumCRC64NVME().empty()) {
+                            partState->SetChecksum(getResult.GetChecksumCRC64NVME());
+                        }
+
```
sbaluja: This logic of retrieving a checksum from an S3Outcome is repeated a few times here. Could be worth trying something like this (maybe with SFINAE) in the header file so we can centralize this logic:

```
            template<typename OutcomeType>
            Aws::String ExtractChecksum(const OutcomeType& outcome) {
              const auto& result = outcome.GetResult();

              if (!result.GetChecksumCRC32().empty()) {
                return result.GetChecksumCRC32();
              }
              if (!result.GetChecksumCRC32C().empty()) {
                return result.GetChecksumCRC32C();
              }
              if (!result.GetChecksumSHA256().empty()) {
                return result.GetChecksumSHA256();
              }
              if (!result.GetChecksumSHA1().empty()) {
                return result.GetChecksumSHA1();
              }
              if (!result.GetChecksumCRC64NVME().empty()) {
                return result.GetChecksumCRC64NVME();
              }
              return {};
            }
```

Then we can just call it:
```
auto actualChecksum = ExtractChecksum(getObjectOutcome);
partState->SetChecksum(ExtractChecksum(outcome))
```

src/aws-cpp-sdk-transfer/source/transfer/TransferManager.cpp
```diff
@@ -1507,5 +1648,65 @@ namespace Aws
...
+                    return "";
+                }
+            }
+            
+            // Simple CRC32 combination for parts in order
+            std::vector<std::pair<int, Aws::String>> sortedParts;
+            for (const auto& part : completedParts) {
+                sortedParts.emplace_back(part.first, part.second->GetChecksum());
+            }
+            std::sort(sortedParts.begin(), sortedParts.end());
```
sbaluja: I could be COMPLETELY wrong here, but is sorting needed since XOR is associative? Might be missing some s3 specific context

src/aws-cpp-sdk-transfer/source/transfer/TransferManager.cpp
pulimsr: Agreed - this refactoring makes sense to eliminate the duplication.

src/aws-cpp-sdk-transfer/source/transfer/TransferManager.cpp
pulimsr: You're right, `!actualChecksum.empty()` is redundant here.

src/aws-cpp-sdk-transfer/source/transfer/TransferManager.cpp
```diff
@@ -1594,5 +1640,56 @@ namespace Aws
             partFunc->second(part, state->GetChecksum());
           }
         }
+
+        bool TransferManager::ValidateDownloadChecksum(const std::shared_ptr<TransferHandle>& handle) {
```
sbiscigl: whenever we return a `bool` from a function we usually want a [`Aws::Utils::Outcome`](https://github.com/aws/aws-sdk-cpp/blob/7196862eb6fd4f92e9624586ad02c274329dbb0f/src/aws-cpp-sdk-core/include/aws/core/utils/Outcome.h#L27) instead. instead of simply returning a true or false fail, we should return a proper error if something happened.

src/aws-cpp-sdk-transfer/source/transfer/TransferManager.cpp
```diff
@@ -1594,5 +1640,56 @@ namespace Aws
...
+                hashCalculator = Aws::MakeShared<Aws::Utils::Crypto::CRC32C>("TransferManager");
+            } else if (m_transferConfig.checksumAlgorithm == S3::Model::ChecksumAlgorithm::SHA1) {
+                hashCalculator = Aws::MakeShared<Aws::Utils::Crypto::Sha1>("TransferManager");
+            } else if (m_transferConfig.checksumAlgorithm == S3::Model::ChecksumAlgorithm::SHA256) {
+                hashCalculator = Aws::MakeShared<Aws::Utils::Crypto::Sha256>("TransferManager");
+            } else {
+                hashCalculator = Aws::MakeShared<Aws::Utils::Crypto::CRC64>("TransferManager");
+            }
+
+            auto fileStream = Aws::MakeShared<Aws::FStream>("TransferManager",
```
sbiscigl, line 1694: So we are opening the file and re-reading the entire stream buffering and then failing the download if the file is different. this gives me two concerns:

* a file that is corrupted is written to disc

> Failure handling: The S3 Transfer Manager SHOULD download to a file in the same directory with .s3tmp.{uniqueIdentifier} (uniqueIdentifier <= 8 chars) suffix to indicate the file is in the process of being downloaded, but not yet complete. After all content has been written, rename the temporary file to the destination file using atomic rename operation. If there is an error, delete the temporary file. The S3 Transfer Manager SHOULD register cleanup handlers to remove the temporary file if the process terminates unexpectedly, if supported. The S3 Transfer Manager MUST NOT pre-validate available space because it can introduce race conditions and false failures.

This concerns me that this would cause an issue where we would leave a file that has failing data intergrity.

* we are reading the stream twice.

on files that are massive think 5TB, this can introduce overhead. i think this is covered in the spec with 

> If full object checksum is calculated inline while downloading source, the S3 Transfer Manager MAY buffer full parts in-memory while flushing is blocked on updating the running checksum object.

[this really means that while we write out to the file we want to be updating the checksum](https://github.com/aws/aws-sdk-cpp/blob/main/src/aws-cpp-sdk-transfer/source/transfer/TransferManager.cpp#L1067-L1070)

src/aws-cpp-sdk-transfer/source/transfer/TransferManager.cpp
```diff
@@ -231,15 +246,19 @@ namespace Aws
                                                                       const DownloadConfiguration& downloadConfig,
                                                                       const std::shared_ptr<const Aws::Client::AsyncCallerContext>& context)
         {
+            Aws::String tempFilePath = GenerateTempFilePath(writeToFile);
```
sbiscigl, line 249: why is this only on one of the download file apis and not both?

src/aws-cpp-sdk-transfer/source/transfer/TransferManager.cpp
```diff
@@ -946,7 +976,12 @@ namespace Aws
                         << "] Failed to download object to Bucket: [" << handle->GetBucketName() << "] with Key: ["
                         << handle->GetKey() << "] " << getObjectOutcome.GetError());
                 handle->ChangePartToFailed(partState);
-                handle->UpdateStatus(DetermineIfFailedOrCanceled(*handle));
+                auto finalStatus = DetermineIfFailedOrCanceled(*handle);
+                handle->UpdateStatus(finalStatus);
+
+                if (finalStatus == TransferStatus::FAILED && !handle->GetTempFilePath().empty()) {
+                    Aws::FileSystem::RemoveFileIfExists(handle->GetTempFilePath().c_str());
```
sbiscigl, line 983: what actually writes out to the real path? this just removes the temp file?

pulimsr: the temp file gets renamed inside `ValidateDownloadChecksum()` [here](https://github.com/aws/aws-sdk-cpp/pull/3629/files/1aac67b3fc9f57df8cea52d15c0415028286fcaf#diff-1e9c00af8f974d27569456eaa8292bc89bf0d8c3aa7b926c19f7dace8575f481R1734)

PR #3633, opened 2025-11-29
Add lombok annotation processor in maven to fix builds on recent jdk versions

apache-hb:
*Description of changes:*

Since jdk23 lombok needs to be registered as an annotation processor explicitly otherwise it will not process java sources. This change applies the [lombok recommended method](https://projectlombok.org/setup/maven) of applying the annotation processor to fix builds on jdk23+

*Check all that applies:*

This doesnt effect generated sdk code, so I don't think any tests should be needed.

- [x] Linux
- [ ] Windows
- [ ] Android
- [ ] MacOS
- [ ] IOS
- [ ] Other Platforms

### DIFF
--- a/tools/code-generation/generator/pom.xml
+++ b/tools/code-generation/generator/pom.xml
@@ -11,6 +11,10 @@
     <name>AWS Client Generator</name>
     <url>http://aws.amazon.com</url>
 
+    <properties>
+        <lombok.version>1.18.42</lombok.version>
+    </properties>
+
     <build>
         <plugins>
             <plugin>
@@ -20,6 +24,13 @@
                 <configuration>
                     <source>1.8</source>
                     <target>1.8</target>
+                    <annotationProcessorPaths>
+                        <path>
+                            <groupId>org.projectlombok</groupId>
+                            <artifactId>lombok</artifactId>
+                            <version>${lombok.version}</version>
+                        </path>
+                    </annotationProcessorPaths>
                 </configuration>
             </plugin>
             <plugin>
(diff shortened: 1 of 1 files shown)


sbiscigl:
Why is this change needed? are you running the code generation standalone with a different JVM? the source in the SDK is the result of the running code gen on top of AWS models. curious how you are using the code gen where you would want this updated

apache-hb:
> are you running the code generation standalone with a different JVM?

I was building the sdk locally and wanted to test a change I made to the code generator, but found I couldn't since I only have jdk25 installed locally.

sbiscigl:
>and wanted to test a change I made to the code generator

what change did you make to the code generator and why? the code generator and code generation process is pretty internal to AWS and the models that we provide. What are you trying to accomplish about editing the codegen? I'm not against it, more so curious why you want to change something about it.

apache-hb:
I noticed that there's alot of padding between members in models because of `bool m_FieldIsSet` being placed after each field, i moved all the bools to the end of each model class and got between a 25-35% reduction in memory usage for each model class. I wanted to test it against the full sdk so needed to do a rebuild which is when I ran into this.

sbiscigl:
I actually love that change, let me give this PR a whirl in our CI and if it passes, will merge it. for adminsitrative reasons I have to open the PR, the commit will still be authored by you.

sbiscigl:
PR for reference - https://github.com/aws/aws-sdk-cpp/pull/3638

sbiscigl:
[merged to main now](https://github.com/aws/aws-sdk-cpp/commit/5d771d45d2760a54f02e5729b00b1c2b42fb465e), give a shout if you have any questions about code gen

Issue #3297, opened 2025-02-13
AWSAuthV4Signer::PayloadSigningPolicy::Never not work


zombee0:
### Describe the bug

from the code
https://github.com/aws/aws-sdk-cpp/blob/8209a3ed365edb4789d6eb9d5763c08f6bd77d11/src/aws-cpp-sdk-core/source/client/AWSClient.cpp#L561
I see the AWSAuthV4Signer::PayloadSigningPolicy::Never won't work any more, so is this a bug? or by designed?
And even in AWSAuthV4Signer
https://github.com/aws/aws-sdk-cpp/blob/8209a3ed365edb4789d6eb9d5763c08f6bd77d11/src/aws-cpp-sdk-core/source/auth/signer/AWSAuthV4Signer.cpp#L219
i see singBody has no effect
so how does AWSAuthV4Signer::PayloadSigningPolicy::Never work?

### Regression Issue

- [ ] Select this option if this issue appears to be a regression.

### Expected Behavior

No signBody

### Current Behavior

signBody

### Reproduction Steps

Any PutObject

### AWS CPP SDK version used

main

### Compiler and Version used

gcc 11.4.0

### Operating System and version

ubuntu

zombee0:
@sbiscigl thank you!
In my system, I have set AWSAuthV4Signer::PayloadSigningPolicy::Never, and we introduced a new HTTPClient that didn't support chunked upload, it works fine for the prior version, but not now, and now i know why.
and we may support chunked upload later in our HTTPClient, but do you think if you split chunked upload code from the HTTPClient that will be better?
In my opinion, chunked upload is more related to AWS, and it will be clearer that HTTPClient is only responsible for data transmission.

sbiscigl:
> but do you think if you split chunked upload code from the HTTPClient that will be better?
In my opinion, chunked upload is more related to AWS, and it will be clearer that HTTPClient is only responsible for data transmission.

yes you are right and that was a [issue in the original implementation](https://github.com/aws/aws-sdk-cpp/commit/01e61b3137f90509e71765ee7e3b4a39e2e8de91) where it was made http clients responsibility to handle chunked. we should only be giving http request stream of bytes, and well thats it, just a stream of bytes.

let me change this issue to a feature request and hopefully we can make that adjustment sooner than later

> it works fine for the prior version

[...]

PR #3635, opened 2025-12-02
Refactor AWS Chunked logic from HTTP clients

kai-ion:
*Issue #, if available:*
https://github.com/aws/aws-sdk-cpp/issues/3297

*Description of changes:*
Our goal is specifically to refactor this logic out of http clients and into a interceptor. We will introduce a core ChunkingInterceptor that owns all aws-chunked behavior for streaming requests and remove the chunking logics from the individual HTTP clients. 

*Check all that applies:*

- [x] Linux
- [x] Windows
- [x] Android
- [x] MacOS
- [ ] IOS
- [ ] Other Platforms

### DIFF
--- a/src/aws-cpp-sdk-core/include/aws/core/client/ClientConfiguration.h
+++ b/src/aws-cpp-sdk-core/include/aws/core/client/ClientConfiguration.h
@@ -78,6 +78,16 @@ namespace Aws
           WHEN_REQUIRED,
         };
 
+        /**
+         * Control HTTP client chunking implementation mode.
+         * DEFAULT: Use SDK's ChunkingInterceptor for aws-chunked encoding
+         * CLIENT_IMPLEMENTATION: Rely on HTTP client's native chunking (default for custom clients)
+         */
+        enum class HttpClientChunkedMode {
+          DEFAULT,
+          CLIENT_IMPLEMENTATION,
+        };
+
         struct RequestCompressionConfig {
           UseRequestCompression useRequestCompression=UseRequestCompression::ENABLE;
           size_t requestMinCompressionSizeBytes = 10240;
@@ -493,6 +503,12 @@ namespace Aws
            * https://docs.aws.amazon.com/sdkref/latest/guide/feature-account-endpoints.html
            */
           Aws::String accountIdEndpointMode = "preferred";
+
+          /**
+           * Control HTTP client chunking implementation mode.
+           * Default is set automatically: CLIENT_IMPLEMENTATION for custom clients, DEFAULT for AWS clients.
+           */
+          HttpClientChunkedMode httpClientChunkedMode = HttpClientChunkedMode::CLIENT_IMPLEMENTATION;
           /**
           * Configuration structure for credential providers in the AWS SDK.
           * This structure allows passing configuration options to credential providers
--- a/src/aws-cpp-sdk-core/include/smithy/client/AwsSmithyClientBase.h
+++ b/src/aws-cpp-sdk-core/include/smithy/client/AwsSmithyClientBase.h
@@ -9,6 +9,7 @@
 #include <smithy/tracing/TelemetryProvider.h>
 #include <smithy/interceptor/Interceptor.h>
 #include <smithy/client/features/ChecksumInterceptor.h>
+#include <smithy/client/features/ChunkingInterceptor.h>
 #include <smithy/client/features/UserAgentInterceptor.h>
 
 #include <aws/crt/Variant.h>
@@ -20,6 +21,7 @@
 #include <aws/core/utils/Outcome.h>
 #include <aws/core/NoResult.h>
 #include <aws/core/http/HttpClientFactory.h>
+#include <aws/core/http/HttpClient.h>
 #include <aws/core/client/AWSErrorMarshaller.h>
 #include <aws/core/AmazonWebServiceResult.h>
 #include <utility>
@@ -99,8 +101,13 @@ namespace client
           m_serviceUserAgentName(std::move(serviceUserAgentName)),
           m_httpClient(std::move(httpClient)),
           m_errorMarshaller(std::move(errorMarshaller)),
-          m_interceptors{Aws::MakeShared<ChecksumInterceptor>("AwsSmithyClientBase", *m_clientConfig)}
+          m_interceptors({
+              Aws::MakeShared<ChecksumInterceptor>("AwsSmithyClientBase", *m_clientConfig),
+              Aws::MakeShared<features::ChunkingInterceptor>("AwsSmithyClientBase", 
+                  m_httpClient->IsDefaultAwsHttpClient() ? Aws::Client::HttpClientChunkedMode::DEFAULT : m_clientConfig->httpClientChunkedMode)
+          })
         {
+            
             baseInit();
         }
 
--- a/src/aws-cpp-sdk-core/include/smithy/client/features/ChunkingInterceptor.h
+++ b/src/aws-cpp-sdk-core/include/smithy/client/features/ChunkingInterceptor.h
@@ -1,0 +1,225 @@
+/**
+ * Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.
+ * SPDX-License-Identifier: Apache-2.0.
+ */
+#pragma once
+#include <aws/core/http/HttpRequest.h>
+#include <aws/core/utils/Array.h>
+#include <aws/core/utils/StringUtils.h>
+#include <aws/core/utils/HashingUtils.h>
+#include <aws/core/utils/logging/LogMacros.h>
+#include <aws/core/utils/memory/stl/AWSStringStream.h>
+#include <aws/core/utils/memory/stl/AWSVector.h>
+#include <smithy/interceptor/Interceptor.h>
+#include <aws/core/client/ClientConfiguration.h>
+#include <aws/core/utils/Outcome.h>
+#include <aws/core/client/AWSError.h>
+#include <memory>
+
+namespace smithy {
+namespace client {
+namespace features {
+
+static const size_t AWS_DATA_BUFFER_SIZE = 65536;
+static const char* ALLOCATION_TAG = "ChunkingInterceptor";
+static const char* CHECKSUM_HEADER_PREFIX = "x-amz-checksum-";
+
+template <size_t DataBufferSize = AWS_DATA_BUFFER_SIZE>
+class AwsChunkedStreamBuf : public std::streambuf {
+public:
+    AwsChunkedStreamBuf(Aws::Http::HttpRequest* request,
+                        const std::shared_ptr<Aws::IOStream>& stream,
+                        size_t bufferSize = DataBufferSize)
+        : m_request(request),
+          m_stream(stream),
+          m_data(bufferSize)
+    {
+        assert(m_stream != nullptr);
+        if (m_stream == nullptr) {
+            AWS_LOGSTREAM_ERROR("AwsChunkedStream", "stream is null");
+        }
+        assert(m_request != nullptr);
+        if (m_request == nullptr) {
+            AWS_LOGSTREAM_ERROR("AwsChunkedStream", "request is null");
+        }
+    }
+
+protected:
+    int_type underflow() override {
+        if (gptr() && gptr() < egptr()) {
+            return traits_type::to_int_type(*gptr());
+        }
+
+        // only read and write to chunked stream if the underlying stream
+        // is still in a valid state and we have buffer space
+        if (m_stream->good() && m_chunkingBufferPos >= m_chunkingBufferSize) {
+            // Reset buffer for new data only when buffer is consumed
+            m_chunkingBufferPos = 0;
+            m_chunkingBufferSize = 0;
+            
+            // Check if we have enough space for worst-case chunk (data + header + footer)
+            size_t maxChunkSize = m_data.GetLength() + 20; // data + hex header + CRLF
+            if (m_chunkingBufferSize + maxChunkSize <= m_chunkingBuffer.GetLength()) {
+                // Try to read in a 64K chunk, if we cant we know the stream is over
+                m_stream->read(m_data.GetUnderlyingData(), m_data.GetLength());
+                size_t bytesRead = static_cast<size_t>(m_stream->gcount());
+                writeChunk(bytesRead);
+
+                // if we've read everything from the stream, we want to add the trailer
+                // to the underlying stream
+                if ((m_stream->peek() == EOF || m_stream->eof()) && !m_stream->bad()) {
+                    writeTrailerToUnderlyingStream();
+                }
+            }
+        }
+
+        // if the chunking buffer is empty there is nothing to read
+        if (m_chunkingBufferPos >= m_chunkingBufferSize) {
+            return traits_type::eof();
+        }
+
+        // Set up buffer pointers to read from chunking buffer
+        size_t remainingBytes = m_chunkingBufferSize - m_chunkingBufferPos;
+        size_t bytesToRead = std::min(remainingBytes, DataBufferSize);
+        
+        setg(m_chunkingBuffer.GetUnderlyingData() + m_chunkingBufferPos, 
+             m_chunkingBuffer.GetUnderlyingData() + m_chunkingBufferPos, 
+             m_chunkingBuffer.GetUnderlyingData() + m_chunkingBufferPos + bytesToRead);
+        
+        m_chunkingBufferPos += bytesToRead;
+        
+        return traits_type::to_int_type(*gptr());
+    }
+
+private:
+    void writeTrailerToUnderlyingStream() {
+        Aws::String trailer = "0\r\n";
+        if (m_request->GetRequestHash().second != nullptr) {
+            trailer += "x-amz-checksum-" + m_request->GetRequestHash().first + ":"
+                     + Aws::Utils::HashingUtils::Base64Encode(m_request->GetRequestHash().second->GetHash().GetResult()) + "\r\n";
+        }
+        trailer += "\r\n";
+        if (m_chunkingBufferSize + trailer.length() <= m_chunkingBuffer.GetLength()) {
+            std::memcpy(m_chunkingBuffer.GetUnderlyingData() + m_chunkingBufferSize, trailer.c_str(), trailer.length());
+            m_chunkingBufferSize += trailer.length();
+        }
+    }
+
+    void writeChunk(size_t bytesRead) {
+        if (m_request->GetRequestHash().second != nullptr) {
+            m_request->GetRequestHash().second->Update(reinterpret_cast<unsigned char*>(m_data.GetUnderlyingData()), bytesRead);
+        }
+
+        if (bytesRead > 0) {
+            Aws::String chunkHeader = Aws::Utils::StringUtils::ToHexString(bytesRead) + "\r\n";
+            size_t totalSize = chunkHeader.length() + bytesRead + 2;
+            if (m_chunkingBufferSize + totalSize <= m_chunkingBuffer.GetLength()) {
+                std::memcpy(m_chunkingBuffer.GetUnderlyingData() + m_chunkingBufferSize, chunkHeader.c_str(), chunkHeader.length());
+                m_chunkingBufferSize += chunkHeader.length();
+                std::memcpy(m_chunkingBuffer.GetUnderlyingData() + m_chunkingBufferSize, m_data.GetUnderlyingData(), bytesRead);
+                m_chunkingBufferSize += bytesRead;
+                std::memcpy(m_chunkingBuffer.GetUnderlyingData() + m_chunkingBufferSize, "\r\n", 2);
+                m_chunkingBufferSize += 2;
+            }
+        }
+    }
+
+    // Buffer for chunked data plus overhead for HTTP chunked encoding headers, trailers, and safety margin
+    Aws::Utils::Array<char> m_chunkingBuffer{DataBufferSize + 128};
+    size_t m_chunkingBufferSize{0};
+    size_t m_chunkingBufferPos{0};
+    Aws::Http::HttpRequest* m_request{nullptr};
+    std::shared_ptr<Aws::IOStream> m_stream;
+    Aws::Utils::Array<char> m_data;
+};
+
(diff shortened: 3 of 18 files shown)


src/aws-cpp-sdk-core/include/aws/core/http/interceptor/ChunkingInterceptor.h
sbiscigl: just use `ClientConfiguration` as a parameter, it saves us down the road from creating a new constructor when we decide we want something else

sbiscigl: since we accept `HttpRequest` as a pointer, we can just forward declare it

src/aws-cpp-sdk-core/source/auth/signer/AWSAuthV4Signer.cpp
sbiscigl: since all of this has to do with checksums and chunked encoding, can we move all of this logic to the interceptors?

src/aws-cpp-sdk-core/source/http/curl/CurlHttpClient.cpp
sbiscigl, line 383: isnt this still true? Also since this is a static, in implementation, and [un-used in the project](https://github.com/search?q=repo%3Aaws%2Faws-sdk-cpp%20SeekBody&type=code), should we just remove `SeekBody` all together?

src/aws-cpp-sdk-core/source/http/interceptor/ChunkingInterceptor.cpp
sbiscigl: why does this need inherit from `Aws:IOStream`? seems to me like `class ChunkedStreamBuf : public std::streambuf` is all that is needed, and everything go in that class.

sbiscigl: lets move all string constants into constants in anonymous namespace

src/aws-cpp-sdk-core/CMakeLists.txt
sbiscigl: this brings up two questions for me:
* [why isnt this in the same place as the other interceptors](https://github.com/aws/aws-sdk-cpp/tree/main/src/aws-cpp-sdk-core/include/smithy/client/features)
* [why isnt this added in the smithy base clients](https://github.com/aws/aws-sdk-cpp/blob/main/src/aws-cpp-sdk-core/include/smithy/client/AwsSmithyClientBase.h#L102)

src/aws-cpp-sdk-core/include/aws/core/client/ClientConfiguration.h
sbiscigl: two points here:
* how will this be disabled for custom http clients by default? I think that was what we discussed about how we wanted that to be by default and opt in.
* this is sort of confusing for customers as this sounds like you are enabling it at the client level. consider renaming to `httpClientChunkedMode`. where the enum values are `DEFAULT`  and `CLIENT_IMPLEMENTATION`. as this is more clear that the configuration turns on/off whether or not the implementation is expected to have a chunked implementation,

src/aws-cpp-sdk-core/source/http/interceptor/ChunkingInterceptor.cpp
sbiscigl: > &m_streambuf

is `m_streambuf` initialized at this point? should we be taking a reference of it?

src/aws-cpp-sdk-core/source/http/curl/CurlHttpClient.cpp
kai-ion: Its still being used as a call back function [here](https://github.com/aws/aws-sdk-cpp/blob/324bd2936a4fc78d8195db183126412dc334cf45/src/aws-cpp-sdk-core/source/http/curl/CurlHttpClient.cpp#L819)

src/aws-cpp-sdk-core/include/aws/core/http/interceptor/ChunkingInterceptor.h
kai-ion: https://github.com/aws/aws-sdk-cpp/blob/b4aaab3b802e948d162eb570cf4cb7b75a003492/src/aws-cpp-sdk-core/include/smithy/client/features/ChunkingInterceptor.h#L58-L62
Update it to use ClientConfiguration as a parameter

kai-ion: In the comment below I see that all the other interceptor are header files only, so changed my interceptor to header only as well

src/aws-cpp-sdk-core/source/auth/signer/AWSAuthV4Signer.cpp
kai-ion: Refactored all the logic from signer to interceptor

src/aws-cpp-sdk-core/source/http/interceptor/ChunkingInterceptor.cpp
kai-ion: Ah, understood, changed the inheritance to from `std::streambuf`

kai-ion: moved the string constants into the anonymous namespace:

CHECKSUM_HEADER_PREFIX = "x-amz-checksum-" 

ALLOCATION_TAG = "ChunkingInterceptor"

src/aws-cpp-sdk-core/CMakeLists.txt
kai-ion: Yes, I see how the other interceptors are constructed now. Moved the interceptor to the correct place and added it to the smithy base client

src/aws-cpp-sdk-core/include/aws/core/client/ClientConfiguration.h
kai-ion: Implemented a check to see if custom http clients are used to turn on/off httpClientChunkedMode
```
if (clientConfig.httpLibOverride == Aws::Http::TransferLibType::DEFAULT_CLIENT) {
        clientConfig.httpClientChunkedMode = HttpClientChunkedMode::CLIENT_IMPLEMENTATION;
    } else {
        clientConfig.httpClientChunkedMode = HttpClientChunkedMode::DEFAULT;
    }
```

src/aws-cpp-sdk-core/source/http/interceptor/ChunkingInterceptor.cpp
kai-ion: Ah yes, realized the mistake of initializing the streambuf before `m_chunkStream` is initialized.

src/aws-cpp-sdk-core/include/smithy/client/features/ChunkingInterceptor.h
sbiscigl: why are we calling setg in the constructor?, specifically to nullptr?

sbiscigl: why is this buffer 8KB? why is this buffer a c array and not a c++ array? why not make this configurable? 

`AwsChunkedStream` has a [buffer of 64KB](https://github.com/aws/aws-sdk-cpp/blob/main/src/aws-cpp-sdk-core/include/aws/core/utils/stream/AwsChunkedStream.h#L15-L18) for double encoding, why are creating a second buffer to buffer that buffer into?

If the answer is "because i need to use the AwsChunkedStream logic to do what i need it to do" then we need to move the logic from chunked stream into this class to avoid a extra buffer, or add additionally APIs to the existing class.

sbiscigl: why do you need a default constructor?

sbiscigl: why do you have this as the last step and not the first?

src/aws-cpp-sdk-core/include/smithy/client/AwsSmithyClientBase.h
sbiscigl: why do you call emplace back instead of using the initializer list?

src/aws-cpp-sdk-core/source/http/curl/CurlHttpClient.cpp
sbiscigl: github search failing then i guess, then i guess my statement still stands, dont we need this? is a chunked stream seekable?

src/aws-cpp-sdk-core/include/smithy/client/features/ChunkingInterceptor.h
sbiscigl, line 205: I really dont like this but for backwards compatibility sake we have to let this continue. We _cannot_ tie checksumming to chunked encoded, something that has a checksum doesnt need to be chunk encoded or vice-versa. the model should allow you say whether or not something should be. but for now we can continue in backward compat way.

src/aws-cpp-sdk-core/source/client/ClientConfiguration.cpp
sbiscigl: consider the following code
```C++
auto main() -> int {
  SDKOptions options{};
  options.httpOptions.httpClientFactory_create_fn = 
    []() -> std::shared_ptr<Aws::Http::HttpClientFactory> {
      Aws::MakeShared<MyHttpClientFactory>();
    };
  SdkContext context{std::move(options)};

  //.. sdk code
  ClientConfiguration configuration{};
  S3Client client{configuration};
  return 0;
}
```

would this operate correctly? or would the custom http client use the correct default?

src/aws-cpp-sdk-core/include/smithy/client/features/ChunkingInterceptor.h
kai-ion: Its to initialize the streambuf as an empty state. 

Its also done the same way in the code snipet you showed me 

```
class chunked_stream : public std::streambuf {
 public:
  chunked_stream(std::unique_ptr<std::istream> is) : source(std::move(is)) {
    setg(nullptr, nullptr, nullptr);
  }

 protected:
  int_type underflow() override {
    if (gptr() == egptr()) {
      if (!trailer.empty()) {
        buffer = trailer[trailer_pos++];
        if (trailer_pos == trailer.size()) {
          trailer.clear();
          trailer_pos = 0;
        }
        setg(&buffer, &buffer, &buffer + 1);
        return traits_type::to_int_type(*gptr());
      }
      int ch = source->get();
      if (ch == traits_type::eof()) {
        return traits_type::eof();
      }
      buffer = static_cast<char>(ch);
      char_count++;
      if (char_count % 10 == 0) {
        trailer = "<trailer>";
        trailer_pos = 0;
      }
      setg(&buffer, &buffer, &buffer + 1);
    }
    return traits_type::to_int_type(*gptr());
  }

 private:
  std::unique_ptr<std::istream> source;
  char buffer;
  std::string trailer;
  size_t trailer_pos = 0;
  int char_count = 0;
};
```

kai-ion: Ah yes you're right, this was originally setup just because we needed a streambuff interface. I think extending AwsChunkedStream API to add a streamBuff interface makes sense.

kai-ion: https://github.com/aws/aws-sdk-cpp/blob/8a4f5e9775a0d5193122f56702c924395f24957e/src/aws-cpp-sdk-core/include/smithy/client/features/ChecksumInterceptor.h#L48
The checksum interceptor also have a default constructor. But I think we actually dont need this default constructor here since we have the explicit clientconfig constructor

kai-ion: Okay moving this up

src/aws-cpp-sdk-core/include/smithy/client/AwsSmithyClientBase.h
kai-ion: Oh yeah, this is redundant logic, changing it to an initializer list

src/aws-cpp-sdk-core/source/http/curl/CurlHttpClient.cpp
kai-ion: Chunkstream are not seekable. adding this logic back

src/aws-cpp-sdk-core/include/smithy/client/features/ChunkingInterceptor.h
sbiscigl: the base constructor already sets them to null, so why are we doing it again?

src/aws-cpp-sdk-core/include/smithy/client/features/ChunkingInterceptor.h
```diff
@@ -0,0 +1,213 @@
...
+            m_chunkingStream->write(m_data.GetUnderlyingData(), bytesRead);
+            *m_chunkingStream << "\r\n";
+        }
+    }
+
+    std::shared_ptr<Aws::IOStream> m_chunkingStream;
+    Aws::Http::HttpRequest* m_request{nullptr};
+    std::shared_ptr<Aws::IOStream> m_stream;
+    Aws::Utils::Array<char> m_data;
+    Aws::Utils::Array<char> m_buffer{DataBufferSize};
```
sbiscigl: This effectively doubles the size of this class by adding another buffer, it seems the only reason `m_buffer` exists is so that we can access the underlying data in `m_chunkingStream`. how can we access the data in `m_chunkingStream` without creating another buffer? seems to me like `m_chunkingStream` should be a different type, perhaps a `Aws::Vector` or a `Aws::Utils::Array<char>` so that you can access the underlying data. You will however change how you add and read from it.

src/aws-cpp-sdk-core/include/smithy/client/AwsSmithyClientBase.h
```diff
@@ -98,9 +100,19 @@ namespace client
           m_serviceName(std::move(serviceName)),
           m_serviceUserAgentName(std::move(serviceUserAgentName)),
           m_httpClient(std::move(httpClient)),
-          m_errorMarshaller(std::move(errorMarshaller)),
-          m_interceptors{Aws::MakeShared<ChecksumInterceptor>("AwsSmithyClientBase", *m_clientConfig)}
+          m_errorMarshaller(std::move(errorMarshaller))
         {
+            // Create modified config for chunking interceptor
+            Aws::Client::ClientConfiguration chunkingConfig(*m_clientConfig);
```
sbiscigl: why do you construct a object instead of referring to `m_clientConfig` directly?

src/aws-cpp-sdk-core/include/smithy/client/AwsSmithyClientBase.h
```diff
@@ -98,9 +100,19 @@ namespace client
...
+            if (!m_httpClient->IsDefaultAwsHttpClient()) {
+                chunkingConfig.httpClientChunkedMode = Aws::Client::HttpClientChunkedMode::CLIENT_IMPLEMENTATION;
+            }
+            
+            m_interceptors = {
```
sbiscigl: why is this being moved out of the initializer list? seems like it should stay in the initailizer list, and you should construct the `ChunkingInterceptor` inplace, perhaps using IIFE experession

src/aws-cpp-sdk-core/source/smithy/client/AwsSmithyClientBase.cpp
```diff
@@ -102,7 +103,17 @@ void AwsSmithyClientBase::baseCopyAssign(const AwsSmithyClientBase& other,
   m_serviceUserAgentName = other.m_serviceUserAgentName;
   m_httpClient = std::move(httpClient);
   m_errorMarshaller = std::move(errorMarshaller);
-  m_interceptors = Aws::Vector<std::shared_ptr<interceptor::Interceptor>>{Aws::MakeShared<ChecksumInterceptor>("AwsSmithyClientBase")};
+  
+  // Create modified config for chunking interceptor
+  Aws::Client::ClientConfiguration chunkingConfig(*m_clientConfig);
```
sbiscigl: same comment as the other class, why create a temporary object, why not just inline the creation of the interceptor

src/aws-cpp-sdk-core/include/smithy/client/AwsSmithyClientBase.h
kai-ion: I was thinking we should not mutate the user's `clientConfig` directly

src/aws-cpp-sdk-core/include/smithy/client/features/ChunkingInterceptor.h
kai-ion: removed

src/aws-cpp-sdk-core/include/smithy/client/features/ChunkingInterceptor.h
kai-ion: I attempted to use `accessible_stream_buf` but it was more complicated then initially thought. (Failing intergration test) 
So I decided to just change the `m_chunkingStream` to a vector to access the underlying data.

src/aws-cpp-sdk-core/include/smithy/client/features/ChunkingInterceptor.h
```diff
@@ -0,0 +1,206 @@
...
+
+        // Set up buffer pointers to read from chunking buffer
+        size_t remainingBytes = m_chunkingBuffer.size() - m_chunkingBufferPos;
+        size_t bytesToRead = std::min(remainingBytes, DataBufferSize);
+        
+        setg(m_chunkingBuffer.data() + m_chunkingBufferPos, 
+             m_chunkingBuffer.data() + m_chunkingBufferPos, 
+             m_chunkingBuffer.data() + m_chunkingBufferPos + bytesToRead);
+        
+        m_chunkingBufferPos += bytesToRead;
```
sbiscigl, line 89: is data ever erased from the vector? otherwise this vector would grow to the size of the body right?

src/aws-cpp-sdk-core/include/smithy/client/AwsSmithyClientBase.h
```diff
@@ -99,8 +101,16 @@ namespace client
           m_serviceUserAgentName(std::move(serviceUserAgentName)),
           m_httpClient(std::move(httpClient)),
           m_errorMarshaller(std::move(errorMarshaller)),
-          m_interceptors{Aws::MakeShared<ChecksumInterceptor>("AwsSmithyClientBase", *m_clientConfig)}
+          m_interceptors({
+              Aws::MakeShared<ChecksumInterceptor>("AwsSmithyClientBase", *m_clientConfig),
+              Aws::MakeShared<features::ChunkingInterceptor>("AwsSmithyClientBase", [this]() {
+                  Aws::Client::ClientConfiguration chunkingConfig = *m_clientConfig;
+                  chunkingConfig.httpClientChunkedMode = m_httpClient->IsDefaultAwsHttpClient() ? m_clientConfig->httpClientChunkedMode : Aws::Client::HttpClientChunkedMode::CLIENT_IMPLEMENTATION;
```
sbaluja: Is this the behavior we want? If it's our implementation (`IsDefaultAwsHttpClient() == true`) then don't we want to force `Aws::Client::HttpClientChunkedMode::DEFAULT` so that the chunking is applied? And if it's not our implementation then we follow whatever the customer has set, ie `m_clientConfig->httpClientChunkedMode`. Seems to me this condition should be reversed:

```
chunkingConfig.httpClientChunkedMode = m_httpClient->IsDefaultAwsHttpClient() ?  Aws::Client::HttpClientChunkedMode::DEFAULT : m_clientConfig->httpClientChunkedMode;
```

I think we should default initialize `clientConfig.httpClientChunkedMode = HttpClientChunkedMode::CLIENT_IMPLEMENTATION;` so the client can _opt-in_ for their HttpClient implementation if they want to, but for our implementation we should ensure it's going to use chunking

src/aws-cpp-sdk-core/source/client/ClientConfiguration.cpp
```diff
@@ -220,6 +220,10 @@ void setLegacyClientConfigurationParameters(ClientConfiguration& clientConfig)
     clientConfig.writeRateLimiter = nullptr;
     clientConfig.readRateLimiter = nullptr;
     clientConfig.httpLibOverride = Aws::Http::TransferLibType::DEFAULT_CLIENT;
+
+    // Users can explicitly set CLIENT_IMPLEMENTATION if their custom client handles chunking
+    clientConfig.httpClientChunkedMode = HttpClientChunkedMode::DEFAULT;
```
sbaluja: This needs to be 
```
clientConfig.httpClientChunkedMode = HttpClientChunkedMode::CLIENT_IMPLEMENTATION;
```
We don't want to change existing client's HttpClient behavior unless they opt-in to our chunking logic.

src/aws-cpp-sdk-core/include/smithy/client/AwsSmithyClientBase.h
```diff
@@ -99,8 +101,17 @@ namespace client
           m_serviceUserAgentName(std::move(serviceUserAgentName)),
           m_httpClient(std::move(httpClient)),
           m_errorMarshaller(std::move(errorMarshaller)),
-          m_interceptors{Aws::MakeShared<ChecksumInterceptor>("AwsSmithyClientBase", *m_clientConfig)}
+          m_interceptors({
+              Aws::MakeShared<ChecksumInterceptor>("AwsSmithyClientBase", *m_clientConfig),
+              Aws::MakeShared<features::ChunkingInterceptor>("AwsSmithyClientBase", [this]() {
+                  Aws::Client::ClientConfiguration chunkingConfig = *m_clientConfig;
```
sbiscigl: ok all of this exists because of my comment that we should be passing the whole client configuration, i'll admit i was wrong, lets use the enum value, that makes this all together better then you dont need any of this and a simple constructor call to  

`m_httpClient->IsDefaultAwsHttpClient() ? Aws::Client::HttpClientChunkedMode::DEFAULT : m_clientConfig->httpClientChunkedMode`

will suffice, lets do that again, thats simpler and makes the better, sorry for the shuffling around there, lets do that though, that will make this better

sbiscigl (approved):
alright its in a good place now, please fix the constructor and do some memory tests and im good to ship

sbiscigl (approved):

PR #3641, opened 2025-12-04
Implementing checksum validation for single-part and multipart downloads

pulimsr:
*Description of changes:* 
Implemented per-part checksum validation for downloads in TransferManager. Added validateChecksums flag (default: true) to TransferManagerConfiguration. Downloads now validate checksums for both single-part (via file read after download) and multipart transfers (per-part validation). Created GetChecksumFromResult() helper to reduce code duplication. Downloads fail immediately on checksum mismatch.                                                                                                                                                                                                                                                 
                                                                             

*Check all that applies:*

- [x] Linux
- [x] Windows
- [ ] Android
- [x] MacOS
- [ ] IOS
- [ ] Other Platforms



src/aws-cpp-sdk-transfer/source/transfer/TransferManager.cpp
```diff
@@ -1202,6 +1280,48 @@ namespace Aws
                     Aws::IOStream* bufferStream = partState->GetDownloadPartStream();
                     assert(bufferStream);
 
+                    // Calculate and validate checksum for this part if validation is enabled
+                    if (m_transferConfig.validateChecksums)
+                    {
+                        auto hash = handle->GetPartChecksum(partState->GetPartId());
+                        if (hash && partState->GetDownloadBuffer())
+                        {
+                            hash->Update(partState->GetDownloadBuffer(), static_cast<size_t>(partState->GetSizeInBytes()));
```
sbiscigl: lets not do this, in the get object we already validate the checksum of the body i think the idea here is more so the compare the checksum returned from get object, combined across parts, to make sure that the combined object has the same full object checksum.

pulimsr: Alright

src/aws-cpp-sdk-transfer/source/transfer/TransferManager.cpp
```diff
@@ -1147,6 +1162,51 @@ namespace Aws
                 handle->UpdateStatus(DetermineIfFailedOrCanceled(*handle));
                 TriggerTransferStatusUpdatedCallback(handle);
             }
+
+            if (handle->GetPendingParts().empty() && m_transferConfig.validateChecksums && !handle->GetChecksum().empty()) {
```
sbiscigl: lets add at least one unit tests around this logic

src/aws-cpp-sdk-transfer/source/transfer/TransferManager.cpp
```diff
@@ -1147,6 +1162,51 @@ namespace Aws
...
+                uint64_t combinedChecksum = 0;
```
sbiscigl: i think we want this logic to go [here](https://github.com/aws/aws-sdk-cpp/blob/main/src/aws-cpp-sdk-transfer/source/transfer/TransferManager.cpp#L1238-L1250) when we change the download status to complete, we can also change it to a error if it fails.

src/aws-cpp-sdk-transfer/source/transfer/TransferManager.cpp
```diff
@@ -1204,6 +1222,19 @@ namespace Aws
 
                     Aws::String errMsg{handle->WritePartToDownloadStream(bufferStream, partState->GetRangeBegin())};
                     if (errMsg.empty()) {
+                        if (!outcome.GetResult().GetChecksumCRC32().empty()) {
+                            partState->SetChecksum(outcome.GetResult().GetChecksumCRC32());
```
sbiscigl: do we still need this part now? we are double reading the stream as required now?

tests/aws-cpp-sdk-transfer-unit-tests/TransferUnitTests.cpp
```diff
@@ -1,13 +1,23 @@
+#ifdef _WIN32
```
sbiscigl: why do we need this? feels like this test shoudlnt need to know if its on windows or not

tests/aws-cpp-sdk-transfer-unit-tests/TransferUnitTests.cpp
```diff
@@ -23,24 +33,100 @@ class MockS3Client : public S3Client {
...
+    Aws::String FULL_OBJECT_CHECKSUM; //"SBi/K+1ooBg="
+    MockMultipartS3Client(Aws::String expected_checksum) : S3Client() {
+        FULL_OBJECT_CHECKSUM = expected_checksum;
+    };
+
+    HeadObjectOutcome HeadObject(const HeadObjectRequest&) const override {
+        HeadObjectResult result;
+        result.SetContentLength(78643200);
+        result.SetChecksumCRC64NVME(FULL_OBJECT_CHECKSUM);
+        result.SetChecksumType(Aws::S3::Model::ChecksumType::FULL_OBJECT);  // This is key!
```
sbiscigl: > // This is key!

Lets improve the comment or remove it

tests/aws-cpp-sdk-transfer-unit-tests/TransferUnitTests.cpp
```diff
@@ -62,14 +149,90 @@ TEST_F(TransferUnitTest, ContentValidationShouldFail) {
...
     EXPECT_EQ(TransferStatus::FAILED, handle->GetStatus());
 }
+
+TEST_F(TransferUnitTest, MultipartDownloadTest) {
+    TransferManagerConfiguration config(executor.get());
+    config.s3Client = mockMultipartS3Client;
+    config.bufferSize = 5242880;  // 5MB to ensure multipart
+    auto transferManager = TransferManager::Create(config);
+
+    // Create a temporary file for download since multipart needs seekable stream
```
sbiscigl: oh we have a type for a "temp file" so you can remove this all together, see [`TempFile`](https://github.com/aws/aws-sdk-cpp/blob/main/src/aws-cpp-sdk-core/include/aws/core/utils/FileSystemUtils.h#L31-L53) for usage. [We use it in tests in loads of places](https://github.com/aws/aws-sdk-cpp/blob/main/tests/aws-cpp-sdk-core-integration-tests/STSWebIdentityProviderIntegrationTest.cpp#L134).

i think you should be able to get rid of your windows specific paths if you use that instead.

src/aws-cpp-sdk-transfer/source/transfer/TransferManager.cpp
pulimsr: I put this back to avoid reading the stream for unit tests. Is there a cleaner approach?

pulimsr: I've updated the tests to handle this properly, so I was able to remove this code.

PR #3642, opened 2025-12-05
Place isSet members at the bottom of model classes to reduce class size

apache-hb:
*Description of changes:*

Moves m_IsSet members to the bottom of generated models so they're more closely packed, which reduces the size of models. As a random sample `Aws::RDS::Model::DBCluster` is currently 2864 bytes, after regenerating with this patch it is 2368 bytes. Which is a saving of 18%. This doesn't affect any APIs but will be an ABI change.

*Check all that applies:*

I ran the usual generated tests and they all continued to pass.

- [x] Linux
- [ ] Windows
- [ ] Android
- [ ] MacOS
- [ ] IOS
- [ ] Other Platforms

### DIFF
--- a/tools/code-generation/generator/src/main/resources/com/amazonaws/util/awsclientgenerator/velocity/cpp/ModelClassMembersAndInlines.vm
+++ b/tools/code-generation/generator/src/main/resources/com/amazonaws/util/awsclientgenerator/velocity/cpp/ModelClassMembersAndInlines.vm
@@ -155,3 +155,14 @@
 
 #end
 #end##if($shape.members.size() > 0)
+#foreach($member in $shape.members.entrySet())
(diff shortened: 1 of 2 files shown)


apache-hb:
I didn't regenerate any classes as part of this changeset as the diff was massive. If it would be desired for this to include regenerated model classes please let me know and I'll add them to this changeset.

sbiscigl:
> I didn't regenerate any classes as part of this changeset as the diff was massive.

i actually prefer this, it makes it easier to digest, we can generate the clients on our end before releasing

sbiscigl:
https://github.com/aws/aws-sdk-cpp/pull/3644 has been merged, closing this out. and thank you for this contribution, its really really quite great

Issue #3464, opened 2025-06-24
Undefined behavior in ClientWithAsyncTemplateMethods constructor


pitrou:
### Describe the bug

`ClientWithAsyncTemplateMethods` has the following constructor:
https://github.com/aws/aws-sdk-cpp/blob/08f5335f43eabe4a3d7fb5454c655e7b86630b4d/src/aws-cpp-sdk-core/include/aws/core/client/AWSClientAsyncCRTP.h#L47-L56

The `static_cast<AwsServiceClientT*>(this)` is meant to cast the base `this` pointer to the corresponding derived `this` pointer (which can have a different address when multiple inheritance is involved).

The problem, though, is that it happens in the base class *constructor*, so the derived object is not yet constructed yet. And, therefore, the `static_cast` to the derived `this` is undefined behavior.

This is flagged by some of the UBSan builds in Apache Arrow:
https://github.com/apache/arrow/issues/46863
```
Start test: S3FileSystem
/var/folders/x7/ch5v91h56_zbvbd1y2f600dm0000gn/T/RtmpjRMkwO/file8951246f6d2e/_deps/aws-sdk-cpp-src/src/aws-cpp-sdk-core/include/aws/core/client/AWSClientAsyncCRTP.h:51:40: runtime error: downcast of address 0x61b00011cd98 which does not point to an object of type 'Aws::S3::S3Client'
0x61b00011cd98: note: object is of type 'Aws::Client::AWSXMLClient'
 00 00 00 00  b0 eb b9 28 01 00 00 00  75 73 2d 65 61 73 74 2d  31 00 e6 8b 01 00 00 00  00 00 00 00
              ^~~~~~~~~~~~~~~~~~~~~~~
              vptr for 'Aws::Client::AWSXMLClient'
    #0 0x0001245e7770 in Aws::Client::ClientWithAsyncTemplateMethods<Aws::S3::S3Client>::ClientWithAsyncTemplateMethods()+0x1f0 (arrow.so:arm64+0x701f770)
[...]

PR #3645, opened 2025-12-09
Add build flag to enable the Undefined Behavior Sanitizer

bjosv:
The flag can be enabled in CMake using `-DENABLE_UB_SANITIZER=ON` similar to the existing flag for the address sanitizer.
Building aws-sdk-cpp (and its dependencies) with this flag indicates a couple of issues, like:
```
aws-sdk-cpp/src/aws-cpp-sdk-core/include/aws/core/client/AWSClientAsyncCRTP.h:51:40: runtime error: downcast of address 0x7ffe7bff5b90 which does not point to an object of type 'Aws::S3::S3Client'
aws-sdk-cpp/src/aws-cpp-sdk-core/include/aws/core/utils/stream/StreamBufProtectedWriter.h:46:67: runtime error: downcast of address 0x7ffd80fa21e8 which does not point to an object of type 'StreamBufProtectedWriter'
aws-sdk-cpp/crt/aws-crt-cpp/crt/aws-c-common/source/hash_table.c:68:12: runtime error: call to function aws_byte_cursor_eq through pointer to incorrect function type 'bool (*)(const void *, const void *)'
aws-sdk-cpp/crt/aws-crt-cpp/crt/aws-c-sdkutils/source/endpoints_util.c:50:24: runtime error: addition of unsigned offset to 0x595033296ca0 overflowed to 0x595033296c9f
aws-sdk-cpp/tests/aws-cpp-sdk-core-tests/utils/crypto/SymmetricCiphersTest.cpp:689:12: runtime error: null pointer passed as argument 1, which is declared to never be null
aws-sdk-cpp/tests/aws-cpp-sdk-core-tests/utils/crypto/SymmetricCiphersTest.cpp:689:43: runtime error: null pointer passed as argument 2, which is declared to never be null
```
Existing PRs to fix above issues are: https://github.com/awslabs/aws-c-http/pull/531 and https://github.com/awslabs/aws-c-sdkutils/pull/58
Related issue for one issue: https://github.com/aws/aws-sdk-cpp/issues/3464

Some of these issues are also seen when a SDK-user builds with UBSan.
Maybe we should add a CI-step building with this flag?

*Check all that applies:*
- [ ] Added proper tests to cover this PR. (Can be added to CI?)

- [x] Linux
- [ ] Windows
- [ ] Android
- [ ] MacOS
- [ ] IOS
- [ ] Other Platforms


sbiscigl:
So i have two high level thoughts on this

I know `ENABLE_ADDRESS_SANITIZER` exists and this would just be adding a mirror to that. However we've been trying to move away from making a cmake option for everything and only making knobs for things that cannot be provided to cmake as a argument. for instance instead of creating this flag one could just use

```
cmake -DCMAKE_CXX_FLAGS="-ggdb -fsanitize=undefined"
```

when running cmake generation instead of relying on a option we define ourselves. infact in our CI step we actually use address sanatizer that way instead of providing the [cmake option directly](https://github.com/aws/aws-sdk-cpp/blob/main/tools/scripts/build-tests/build-al2-debug-default.sh#L25).

> Maybe we should add a CI-step building with this flag?

I whole heartily agree on this sentiment, but until your PRs are merged and the other issue dealt with the build will actually fail so i would be against adding a flag until its passing.


so summarizing:
* i would prefer to stop the adding of specific cmake options for things that can be provided at generation time.
* we definitely need to add a CI step. likely just updating this [argument in our CI scripts](https://github.com/aws/aws-sdk-cpp/blob/main/tools/scripts/build-tests/build-al2-debug-default.sh#L25) once your changes are merged.

bjosv:
> * i would prefer to stop the adding of specific cmake options for things that can be provided at generation time.

Sounds reasonable. The problem I faced was how to get the same flags in crt dependencies when building with the default `BUILD_DEPS=ON`. We want to enable UBSan on dependencies to avoid false positives.
This was solved in this PR by using cached CMake variables, but maybe there is some other way..  I will take a look.

PR #3658, opened 2025-12-17
Add cstdint #include for uint64_t

tzik:
*Description of changes:*
`aws/core/utils/crypto/SecureRandom.h` uses `uint64_t` without including `cstdint`, that causes a compilation error on gcc-15.
This PR is adding the `#include` to fix the compile error.

*Check all that applies:*
  - I believe this PR is simple enough to skip the test.

- [x] Linux
- [ ] Windows
- [ ] Android
- [ ] MacOS
- [ ] IOS
- [ ] Other Platforms

### DIFF
--- a/src/aws-cpp-sdk-core/include/aws/core/utils/crypto/SecureRandom.h
+++ b/src/aws-cpp-sdk-core/include/aws/core/utils/crypto/SecureRandom.h
@@ -6,6 +6,7 @@
 #include <type_traits>
 #include <memory>
 #include <cassert>
+#include <cstdint>
 
 namespace Aws
 {


sbiscigl:
> to fix the compile error.

~~out of curiosity what platform and toolchain does this cause a failure for? more so looking for gaps in our CI~~

>  that causes a compilation error on gcc-15.

sorry you did mention

tzik:
> > to fix the compile error.
> 
> ~out of curiosity what platform and toolchain does this cause a failure for? more so looking for gaps in our CI~
> 
> > that causes a compilation error on gcc-15.
> 
> sorry you did mention

I'm on Ubuntu 25.10, and using gcc-15 as it's the default compiler.
According to [an announcement about gcc-15](https://gcc.gnu.org/gcc-15/porting_to.html#header-dep-changes), it seems to have removed some implicit \#include to `stdint.h` to comply a spec change.
> Header dependency changes
> Some C++ Standard Library headers have been changed to no longer include other headers that were being used internally by the library. As such, C++ programs that used standard library components without including the right headers will no longer compile.
>
> In particular, the following headers are used less widely within libstdc++ and may need to be included explicitly when compiling with GCC 15:
>
> * <stdint.h> (for int8_t, int32_t etc.) and <cstdint> (for std::int8_t, std::int32_t etc.)
> * <ostream> (for std::endl, std::flush etc.)

sbiscigl:
for administrative reasons i have to merge your commit from a branch that I own, so i created a PR with your commit on it https://github.com/aws/aws-sdk-cpp/pull/3665, will run CI on it and merge it

sbiscigl:
fix has been pushed, thanks for the contribution!

PR #3662, opened 2025-12-18
ThreadTask: fix thread lifetime ordering and make detach flag atomic

rajanyadav0307:
*Issue #, if available:*
N/A

*Description of changes:*
This PR improves the thread-safety and lifetime correctness of `ThreadTask`.

Changes include:
- Reordering data members so `std::thread` is destroyed last, ensuring all
  members accessed by the worker thread remain valid during destruction.
- Promoting `m_detached` to `std::atomic<bool>` to prevent data races when
  accessed concurrently by multiple threads.
- Updating the constructor initializer list to match the member declaration
  order and using a lambda for thread startup for improved clarity.

These changes do not modify the public API and are purely internal
correctness and safety improvements.

*Check all that applies:*
- [x] Added proper tests to cover this PR. (Existing threading behavior is exercised; no new tests required.)

- [x] Linux
- [x] Windows
- [ ] Android
- [x] MacOS
- [ ] IOS
- [ ] Other Platforms

By submitting this pull request, I confirm that you can use, modify, copy, and
redistribute this contribution, under the terms of your choice.

### DIFF
--- a/src/aws-cpp-sdk-core/include/aws/core/utils/threading/ThreadTask.h
+++ b/src/aws-cpp-sdk-core/include/aws/core/utils/threading/ThreadTask.h
@@ -41,10 +41,10 @@ namespace Aws
                 void MainTaskRunner();
 
             private:
-                std::atomic<bool> m_continue;
                 PooledThreadExecutor& m_executor;
+                std::atomic<bool> m_continue;
+                std::atomic<bool> m_detached;
                 std::thread m_thread;
-                bool m_detached = false;
             };
         }
     }
--- a/src/aws-cpp-sdk-core/source/utils/threading/ThreadTask.cpp
+++ b/src/aws-cpp-sdk-core/source/utils/threading/ThreadTask.cpp
@@ -9,7 +9,11 @@
 using namespace Aws::Utils;
 using namespace Aws::Utils::Threading;
 
-ThreadTask::ThreadTask(PooledThreadExecutor& executor) : m_continue(true), m_executor(executor), m_thread(std::bind(&ThreadTask::MainTaskRunner, this))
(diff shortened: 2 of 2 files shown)


rajanyadav0307:
@sbiscigl ,

Please can you review the changes?

Thanks

sbiscigl:
> Added proper tests to cover this PR. (Existing threading behavior is exercised; no new tests required.)

So while this is true, can you write any tests or CI checks that would fail before your change that would show improved correctness and saftey?

rajanyadav0307:
Thanks @sbiscigl ,

Good question. ThreadTask isn’t referenced directly in tests, but it’s used internally by PooledThreadExecutor, so executor tests already exercise it indirectly.

My plan is to add a TSAN-gated stress test to PooledThreadExecutorTest.cpp that races executor shutdown with concurrent Submit() calls to hit the detach/shutdown path. This should surface the existing race on m_detached under ThreadSanitizer and show the benefit of making it atomic. I’ll follow up once that’s in place.

The std::thread reordering is a small construction-order safety improvement to avoid partially constructed state being observed by the worker.

sbiscigl:
So in general when we accept PR's we want reproducible issues that are impacting customers. It is a slippery slope to change implementation of key components of the SDK without real proven impact.

So from a user standpoint how could you use the SDK in way where this issue would come up? i.e. how could invoking a client API trigger this behavior and can you trigger it in a reproducible way? That is the kind of test I would want to see along this change. One that i can verify that this change is fixing something for users.

rajanyadav0307:
@sbiscigl 
It's true changing code that is in production without any evidence can bring impacts. Will try to see if I could design any test case around something you described. If I get any, will update it.

Thanks.

PR #3675, opened 2025-12-30
Smithy BiDi streaming support + Bedrock Migration 

sbaluja:
*Issue #, if available:*

*Description of changes:*
- Added support for bidirectional streaming in smithy clients
- Migrated bedrock, bedrock-runtime, bedrock-agent, bedrock-agent-runtime to be smithy clients

*Check all that applies:*

- [ ] Linux
- [ ] Windows
- [ ] Android
- [ ] MacOS
- [ ] IOS
- [ ] Other Platforms

### DIFF
--- a/generated/src/aws-cpp-sdk-bedrock-runtime/include/aws/bedrock-runtime/BedrockRuntimeClient.h
+++ b/generated/src/aws-cpp-sdk-bedrock-runtime/include/aws/bedrock-runtime/BedrockRuntimeClient.h
@@ -37,7 +37,8 @@ class AWS_BEDROCKRUNTIME_API BedrockRuntimeClient
   typedef BedrockRuntimeEndpointProvider EndpointProviderType;
 
   /**
-   * Initializes client to use BearerTokenAuthSignerProvider, with default http client factory, and optional client config.
+   * Initializes client to use mapped auth providers for different auth schemes, with default http client factory, and optional client
+   * config.
    */
   BedrockRuntimeClient(
       const Aws::UnorderedMap<Aws::String, Aws::Crt::Variant<smithy::SigV4AuthScheme, smithy::BearerTokenAuthScheme>> authSchemeMap,
@@ -71,11 +72,6 @@ class AWS_BEDROCKRUNTIME_API BedrockRuntimeClient
                            Aws::BedrockRuntime::BedrockRuntimeClientConfiguration());
 
   /* Legacy constructors due deprecation */
-  /**
-   * Initializes client to use BearerTokenAuthSignerProvider, with default http client factory, and optional client config.
-   */
-  BedrockRuntimeClient(const Aws::Auth::BearerTokenAuthSignerProvider& bearerTokenProvider,
-                       const Aws::Client::ClientConfiguration& clientConfiguration);
   /**
    * Initializes client to use DefaultCredentialProviderChain, with default http client factory, and optional client config. If client
    * config is not specified, it will be initialized to default values.
--- a/generated/src/aws-cpp-sdk-bedrock-runtime/source/BedrockRuntimeClient.cpp
+++ b/generated/src/aws-cpp-sdk-bedrock-runtime/source/BedrockRuntimeClient.cpp
@@ -67,22 +67,6 @@ BedrockRuntimeClient::BedrockRuntimeClient(
                                                                      smithy::BearerTokenAuthSchemeOption::bearerTokenAuthSchemeOption})),
           authSchemeMap) {}
 
-BedrockRuntimeClient::BedrockRuntimeClient(const Aws::Auth::BearerTokenAuthSignerProvider& bearerTokenProvider,
-                                           const Aws::Client::ClientConfiguration& clientConfiguration)
-    : AwsSmithyClientT(
-          clientConfiguration, GetServiceName(), "Bedrock Runtime", Aws::Http::CreateHttpClient(clientConfiguration),
-          Aws::MakeShared<BedrockRuntimeErrorMarshaller>(ALLOCATION_TAG), Aws::MakeShared<BedrockRuntimeEndpointProvider>(ALLOCATION_TAG),
-          Aws::MakeShared<smithy::GenericAuthSchemeResolver<>>(
-              ALLOCATION_TAG, Aws::Vector<smithy::AuthSchemeOption>({smithy::SigV4AuthSchemeOption::sigV4AuthSchemeOption,
-                                                                     smithy::BearerTokenAuthSchemeOption::bearerTokenAuthSchemeOption})),
-          {
-              {smithy::SigV4AuthSchemeOption::sigV4AuthSchemeOption.schemeId,
-               smithy::SigV4AuthScheme{GetServiceName(), clientConfiguration.region}},
-              {smithy::BearerTokenAuthSchemeOption::bearerTokenAuthSchemeOption.schemeId,
-               smithy::BearerTokenAuthScheme{Aws::MakeShared<smithy::AwsBearerTokenIdentityResolver>(ALLOCATION_TAG, bearerTokenProvider),
-                                             GetServiceName(), clientConfiguration.region}},
-          }) {}
-
 BedrockRuntimeClient::BedrockRuntimeClient(const BedrockRuntime::BedrockRuntimeClientConfiguration& clientConfiguration,
                                            std::shared_ptr<BedrockRuntimeEndpointProviderBase> endpointProvider)
     : AwsSmithyClientT(
--- a/generated/src/aws-cpp-sdk-bedrock/include/aws/bedrock/BedrockClient.h
+++ b/generated/src/aws-cpp-sdk-bedrock/include/aws/bedrock/BedrockClient.h
@@ -37,7 +37,8 @@ class AWS_BEDROCK_API BedrockClient
   typedef BedrockEndpointProvider EndpointProviderType;
 
   /**
-   * Initializes client to use BearerTokenAuthSignerProvider, with default http client factory, and optional client config.
+   * Initializes client to use mapped auth providers for different auth schemes, with default http client factory, and optional client
+   * config.
    */
   BedrockClient(
       const Aws::UnorderedMap<Aws::String, Aws::Crt::Variant<smithy::SigV4AuthScheme, smithy::BearerTokenAuthScheme>> authSchemeMap,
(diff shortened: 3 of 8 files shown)


generated/src/aws-cpp-sdk-bedrock-runtime/include/aws/bedrock-runtime/BedrockRuntimeClient.h
```diff
@@ -54,6 +70,11 @@ class AWS_BEDROCKRUNTIME_API BedrockRuntimeClient : public Aws::Client::AWSJsonC
                            Aws::BedrockRuntime::BedrockRuntimeClientConfiguration());
 
   /* Legacy constructors due deprecation */
+  /**
```
sbiscigl: do we want to add a "legacy" constructor? if we're adding one constructor i would say just add the first one, and if someone wants to provide a bearer auth provider, they have to use the newer version of constructor.

generated/src/aws-cpp-sdk-bedrock-runtime/include/aws/bedrock-runtime/BedrockRuntimeClient.h
```diff
@@ -4,29 +4,45 @@
...
   static const char* GetAllocationTag();
+  inline const char* GetServiceClientName() const override { return "Bedrock Runtime"; }
 
   typedef BedrockRuntimeClientConfiguration ClientConfigurationType;
   typedef BedrockRuntimeEndpointProvider EndpointProviderType;
 
+  /**
+   * Initializes client to use BearerTokenAuthSignerProvider, with default http client factory, and optional client config.
+   */
+  BedrockRuntimeClient(const Aws::Auth::BearerTokenAuthSignerProvider& bearerTokenProvider,
```
sbiscigl: just thinking out loud here for a second, this constructor allows someone to use the `BearerTokenAuthSignerProvider` in the bedrock client. IF the operation allows for it, it will use bearer auth, otherwise it will fail? 

Im trying to think of a world where a customer whats to mix sigv4 and bearer auth in the same client and how those could live together. is it possible to create a constructor that take a list of "signer providers" somehow and resolves at runtime?

generated/src/aws-cpp-sdk-bedrock-runtime/source/BedrockRuntimeClient.cpp
```diff
@@ -46,111 +47,150 @@ using ResolveEndpointOutcome = Aws::Endpoint::ResolveEndpointOutcome;
...
+                                           std::shared_ptr<BedrockRuntimeEndpointProviderBase> endpointProvider,
+                                           const BedrockRuntime::BedrockRuntimeClientConfiguration& clientConfiguration)
+    : AwsSmithyClientT(
+          clientConfiguration, GetServiceName(), "Bedrock Runtime", Aws::Http::CreateHttpClient(clientConfiguration),
+          Aws::MakeShared<BedrockRuntimeErrorMarshaller>(ALLOCATION_TAG),
+          endpointProvider ? endpointProvider : Aws::MakeShared<BedrockRuntimeEndpointProvider>(ALLOCATION_TAG),
+          Aws::MakeShared<smithy::GenericAuthSchemeResolver<>>(
+              ALLOCATION_TAG, Aws::Vector<smithy::AuthSchemeOption>({smithy::SigV4AuthSchemeOption::sigV4AuthSchemeOption,
+                                                                     smithy::BearerTokenAuthSchemeOption::bearerTokenAuthSchemeOption})),
+          {
```
sbiscigl: OK this is sort of what i was alluding to earlier should we just allow a user to override the whole map and say "bring whatever you want", so that we arent bearer token specific, just, if you want to override this mapping, bring your own mapping.

generated/src/aws-cpp-sdk-bedrock-runtime/source/BedrockRuntimeClient.cpp
```diff
@@ -411,32 +416,25 @@ void BedrockRuntimeClient::InvokeModelWithBidirectionalStreamAsync(
...
+  auto authCallback = [&](std::shared_ptr<smithy::client::AwsSmithyClientAsyncRequestContext> ctx) -> void {
+    eventEncoderStream->SetSigningCallback([this, ctx, eventEncoderStream](Aws::Utils::Event::Message& message, Aws::String& seed) -> bool {
+      auto outcome = SignEventMessage(message, seed, ctx);
+      return outcome.IsSuccess();
+    });
+  };
   auto requestCopy = Aws::MakeShared<InvokeModelWithBidirectionalStreamRequest>("InvokeModelWithBidirectionalStream", request);
   requestCopy->SetBody(eventEncoderStream);  // this becomes the body of the request
-  request.SetBody(eventEncoderStream);
+  request.SetBody(eventEncoderStream);       // this becomes the body of the request
```
sbiscigl, line 413: nit extra comment

src/aws-cpp-sdk-core/include/aws/core/client/AWSClientEventStreamingAsyncTask.h
```diff
@@ -125,6 +132,92 @@ class AWS_CORE_LOCAL BidirectionalEventStreamingTask final {
   std::shared_ptr<Aws::Utils::Threading::Semaphore> m_sem;
 };
 
+/**
+ * Smithy-compatible bi-directional streaming task for modern AWS clients
+ */
+template <typename OutcomeT, typename ClientT, typename RequestT, typename HandlerT>
+class AWS_CORE_LOCAL SmithyBidirectionalStreamingTask final {
```
sbiscigl: maybe we want to put this in the smithy namespace and not in core as this is the sra client specific

src/aws-cpp-sdk-core/include/aws/core/client/AWSClientEventStreamingAsyncTask.h
```diff
@@ -125,6 +132,92 @@ class AWS_CORE_LOCAL BidirectionalEventStreamingTask final {
...
+
+  SmithyBidirectionalStreamingTask(const ClientT* client, const std::shared_ptr<RequestT>& request, const HandlerT& handler,
+                                   const std::shared_ptr<const Aws::Client::AsyncCallerContext>& context,
+                                   const std::shared_ptr<Utils::Event::EventEncoderStream>& stream,
+                                   EndpointUpdateCallback&& endpointCallback,
+                                   AuthResolvedCallback&& authCallback)
+      : m_client(client), m_request(request), m_handler(handler), m_context(context), m_stream(stream),
+        m_sem(Aws::MakeShared<Aws::Utils::Threading::Semaphore>("SmithyBidirectionalStreamingTask", 0, 1)) {
+
+    m_authCallback = std::move(authCallback);
```
sbiscigl: moving this call back scares me a little bit about lifetimes, consider the following _greatly_ simplified example

```c++
#include <future>
#include <iostream>

class bedrock_runtime_client {
public:
  std::future<void> invoke_bidi_async() {
    std::string endpoint{"https://aws.amazon.com"};
    auto value_callback = [&endpoint]() -> void {
      std::this_thread::sleep_for(std::chrono::seconds(1));
      std::cout << endpoint << std::endl;
    };
    return std::async(std::launch::async, std::move(value_callback));
  }
};

auto main() -> int {
  bedrock_runtime_client client{};
  auto run_future = client.invoke_bidi_async();
  run_future.wait();
  return 0;
}
```

Since we construct the callable async task with local variables does this cause lifetime issues? i could have simplified this wrong, but it seems like there might be lifetime issues. looking at where you call this it seems like you pass by value which will copy the relevant parameters.

At the end of the day, is it possible that we run into lifetime issues, and if so, we cant do that

src/aws-cpp-sdk-core/source/auth/signer/AWSAuthEventStreamV4Signer.cpp
```diff
@@ -214,13 +221,19 @@ bool AWSAuthEventStreamV4Signer::SignEventMessage(Event::Message& message, Aws::
...
+      AWS_LOGSTREAM_TRACE(v4StreamingLogTag, "EventStream Event Signing Event With Empty Credentials");
+    }
+    Aws::Utils::ByteBuffer finalSignatureDigest = GenerateSignature(credentials, canonicalRequestString, simpleDate, m_region, m_serviceName);
     const auto finalSignature = HashingUtils::HexEncode(finalSignatureDigest);
     AWS_LOGSTREAM_DEBUG(v4StreamingLogTag, "Final computed signing hash: " << finalSignature);
     priorSignature = finalSignature;
 
     message.InsertEventHeader(EVENTSTREAM_DATE_HEADER, EventHeaderValue(now.Millis(), EventHeaderValue::EventHeaderType::TIMESTAMP));
     message.InsertEventHeader(EVENTSTREAM_SIGNATURE_HEADER, std::move(finalSignatureDigest));
+    for (auto&& header : message.GetEventHeaders()) {
```
sbiscigl, line 234: nit use `std::for_each` for a nice one liner

kai-ion:
Test

sbiscigl (changes requested):

src/aws-cpp-sdk-core/include/aws/core/client/AWSClientEventStreamingAsyncTask.h
sbaluja: I think this is actually fine. Since std::move is passing ownership, this isn't much different from what [we currently do](https://github.com/aws/aws-sdk-cpp/blob/2020c03ddea07ea7226d5ac72358ed66ff031e1b/generated/src/aws-cpp-sdk-dynamodb/source/DynamoDBClient.cpp#L1710) with the endpoint callbacks in `MakeRequestDeserialize`, with the difference being ownership transferred to an intermediary class (`SmithyBidirectionalStreamingTask`), before making it's way to an rvalue assignment to `MakeRequestDeserialize`. I think the bigger concern is the fact that the authCallback captures a local variable `eventEncoderStream` since that reference is pointing to the original scope. BUT this is also fine because `eventEncoderStream` is a shared pointer, and the pass by copy will increment the use count, and it also lives throughout the lifetime of the request (SetBody holds a reference to this too).

generated/src/aws-cpp-sdk-bedrock-runtime/include/aws/bedrock-runtime/BedrockRuntimeClient.h
sbaluja: This is a fair point, do we want to maintain consistency with legacy [clients that support bearer auth](https://github.com/aws/aws-sdk-cpp/blob/2020c03ddea07ea7226d5ac72358ed66ff031e1b/generated/src/aws-cpp-sdk-codecatalyst/include/aws/codecatalyst/CodeCatalystClient.h#L107)? 
If we choose not to add legacy constructors, we'll need to:
- Add a codegen opt-out exception specifically for Bedrock's bearer auth legacy constructor
- Monitor and add similar exceptions for any Smithy-migrated services that later add new auth scheme support

It's worth noting that this approach would create inconsistency across our clients: Bedrock's bearer constructor would be uniquely different since we'd have legacy constructors for SigV4 but not for bearer.

I'm biased toward keeping the legacy constructors to maintain a consistent customer experience and avoid creating a special case for Bedrock's bearer token constructors.

sbiscigl (approved):

PR #3678, opened 2026-01-06
Bedrock API key auth support 

sbaluja:
*Issue #, if available:*

*Description of changes:*
Add support for Bedrock Environment bearer tokens & add client configuration customization for bedrock so preference for bearer is set automatically

*Check all that applies:*

- [x] Linux
- [x] Windows
- [x] Android
- [x] MacOS
- [ ] IOS
- [ ] Other Platforms

### DIFF
--- a/src/aws-cpp-sdk-core/include/smithy/identity/identity/AwsIdentity.h
+++ b/src/aws-cpp-sdk-core/include/smithy/identity/identity/AwsIdentity.h
@@ -6,6 +6,7 @@
 
 #include <aws/core/utils/DateTime.h>
 #include <aws/crt/Optional.h>
+#include <aws/core/utils/memory/stl/AWSSet.h>
 
 namespace Aws {
     namespace Client {


src/aws-cpp-sdk-core/include/smithy/identity/resolver/AwsBearerTokenIdentityResolver.h
```diff
@@ -107,7 +112,39 @@ class DefaultAwsBearerTokenIdentityResolver
...
         : AwsBearerTokenIdentityResolver(Aws::Vector<std::shared_ptr<Aws::Auth::AWSBearerTokenProviderBase>>{
-              Aws::MakeShared<Aws::Auth::SSOBearerTokenProvider>("SSOBearerTokenProvider")}){};
+            Aws::MakeShared<Aws::Auth::SSOBearerTokenProvider>("SSOBearerTokenProvider")}){};
+
+    DefaultAwsBearerTokenIdentityResolver(const Aws::Client::ClientConfiguration::CredentialProviderConfiguration& config)
+        : AwsBearerTokenIdentityResolver(Aws::Vector<std::shared_ptr<Aws::Auth::AWSBearerTokenProviderBase>>{
+            Aws::MakeShared<Aws::Auth::SSOBearerTokenProvider>("SSOBearerTokenProvider", config.profile)}){};
+};
+
+class BedrockAwsBearerTokenIdentityResolver
```
sbiscigl: could we move this under the actual service that uses it as `BedrockAwsBearerTokenIdentityResolver.h` and have it codgenned? 

Generally we dont want to put things in core that are service specific. I dont think this type is actually used in core either so moving it out to the service library sounds like it would be cleaner. This feels very similar to how we did [S3ExpressIdentityProvider](https://github.com/aws/aws-sdk-cpp/blob/main/generated/src/aws-cpp-sdk-s3/include/aws/s3/S3ExpressIdentityProvider.h) and have it live in S3 as opposed to core.

Issue #3666, opened 2025-12-20
Can't use General HTTP Provider: AWS_CONTAINER_CREDENTIALS_FULL_URI is not HTTPS and is not within loop back CIDR: [fd00:ec2::23]


artklochkov:
https://github.com/aws/aws-sdk-cpp/blame/f5309cf6692124b6a20564615a8dab30e974e9d3/src/aws-cpp-sdk-core/source/auth/GeneralHTTPCredentialsProvider.cpp#L36

Looks like the problem is that in case of IPv6 authority part of URI is [fd00:ec2::23], with IPv6 inside [], while in this comparison we are trying to compare it with loopback address fd00:ec2::23, without []. So it leads to an error 

> Can't use General HTTP Provider: AWS_CONTAINER_CREDENTIALS_FULL_URI is not HTTPS and is not within loop back CIDR: [fd00:ec2::23]

When using with EKS Pod Identity agent, since it's URL (which couldn't be changed with configuration) is:

`http://[fd00:ec2::23]/v1/credentials`

sbiscigl:
@artklochkov / @rajanyadav0307 not sure if yall are on the same team working on this but took at look at @rajanyadav0307 's PR and left a comment. This is actually identical to a similar issue we saw [a while ago](https://github.com/aws/aws-sdk-cpp/pull/3398) but just for ipv6 endpoints in general. This introduced a new API called `GetHost` which is the correct RFC 3986 encoding. so in the PR if you just swap `GetAuthority` use for `GetHost` I will approve.

I can also push a updated PR if you are unable to make update to yours

rajanyadav0307:
@sbiscigl ,
Thanks for pointing it out. I think it makes more sense to use`GetHost` instead of calling `GetAuthority` as you stated. 

I see the code handles for IPv6:

<img width="722" height="206" alt="Image" src="https://github.com/user-attachments/assets/fd6b7f51-648e-45cc-8651-ab5a7299dc88" />

I will update the PR.

rajanyadav0307:
@sbiscigl ,
I hope we don't need a new PR, last [PR](https://github.com/aws/aws-sdk-cpp/pull/3680) should be good to merge?

sbiscigl:
For process reasons i need to pull your branch onto a [branch of the main repo/made by me](https://github.com/aws/aws-sdk-cpp/pull/3688). The commit will still be credited to you. Will merge that PR once CI passes.

[...]

PR #3680, opened 2026-01-07
Issue-3666: fixes IPv6 loopback handling for EKS Pod Identity agent URLs

rajanyadav0307:
**Issue #** [3666](https://github.com/aws/aws-sdk-cpp/issues/3666) 

**Description of changes:**  

Optimized IPv6 address handling in `GeneralHTTPCredentialsProvider.cpp`.  

When the authority part of a URI contains IPv6 brackets (e.g., `[fd00:ec2::23]`), brackets are removed safely and normalized before any loopback or allowed-host checks.  

Single-call construction of `Aws::String` from the normalized authority to avoid temporary assignments.  

Preserves all existing behavior for IPv4, HTTPS, and already-valid IPv6 URIs.  

Added clear comments explaining the bracket removal and why it is necessary for loopback checks.  

Verified using existing unit tests in `GeneralHTTPCredentialsProviderTest.cpp`, specifically:  

- `http loopback(v6) URI` test  
- `http link-local EKS URI` test  

These tests cover bracketed IPv6 addresses and ensure the provider no longer incorrectly rejects valid URIs.  

No functional changes beyond fixing invalid rejection of valid HTTP URIs with bracketed IPv6 addresses.  

**Check all that applies:**  
- [x] Added proper tests to cover this PR. (Existing unit tests verify correctness; no new tests required.)  

**Check which platforms you have built SDK on to verify the correctness of this PR:**  
- [x] Linux  
- [x] Windows  
- [x] Android  
- [x] MacOS  
- [x] IOS  
- [ ] Other Platforms  

### DIFF
--- a/src/aws-cpp-sdk-core/source/auth/GeneralHTTPCredentialsProvider.cpp
+++ b/src/aws-cpp-sdk-core/source/auth/GeneralHTTPCredentialsProvider.cpp
@@ -90,7 +90,7 @@ bool GeneralHTTPCredentialsProvider::ShouldCreateGeneralHTTPProvider(const Aws::
         }
 
         Aws::Http::URI absUri(absoluteUri);
-        const Aws::String& authority = absUri.GetAuthority();
+        const Aws::String& authority = absUri.GetHost();
 
         // Otherwise, implementations MUST fail to resolve when the URI hostname does not satisfy any of the following conditions
         if (IsAllowedIp(authority))


src/aws-cpp-sdk-core/source/auth/GeneralHTTPCredentialsProvider.cpp
```diff
@@ -90,7 +90,18 @@ bool GeneralHTTPCredentialsProvider::ShouldCreateGeneralHTTPProvider(const Aws::
         }
 
         Aws::Http::URI absUri(absoluteUri);
-        const Aws::String& authority = absUri.GetAuthority();
```
sbiscigl, line 93: We already went about this difference in a [separate issue](https://github.com/aws/aws-sdk-cpp/commit/66712fa81901fd4a3e1f91b8c8e9e6dc39027ab8) related to ipv6 where we added the API 

```
/**
 * The host portion of the URI as described in rfc3986.
 *
 * The host subcomponent of authority is identified by an IPv6 literal
 * encapsulated within square brackets, an IPv4 address in dotted-
 * decimal form, or a registered name.
 *
 * @return The host portion of the URI
 */
Aws::String GetHost() const;
```

and called out in the `GetAuthority` API 
```
/**
 * Gets the authority portion of the URI. Differs from the authority definition in 
 * RFC 3986. user information and the port information are not included. It is 
 * functionally the host as defined by  RFC 3986 with encoding included.
 */
inline const Aws::String& GetAuthority() const { return m_authority; }
```

so instead of hand coding the logic again just change this API to `GetHost`

rajanyadav0307: @sbiscigl ,

I have updated. 
Thanks

sbiscigl:
merged in https://github.com/aws/aws-sdk-cpp/pull/3688 closing

rajanyadav0307:
@sbiscigl 
Thanks.

PR #3697, opened 2026-01-22
validate region EP param is a valid host label

sbiscigl:
*Description of changes:*

Adds a check to the region builtin endpoint parameter to make sure the region is a valid host label. This will fail to set the region EP parameter failing endpoints resolution for the client so no call is made.

*Check all that applies:*

- [x] Linux
- [x] Windows
- [x] Android
- [x] MacOS
- [ ] IOS
- [ ] Other Platforms

### DIFF
--- a/src/aws-cpp-sdk-core/source/endpoint/BuiltInParameters.cpp
+++ b/src/aws-cpp-sdk-core/source/endpoint/BuiltInParameters.cpp
@@ -4,6 +4,7 @@
  */
 
 #include <aws/core/endpoint/BuiltInParameters.h>
+#include <aws/core/utils/DNS.h>
 #include <aws/core/utils/logging/LogMacros.h>
 #include <aws/core/config/EndpointResolver.h>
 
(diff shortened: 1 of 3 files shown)


src/aws-cpp-sdk-core/source/endpoint/BuiltInParameters.cpp
```diff
@@ -41,7 +42,9 @@ namespace Endpoint
         if (!config.region.empty()) {
             static const char* FIPS_PREFIX = "fips-";
             static const char* FIPS_SUFFIX = "-fips";
-            if (config.region.rfind(FIPS_PREFIX, 0) == 0) {
+            if (!Aws::Utils::IsValidDnsLabel(config.region)) {
+                AWS_LOGSTREAM_ERROR(ENDPOINT_BUILTIN_LOG_TAG, "Invalid region name: " << config.region);
```
DmitriyMusatkin, line 46: sanity check: logstream error causes the function to error?

sbiscigl: No it just logs a failure, since the function returns void, and we have exceptions disabled we cannot return a real "error" from the function. the failure mode is that the region built in parameter region is not set. It fails in endpoint resolution because region is not set. Hypothetically if a endpoints rule set did not require region, resolution would still work.

Issue #3701, opened 2026-01-27
ZLib declared as optional depencency while in fact it is mandatory


DSysoletin:
### Describe the bug

Hello!
While Readme states ZLib is **optional** and required only in certain cases, in fact build will fail on Linux if there is no ZLib in the system. This happens because in file external_dependencies.cmake we have check if platform is NOT windows and it is NOT custom - then throw an error if ZLib is not found, even if -DENABLE_ZLIB_REQUEST_COMPRESSION is set to "OFF" or just absent.
This seems like a bug because if this dependency is optional and we don't use features that require ZLib - build should not fail because of "Could not find zlib".

### Regression Issue

- [ ] Select this option if this issue appears to be a regression.

### Expected Behavior

Build on Linux without -DENABLE_ZLIB_REQUEST_COMPRESSION flag OR with -DENABLE_ZLIB_REQUEST_COMPRESSION="OFF" should pass even if there is no ZLib in the build system(container).

### Current Behavior

Build on Linux will fail if there is no ZLib in the build system(container).

### Reproduction Steps

1. Use Docker based on Ubuntu 20.04 
2. cd /tmp && git clone --recurse-submodules https://github.com/aws/aws-sdk-cpp && cd ./aws-sdk-cpp && mkdir build && cd ./build
3. cmake .. \
	-DCMAKE_C_COMPILER=arm-none-linux-gnueabihf-gcc \
	-DCMAKE_CXX_COMPILER=arm-none-linux-gnueabihf-g++ \
	-DBUILD_ONLY="s3" \
	-DENABLE_ZLIB_REQUEST_COMPRESSION=OFF \
	-DCMAKE_CROSSCOMPILING=TRUE

### Possible Solution

Alter checks in external_dependencies.cmake file to take -DENABLE_ZLIB_REQUEST_COMPRESSION flag into account.

### Additional Information/Context

This is only observed during cross-compiling when we don't have zlib in cross-compilation sysroot. This is NOT observed on normal (non-cross) compilation because zilb from the host is used.

### AWS CPP SDK version used

1.11.733

### Compiler and Version used

It fails on CMake phase, not on compile phase

### Operating System and version

Ubuntu 20.04 (dockerized)

kai-ion:
@DSysoletin thank you for the report. I was able to reproduce the error on Ubuntu 20.04. Looks like on that specific distribution  zlib is installed only for runtime, causing the cross-compilation to fail.

[...]

PR #3704, opened 2026-01-28
added another check for ENABLE_ZLIB_REQUEST_COMPRESSION=OFF

kai-ion:
*Issue #, if available:*
[Issue 3701](https://github.com/aws/aws-sdk-cpp/issues/3701)

*Description of changes:*
Make zlib dependency optional when ENABLE_ZLIB_REQUEST_COMPRESSION=OFF

Fixes build failure when zlib is not available in cross-compile environments.

Previously, the SDK unconditionally required zlib on non-Windows/non-custom
platforms regardless of the ENABLE_ZLIB_REQUEST_COMPRESSION flag setting.
This caused builds to fail with "Could not find zlib" even when users
explicitly disabled request compression.

This change implements 3-state behavior for zlib dependency:
- Explicit ON (-DENABLE_ZLIB_REQUEST_COMPRESSION=ON): Requires zlib, fails if missing
- Explicit OFF (-DENABLE_ZLIB_REQUEST_COMPRESSION=OFF): Skips zlib check entirely
- Default (ON): Tries to find zlib, continues without it if unavailable

*Check all that applies:*

- [x] Linux
- [x] Windows
- [x] Android
- [x] MacOS
- [ ] IOS
- [ ] Other Platforms

### DIFF
--- a/tests/aws-cpp-sdk-bedrock-runtime-integration-tests/IntegrationTests.cpp
+++ b/tests/aws-cpp-sdk-bedrock-runtime-integration-tests/IntegrationTests.cpp
@@ -86,7 +86,7 @@ TEST_F(BedrockRuntimeTests, TestInvokeModel)
   bedrockRequest.SetContentType("application/json");
 
   Aws::BedrockRuntime::Model::InvokeModelOutcome outcome = m_client->InvokeModel(bedrockRequest);
-  ASSERT_TRUE(outcome.IsSuccess());
+  EXPECT_TRUE(outcome.IsSuccess()) << outcome.GetError().GetExceptionName() << " - " << outcome.GetError().GetMessage();
   Aws::StringStream ss;
   ss << outcome.GetResult().GetBody().rdbuf();
   ASSERT_FALSE(ss.str().empty());


cmake/external_dependencies.cmake
```diff
@@ -3,12 +3,23 @@ if(PLATFORM_ANDROID AND ANDROID_BUILD_ZLIB)
     set(BUILD_ZLIB 1)
     message(STATUS "  Building Zlib as part of AWS SDK")
 elseif(NOT PLATFORM_WINDOWS AND NOT PLATFORM_CUSTOM)
-    #If zlib is required either by openssl and curl in their linking chain, we should find it.
-    include(FindZLIB)
-    if(NOT ZLIB_FOUND)
-        message(FATAL_ERROR "Could not find zlib")
+    # Check if user explicitly disabled compression
+    if(ENABLE_ZLIB_REQUEST_COMPRESSION STREQUAL "OFF")
```
sbiscigl: this if statement seems a little weird
```
else if( _statement_ ) {
  if( _other_statement_ ) {
    // _do nothing_
  } else {
    _logic_
  }
}
```

seems like we're adding some branches to the if statement. should likely be
```
else if( _statement_ AND NOT _other_statement_ ) {
    _logic_
}
```

infact if this was [c++ it would compile with a warning](https://clang.llvm.org/docs/DiagnosticsReference.html#wempty-body). we should avoid doing things in cmake that would be considered bad practice in normal code.

sbiscigl (approved):

Issue #3613, opened 2025-11-06
Pkgconfig file injects -std=c++11 into downstream projects


WillAyd:
### Describe the bug

When compiling a project that uses the SDK as a dependency and resolves said dependency with pkgconfig, the AWS SDK ends up injecting -std=c++11 into the compilation flags of the main project. This appears hard-coded in the configuration:

https://github.com/aws/aws-sdk-cpp/blob/095b07c61ac1e84c181cd077a12218d1629742d9/cmake/compiler_settings.cmake#L39

### Regression Issue

- [ ] Select this option if this issue appears to be a regression.

### Expected Behavior

The dependency should not inject any c++ standard requirements into the parent project

### Current Behavior

Compilation fails for projects newer than C++11

### Reproduction Steps

Create a module that uses a C++14 feature, like so:

```cpp
auto doNothing() {
}

int main() {
  return 0;
}
```

Attempting to compile this with a dependency on aws-cpp-sdk-core will fail:

```
$ gcc `pkg-config --cflags aws-cpp-sdk-core` main.cc
main.cc:1:1: error: ‘doNothing’ function uses ‘auto’ type specifier without trailing return type
    1 | auto doNothing() {
      | ^~~~
main.cc:1:1: note: deduced return type only available with ‘-std=c++14’ or ‘-std=gnu++14’
```

### AWS CPP SDK version used

1.11.68

### Compiler and Version used

gcc (Ubuntu 15.2.0-4ubuntu4) 15.2.0

### Operating System and version

Ubuntu 25.10

pulimsr:
Hi @WillAyd,
                                                                                                                                                                                                                                                                                                                          
The recommended approach for using the AWS SDK for C++ is building from source rather than using pre-built packages via pkgconfig. This gives you full control over the build configuration.                                                                                                                              
                                                                                                                                                                                                                                                                                                                          
Please refer to our [Building From Source documentation](https://github.com/aws/aws-sdk-cpp#building-from-source) and use the `CPP_STANDARD` parameter documented in our [CMake Parameters guide](https://github.com/aws/aws-sdk-cpp/blob/main/docs/CMake_Parameters.md) to set your desired C++ standard (e.g., `-DCPP_STANDARD=14`).                  
                                                                                                                                                                                                                                                                                                                          
Building from source with the appropriate `CPP_STANDARD` setting will generate pkgconfig files compatible with your project's C++ version and resolve your compilation issue.

[...]

PR #3709, opened 2026-02-01
Issue-3613: Fix

rajanyadav0307:
***Issue #, if available:***
#3613

***Description of changes:***
This PR adds an opt-out CMake option to prevent the AWS SDK for C++ from exporting
the C++ language standard via pkg-config `Cflags`.

Currently, the generated `.pc` files include compiler policy flags (e.g.
`-std=c++<N>`, `-fno-exceptions`) derived from `AWS_COMPILER_FLAGS`. Downstream
projects that consume the SDK using pkg-config (for example, Meson or custom
build systems) blindly apply these flags, which can override the consumer’s
chosen C++ standard and cause build failures.

To address this without breaking existing users, this PR introduces a new option:

> -DNO_CPPSTD_IN_PKG_CONFIG=ON

When enabled, the generated pkg-config `Cflags` will no longer include:
- `-std=c++<N>` (primary issue reported)
- `-fno-exceptions` (also identified as leaking semantics into downstream builds)

The default behavior remains unchanged to preserve backward compatibility.
CMake-based consumers are unaffected, as these flags remain `PRIVATE` to SDK
targets.

*Check all that applies:*
- [x] Added proper tests to cover this PR.  
      Two Docker-based tests were used:
      - One Dockerfile builds the upstream SDK and a minimal downstream project
        to reliably reproduce the reported failure. [Docker_Issue_Repro.zip](https://github.com/user-attachments/files/24989369/Docker_Issue_Repro.zip)
      - A second Dockerfile builds the SDK with this patch enabled and verifies
        that the downstream project builds successfully.[Docker-Fix-Verify.zip](https://github.com/user-attachments/files/24989385/Docker-Fix-Verify.zip)

- [x] Linux
- [ ] Windows
- [ ] Android
- [ ] MacOS
- [ ] IOS
- [ ] Other Platforms

**Screenshots:**

**Issue Reproduced:**
Screenshot 2026-01-31 at 11 32 42 PM

**Fix-Verify:**
Screenshot 2026-01-31 at 11 32 10 PM

By submitting this pull request, I confirm that you can use, modify, copy, and
redistribute this contribution, under the terms of your choice.

Sign-off: Rajan Y.(rajanyadav0307@gmail.com)


cmake/compiler_settings.cmake
```diff
@@ -23,7 +32,19 @@ function(set_compiler_flags target)
...
+        if(NO_CPPSTD_IN_PKG_CONFIG)
+            # pkg-config has no PRIVATE/PUBLIC separation: everything in Cflags leaks.
+            # Remove policy flags that should be controlled by the downstream project.
+
+            # Remove any explicit -std=c++NN
+            string(REGEX REPLACE "(^|[ \t])(-std=c\\+\\+[0-9]+)([ \t]|$)" " " _TMP " ${_TMP} ")
+
+            # Also remove -fno-exceptions since it changes downstream semantics
+            string(REGEX REPLACE "(^|[ \t])(-fno-exceptions)([ \t]|$)" " " _TMP " ${_TMP} ")
+
```
sbaluja: Don't think we need regex here, we know exactly the substring we're looking for. Ex. `std=c++${CPP_STANDARD}`

rajanyadav0307: Thanks @sbaluja , I have updated without regex.

rajanyadav0307:
@sbaluja ,
Please help to review and merge the change.

Thanks

rajanyadav0307:
@sbaluja ,

Please let me know if we are expecting more for the PR.

Thanks

cmake/compiler_settings.cmake
```diff
@@ -3,6 +3,15 @@ set(COMPILER_MSVC 0)
 set(COMPILER_GCC 0)
 set(COMPILER_CLANG 0)
 
+# pkg-config consumers blindly apply Cflags to downstream compilation. Historically
+# aws-sdk-cpp exported -std=c++${CPP_STANDARD} (and other policy flags) in .pc files,
+# which can override a consumer's chosen language level. Keep default behavior for
+# backward compatibility, but allow packagers/consumers to disable exporting the C++
+# standard (and related policy flags) via pkg-config.
+option(NO_CPPSTD_IN_PKG_CONFIG
```
sbiscigl: This option is in the wrong place, this option should appear and act identically to `ENABLE_RTTI` which is another compiler setting turned on and off.

this is found in the [top level `CMakeLists.txt`](https://github.com/aws/aws-sdk-cpp/blob/main/CMakeLists.txt#L40)
```cmake
option(ENABLE_RTTI "Flag to enable/disable rtti within the library" ON)
```

in that list of options we should include one for this case
```cmake
option(NO_CPPSTD_IN_PKG_CONFIG "Removes cpp language standard from being included in pkg config installs" OFF)
```

then this can be used more easily where the actual compile flags are [being set for gcc](https://github.com/aws/aws-sdk-cpp/blob/main/cmake/compiler_settings.cmake#L38-L59) and this can look exactly like RTTI i.e.
```cmake
macro(set_gcc_flags)
    list(APPEND AWS_COMPILER_FLAGS "-fno-exceptions")

    if(NOT NO_CPPSTD_IN_PKG_CONFIG)
         list(APPEND AWS_COMPILER_FLAGS "-std=c++${CPP_STANDARD}")
    endif()

    if(NOT ENABLE_RTTI)
        list(APPEND AWS_COMPILER_FLAGS "-fno-rtti")
    endif()
    # other code  ...
endmacro()
```

cmake/compiler_settings.cmake
```diff
@@ -23,7 +32,19 @@ function(set_compiler_flags target)
         set_gcc_flags()
         target_compile_options(${target} PRIVATE "${AWS_COMPILER_FLAGS}")
         string(REPLACE ";" " " _TMP "${AWS_COMPILER_FLAGS}")
-        set(PKG_CONFIG_CFLAGS "${_TMP}" CACHE INTERNAL "C++ compiler flags which affect the ABI")
+        if(NO_CPPSTD_IN_PKG_CONFIG)
```
sbiscigl: please refer to the other comment, thats likely where and how we want to handle this option

sbiscigl:
created https://github.com/aws/aws-sdk-cpp/pull/3792 to run CI on and merge

sbiscigl:
merged, sorry for the long wait and thanks for contributing!

rajanyadav0307:
> merged, sorry for the long wait and thanks for contributing!

Thank you @sbiscigl for merging the changes. I really appreciate it. Looking forward for contributing more.

Issue #3440, opened 2025-05-30
Porting aws-sdk-cpp for AIX OS


patel-parth7:
### Describe the feature

To port aws-sdk-cpp for AIX OS, i have made changes for below modules and wanted to upstream these changes to community.  So before creating PR's under these sub-modules, i wanted to know any guideline/input/concerns which i have to consider.

- crt/aws-c-cal
- crt/aws-c-common
- crt/aws-c-io
- crt/s2n

For initial port we are building for "s3;transfer;sts" 

### Use Case

AWS S3 service integration with IBM AIX.

### Acknowledgements

- [x] I may be able to implement this feature request
- [ ] This feature might incur a breaking change

DmitriyMusatkin:
This submodules are maintained by 2 other teams - CRT and S2N. From CRT side we are open to external contributions, but might have limited bandwidth to support debugging if things do not work. And the usual thing of every pr will need to have sufficient testing (on aix here) before we can accept it.

kai-ion:
The aws-sdk-cpp changes for AIX support were merged in #3794 and are on main. Closing this as completed for the aws-sdk-cpp repo.

Note the CRT/S2N submodule changes (aws-c-cal, aws-c-common, aws-c-io, s2n) are maintained in their own repositories, per the earlier note on this thread, those need PRs (with AIX testing) against each of those repos rather than here. Please open issues/PRs in the respective CRT and S2N repos for any remaining submodule work.

PR #3714, opened 2026-02-09
Necessary changes were made in the aws-sdk-cpp to support AIX OS.

patel-parth7:
*Issue #3440 *

*Description of changes:*

*Check all that applies:*
- [ Yes] Did a review by yourself.
- [ NA] Checked if this PR is a breaking (APIs have been changed) change.
- [Yes] Checked if this PR would require a ReadMe/Wiki update.
    Require to add support for AIX platform.

- [ ] Linux
- [ ] Windows
- [ ] Android
- [ ] MacOS
- [ ] IOS
- [AIX ] Other Platforms


src/aws-cpp-sdk-core/source/platform/aix/TimeHelper.cpp
```diff
@@ -0,0 +1,80 @@
...
+
+#include <time.h>
+
+namespace Aws
+{
+namespace Time
+{
+time_t timegm (struct tm *tm)
+{
+    static const int msum [2][12] = {
```
sbiscigl: while this pattern may be common for C. lets use C++ patterns. lets use anonymous name spacing for static. additionally lets use C++ arrays not C arrays.

```c++
namespace {
const Aws::Array<Aws::Array<int, 12>, 2> msum{{
    {{ 0, 31, 59, 90, 120, 151, 181, 212, 243, 273, 304, 334}},        /* normal years */
    {{ 0, 31, 60, 91, 121, 152, 182, 213, 244, 274, 305, 335}}        /* leap years */
}};
const Aws::Array<Aws::Array<int, 12>, 2> msummlen{{
  {{ 31, 28, 31, 30,  31,  30,  31,  31,  30,  31,  30, 31}},
  {{ 31, 29, 31, 30,  31,  30,  31,  31,  30,  31,  30, 31}}
}};
const int tmstr_year= 1900; /* base of 'tm_year' in 'struct tm' */
const int epoch_year= 1970; /* unix timestamp epoch */
const int base_year=  -9999;  /* start of a 400-year period: used to be 1601,
                                 but this allows larger range (in 64 bit)
                                 mind you, this is proleptic Gregorian */
}
```

src/aws-cpp-sdk-core/source/platform/aix/TimeHelper.cpp
```diff
@@ -0,0 +1,80 @@
+/* TimeHelper.cpp
```
sbiscigl: did you copy this implementation from [here](https://github.com/lzsiga/timegm/blob/main/aix-timegm.c)? i see it is under [The Unlicense](https://github.com/lzsiga/timegm/blob/main/LICENSE), which doesnt require attribution, but lets add attribution just so the original author gets credit. something along the lines of how we do it in [cjson](https://github.com/aws/aws-sdk-cpp/blob/main/src/aws-cpp-sdk-core/include/aws/core/external/cjson/cJSON.h#L1-L21) but indicating where this came from.

src/aws-cpp-sdk-core/source/platform/aix/TimeHelper.cpp
```diff
@@ -0,0 +1,80 @@
...
+
+#include <aws/core/platform/Time.h>
+
+#include <time.h>
+
+namespace Aws
+{
+namespace Time
+{
+time_t timegm (struct tm *tm)
```
sbiscigl: im confused here are you trying to implement `timegm` for [SDK interface](https://github.com/aws/aws-sdk-cpp/blob/main/src/aws-cpp-sdk-core/include/aws/core/platform/Time.h#L19) or are you trying to implement the interface for [linux](https://man7.org/linux/man-pages/man3/timegm.3.html), and then subsequently use the [linux implementation](https://github.com/aws/aws-sdk-cpp/blob/main/src/aws-cpp-sdk-core/source/platform/linux-shared/Time.cpp#L17)?

src/aws-cpp-sdk-core/source/platform/aix/TimeHelper.cpp
```diff
@@ -0,0 +1,80 @@
...
+    static const int tmstr_year= 1900; /* base of 'tm_year' in 'struct tm' */
+    static const int epoch_year= 1970; /* unix timestamp epoch */
+    static const int base_year=  -9999;  /* start of a 400-year period: used to be 1601,
+                                            but this allows larger range (in 64 bit)
+                                            mind you, this is proleptic Gregorian */
+    int year, ytmp, dtmp, ytmpe, dtmpe;
+    int isleapyear;
+    long long t;
+
+    if (!tm) return -1;
```
sbiscigl: from the `timegm` documentation

> On success, these functions return the calendar time (seconds since the Epoch), expressed as a value of type time_t.  On error, they return the value (time_t) -1 and set [errno](https://man7.org/linux/man-pages/man3/errno.3.html) to indicate the error.

dont you need to set `errno` here? additionally dont you need to set it everywhere you are returning with err?

src/aws-cpp-sdk-core/source/platform/aix/TimeHelper.cpp
```diff
@@ -0,0 +1,80 @@
...
+
+    year = tm->tm_year + tmstr_year;
+    isleapyear= (year%4==0) - (year%100==0) + (year%400==0);
+
+    if (year<-9999    || year>9999  ||
+        tm->tm_mon<0  || tm->tm_mon>11 ||
+        tm->tm_mday<1 || tm->tm_mday>mlen[isleapyear][tm->tm_mon] ||
+        tm->tm_hour<0 || tm->tm_hour>23 ||
+        tm->tm_min<0  || tm->tm_min>59 ||
+        tm->tm_min<0  || tm->tm_sec>59) return -1;
```
sbiscigl: isnt this a typo? shouldnt this be
```c++
tm->tm_sec<0
```

src/aws-cpp-sdk-core/source/platform/aix/TimeHelper.cpp
```diff
@@ -0,0 +1,80 @@
...
+    ytmpe = epoch_year - base_year;
+    dtmpe = ytmpe*365 + ytmpe/4 - ytmpe/100 + ytmpe/400;
+
+/* 3. days between 'current year' and 'epoch_year' */
+    t  = dtmp - dtmpe;
+
+    t += msum[isleapyear][tm->tm_mon];
+    t += tm->tm_mday-1;
+
+    tm->tm_wday = (t+4)%7;/* 0..6=Sun..Sat; adding 4 to adjust 1970.01.01.=Thursday; */
```
sbiscigl: Why are you modifying input? In a conversion function i think this should be const?

CMakeLists.txt
```diff
@@ -173,6 +168,13 @@ if (LEGACY_BUILD)
     # build the sdk targets
     project("aws-cpp-sdk-all" VERSION "${PROJECT_VERSION}" LANGUAGES CXX)
 
+    if(NOT CMAKE_SYSTEM_NAME STREQUAL "AIX")
```
sbiscigl, line 171: why does this need to be moved?

tests/aws-cpp-sdk-core-tests/utils/FileSystemUtilsTest.cpp
```diff
@@ -12,6 +12,9 @@
 #include <unistd.h>
 #include <climits>
 #endif
+#ifdef _AIX
+#define NAME_MAX     255
```
sbiscigl, line 16: what is this define used for?

src/aws-cpp-sdk-core/source/platform/aix/TimeHelper.cpp
patel-parth7: I will address this in the next commit.

src/aws-cpp-sdk-core/source/platform/aix/TimeHelper.cpp
patel-parth7: I will implement timegm() as a standalone C-style API, so I will not use any C++ constructs.

src/aws-cpp-sdk-core/source/platform/aix/TimeHelper.cpp
patel-parth7: As mentioned in the previous comment, I will not use any C++ constructs.

src/aws-cpp-sdk-core/source/platform/aix/TimeHelper.cpp
patel-parth7: Addressed in next commit.

src/aws-cpp-sdk-core/source/platform/aix/TimeHelper.cpp
patel-parth7: Addressed in next commit.

tests/aws-cpp-sdk-core-tests/utils/FileSystemUtilsTest.cpp
patel-parth7: NAME_MAX is used in the same file at line 281.
On AIX, this macro is undefined. So need to define it with respect to AIX.

CMakeLists.txt
patel-parth7: CMake response file support is not enabled on AIX, so it must be blocked.
To block it on AIX, we need to use CMAKE_SYSTEM_NAME, which is evaluated after the project() call.

src/aws-cpp-sdk-core/source/platform/aix/TimeHelper.cpp
patel-parth7: I was going through man page of mktime() as mktime and timegm functionality is same except they differ only on how they interpret the input parameter.
mktime interprets the input as a local time, while timegm interprets it as Coordinated Universal Time 
(UTC)

So as per the man page:
https://man7.org/linux/man-pages/man3/mktime.3p.html
```
Upon successful completion, the values of the tm_wday and tm_yday
components of the structure shall be set appropriately, and the
other components shall be set to represent the specified time
since the Epoch, but with their values forced to the ranges
indicated in the <time.h> entry; the final value of tm_mday shall
not be set until tm_mon and tm_year are determined.
```

src/aws-cpp-sdk-core/source/platform/aix/TimeHelper.cpp
sbiscigl: this doesnt really answer my question -- what is it that you are trying to accomplish here. are you trying to make a SDK implementation for AIX or are you trying to implement the C API so that you can use the linux implementation. what im really getting is how does this actually work for you. are you using the [linux implementation](https://github.com/aws/aws-sdk-cpp/blob/main/src/aws-cpp-sdk-core/source/platform/linux-shared/Time.cpp#L17) which requires `timegm` but using your implementation of timegm?

src/aws-cpp-sdk-core/source/platform/aix/TimeHelper.cpp
sbiscigl: the file is called "TimeHelper.cpp", this is a C++ source file, compiled with a C++ compiler. we wont be accepting this as-is. This is a C++ project. We use C++ constructs.

src/aws-cpp-sdk-core/source/platform/aix/TimeHelper.cpp
patel-parth7: My plan is to use the Linux implementation, and since timegm is not available on AIX, I am providing a timegm implementation in the newly introduced TimeHelper.cpp

src/aws-cpp-sdk-core/source/platform/aix/TimeHelper.cpp
patel-parth7: In the first commit, I created a C++ source file because I was using a namespace. Since we are now keeping this as a C API, we can rename the file to .c.
Please let me know if you are okay with renaming the file and keeping the timegm implementation as C source code.

src/aws-cpp-sdk-core/source/platform/aix/TimeHelper.cpp
sbiscigl: so i don't really like this, this is sort of tricking the runtime into thinking `timegm` exists. what you should do is create a implementation specific to AIX for the sdk interface. i.e. in the [platform specific directory](https://github.com/aws/aws-sdk-cpp/tree/main/src/aws-cpp-sdk-core/source/platform). create a `aix` directory. and then implement `Time.cpp` using AIX interfaces i.e.

```c++
/**
 * Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.
 * SPDX-License-Identifier: Apache-2.0.
 */

#include <aws/core/platform/Time.h>

#include <time.h>

namespace Aws
{
namespace Time
{

time_t TimeGM(struct tm* const t)
{
  //TODO: your AIX code
}

void LocalTime(tm* t, std::time_t time)
{
  //TODO: your AIX code
}

void GMTime(tm* t, std::time_t time)
{
  //TODO: your AIX code
}

} // namespace Time
} // namespace Aws
```

This way you aren't messing with the runtime or trying to piggy back back off of the linux implementation. the platform support is cleaner and compartmentalized.

patel-parth7: Updated the code in the subsequent commit.

src/aws-cpp-sdk-core/source/platform/aix/Time.cpp
```diff
@@ -0,0 +1,132 @@
...
+        return (time_t)-1;
+    }
+#endif
+
+    return (time_t)ret_val;
+}
+
+void LocalTime(tm* t, std::time_t time)
+{
+    localtime_r(&time, t);
```
sbiscigl, line 123: are `localtime_r` and `gmtime_r` supported in AIX?

CMakeLists.txt
```diff
@@ -173,6 +168,13 @@ if (LEGACY_BUILD)
...
+        # use response files to prevent command-line-too-big errors for large libraries like iam
+        set(CMAKE_CXX_USE_RESPONSE_FILE_FOR_OBJECTS 1)
```
sbiscigl, line 173: why do these cmake flags need to be turned off for AIX?

src/aws-cpp-sdk-core/source/platform/aix/TimeHelper.cpp
sbiscigl: yeah looks super clean now thanks

tests/aws-cpp-sdk-core-tests/utils/FileSystemUtilsTest.cpp
sbiscigl: on [line 281 NAME_MAX is used as](https://github.com/aws/aws-sdk-cpp/blob/03b3ebe48d6ec435227d505702b280ed41c8a76d/tests/aws-cpp-sdk-core-tests/utils/FileSystemUtilsTest.cpp#L272-L285)
```c++
#if defined(HAS_PATHCONF)
    errno = 0;
    long longNameLength = pathconf(".", _PC_NAME_MAX);
    ASSERT_TRUE(longNameLength >= 0 || errno == 0);
    if (longNameLength <= 0)
    {
        longNameLength = NAME_MAX;
    }
#else
    // Path compoments on Windows can't exceed 255(_MAX_FNAME) chars.
    // To cover the Windows case where the path with length over 260(MAX_PATH) chars,
    // set one path part to be 255 characters, so dir1/dir2/dir3/[longDirName] is over 260 chars.
    long longNameLength = 255;
#endif
```

so which one does AIX fall into? does it have `HAS_PATHCONF` in which case why can you not use it? if it does not why does the default value of 255 not work for you?

patel-parth7: AIX defines `HAS_PATHCONF`, so the code enters the corresponding conditional block. However, the `NAME_MAX` macro is not defined in the AIX system header (`sys/limits.h`). Therefore, this macro must be explicitly defined for AIX.

src/aws-cpp-sdk-core/source/platform/aix/Time.cpp
patel-parth7: Yes. `localtime_r` and `gmtime_r` are supported in AIX.

CMakeLists.txt
patel-parth7: CMake response file support is not enabled on AIX; therefore, it must be blocked.

tests/aws-cpp-sdk-core-tests/utils/FileSystemUtilsTest.cpp
patel-parth7: Please let me know if any action is required from my side regarding these changes.

sbiscigl:
sorry for the wait, looks good now though! will stage it in a PR that we can run CI on

sbiscigl:
https://github.com/aws/aws-sdk-cpp/pull/3794 is merged, closing, thanks for the contribution!

PR #3720, opened 2026-02-11
Waiter template + response code access for results

sbaluja:
*Issue #, if available:*

*Description of changes:*
- All result objects now contain the HttpResponse code and a getter function
- Base waiter template 

*Check all that applies:*

- [x] Linux
- [x] Windows
- [x] Android
- [x] MacOS
- [ ] IOS
- [ ] Other Platforms

### DIFF
--- a/generated/src/aws-cpp-sdk-signer-data/include/aws/signer-data/model/GetRevocationStatusResult.h
+++ b/generated/src/aws-cpp-sdk-signer-data/include/aws/signer-data/model/GetRevocationStatusResult.h
@@ -4,6 +4,7 @@
  */
 
 #pragma once
+#include <aws/core/http/HttpResponse.h>
 #include <aws/core/utils/memory/stl/AWSString.h>
 #include <aws/core/utils/memory/stl/AWSVector.h>
 #include <aws/signer-data/SignerData_EXPORTS.h>
@@ -71,10 +72,13 @@ class GetRevocationStatusResult {
     return *this;
   }
   ///@}
+  inline Aws::Http::HttpResponseCode GetHttpResponseCode() const { return m_HttpResponseCode; }
+
  private:
   Aws::Vector<Aws::String> m_revokedEntities;
 
   Aws::String m_requestId;
+  Aws::Http::HttpResponseCode m_HttpResponseCode;
   bool m_revokedEntitiesHasBeenSet = false;
   bool m_requestIdHasBeenSet = false;
 };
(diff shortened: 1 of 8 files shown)


src/aws-cpp-sdk-core/include/aws/core/utils/Waiter.h
```diff
@@ -0,0 +1,127 @@
...
+  MatcherType matcher;
+  ExpectedValue expected;
+};
+
+template <typename OutcomeT>
+using PathMatcher = std::function<bool(const OutcomeT&, const ExpectedValue&)>;
+
+template <typename RequestT, typename OutcomeT>
+class Waiter {
+  int delay;
```
sbiscigl: nit: lets move instance variables to `private` visibility and name them `m_*`

src/aws-cpp-sdk-core/include/aws/core/utils/Waiter.h
```diff
@@ -0,0 +1,127 @@
...
+      case MatcherType::PATH_ALL:
+      case MatcherType::PATH_ANY:
+        return pathMatcher ? pathMatcher(outcome, acceptor.expected) : false;
+    }
+    return false;
+  }
+
+ public:
+  Waiter(int delay, int maxAttempts, std::vector<Acceptor> acceptors, std::function<OutcomeT(const RequestT&)> op,
+         const Aws::String& waiterName = "Waiter", PathMatcher<OutcomeT> pathMatcher = nullptr)
```
sbiscigl: > PathMatcher<OutcomeT> pathMatcher = nullptr

you know i hate my null pointers. instead of using `nullptr` which will throw a exception if invoked (we wont see it because we have exceptions disabled) should we use a function that always false? that accurately matches the idea of "passing no path matcher, so nothing will ever be matched".

src/aws-cpp-sdk-core/include/aws/core/utils/Waiter.h
```diff
@@ -0,0 +1,127 @@
...
+        acceptors(std::move(acceptors)),
+        operation(std::move(op)),
+        name(waiterName),
+        pathMatcher(std::move(pathMatcher)) {}
+
+  WaiterOutcome<OutcomeT> Wait(const RequestT& request) {
+    for (int attempt = 0; attempt < maxAttempts; ++attempt) {
+      auto outcome = operation(request);
+
+      for (const auto& acceptor : acceptors) {
```
sbiscigl: nit: use prefer std algorithms over writing loop logic i.e. [std::find](https://en.cppreference.com/w/cpp/algorithm/find.html) so this would become
```c++
const auto matched = std::find_if(acceptors.begin(), 
  acceptors.end(), 
  [&outcome](const Acceptor& acceptor) -> bool { return Matches(acceptor, outcome} );
if(matched != acceptors.end()) {
  return outcome;
}
```

PR #3726, opened 2026-02-20
install test

kai-ion:
*Issue #, if available:*

*Description of changes:*

*Check all that applies:*

- [x] Linux
- [x] Windows
- [x] Android
- [x] MacOS
- [ ] IOS
- [ ] Other Platforms


tools/scripts/codegen/install_test_gen.py
```diff
@@ -0,0 +1,140 @@
...
+        self.debug = debug
+
+    def generate(self, clients_to_build: set):
+        """Generate install tests for the given client list
+        
+        :param clients_to_build: Set of client names to generate tests for
+        :return: 0 if success, non-zero otherwise
+        """
+        try:
+            client_list = list(clients_to_build) if clients_to_build else None
```
sbaluja: nit: leave as a set for faster lookup in _find_client_headers

tools/scripts/codegen/install_test_gen.py
```diff
@@ -0,0 +1,140 @@
...
+            return -1
+
+    def _find_client_headers(self, client_list=None):
+        headers = []
+        for dirpath, _, filenames in os.walk(GENERATED_SRC):
+            for fname in filenames:
+                if fname.endswith("Client.h"):
+                    full = os.path.join(dirpath, fname)
+                    parts = full.split(os.sep + "include" + os.sep, 1)
+                    if len(parts) == 2:
```
sbaluja: nit: can flatten the nesting here with an early continue:
```
if len(parts) != 2:
  continue
```

tools/scripts/codegen/install_test_gen.py
```diff
@@ -0,0 +1,140 @@
...
+                    print(f"Generated compilation test for {len(headers)} clients: {', '.join(client_list)}")
+                else:
+                    print(f"Generated compilation test for all {len(headers)} clients")
+            return 0
+        except Exception as e:
+            if self.debug:
+                print(f"Error generating install tests: {e}")
+            return -1
+
+    def _find_client_headers(self, client_list=None):
```
sbaluja: might want to add types to function params like you did for the above functions (this is missing from some other funcs too in this file). ie `client_list: List[str] = None`

tools/scripts/codegen/install_test_gen.py
```diff
@@ -0,0 +1,140 @@
...
+        headers = []
+        for dirpath, _, filenames in os.walk(GENERATED_SRC):
+            for fname in filenames:
```
sbaluja: do we need to loop through all the filenames in the generated/src directory? can we use a [`glob`](https://docs.python.org/3/library/glob.html#glob.glob) recursive filter to only loop through `*Client.h` files?
Ex.
```
for path in Path(GENERATED_SRC).rglob("*Client.h"):
# Path is now a Client.h filepath
# logic here 
```

tools/scripts/codegen/install_test_gen.py
kai-ion: good catch! removed list

tools/scripts/codegen/install_test_gen.py
kai-ion: Yup, adding an early continue ontop of the [glob](https://docs.python.org/3/library/glob.html#glob.glob) recursive filter to only loop through *Client.h files

PR #3740, opened 2026-03-05
sparse codegen 

kai-ion:
*Issue #, if available:*

*Description of changes:*
Change in codege so that services may return a null object for a type in a collection of objects.

*Check all that applies:*

- [x] Linux
- [x] Windows
- [x] Android
- [x] MacOS
- [ ] IOS
- [ ] Other Platforms


tools/code-generation/generator/src/main/resources/com/amazonaws/util/awsclientgenerator/velocity/cpp/json/ModelInternalMapOrListJsonDeserializer.vm
sbaluja, line 1: Is it fine that we're only worried about JSON services?

src/aws-cpp-sdk-core/include/aws/core/utils/json/JsonSerializer.h
```diff
@@ -94,6 +94,12 @@ namespace Aws
                 JsonValue& WithString(const Aws::String& key, const Aws::String& value);
                 JsonValue& WithString(const char* key, const Aws::String& value);
 
+                /**
+                 * Adds a null value to the top level of this node with key.
+                 */
+                JsonValue& WithNull(const Aws::String& key);
```
sbiscigl, line 100: [can you add unit tests that handle this code path](https://github.com/aws/aws-sdk-cpp/blob/main/tests/aws-cpp-sdk-core-tests/utils/json/JsonSerializerTest.cpp)?

tools/code-generation/generator/src/main/java/com/amazonaws/util/awsclientgenerator/domainmodels/codegeneration/cpp/CppViewHelper.java
```diff
@@ -225,12 +225,18 @@ else if(shape.isStructure() || shape.isEnum())
 
         else if(shape.isList()) {
             String type = computeCppTypeInternal(shape.getListMember().getShape(), typeMapping);
+            if (shape.isSparse()) {
```
sbiscigl: can a sparse shape _ONLY_ occur in a list or a map? if so, how can we fail gracefully if we encounter a case when it is not?

tools/scripts/run_code_generation.py
```diff
@@ -156,7 +156,10 @@ def main():
             return -1
 
         if args["generate_protocol_tests"]:
-            protocol_tests_generator = ProtocolTestsGen(args)
+            # Disable smithy generation for protocol tests
```
sbiscigl, line 164: why does this need a codegen change?

tools/code-generation/generator/src/main/java/com/amazonaws/util/awsclientgenerator/domainmodels/codegeneration/cpp/CppViewHelper.java
kai-ion: added a graceful fail where if shape is sparse and not list/map, return an exception during shape transformation
```
shape.setSparse(c2jShape.isSparse());
        if (shape.isSparse() && !shape.isList() && !shape.isMap()) {
            throw new SourceGenerationFailedException("The sparse trait is only applicable to list and map shapes, but was found on shape: " + shape.getName());
        }
```

src/aws-cpp-sdk-core/include/aws/core/utils/json/JsonSerializer.h
kai-ion: added a new test

tools/code-generation/generator/src/main/java/com/amazonaws/util/awsclientgenerator/domainmodels/codegeneration/cpp/CppViewHelper.java
```diff
@@ -225,12 +225,18 @@ else if(shape.isStructure() || shape.isEnum())
...
+                return String.format("Aws::Vector<Aws::Crt::Optional<%s>>", type);
```
sbiscigl: I think it would be cleaner and more correct to update `type` than creating branching return statements i.e.
```c++
else if(shape.isList()) {
  String type = computeCppTypeInternal(shape.getListMember().getShape(), typeMapping);
  type = shape.isSparse() ? String.format("Aws::Crt::Optional<%s>",  type) : type
  return String.format("Aws::Vector<%s>", type); 
}
```
ideally this should go into `computeCppTypeInternal` but that would require the signature for the method which is a bit much.

the operation on map can follow roughly the same logic.

tools/code-generation/generator/src/main/java/com/amazonaws/util/awsclientgenerator/generators/cpp/CppProtocolTestGenerator.java
```diff
@@ -120,6 +120,7 @@ protected final VelocityContext createContext() {
         } else {
             context.put("CppViewHelper", CppViewHelper.class);
         }
+        context.put("CppRequestViewHelper", CppViewHelper.class);
```
sbiscigl: im not following this, in the event of non-cbor protocols we will run
```java
context.put("CppViewHelper", CppViewHelper.class);
context.put("CppRequestViewHelper", CppViewHelper.class);
```
why would we add the same class under two different keys? we should only have to add this class to the context once. You likely want to add `CppViewHelper` regardless of the protocol now, so either you have to reconcile not including it twice, or reconcile why `CppCborViewHelper` exists and doesn't have the logic

tools/code-generation/generator/src/main/java/com/amazonaws/util/awsclientgenerator/domainmodels/codegeneration/cpp/CppViewHelper.java
kai-ion: Good call out, reverted the branching return and instead update type. Made the same change to update value in list

tools/code-generation/generator/src/main/java/com/amazonaws/util/awsclientgenerator/generators/cpp/CppProtocolTestGenerator.java
kai-ion: So originally I added the `Cpprequerstviewhelper` class because we can't use a single CppViewHelper class swap (setting it to CppCborViewHelper for CBOR) because request and result shapes intentionally use different type mappings for CBOR. Request classes use base types (int) to preserve API compatibility with existing CBOR services like CloudWatch, while result classes use CBOR types (int64_t). 

I agree we should only add the context once, as a workaround for the above issue I made the change to add `computeResultCppType(Shape shape, String protocol)` which picks the right type mapping internally:
```
public static String computeResultCppType(Shape shape, String protocol) {
    return "smithy-rpc-v2-cbor".equals(protocol)
            ? computeCppTypeInternal(shape, CORAL_TYPE_TO_CBOR_CPP_TYPE_MAPPING)
            : computeCppTypeInternal(shape, CORAL_TYPE_TO_CPP_TYPE_MAPPING);
}
```
Templates will now call `$CppViewHelper.computeResultCppType($shape, $serviceModel.metadata.protocol)`

tools/code-generation/generator/src/main/java/com/amazonaws/util/awsclientgenerator/domainmodels/codegeneration/cpp/CppViewHelper.java
```diff
@@ -56,6 +57,22 @@ public class CppViewHelper {
         CORAL_TYPE_TO_CPP_TYPE_MAPPING.put("sensitive_double", "double");
         CORAL_TYPE_TO_CPP_TYPE_MAPPING.put("sensitive_timestamp", "Aws::Utils::DateTime");
 
+        CORAL_TYPE_TO_CBOR_CPP_TYPE_MAPPING.put("long", "long long");
+        CORAL_TYPE_TO_CBOR_CPP_TYPE_MAPPING.put("integer", "int64_t");
+        CORAL_TYPE_TO_CBOR_CPP_TYPE_MAPPING.put("string", "Aws::String");
+        CORAL_TYPE_TO_CBOR_CPP_TYPE_MAPPING.put("timestamp", "Aws::Utils::DateTime");
+        CORAL_TYPE_TO_CBOR_CPP_TYPE_MAPPING.put("boolean", "bool");
+        CORAL_TYPE_TO_CBOR_CPP_TYPE_MAPPING.put("sensitive_boolean", "bool");
```
sbiscigl: summarizing what we talked about offline:

This logic is [copy and pasted from `CppCborViewHelper`](https://github.com/aws/aws-sdk-cpp/blob/main/tools/code-generation/generator/src/main/java/com/amazonaws/util/awsclientgenerator/domainmodels/codegeneration/cpp/CppCborViewHelper.java). There should only be one copy of this code, so it would make sense to remove `CppCborViewHelper` and move everything into `CppCborViewHelper`, so you dont need branching everything can use the same utils class.

Second `CORAL_TYPE_TO_CPP_TYPE_MAPPING` and `CORAL_TYPE_TO_CBOR_CPP_TYPE_MAPPING` differ only in the mapping of `integer` and `sensitive_integer`. to this end we should not duplicate the entire map you should use a tiered lookup and only have two values in the `CORAL_TYPE_TO_CBOR_CPP_TYPE_MAPPING` map. this would end up looking like

```java
CORAL_TYPE_TO_CBOR_CPP_TYPE_MAPPING. getOrDefault("key", 
  CORAL_TYPE_TO_CPP_TYPE_MAPPING.get("key"))
```

PR #3745, opened 2026-03-10
Cbor header fix for empty vs no payload

sbaluja:
*Issue #, if available:*

*Description of changes:*
- Send content type header if we are sending a payload, else we exclude it 

*Check all that applies:*

- [x] Linux
- [x] Windows
- [x] Android
- [x] MacOS
- [] IOS
- [ ] Other Platforms


tools/code-generation/generator/src/main/resources/com/amazonaws/util/awsclientgenerator/velocity/cpp/cbor/CborRequestSource.vm
```diff
@@ -80,6 +80,9 @@
         #foreach($headerEntry in $requestSpecificHeaders.entrySet())
             headers.emplace(${headerEntry.key}, ${headerEntry.value});
         #end
+        #if($operation.hasRequest())
```
sbiscigl, line 80: small nit - can we move this to above the for loop so there is no git diff on unchanged clients?

PR #3459, opened 2025-06-20
Place generated files in `CMAKE_BINARY_DIR`.

teo-tsirpanis:
*Issue #, if available:*

*Description of changes:*

This PR updates the CMake project to place generated files in the binary directory, instead of the source directory. This follows best practices, and unblocks the vcpkg port to support configuring `aws-sdk-cpp` in parallel.

*Check all that applies:*
- [x] Did a review by yourself.
- [ ] Added proper tests to cover this PR. (If tests are not applicable, explain.)
- [X] Checked if this PR is a breaking (APIs have been changed) change.
- [X] Checked if this PR will _not_ introduce cross-platform inconsistent behavior.
- [X] Checked if this PR would require a ReadMe/Wiki update.

Check which platforms you have built SDK on to verify the correctness of this PR.
- [ ] Linux
- [X] Windows
- [ ] Android
- [ ] MacOS
- [ ] IOS
- [ ] Other Platforms

By submitting this pull request, I confirm that you can use, modify, copy, and redistribute this contribution, under the terms of your choice.

PR #3747, opened 2026-03-12
Place generated files in `CMAKE_BINARY_DIR`.

teo-tsirpanis:
*Issue #, if available:*

*Description of changes:*

> [!NOTE]
> This is a reopening of #3459, which was closed for reasons unknown to me.

This PR updates the CMake project to place generated files in the binary directory, instead of the source directory. This follows best practices, and unblocks the vcpkg port to support configuring `aws-sdk-cpp` in parallel.

*Check all that applies:*

- [ ] Linux
- [X] Windows
- [ ] Android
- [ ] MacOS
- [ ] IOS
- [ ] Other Platforms

### DIFF
--- a/cmake/initialize_project_version.cmake
+++ b/cmake/initialize_project_version.cmake
@@ -6,9 +6,8 @@ if(GIT_FOUND)
 endif()
 
 if(NOT VERSION_STRING)
-    # extract it from the existing generated header file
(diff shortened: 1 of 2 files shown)


kai-ion:
@teo-tsirpanis 
I realize I closed this PR earlier, apologies for that. The comment I left was in response to another issue we had.

After taking another look at the original issue, I wanted to follow up on when you said 

> unblocks the vcpkg port to support configuring aws-sdk-cpp in parallel.

Do we have an example where the current behavior causes a failure? Since the change also touches include paths and install behavior, it would be helpful to understand whether this is addressing a concrete issue.

teo-tsirpanis:
The patch is being used in vcpkg since microsoft/vcpkg#46105 for more than half a year without problems.

Vcpkg builds ports in both debug and release mode, and to save time it configures them by default in parallel. Configuring from the same source tree in parallel doesn't work if a port writes to it, which according to my understanding is an anti-pattern either way.

PR #3748, opened 2026-03-14
Reduce size of generated client libraries

apache-hb:
Reduce the size of generated client libraries by deduplicating operation bodies. In my testing this reduces the size of client libraries by between 25% and 50% depending on the number of operations a service has. As an example `23M libaws-cpp-sdk-ec2-before.so` becomes `14M libaws-cpp-sdk-ec2-after.so`.

I didn't touch the code generation for streaming methods as they're not very common and are alot harder to test than the request response operations.


sbiscigl:
Hey @apache-hb I really like this. Let me run this through our CI and see if it flags anything, otherwise i dont see anything blocking us from accepting it.

sbiscigl:
running CI on it in https://github.com/aws/aws-sdk-cpp/pull/3751 will check back in with any failures

sbiscigl:
merged in https://github.com/aws/aws-sdk-cpp/pull/3751, thank you so much for this contribution!

PR #3798, opened 2026-04-16
Fix -Wunused-parameter error in codegen

kai-ion:
*Issue #, if available:*

*Description of changes:*
Fix -Wunused-parameter error in generated ClientConfiguration for services with no specific config

The LoadSpecificConfig codegen template produces an empty function body for services that don't match any conditional block (e.g. DynamoDB Streams), leaving inputProfileName unused. This fails compilation with -Werror.

Add AWS_UNREFERENCED_PARAM(inputProfileName) unconditionally to the template.

*Check all that applies:*

- [x] Linux
- [x] Windows
- [x] Android
- [x] MacOS
- [ ] IOS
- [ ] Other Platforms


tools/code-generation/generator/src/main/resources/com/amazonaws/util/awsclientgenerator/velocity/cpp/common/ServiceClientConfigurationSource.vm
```diff
@@ -73,6 +74,7 @@ bool IsEndpointDiscoveryEnabled(const Aws::String& endpointOverride, const Aws::
 #end
 void ${metadata.classNamePrefix}ClientConfiguration::Load${serviceNamespace}SpecificConfig(const Aws::String& inputProfileName)
 {
+  AWS_UNREFERENCED_PARAM(inputProfileName);
```
sbiscigl: you put this regardless if it is used or not, for instance in S3 you will put `AWS_UNREFERENCED_PARAM(inputProfileName);` then immediately use it. you should only use this if its truly unused.

tools/code-generation/generator/src/main/resources/com/amazonaws/util/awsclientgenerator/velocity/cpp/common/ServiceClientConfigurationSource.vm
```diff
@@ -151,6 +156,9 @@ void ${metadata.classNamePrefix}ClientConfiguration::Load${serviceNamespace}Spec
         authPreferences.insert(authPreferences.begin(), "bearer");
     }
 #end
+#if(!$inputProfileNameUsed)
+  (void)(inputProfileName);
```
sbiscigl: so in the [actual definition for the removed include]([ServiceClientConfigurationSource.vm‎](https://github.com/aws/aws-sdk-cpp/pull/3798/changes#diff-4df9801a18c629b976dcfe8786c594554e6780b05dad92e40d29800c53fa4944)tools/code-generation/generator/src/main/resources/com/amazonaws/util/awsclientgenerator/velocity/cpp/common/ServiceClientConfigurationSource.vm)
```
#if defined (_MSC_VER)

    #define AWS_UNREFERENCED_PARAM(x) (&reinterpret_cast<const int &>(x))

#else

    #define AWS_UNREFERENCED_PARAM(x) ((void)(x))

#endif // _MSC_VER
```

we treat missing argument differently on windows and unix systems, by only coying the unix bit, does this work on windows?

tools/code-generation/generator/src/main/resources/com/amazonaws/util/awsclientgenerator/velocity/cpp/common/ServiceClientConfigurationSource.vm
```diff
@@ -151,6 +157,14 @@ void ${metadata.classNamePrefix}ClientConfiguration::Load${serviceNamespace}Spec
         authPreferences.insert(authPreferences.begin(), "bearer");
     }
 #end
+#if(!$inputProfileNameUsed)
+\#if defined(_MSC_VER)
```
sbiscigl, line 160: you cant re-define the macro. you can either:
* include the header and use the macro ONLY if the parameter is unused.
* make this do the same thing as the macro which would be
```c++
#if defined(_MSC_VER)
  &reinterpret_cast<const int &>(inputProfileName)
#else
  (void)(inputProfileName)
#end
```

sbiscigl (approved):

PR #3800, opened 2026-04-17
Add expect100ContinueTimeoutMs to ClientConfiguration

kai-ion:
*Issue #, if available:*

*Description of changes:*

Expose CURLOPT_EXPECT_100_TIMEOUT_MS through ClientConfiguration so customers can configure the 100-Continue timeout without needing a custom HTTP client factory. This is useful when operating behind proxies that introduce network delays, where the default 1000ms timeout can cause IncompleteBody errors on uploads.

Default of 0 preserves existing behavior (curl's built-in 1000ms).

*Check all that applies:*

- [x] Linux
- [x] Windows
- [x] Android
- [x] MacOS
- [ ] IOS
- [ ] Other Platforms

### DIFF
--- a/src/aws-cpp-sdk-core/source/http/curl/CurlHandleContainer.cpp
+++ b/src/aws-cpp-sdk-core/source/http/curl/CurlHandleContainer.cpp
@@ -16,10 +16,10 @@ static const char* CURL_HANDLE_CONTAINER_TAG = "CurlHandleContainer";
 
 CurlHandleContainer::CurlHandleContainer(unsigned maxSize, long httpRequestTimeout, long connectTimeout, bool enableTcpKeepAlive,
                                         unsigned long tcpKeepAliveIntervalMs, long lowSpeedTime, unsigned long lowSpeedLimit,
-                                        Version version) :
+                                        Version version, long expect100ContinueTimeoutMs) :
                 m_maxPoolSize(maxSize), m_httpRequestTimeout(httpRequestTimeout), m_connectTimeout(connectTimeout), m_enableTcpKeepAlive(enableTcpKeepAlive),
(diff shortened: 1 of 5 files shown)


src/aws-cpp-sdk-core/source/http/curl/CurlHandleContainer.cpp
```diff
@@ -174,6 +174,10 @@ void CurlHandleContainer::SetDefaultOptionsOnHandle(CURL* handle)
     curl_easy_setopt(handle, CURLOPT_TCP_KEEPIDLE, m_tcpKeepAliveIntervalMs / 1000);
     curl_easy_setopt(handle, CURLOPT_HTTP_VERSION, ConvertHttpVersion(m_version));
     curl_easy_setopt(handle, CURLOPT_MAXCONNECTS, m_maxPoolSize);
+    if (m_expect100ContinueTimeoutMs > 0)
```
sbiscigl: according to [curl documentation](https://curl.se/libcurl/c/CURLOPT_EXPECT_100_TIMEOUT_MS.html) the default value is 1000, so why not just use that instead of using 0 as representation of value absent

src/aws-cpp-sdk-core/source/http/curl/CurlHandleContainer.cpp
```diff
@@ -174,6 +174,10 @@ void CurlHandleContainer::SetDefaultOptionsOnHandle(CURL* handle)
...
+    {
+        curl_easy_setopt(handle, CURLOPT_EXPECT_100_TIMEOUT_MS, m_expect100ContinueTimeoutMs);
+    }
```
sbaluja: I'd rather we just default the value to 1000ms in clientConfiguration so we can avoid this check

Issue #3257, opened 2025-01-17
High CPU Usage (100% per stream) in AWS Transcribe Streaming


blundercode:
### Describe the bug

The AWS Transcribe Streaming SDK C++ implementation is consuming excessive CPU resources when processing audio streams. Each individual stream consumes approximately 100% CPU usage, scaling linearly with multiple streams (e.g., 3 streams = 300% CPU usage). This appears inefficient for an operation that should primarily be handling audio data transmission to AWS Transcribe service.

I have tested using the CRT-HTTP version also and I get similar results. Will follow up with CRT-HTTP Docker version if requsted.

### Regression Issue

- [ ] Select this option if this issue appears to be a regression.

### Expected Behavior

1. Minimal CPU usage for streaming audio to AWS Transcribe service
2. Efficient handling of multiple concurrent streams without linear CPU scaling
3. CPU usage should primarily be focused on audio data transmission rather than processing

### Current Behavior

1. Each individual stream consumes 100% CPU
2. Multiple streams scale linearly (e.g., 3 streams = 300% CPU)
3. CPU usage monitored through top command shows excessive utilization
4. The high CPU usage persists throughout the entire streaming session
5. Behavior is consistent across multiple test runs

### Reproduction Steps

Here is the minimal reproduction steps in a single Dockerfile using the sample code.

Dockerfile
```Dockerfile
FROM public.ecr.aws/lts/ubuntu:22.04_stable

RUN apt-get update && \
  apt-get install build-essential cmake git libcurl4-openssl-dev zlib1g-dev libssl-dev curl ffmpeg -y

#Build sdk from source
RUN git clone --recurse-submodules https://github.com/aws/aws-sdk-cpp && \
    cd aws-sdk-cpp && \
    mkdir build && \
    cd build && \
    cmake .. -G "Unix Makefiles" -DBUILD_ONLY="transcribestreaming;transcribe" && \
    make install

#Build transcribe samples
RUN git clone https://github.com/awsdocs/aws-doc-sdk-examples.git && \
    cd aws-doc-sdk-examples/cpp/example_code/transcribe-streaming && \
    mkdir build && \
    cd build && \
    cmake .. -G "Unix Makefiles" && \
    make

# Download and convert the test file
RUN cd /aws-doc-sdk-examples/cpp/example_code/transcribe-streaming/.media && \
    rm -f transcribe-test-file.wav && \
[...]

PR #3809, opened 2026-04-27
use write data API in bidirectional streaming

sbiscigl:
*Issue #, if available:*

https://github.com/aws/aws-sdk-cpp/issues/3257

*Description of changes:*

This changes the implementation of bidirectional streaming when using the CRT HTTP client to use a push-based API rather than a polling-based API. Previously, bidirectional streaming used a `ConcurrentStreamBuf` that the HTTP client polled in a hot loop waiting for data/ Now, when built with `-DUSE_CRT_HTTP_CLIENT=ON`, the SDK uses `HttpWriteDataStreamBuf` which pushes data directly to the HTTP/2 connection the moment it's written. This is fully backwards compatible. No code changes required. The same streaming code automatically benefits:

```cpp
// HttpWriteDataStreamBuf is a std::streambuf that writes to an HTTP/2 connection
class HttpWriteDataStreamBuf : public std::streambuf {
  // When the buffer fills, push it to the wire immediately
  bool SendBuffer(bool endStream) {
    m_stream->WriteData(data, onComplete, endStream);  // CRT push API
  }
};

// AudioStream uses it as its underlying streambuf
class AudioStream : public EventEncoderStream {
  AudioStream(shared_ptr<HttpWriteDataStreamBuf> buf)
    : EventEncoderStream(buf) {}  // writes go directly to wire

  void WriteAudioEvent(const AudioEvent& event) {
    // encode + sign event, write to streambuf → WriteData → wire
    WriteEvent(encode(event));
  }
}

auto OnStreamReady = [](AudioStream& stream) {
  stream.WriteAudioEvent(audioEvent);  // now pushes directly to wire
  stream.Close();
};

client.StartStreamTranscriptionAsync(request, OnStreamReady, OnResponseCallback, nullptr);
```

This has been verified against the current [transcribe example](https://github.com/awsdocs/aws-doc-sdk-examples/blob/main/cpp/example_code/transcribe-streaming/get_transcript.cpp) where CPU usage dropped from 100% to background noise levels. This applies to all bidirectional streaming operations across all services (Transcribe, Bedrock, QBusiness, Polly, ConnectHealth, etc.) and is only active when built with `-DUSE_CRT_HTTP_CLIENT=ON`. The legacy path (libcurl/winhttp) is unchanged.

Additionally this is an ongoing effort where the CRT HTTP client will eventually become the default client of the SDK. This fix will not be ported to libcurl or winhttp.
 
*Check all that applies:*

- [x] Linux
- [x] Windows
- [x] Android
- [x] MacOS
- [ ] IOS
- [ ] Other Platforms

### DIFF
--- a/src/aws-cpp-sdk-core/include/aws/core/utils/stream/HttpWriteDataStreamBuf.h
+++ b/src/aws-cpp-sdk-core/include/aws/core/utils/stream/HttpWriteDataStreamBuf.h
@@ -1,0 +1,118 @@
+/**
+ * Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.
+ * SPDX-License-Identifier: Apache-2.0.
+ */
+#pragma once
+
+#include <aws/core/Core_EXPORTS.h>
+#include <aws/core/client/AWSError.h>
+#include <aws/core/client/CoreErrors.h>
+#include <aws/core/utils/Array.h>
+#include <aws/crt/Types.h>
+
+#include <condition_variable>
+#include <memory>
+#include <mutex>
+
+namespace Aws {
+namespace Http {
+class HttpClient;
+class HttpRequest;
+class HttpResponse;
+class Connection;
+class ClientStream;
+}  // namespace Http
+namespace Utils {
+namespace Stream {
+
+/**
+ * Write-only streambuf that pushes data to an HTTP/2 connection via ClientStream::WriteData.
+ *
+ * Lifecycle: construct → Initialize() → write via ostream → Close() → WaitForStreamComplete().
+ * Close() sends endStream=true. Destructor calls Close() + WaitForStreamComplete().
+ *
+ * Not copyable or movable. CRT HTTP client only.
+ */
+class AWS_CORE_API HttpWriteDataStreamBuf : public std::streambuf {
+ public:
+  explicit HttpWriteDataStreamBuf(const std::shared_ptr<Aws::Http::HttpClient>& client, size_t bufferLength = 8 * 1024);
+  HttpWriteDataStreamBuf(const HttpWriteDataStreamBuf& other) = delete;
+  HttpWriteDataStreamBuf(HttpWriteDataStreamBuf&& other) noexcept = delete;
+  HttpWriteDataStreamBuf& operator=(const HttpWriteDataStreamBuf& other) = delete;
+  HttpWriteDataStreamBuf& operator=(HttpWriteDataStreamBuf&& other) noexcept = delete;
+  ~HttpWriteDataStreamBuf() override;
+
+  /**
+   * Acquires a connection and creates a client stream. Must be called before any writes.
+   * Returns an error if connection acquisition or stream creation fails.
+   */
+  Aws::Crt::Optional<Aws::Client::AWSError<Aws::Client::CoreErrors>> Initialize(const std::shared_ptr<Aws::Http::HttpRequest>& request);
+
+  /**
+   * Flushes remaining buffer with endStream=true. Idempotent.
+   * Does not block on stream completion — call WaitForStreamComplete() for that.
+   */
+  void Close();
+
+  /**
+   * Returns the HTTP response from the underlying stream. Must be called after Initialize().
+   */
+  std::shared_ptr<Aws::Http::HttpResponse> GetResponse() const;
+
+  /**
+   * Blocks until the CRT fires the stream-complete callback.
+   * No-op if Initialize() was never called. Close() must be called first.
+   */
+  void WaitForStreamComplete();
+
+ protected:
+  // Write support
+  int_type overflow(int_type c) override;
+  std::streamsize xsputn(const char* s, std::streamsize n) override;
+  int sync() override;
+
+  // Disable reads
+  int_type underflow() override { return traits_type::eof(); }
+  int_type uflow() override { return traits_type::eof(); }
+  std::streamsize xsgetn(char*, std::streamsize) override { return 0; }
+  int_type pbackfail(int_type) override { return traits_type::eof(); }
+
+  // Disable seeking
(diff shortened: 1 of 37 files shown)


src/aws-cpp-sdk-core/source/utils/event/EventEncoderStream.cpp
```diff
@@ -2,10 +2,57 @@
...
 #include <aws/core/utils/event/EventEncoderStream.h>
-#include <iostream>
+#include <aws/core/utils/stream/HttpWriteDataStreamBuf.h>
 
-#include <aws/core/utils/HashingUtils.h>
+namespace Aws {
+namespace Utils {
+namespace Event {
+
+class DecoratedStreamBuf {
```
DmitriyMusatkin: nit: what does decorated mean in this context?

DmitriyMusatkin (approved):

src/aws-cpp-sdk-core/source/utils/event/EventEncoderStream.cpp
sbiscigl: its called out in the forward declaration in a comment in the header

```
            /**
             * Adapter that virtualizes streambuf lifecycle operations (close, drain)
             * so EventEncoderStream can work with different underlying streambufs
             * (e.g. ConcurrentStreamBuf for the pull model, HttpWriteDataStreamBuf
             * for the push model).
             */
```
i can call it `CloseableStreamBuf` so it matches better what its doing

src/aws-cpp-sdk-core/include/aws/core/utils/stream/HttpWriteDataStreamBuf.h
```diff
@@ -0,0 +1,117 @@
...
+  // Write support
+  int_type overflow(int_type c) override;
+  std::streamsize xsputn(const char* s, std::streamsize n) override;
+
+  // Disable reads
+  int_type underflow() override { return traits_type::eof(); }
+  int_type uflow() override { return traits_type::eof(); }
+  std::streamsize xsgetn(char*, std::streamsize) override { return 0; }
+  int_type pbackfail(int_type) override { return traits_type::eof(); }
+
```
sbaluja, line 79: `HttpWriteDataStreamBuf` doesn't override `sync()`, so `flush()` is a no-op and data stays in the put buffer until it fills (8kb) or `Close()` is called. `EventEncoderStream::WriteEvent` calls `flush()` after each event expecting it to push data to the wire. We should add:
```
int sync() override { return SendBuffer(false) ? 0 : -1; }
```
This would probably impact a scenario where we have data in the buffer less than 8kb, and the next input of data takes awhile: would be added latency in best case and a closed connection with not all data sent in the worst case. Note, we actually have overriden `sync()` for the current implementation of the [`ConcurrentStreamBuf`](https://github.com/aws/aws-sdk-cpp/blob/e7714de7a93f86b0a5d772bc5643f408b86ff18f/src/aws-cpp-sdk-core/include/aws/core/utils/stream/ConcurrentStreamBuf.h#L56) class for this reason

src/aws-cpp-sdk-core/source/utils/event/EventEncoderStream.cpp
```diff
@@ -2,10 +2,57 @@
...
 #include <aws/core/utils/event/EventEncoderStream.h>
-#include <iostream>
+#include <aws/core/utils/stream/HttpWriteDataStreamBuf.h>
 
-#include <aws/core/utils/HashingUtils.h>
+namespace Aws {
+namespace Utils {
+namespace Event {
+
+class CloseableStreamBuf {
```
sbaluja: nit: should we have these classes in a separate header/tu?

src/aws-cpp-sdk-core/include/aws/core/utils/stream/HttpWriteDataStreamBuf.h
sbiscigl: yeah good callout, fixing

src/aws-cpp-sdk-core/source/utils/event/EventEncoderStream.cpp
sbiscigl: yeah thought about that while doing this let me place these in a different header that isnt exported outside of the SDK, will be cleaner compile and still wont be exposed

sbaluja (approved):

src/aws-cpp-sdk-core/include/smithy/client/SmithyBidirectionalStreamingWriteDataTask.h
```diff
@@ -0,0 +1,197 @@
...
+        m_request.get(), m_request->GetServiceRequestName(), pExecutor, authResolver);
+    ctx->m_method = Aws::Http::HttpMethod::HTTP_POST;
+
+    bool resolved = false;
+    m_client->ResolveIdentityAuth(ctx,
+        [&resolved](AwsSmithyClientBase::HttpResponseOutcome&& outcome) {
+          if (!outcome.IsSuccess()) {
+            AWS_LOGSTREAM_ERROR("SmithyBidiWriteData", "Failed to resolve identity/auth");
+          }
+          resolved = true;
```
pulimsr: **nit**: `resolved` is set but never read. 
Also noticed the return value of `ResolveIdentityAuth` is discarded — it can return `false` after `m_awsIdentity` is already set (e.g. endpoint resolution or host validation failure) like in `AwsSmithyClientBase.cpp` [link](https://github.com/aws/aws-sdk-cpp/blob/main/src/aws-cpp-sdk-core/source/smithy/client/AwsSmithyClientBase.cpp#L190)

pulimsr (approved):

src/aws-cpp-sdk-core/include/smithy/client/SmithyBidirectionalStreamingWriteDataTask.h
sbiscigl: good catch! updating

PR #3825, opened 2026-05-08
aws-cpp-sdk-s3-crt: raise error when failing to write download data

### LINKED ISSUE #3824: [aws-cpp-sdk-s3-crt] downloads fail silently on local or streambuf write error
### Problem description

When downloading files with the `S3CrtClient`, write errors to the `Aws::Iostream` body are ignored (not even logged).
This silent failure is problematic, in particular when using a `TransferManager`, since it requires external validation to check the transfer.

The problem is that the `failbit/badbit` are not checked after `write`/`flush`:
```c++
// generated/src/aws-cpp-sdk-s3-crt/source/S3CrtClient.cpp

int S3CrtClient::S3CrtRequestGetBodyCallback(struct aws_s3_meta_request *meta_request, const struct aws_byte_cursor *body, uint64_t range_start, void *user_data)
{
  AWS_UNREFERENCED_PARAM(range_start);

  auto *userData = static_cast<S3CrtClient::CrtRequestCallbackUserData*>(user_data);
  if (!userData || !userData->response || !userData->request) {
    return AWS_OP_ERR;
  }
  auto& bodyStream = userData->response->GetResponseBody();

  bodyStream.write(reinterpret_cast<char*>(body->ptr), static_cast<std::streamsize>(body->len));
  if (userData->request->IsEventStreamRequest())
  {
    bodyStream.flush();
  }

   // Continues here without checking for errors.
   // ...
}
```

We realized this while testing downloads in parallel where one of the downloads ran into a `streambuf` error.

### What to do

Check `bodyStream.fail()`  and if `true`, return `AWS_OP_ERR`, logging the cause of the error.

###### AWS CPP SDK version used

Problem is present on `main`.

###### Compiler and Version used

clang >= 15

###### Operating System and version

ubuntu 22.x

grrtrr:
Currently the `S3CrtClient` neither logs nor raises an error when failing to write download data.

To avoid the problem of silent failure, check the `failbit/badbit` after the calls to `write()` and `flush()`.
Raise and log the error cause.
Return `AWS_OP_ERR` to mark the download as failed.

Resolves #3824.

Tested on Linux/ubuntu 22.04.

### DIFF
--- a/generated/src/aws-cpp-sdk-s3-crt/source/S3CrtClient.cpp
+++ b/generated/src/aws-cpp-sdk-s3-crt/source/S3CrtClient.cpp
@@ -577,6 +577,11 @@ int S3CrtClient::S3CrtRequestGetBodyCallback(struct aws_s3_meta_request *meta_re
   {
     bodyStream.flush();
   }
+  if (bodyStream.fail()) {
+    AWS_LOGSTREAM_ERROR(ALLOCATION_TAG, "Failed to write download data");
+    aws_raise_error(AWS_ERROR_FILE_WRITE_FAILURE);
+    return AWS_OP_ERR;
+  }
 
   // Replenish flow-control window (no-op if enable_read_backpressure is not set):
   aws_s3_meta_request_increment_read_window(meta_request, body->len);


grrtrr:
### Log output after applying this patch

With my failing download and this patch, I am now seeing this in the logs:
```console
[ERROR] 2026-05-08 19:54:41.884 S3CrtClient [139836522079808] Failed to write download data
[ERROR] 2026-05-08 19:54:41.884 S3MetaRequest [139836522079808] id=0x7f2e44c62400 Response body callback raised error 59 (Failed writing to file.).
```

sbiscigl:
closing in favor of https://github.com/aws/aws-sdk-cpp/pull/3826 where the commit was moved to thank you for the contribution!

PR #3836, opened 2026-06-01
[Feature] New Retry Behavior 

kai-ion:
*Issue #, if available:*

*Description of changes:*
Retry Behavior 2.1 core logic behind AWS_NEW_RETRIES_2026 flag

*Check all that applies:*

- [x] Linux
- [x] Windows
- [x] Android
- [x] MacOS
- [ ] IOS
- [ ] Other Platforms


src/aws-cpp-sdk-core/include/aws/core/client/RetryStrategy.h
sbiscigl: so can `m_retryCost` AND `m_throttlingRetryCost` ever be set and used at the same time? judging from `RetryCostClassification` you use that enum value to switch between cost implementations, when in reality you should be using a concept like poly morphism to swap implementations. i.e. how you have it now

```
enum class choice { A, B };

class implementation_choice {
public:
  implementation_choice(int a_impl_value, int b_impl_value, choice choice)
      : a_impl_value_(a_impl_value), b_impl_value_(b_impl_value), choice_(choice) {}
  void do_something() {
    if (choice_ == choice::A) {
      std::cout << a_impl_value_ << "\n";
    } else {
      std::cout << b_impl_value_ << "\n";
    }
  }
private:
  int a_impl_value_;
  int b_impl_value_;
  choice choice_;
};
```

this is storing both the variant of A and B on the object and making a runtime decesion. something much cleaner would be

```
class cost_applier {
public:
  virtual ~cost_applier() = default;
  virtual void apply_cost() = 0;
};

class a_cost_applier: public cost_applier {
 public:
  ~a_cost_applier() override;
  void apply_cost() override {
    std::cout << "applied a cost\n";
  }
};

class b_cost_applier: public cost_applier {
 public:
  ~b_cost_applier() override;
  void apply_cost() override {
    std::cout << "applied b cost\n";
  }
};

class implementation_choice {
public:
  implementation_choice(std::unique_ptr<cost_applier> cost_applier)
      : cost_applier_(std::move(cost_applier)) {}
  void do_something() {
    cost_applier_->apply_cost();
  }
private:
  std::unique_ptr<cost_applier> cost_applier_;
};
```

when we are looking for branching implementations we should lean on polymorphism instead of branch statments

src/aws-cpp-sdk-core/source/client/RetryStrategy.cpp
sbiscigl: why are you setting these items and they are not part of the initialization list?

sbiscigl: please remove AI generated comments

src/aws-cpp-sdk-core/include/aws/core/client/RetryStrategy.h
sbiscigl: RetryImpl is 
```
        struct StandardRetryStrategy::RetryImpl
        {
            bool newRetriesEnabled = false;
            double transientBackoffBaseSec = 0.05;
        };
```
why are we using Pimpl for this? typically Pimpl is used to point at a object that is expected to implement methods, this is just data members

src/aws-cpp-sdk-core/source/client/RetryStrategy.cpp
sbiscigl, line 65: formatting here is really off, should prefer to change only the code, not formatting so there is a clear dif

src/aws-cpp-sdk-core/source/client/ClientConfiguration.cpp
sbiscigl: why are we adding a new constructor specifically for the hardcoded value of ` 0.05`, that seems like it should be a implementation detail

sbiscigl: this really doesnt belong in `ClientConfiguration.cpp` this should be included somewhere in the core retry directory. _however_ you should mark the class as [`AWS_CORE_LOCAL`](https://github.com/aws/aws-sdk-cpp/blob/58442a4d2516aa235f86527b39ec60cf72cfc818/src/aws-cpp-sdk-core/include/aws/core/Core_EXPORTS.h#L31) this is local linkage within the core package, this should be in a header that _is not_ installed with the SDK however it should be available to build the SDK, making the symbol internal to the SDK.

src/aws-cpp-sdk-core/source/client/RetryStrategy.cpp
sbiscigl: so you are still using a bool instead of a proper design pattern for this we want to _change the implementation_ based on the value of the environment variable. for example consider a class that uses pimpl.
```c++
class Widget {
public:
    Widget();
    ~Widget();

    struct Impl;
private:
    std::unique_ptr<Impl> impl;
};
```

we want the `Impl` type to change based on the environment variables. i,e in widget.cpp

```c++
struct Widget::Impl {
    virtual ~Impl() = default;
};

namespace {

struct ImplA : Widget::Impl {};

struct ImplB : Widget::Impl {};
}

Widget::Widget() {
    const char* mode = std::getenv("WIDGET_MODE");
    std::string m = mode ? mode : "";

    if (m == "fast") {
        impl = std::make_unique<ImplA>();
    } else {
        impl = std::make_unique<ImplB>();
    }
}
```

where in this case `ImplA` is the old standard retry strategy defintion and `ImplB` is the new retry strategy definition

sbiscigl: nit: prefer anonymous namespace over static functions https://clang.llvm.org/extra/clang-tidy/checks/misc/use-anonymous-namespace.html

sbiscigl: nit use class over struct

tests/aws-cpp-sdk-core-tests/aws/client/RetryBehaviorTest.cpp
sbiscigl: why are we testing a class that we create in the test? we should be testing standard retry strategy/adaptive retry strategy that is where behavior changes

src/aws-cpp-sdk-core/include/aws/core/internal/RetryStrategyImpl.h
sbiscigl: remove static constants from the namespace unless they are acutally used in other places, we dont want to pollute the namespace

sbiscigl: why have to arguments and not three? they all have default values. a struct with default values seems like a better choice

```c++
struct QuotaConfig {
  int retryCost = 14;
  int throttlingCost = 5;
  int initialTokens = 500;
}

class ThrottleBasedRetryQuotaContainer {
  public:
    ThrottleBasedRetryQuotaContainer(const QuotaConfig& config): m_config{config};
  private:
    QuotaConfig m_config{};
}
```

src/aws-cpp-sdk-core/source/client/RetryStrategy.cpp
sbiscigl: why didnt this one get move to anon namepsace

sbiscigl: gotta be `(std::min)`  ([example](https://github.com/aws/aws-sdk-cpp/blob/6178e0532efd74fa8705784847211e0f9c74bcc6/src/aws-cpp-sdk-core/source/client/RetryStrategy.cpp#L89)) or otherwise windows might complain if this ends up in the same compile path as `windows.h`

sbiscigl: if its lower than 0 what should we realistically do? proceeding with math is definitely not what we should do. i.e what if "-999999999999999999" is returned?

src/aws-cpp-sdk-core/include/aws/core/internal/RetryStrategyImpl.h
sbiscigl: shouldn't you wrap this in a [reader lock](https://github.com/aws/aws-sdk-cpp/blob/main/src/aws-cpp-sdk-core/include/aws/core/utils/threading/ReaderWriterLock.h#L65)? since you are writing with a writer lock, seems it would be symmetrical to have this as a reader lock

src/aws-cpp-sdk-core/source/client/RetryStrategy.cpp
sbiscigl, line 27: so this new PIMPL only is a pointer for `CalculateDelay`, while im not against this, it would make sense if the this was properly PIMPL and it was the actual implementation of the retry strategy

sbiscigl (approved):

PR #3837, opened 2026-06-02
Schema types and ShapeSerializer interface

pulimsr:
Initial scaffolding for schema-based serialization adds the core Schema/FieldSchema data types and ShapeSerializer interface

*Check all that applies:*

- [x] Linux
- [x] Windows
- [ ] Android
- [x] MacOS
- [ ] IOS
- [ ] Other Platforms


src/aws-cpp-sdk-core/include/smithy/client/schema/CborShapeSerializer.h
```diff
@@ -0,0 +1,43 @@
...
+            void EndMap() override;
+
+            void BeginNestedStructure(const FieldSchema& field) override;
+            void EndNestedStructure() override;
+
+            Aws::String GetPayload() const;
+        private:
+            void WriteKey(const chat* key);
+
+            Aws::Crt::Cbor::CborEncoder m_encoder;
```
sbiscigl: so this would be a _prime_ example to use PIMPL to avoid including a crt header, i.e 
```c++
class AWS_CORE_API CborShapeSerializer {
  public:
    // ...
  private:
    struct CborEncoderImpl;
    std::unique_ptr<CborEncoderImpl> impl_;
}
```

then in the yet to be defined implementation you can do
```c++
struct CborShapeSerializer::CborEncoderImpl {
  Aws::Crt::Cbor::CborEncoder m_encoder;
};
```

this way consumers don't have a direct dependency on the CRT via the header

src/aws-cpp-sdk-core/include/smithy/client/schema/JsonShapeSerializer.h
```diff
@@ -0,0 +1,55 @@
...
+            void BeginMap(const FieldSchema& field,size_t count) override;
+            void WriteMapKey(const Aws::String& key) override;
+            void EndMap() override;
+
+            void BeginNestedStructure(const FieldSchema& field) override;
+            void EndNestedStructure() override;
+
+            Aws::Utils::Json::JsonValue GetResult() const { return m_root; }
+            Aws::String GetPayload() const;
+        private:
```
sbiscigl: same as cbor, lets make this PIMPL, should have one member that is a pointer, then in the .cpp define the actual implementation of it

src/aws-cpp-sdk-core/include/smithy/client/schema/JsonShapeSerializer.h
```diff
@@ -0,0 +1,55 @@
...
+            Aws::Utils::Json::JsonValue m_root;
+
+            struct StackEntry {
+                Aws::Utils::Json::JsonValue object;
+                const char* fieldName;
+                bool isList;
+                Aws::Utils::Array<Aws::Utils::Json::JsonValue> listItems;
+                size_t listIndex;
+            };
+            Aws::Vector<StackEntry> m_stack;
```
sbiscigl: this scares me a little bit but we can wait for a full implementation review. what is the stack used for, are we keeping a in memory representation of the json object? keeping a in memory representation of the object is something that we want to avoid that came up as we profiled this. we should be pre-allocating objects as well so that we dont pay a hefty alloc per object added to the stack

src/aws-cpp-sdk-core/include/smithy/client/schema/Schema.h
```diff
@@ -0,0 +1,74 @@
...
+            Blob,
+            Enum,
+            List,
+            Map,
+            Structure,
+            Union,
+            Document
+        };
+
+        enum class TimestampFormate : uint8_t {
```
sbiscigl: nit: typo `s/Formate/Format/g`

src/aws-cpp-sdk-core/include/smithy/client/schema/Schema.h
```diff
@@ -0,0 +1,74 @@
...
+            EpochSeconds,
+            DateTime,
+            HttpData
+        };
+
+        enum class FieldLocation : uint8_t {
+            Bode,
```
sbiscigl: typo: `Node`?

src/aws-cpp-sdk-core/source/smithy/client/schema/JsonShapeSerializer.cpp
```diff
@@ -0,0 +1,3 @@
+//
```
sbiscigl: lets remove this file until we actually use it

src/aws-cpp-sdk-core/include/smithy/client/schema/Schema.h
```diff
@@ -0,0 +1,74 @@
...
+            QueryString,
+            Header,
+            UriLabel,
+            StatusCode
+        };
+
+        struct Schema;
```
sbiscigl: so in the original design we only had `Schema` and that looked roughly like
```
class Schema { // Schema.h — lightweight shape descriptor 
public: 
    ShapeId getId() const; 
    ShapeType getType() const; 
    Schema getMember(const char* name) const; 
    Schema getMember(int position) const; 
    // traits, member iteration, etc. 
};
```

where a serializer would interact with it like
```
class ShapeSerializer { // ShapeSerializer.h — abstract, format-agnostic 
public: 
    virtual void writeString(const Schema& schema, const Aws::String& value) = 0; 
    virtual void writeInteger(const Schema& schema, int value) = 0; 
    virtual void writeStruct(const Schema& schema, const SerializableShape& value) = 0; 
    virtual void writeMap(const Schema& schema, ...) = 0; 
    virtual Aws::String flush() = 0; 
}; 
```

why do we now have two types FieldSchema and Schema? [lets just copye how java does it](https://github.com/smithy-lang/smithy-java/blob/main/codecs/json-codec/src/main/java/software/amazon/smithy/java/json/smithy/SmithyJsonSerializer.java#L214-L454) where all of their serializers take a schema object, for now, for this PR that schema object can be empty, and we can work out the implementation when we add out first implementation

src/aws-cpp-sdk-core/include/smithy/client/schema/CborShapeSerializer.h
```diff
@@ -0,0 +1,47 @@
...
+
+#include <memory>
+
+namespace smithy {
+namespace schema {
+
+class AWS_CORE_API CborShapeSerializer : public ShapeSerializer {
+ public:
+  CborShapeSerializer();
+  ~CborShapeSerializer();
```
sbiscigl, line 14: destructors need to be virtual or inheritance needs to be final

sbiscigl (approved):
approved, address my comment in the next iteration nsince this is very "define interfaces" step

src/aws-cpp-sdk-core/include/smithy/client/schema/CborShapeSerializer.h
pulimsr: making the inheritance final

PR #3842, opened 2026-06-09
adding the JsonShapeSerializer

pulimsr:
Adding the implementation for JsonShapeSerializer

*Check all that applies:*

- [x] Linux
- [x] Windows
- [ ] Android
- [x] MacOS
- [ ] IOS
- [ ] Other Platforms


src/aws-cpp-sdk-core/source/smithy/client/schema/JsonShapeSerializer.cpp
```diff
@@ -0,0 +1,209 @@
...
+    if (m_stack.empty()) return m_root;
+    return m_stack.back().object;
+  }
+
+  const char* CurrentKey(const Schema& schema) {
+    if (!m_stack.empty() && m_stack.back().isMap) return m_currentMapKey.c_str();
+    return schema.GetMemberName();
+  }
+
+  void BeginStructure(const Schema&) {}
```
sbiscigl: whats goin on here, shouldnt we be adding something to the stack when we begin/end a structure?

src/aws-cpp-sdk-core/source/smithy/client/schema/JsonShapeSerializer.cpp
```diff
@@ -0,0 +1,209 @@
...
+  void EndStructure() {}
+
+  void WriteBoolean(const Schema& schema, bool value) {
+    if (!m_stack.empty() && m_stack.back().isList) {
+      JsonValue v;
```
sbiscigl: i dont think we wanna declare a `JsonValue` value here because thats gunna make a heap allocation in cjson

src/aws-cpp-sdk-core/source/smithy/client/schema/JsonShapeSerializer.cpp
```diff
@@ -0,0 +1,209 @@
...
+
+  struct StackEntry {
+    JsonValue object;
+    const char* fieldName = nullptr;
+    bool isList = false;
+    bool isMap = false;
+    Aws::Vector<JsonValue> listItems;
+  };
+
+  Aws::Vector<StackEntry> m_stack;
```
sbiscigl: OK i see the grand picture of this and this matches our performance profile issues. This is actually much more simple than you are making it, lets start with what we are trying to do:
* we are prioritizing code gen to avoid a DOM/in memory tree representation of the object like cjson does
* we are allocating all of the memory up front (or at least try to) so that that we're only allocating and deallocating memory once.

So to start `JsonValue` will NOT be used in this class. `JsonValue` calls cjson which will have a multiple heap allocations as it traverses the tree. so that is out the window. 

We are going to make it simple we have a character buffer(your calling it a stack in this case), and we just write to it, nothing more.

The basic structure of this would look like
```c++
#include <vector>

class json_parser_no_dom {
public:
    json_parser_no_dom(size_t buffer_size) {
        buffer_.reserve(buffer_size);
    }
    json_parser_no_dom(const json_parser_no_dom &other) = delete;
    json_parser_no_dom(json_parser_no_dom &&other) noexcept = default;
    json_parser_no_dom & operator=(const json_parser_no_dom &other) = delete;
    json_parser_no_dom & operator=(json_parser_no_dom &&other) noexcept = default;

    //...
    std::string serialize() { return buffer_; }
private:
    std::string buffer_{};
};
```
where in this case the parser has a pre-allocated buffer to write stuff into. now we just put stuff into it.
```c++
class json_parser_no_dom {
public:
    void write_string(const std::string& str, const std::string& value) {
        buffer_.append("\"");
        buffer_.append(str);
        buffer_.append("\":");
        buffer_.append("\"");
        buffer_.append(value);
        buffer_.append("\"");
    }
};
```
where you could do something like
```c++
auto main() -> int {
    json_parser_no_dom json_parser{100};
    json_parser.begin_structure();
    json_parser.write_string("foo", "bar");
    json_parser.end_structure();
    std::cout << json_parser.serialize() << std::endl;
    return 0;
}
```

and see
```
{"foo":"bar"}
```

so at a very high level I think the problem here is that you are relying on `JsonValue` and cjson to do the hard parts, and that actually means you are performing heap allocations. what you want to do is rely on the schema for tree walking, and then you just dump the tree to a buffer.

src/aws-cpp-sdk-core/include/smithy/client/schema/JsonShapeSerializer.h
```diff
@@ -35,7 +35,7 @@ class JsonShapeSerializer final : public ShapeSerializer {
   void BeginNestedStructure(const Schema& schema) override;
   void EndNestedStructure() override;
 
-  Aws::String GetPayload() const;
+  const Aws::String& GetPayload() const;
```
sbiscigl: while returning by `const Aws::String&`avoids a copy, this would create a lifetime issues upstream where if `JsonShapeSerializer` goes out of scope you have a dangling ref. this also correlates to another comment i have which is that i dont like the temporary variables that we use to read into std::vector. so at a high level i think we have a few requirements:
* we want to avoid a copy when we call get payload
* after get payload is called, we can call get payload again, however the buffer at that point is "finalized" and we want to transfer ownership for the buffer to the called.
* we want to avoid temporary variables while writing values to underlying buffer.

So to solve all of these things i suggest that we:
* use std::string as the buffer, pre-allocated to 8KiB on the heap to avoid re-allocation.
*  we create a operation that returns the buffer string by move that invalidates the serializer object however avoid a copy.

```c++
enum class error {
    ALREADY_FINALIZED,
};

class buffer_builder {
public:
    buffer_builder() {
        buffer_.reserve(8192);
    }

    //methods to write to buffer

    std::expected<std::string, error> serialize() {
        if (state_ == state::FINALIZED) {
            return std::unexpected(error::ALREADY_FINALIZED);
        }
        state_ = state::FINALIZED;
        return std::move(buffer_);
    }

private:
    enum class state: uint8_t {
        INITIALIZED,
        FINALIZED,
    } state_{state::INITIALIZED};

    std::string buffer_;
};
```

src/aws-cpp-sdk-core/source/smithy/client/schema/JsonShapeSerializer.cpp
```diff
@@ -0,0 +1,193 @@
...
+#include <smithy/client/schema/JsonShapeSerializer.h>
+
+#include <cassert>
+
+using namespace smithy::schema;
+using namespace Aws::Utils;
+
+static const int MAX_DEPTH = 64;
+
+struct JsonShapeSerializer::Impl {
```
sbiscigl: should probably be a class that has a constructor and fields that are private to ensure encapsulation of the buffer.

src/aws-cpp-sdk-core/source/smithy/client/schema/JsonShapeSerializer.cpp
```diff
@@ -0,0 +1,193 @@
...
+ */
+#include <aws/core/utils/HashingUtils.h>
+#include <smithy/client/schema/JsonShapeSerializer.h>
+
+#include <cassert>
+
+using namespace smithy::schema;
+using namespace Aws::Utils;
+
+static const int MAX_DEPTH = 64;
```
sbiscigl: [cjson has this set at 1000](https://github.com/DaveGamble/cJSON/blob/master/cJSON.h#L134-L139), we should also likely set this the same. so we avoid any surprises

src/aws-cpp-sdk-core/source/smithy/client/schema/JsonShapeSerializer.cpp
```diff
@@ -0,0 +1,193 @@
...
+
+struct JsonShapeSerializer::Impl {
+  Aws::String m_buf;
+  int m_depth = 0;
+  bool m_needsComma[MAX_DEPTH] = {};
```
sbiscigl: perfer c++ `std::array` or `Aws::Array` over c arrays

src/aws-cpp-sdk-core/source/smithy/client/schema/JsonShapeSerializer.cpp
```diff
@@ -0,0 +1,193 @@
...
+  void WriteBoolean(const Schema& schema, bool value) {
+    WriteFieldName(schema);
+    m_buf += value ? "true" : "false";
+  }
+
+  void WriteInteger(const Schema& schema, int value) {
+    WriteFieldName(schema);
+    auto start = m_buf.size();
+    m_buf.resize(start + 11);
+    int written = snprintf(&m_buf[start],12,"%d", value);
```
sbiscigl: i think this goes for all of the operations but we should stop using `snprintf` and `.resize` alltogether. this is a string buffer. you should just use the until `to_string` https://github.com/aws/aws-sdk-cpp/blob/main/src/aws-cpp-sdk-core/include/aws/core/utils/StringUtils.h#L189-L198 and just append the the string to the buffer, allowing the string implementation to handle the underlying buffer resize not us.

tests/aws-cpp-sdk-core-tests/smithy/client/schema/JsonShapeSerializerTest.cpp
```diff
@@ -0,0 +1,364 @@
...
+ * Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.
+ * SPDX-License-Identifier: Apache-2.0.
+ */
+#include <aws/core/utils/DateTime.h>
+#include <aws/testing/AwsCppSdkGTestSuite.h>
+#include <smithy/client/schema/JsonShapeSerializer.h>
+
+using namespace smithy::schema;
+
+class JsonShapeSerializerTest : public Aws::Testing::AwsCppSdkGTestSuite {};
```
sbiscigl, line 11: Test cases we don't have that we should have:

* `MAX_DEPTH` enforcement
* json escaped data

src/aws-cpp-sdk-core/source/smithy/client/schema/JsonShapeSerializer.cpp
```diff
@@ -0,0 +1,193 @@
...
+
+  void WriteDouble(const Schema& schema, double value) {
+    WriteFieldName(schema);
+    auto start = m_buf.size();
+    m_buf.resize(start+32);
+    int writter = snprintf(&m_buf[start],32,"%f", value);
+    m_buf.resize(start + writter);
+  }
+
+  void WriteString(const Schema& schema, const Aws::String& value) {
```
sbiscigl, line 66: so in this case, we need to account for "json escaping strings", and if you dont, crafted json strings can break the string. smithy java extracts this to a utils file called [JsonWriteUtils](https://github.com/smithy-lang/smithy-java/blob/main/codecs/json-codec/src/main/java/software/amazon/smithy/java/json/smithy/JsonWriteUtils.java). and in the utils file the do json encoding. we idealliy should have a identical class called `JsonWriteUtils` that works the same way but is only internally visible to the SDK.

src/aws-cpp-sdk-core/source/smithy/client/schema/JsonShapeSerializer.cpp
```diff
@@ -0,0 +1,193 @@
...
+    m_isList[m_depth] = false;
+  }
+
+  void EndNestedStructure() {
+    m_depth--;
+    m_buf += '}';
+  }
+
+  Aws::String GetPayload() {
+    assert(!m_finalized);
```
sbiscigl: assertions are only enabled in debug mode, what happens in release mode? shouldnt we return a `Aws::Outcome` to represent error?

src/aws-cpp-sdk-core/include/smithy/client/schema/Schema.h
```diff
@@ -35,6 +35,8 @@ enum class ShapeType : uint8_t {
 class Schema {
  public:
   Schema() = default;
+  Schema(const char* memberName, ShapeType type)
```
sbiscigl: why not `const Aws::String& memberName`? lets avoid using c

sbiscigl (changes requested):

src/aws-cpp-sdk-core/include/smithy/client/schema/Schema.h
pulimsr: Done

src/aws-cpp-sdk-core/source/smithy/client/schema/JsonShapeSerializer.cpp
pulimsr: updated max depth to 1000

src/aws-cpp-sdk-core/source/smithy/client/schema/JsonShapeSerializer.cpp
pulimsr: `JsonShapeSerializer::Impl` is now a class

src/aws-cpp-sdk-core/source/smithy/client/schema/JsonShapeSerializer.cpp
pulimsr: now using `Aws::Array`

src/aws-cpp-sdk-core/source/smithy/client/schema/JsonShapeSerializer.cpp
pulimsr: using `to_string`

src/aws-cpp-sdk-core/source/smithy/client/schema/JsonShapeSerializer.cpp
pulimsr: added type `Aws::Outcome` as the result for `GetPayload()`

tests/aws-cpp-sdk-core-tests/smithy/client/schema/JsonShapeSerializerTest.cpp
pulimsr: added tests [here](https://github.com/aws/aws-sdk-cpp/pull/3842/changes#diff-df73add0e568608b5f780f065b627c3681f998898ed7713476cd82bb6a41ad3aR370)

src/aws-cpp-sdk-core/source/smithy/client/schema/JsonShapeSerializer.cpp
pulimsr: now using `JsonWriteUtils::WriteQuotedString(m_buf, value);`

src/aws-cpp-sdk-core/include/smithy/client/schema/JsonShapeSerializer.h
```diff
@@ -2,17 +2,20 @@
 
 #include <smithy/client/schema/ShapeSerializer.h>
 
-#include <memory>
+#include <aws/core/utils/memory/AWSMemory.h>
+
+#include "aws/core/client/AWSError.h"
```
sbiscigl: nit use: `#include <aws/core/client/AWSError.h>`

src/aws-cpp-sdk-core/source/smithy/client/schema/JsonShapeSerializer.cpp
```diff
@@ -0,0 +1,226 @@
...
+        false);
+    }
+    m_finalized = true;
+    return std::move(m_buf);
+  }
+
+private:
+  Aws::String m_buf;
+  int m_depth = 0;
+ Aws::Array<bool, MAX_DEPTH> m_needsComma{};
```
sbiscigl: nit: spacing

src/aws-cpp-sdk-core/CMakeLists.txt
```diff
@@ -543,7 +543,17 @@ check_cxx_source_compiles("
             return 0;
         }" AWS_HAS_ALIGNED_ALLOC)
 
-add_library(${PROJECT_NAME} ${AWS_NATIVE_SDK_SRC})
+add_library(${PROJECT_NAME} ${AWS_NATIVE_SDK_SRC}
+        include/smithy/client/schema/ShapeSerializer.h
```
sbiscigl: this isnt the right thing to do, look at how we do this for [other headers in the smithy directory](https://github.com/aws/aws-sdk-cpp/blob/main/src/aws-cpp-sdk-core/CMakeLists.txt#L80-L99)

src/aws-cpp-sdk-core/include/smithy/client/schema/JsonWriteUtils.h
```diff
@@ -0,0 +1,16 @@
...
+
+#include <aws/core/utils/memory/stl/AWSString.h>
+
+namespace smithy {
+    namespace schema {
+        class JsonWriteUtils {
+        public:
+            static void WriteQuotedString(Aws::String& buf, const Aws::String& value);
+        private:
+            static const bool NEEDS_ESCAPE[128];
```
sbiscigl: if these are static const, why do they need to be static in the `WriteQuotedString` class? seems like it would be a better fit for them to be a in anonymous namespace in `JsonWriteUtils.cpp`. also, again, prefer `std::array` to c style arrays

src/aws-cpp-sdk-core/include/smithy/client/schema/Schema.h
```diff
@@ -35,12 +35,14 @@ enum class ShapeType : uint8_t {
 class Schema {
  public:
   Schema() = default;
+  Schema(const Aws::String& memberName, ShapeType type)
+      : m_type(type), m_memberName(memberName) {}
 
   ShapeType GetType() const { return m_type; }
   const char* GetId() const { return m_id; }
-  const char* GetMemberName() const { return m_memberName; }
+  const Aws::String& GetMemberName() const { return m_memberName; }
```
sbiscigl: kind of sketched by `Aws::String&`, it could possibly create lifetime issues however, the way you use it wont. i would err on caution and return `Aws::String` because likely member name will be short string optimized and removes the lifetime issue not a NEEDS to happen but something to think about.

src/aws-cpp-sdk-core/source/smithy/client/schema/JsonWriteUtils.cpp
```diff
@@ -0,0 +1,80 @@
+#include <smithy/client/schema/JsonWriteUtils.h>
+
+using namespace smithy::schema;
+
+const char JsonWriteUtils::HEX[16] = { '0', '1', '2', '3', '4', '5', '6', '7', '8', '9', 'a', 'b', 'c', 'd', 'e', 'f'};
+
+const bool JsonWriteUtils::NEEDS_ESCAPE[128] = {
```
sbiscigl: so we create a table for both needs escape and then the escape value, im not opposed to this however might make more sense just to copy the [switch statement from cjson](https://github.com/DaveGamble/cJSON/blob/master/cJSON.c#L872-L924). correct me if im wrong but you more or less could just do that loop in our code, and avoid the size penalty (albeit small) on the translation unit.

src/aws-cpp-sdk-core/include/smithy/client/schema/JsonShapeSerializer.h
pulimsr: fixed

src/aws-cpp-sdk-core/source/smithy/client/schema/JsonShapeSerializer.cpp
pulimsr: ran clang format

src/aws-cpp-sdk-core/CMakeLists.txt
pulimsr: fixed this

src/aws-cpp-sdk-core/include/smithy/client/schema/JsonWriteUtils.h
pulimsr: moved to anonymous namespace in .cpp, using `Aws::Array` and switch statement

src/aws-cpp-sdk-core/include/smithy/client/schema/Schema.h
pulimsr: changed it to `Aws::String`

src/aws-cpp-sdk-core/source/smithy/client/schema/JsonWriteUtils.cpp
pulimsr: fixed

sbiscigl (approved):

PR #3843, opened 2026-06-10
Added Transfer Manager 2.0 public API types (Phase 1)

narahavi:
*Issue #, if available:*

*Description of changes:*

*Check all that applies:*

- [x] Linux
- [x] Windows
- [x] Android
- [x] MacOS
- [ ] IOS
- [ ] Other Platforms


src/aws-cpp-sdk-s3-transfer/source/s3-transfer/internal/CrtOperations.h
kai-ion, line 1: Header files should be in include/

kai-ion, line 5: You should use `#pragma once` like what you did above

kai-ion, line 2: `// Created by Narahari, Vinay on 6/5/26.`
yeahhhh, i dont think we want this comment

src/aws-cpp-sdk-s3-transfer/include/aws/s3-transfer/S3TransferManagerConfiguration.h
kai-ion, line 12: Is this an empty class? if we need to add members in the future we should make this into a pimpl

src/aws-cpp-sdk-s3-transfer/include/aws/s3-transfer/UploadResponse.h
kai-ion, line 31: we should add IsSuccess()/GetError() on the response to let customer know whether their transfer failed or succeeded

src/aws-cpp-sdk-s3-transfer/include/aws/s3-transfer/S3TransferManager.h
kai-ion, line 30: TM v1 uses a private constructor and static `shared_ptr<TransferManager> Create() factory`, which enables `shared_from_this` internally to prevent destruction while async CRT callbacks are in flight. Making this constructor public and deleting moves locks you out of that pattern later without a breaking API change. We should use a factory here

PR #3844, opened 2026-06-10
TM Header Update

narahavi:
*Issue #, if available:*

*Description of changes:*

*Check all that applies:*

- [x] Linux
- [x] Windows
- [x] Android
- [x] MacOS
- [ ] IOS
- [ ] Other Platforms


src/aws-cpp-sdk-s3-transfer/include/aws/s3-transfer/S3TransferManager.h
sbiscigl: why `std::enable_shared_from_this`, what we are doing here is saying "the only way to create a transfer manager is via a factory that creates a shared pointer". we do that in the previous transfer manager, why are we doing that again?

if possible we should allow to create a stack variable instead of requiring shared pointer

sbiscigl, line 27: copy ctor delete foresure, but why delete move?

src/aws-cpp-sdk-s3-transfer/include/aws/s3-transfer/UploadRequest.h
sbiscigl: so looking at this class you modeled it after most of our model classes that are codegenned. what i want to push back here on is do we need that? why are we following a builder patter when you could just have a constructor with default values? i.e. 

```c++
class UploadRequest {
  public:
    explicit UploadRequest(...): m_s3Request{...}
  private:
     Aws::S3::Model::PutObjectRequest m_s3Request;
     Aws::String m_sourceFilePath;
     std::shared_ptr<Aws::IOStream> m_body;
     Aws::Vector<std::shared_ptr<UploadProgressListener>> m_transferListeners;
}
```

because what you are doing right now is allowing someone to default construct this without a request, source file path or body, and if those are not set the request should fail. if possible we should cover up ways to mis-use this with design patterns.

src/aws-cpp-sdk-s3-transfer/include/aws/s3-transfer/DownloadRequest.h
sbiscigl: why are we calling this `SetBody` its not really setting the body, its setting a output stream factory, the reason this factory exists is so that we have a 0 copy option when downloading data from remote. this sort of goes back to our conversation of "how do this interact with aws-c-s3 to achieve 0 copy", so what is your plan with using this stream factory?

src/aws-cpp-sdk-s3-transfer/include/aws/s3-transfer/DownloadProgressListener.h
sbiscigl: clang-tidy: https://clang.llvm.org/extra/clang-tidy/checks/readability/named-parameter.html all parameters must be named

src/aws-cpp-sdk-s3-transfer/include/aws/s3-transfer/DownloadHandle.h
sbiscigl: why `shared_future`? we typically use use `std::future` to enforce that the thread calling it has to resolve it

sbiscigl: why shared_ptr and not unqiue?

sbiscigl, line 23: theres a rule here either declare the destructor virtual, or make the class final so i cannot be inherited from. if someone were to subclass this and refer to it at the base, it would get sliced. so as a rule of the thumb, always add a virutal destructor or make the class final

src/aws-cpp-sdk-s3-transfer/include/aws/s3-transfer/UploadResponse.h
sbiscigl, line 39: kinda weird to have a class that is only one member, since this is the first thing we do ill let it go for now but we should revist this existence if it only has one member

sbiscigl: use `using` over type def [cpp core guidelines](https://isocpp.github.io/CppCoreGuidelines/CppCoreGuidelines#t43-prefer-using-over-typedef-for-defining-aliases)

src/aws-cpp-sdk-s3-transfer/include/aws/s3-transfer/DownloadProgressSnapshot.h
sbiscigl: DownloadProgressSnapshot and UploadProgressSnapshot are about the same thing, we should likely break that out to a template i.e.

```c++
template<typename ReponseT>
class ProgressShapshot {
  ....
}

using DownloadProgressSnapshot = ProgressShapshot<ownloadResponse>;
```
etc

src/aws-cpp-sdk-s3-transfer/include/aws/s3-transfer/DownloadHandle.h
sbiscigl: needs to be `Aws::UniquePtr" because unique ptr is allocator aware

sbiscigl (approved):

src/aws-cpp-sdk-s3-transfer/include/aws/s3-transfer/S3TransferManagerConfiguration.h
sbaluja: Should this not be a struct that inherits from the `Aws::Client::GenericClientConfiguration`? See how we do this for other clients (ex. [dynamodb](https://github.com/aws/aws-sdk-cpp/blob/531fc621e620ee190dc0296806ff7aa0909748da/generated/src/aws-cpp-sdk-dynamodb/include/aws/dynamodb/DynamoDBClientConfiguration.h))

src/aws-cpp-sdk-s3-transfer/include/aws/s3-transfer/UploadHandle.h
sbaluja: nit: both the uploadhandle and downloadhandle have this comment saying it's freely copyable, but they are both move-only types (not copyable)

sbaluja (approved):

src/aws-cpp-sdk-s3-transfer/include/aws/s3-transfer/UploadRequest.h
sbaluja, line 27: One thing I want you to double check is that there are no additional types in the mapping-reference json that don't already exist in the PutObjectRequest and response types (UploadResponse)

PR #3846, opened 2026-06-11
copy tags and metadata in s3crt copy object when directive is defined

sbiscigl:
*Description of changes:*

Copies metadata and tags on a object. this is specifically helpful when CRT performs a MPU Copy Object.

*Check all that applies:*

- [x] Linux
- [x] Windows
- [x] Android
- [x] MacOS
- [ ] IOS
- [ ] Other Platforms

### DIFF
--- a/generated/src/aws-cpp-sdk-s3-crt/source/S3CrtClient.cpp
+++ b/generated/src/aws-cpp-sdk-s3-crt/source/S3CrtClient.cpp
@@ -819,6 +819,126 @@ void S3CrtClient::InitCommonCrtRequestOption(CrtRequestCallbackUserData* userDat
   options->finish_callback = S3CrtRequestFinishCallback;
 }
 
+Model::CopyObjectOutcome S3CrtClient::PopulateCopyObjectProperties(const Model::CopyObjectRequest& request,
+                                                                   const std::shared_ptr<Aws::Http::HttpRequest>& httpRequest) const {
+  const bool copyMetadata = request.MetadataDirectiveHasBeenSet() && request.GetMetadataDirective() == Model::MetadataDirective::COPY;
+  const bool copyTags = request.TaggingDirectiveHasBeenSet() && request.GetTaggingDirective() == Model::TaggingDirective::COPY;
+
+  if (!copyMetadata && !copyTags) {
+    return Model::CopyObjectOutcome(Model::CopyObjectResult());
+  }
+
+  // Parse "[/]bucket/key[?versionId=...]" from x-amz-copy-source. rfind: keys may contain '?'.
+  Aws::String copySource = request.GetCopySource();
+  Aws::String sourceVersionId;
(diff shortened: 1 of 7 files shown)


DmitriyMusatkin (approved):

pulimsr (approved):

generated/src/aws-cpp-sdk-s3-crt/source/S3CrtClient.cpp
```diff
@@ -915,6 +1030,12 @@ void S3CrtClient::CopyObjectAsync(const CopyObjectRequest& request, const CopyOb
                                                                         "Unable to create s3 meta request", false)),
                    handlerContext);
   }
+  {
+    auto copyPropertiesOutcome = PopulateCopyObjectProperties(request, userData->request);
+    if (!copyPropertiesOutcome.IsSuccess()) {
```
kai-ion, line 1040: If `PopulateCopyObjectProperties` fails and returns early, isnt userData never freed? I'm not seeing a cleanup of userData for early returns.

generated/src/aws-cpp-sdk-s3-crt/source/S3CrtClient.cpp
```diff
@@ -819,6 +819,121 @@ void S3CrtClient::InitCommonCrtRequestOption(CrtRequestCallbackUserData* userDat
...
+    }
+    for (const auto& item : head.GetMetadata()) {
+      httpRequest->SetHeaderValue("x-amz-meta-" + item.first, item.second);
+    }
+    // REPLACE so CreateMultipartUpload emits these headers.
+    httpRequest->SetHeaderValue("x-amz-metadata-directive", "REPLACE");
+  }
+
+  if (copyTags) {
+    Model::GetObjectTaggingRequest taggingRequest;
```
kai-ion, line 913: For a 1KB file that would have been a simple single-part copy, we would now be running `HeadObject` and `GetObjectTagging` before the CRT decides single-part vs multipart, right. And it would add extra latency.

Since for small objects the CRT arleady does a single-part copy where S3 already honors COPY directives natively. Do you think it would be a good idea if we check content-length from the HeadObject response vs the multipart threshold to skip the header injection for small objects, so we can reduce the extra round trip latency?

generated/src/aws-cpp-sdk-s3-crt/source/S3CrtClient.cpp
sbiscigl: yeah, let me add a delete

generated/src/aws-cpp-sdk-s3-crt/source/S3CrtClient.cpp
sbiscigl: updated

kai-ion (approved):

PR #3847, opened 2026-06-15
[Feature] Honor requestTimeoutMs for event stream writes via wait_until deadline

kai-ion:
*Issue #, if available:*

*Description of changes:*
This change adds a request-level deadline to HttpWriteDataStreamBuf.
  When requestTimeoutMs > 0, a deadline is computed at construction
  (now + requestTimeoutMs) and each SendBuffer call uses wait_until
  instead of wait. If the deadline passes, the write fails immediately.

  When requestTimeoutMs = 0 (default), behavior is unchanged.

*Check all that applies:*

- [x] Linux
- [x] Windows
- [x] Android
- [x] MacOS
- [ ] IOS
- [ ] Other Platforms

### DIFF
--- a/src/aws-cpp-sdk-core/include/aws/core/utils/stream/HttpWriteDataStreamBuf.h
+++ b/src/aws-cpp-sdk-core/include/aws/core/utils/stream/HttpWriteDataStreamBuf.h
@@ -10,6 +10,7 @@
 #include <aws/core/utils/Array.h>
 #include <aws/crt/Types.h>
 
+#include <chrono>
 #include <condition_variable>
 #include <memory>
 #include <mutex>
@@ -35,7 +36,7 @@ namespace Stream {
  */
 class AWS_CORE_API HttpWriteDataStreamBuf : public std::streambuf {
  public:
-  explicit HttpWriteDataStreamBuf(const std::shared_ptr<Aws::Http::HttpClient>& client, size_t bufferLength = 8 * 1024);
+  explicit HttpWriteDataStreamBuf(const std::shared_ptr<Aws::Http::HttpClient>& client, size_t bufferLength = 8 * 1024, size_t requestTimeoutMs = 0);
   HttpWriteDataStreamBuf(const HttpWriteDataStreamBuf& other) = delete;
   HttpWriteDataStreamBuf(HttpWriteDataStreamBuf&& other) noexcept = delete;
   HttpWriteDataStreamBuf& operator=(const HttpWriteDataStreamBuf& other) = delete;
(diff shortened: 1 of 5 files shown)


src/aws-cpp-sdk-core/source/utils/stream/HttpWriteDataStreamBuf.cpp
```diff
@@ -163,7 +168,18 @@ bool Aws::Utils::Stream::HttpWriteDataStreamBuf::SendBuffer(bool endStream) {
       endStream);
 
   std::unique_lock<std::mutex> lock{m_writeMutex};
-  m_writeComplete.wait(lock, [this]() -> bool { return !m_writeInProgress; });
+  if (m_requestTimeoutMs > 0) {
```
sbiscigl: why are we checking this again if we already check in the constructor?

src/aws-cpp-sdk-core/source/utils/stream/HttpWriteDataStreamBuf.cpp
```diff
@@ -5,15 +5,20 @@
...
 const char* WRITE_DATA_BUF_LOG_NAME = "HttpWriteDataStreamBuf";
 }
 
 Aws::Utils::Stream::HttpWriteDataStreamBuf::HttpWriteDataStreamBuf(const std::shared_ptr<Aws::Http::HttpClient>& client,
-                                                                   size_t bufferLength)
-    : m_client{client}, m_buffer{bufferLength} {
+                                                                   size_t bufferLength,
+                                                                   long requestTimeoutMs)
+    : m_client{client}, m_buffer{bufferLength}, m_requestTimeoutMs{requestTimeoutMs} {
+  if (m_requestTimeoutMs > 0) {
```
sbiscigl: so the c++ term for this would be a `invariant`, m_requestTimeoutMs cannot be less than 0. so you have two options:
* make `HttpWriteDataStreamBuf` a template. make  requestTimeoutMs a template parameter, and `static_assert` on the size so that it is caught at compile time.
* add a runtime assert on requestTimeoutMs, so on debug at runtime a assertion would be caught

src/aws-cpp-sdk-core/include/aws/core/utils/stream/HttpWriteDataStreamBuf.h
```diff
@@ -35,7 +36,7 @@ namespace Stream {
  */
 class AWS_CORE_API HttpWriteDataStreamBuf : public std::streambuf {
  public:
-  explicit HttpWriteDataStreamBuf(const std::shared_ptr<Aws::Http::HttpClient>& client, size_t bufferLength = 8 * 1024);
+  explicit HttpWriteDataStreamBuf(const std::shared_ptr<Aws::Http::HttpClient>& client, size_t bufferLength = 8 * 1024, long requestTimeoutMs = 0);
```
sbiscigl: how would 0/no timeout work here? seems like we would create a deadline that was right now

kai-ion: good point, we can address this by making `requestTimeoutMs = 0` set `m_hasDeadline = false`

sbiscigl (approved):

PR #3850, opened 2026-06-17
value-less query parameter parsing in URI canonicalization

pulimsr:
*Check all that applies:*

- [x] Linux
- [x] Windows
- [ ] Android
- [x] MacOS
- [ ] IOS
- [ ] Other Platforms


sbiscigl:
some general feedback:
* lets actually remove [this customization](https://github.com/aws/aws-sdk-cpp/blob/main/tools/code-generation/generator/src/main/java/com/amazonaws/util/awsclientgenerator/generators/cpp/s3/S3RestXmlCppClientGenerator.java#L177-L181) and test it to make sure that we do it right, can be a manual test or integration test.
* lets add a [URI test](https://github.com/aws/aws-sdk-cpp/blob/main/tests/aws-cpp-sdk-core-tests/http/URITest.cpp) that fails before and passes after

sbiscigl (approved):

PR #3855, opened 2026-06-30
Xml Shape Serializer - Schema serde 

pulimsr:
*Check all that applies:*

- [x] Linux
- [x] Windows
- [ ] Android
- [x] MacOS
- [ ] IOS
- [ ] Other Platforms


src/aws-cpp-sdk-core/include/smithy/client/schema/Trait.h
```diff
@@ -0,0 +1,12 @@
+#pragma once
+
+#include <smithy/Smithy_EXPORTS.h>
+
+namespace smithy {
+namespace schema {
+class SMITHY_API Trait {
```
sbiscigl, line 7: im always super skeptical of a interface that can be anything, can this actually be anything?

src/aws-cpp-sdk-core/source/smithy/client/schema/TraitKey.cpp
```diff
@@ -0,0 +1,10 @@
+#include <smithy/client/schema/TraitKey.h>
+
+namespace smithy {
+namespace schema {
+
+static int s_traitIdCounter = 0;
```
sbiscigl: this is a global counter between all `TraitKey`'s? what happens if two Trait keys are used in two threads?

src/aws-cpp-sdk-core/include/smithy/client/schema/TraitKey.h
```diff
@@ -0,0 +1,32 @@
...
+  int m_id;
+};
+
+SMITHY_API int NextTraitKey();
+
+template <typename T>
+class TraitKey : public TraitKeyBase {
+ public:
+  static const TraitKey& Instance() {
+    static const TraitKey instance(NextTraitKey());
```
sbiscigl: wait so you are making instance a monotonically growing int id to be unique? that seems sus, let me try to figure out what is going on here

src/aws-cpp-sdk-core/include/smithy/client/schema/TraitMap.h
```diff
@@ -0,0 +1,39 @@
...
+#include <smithy/client/schema/TraitKey.h>
+
+namespace smithy {
+namespace schema {
+class SMITHY_API TraitMap {
+ public:
+  TraitMap() = default;
+
+  template <typename T>
+  const T* Get(const TraitKey<T>& key) const {
```
sbiscigl: why does Get take a ` TraitKey<T>&` and has takes a `TraitKeyBase`? who own `T*` after its returned? if the map is deleted it is a segmentation fault

sbiscigl (commented):
before moving forward with any of the comments, lets try to figure out what the outcome here is that we want

src/aws-cpp-sdk-core/source/smithy/client/schema/TraitKey.cpp
sbiscigl: this should be std::atomic<int> to match how java does it

src/aws-cpp-sdk-core/include/smithy/client/schema/TraitMap.h
sbiscigl: lets return by value and copy or shared pointer if we want to share lifetimes, raw pointer or ref will create lifetime issues

sbiscigl (approved):
fix and ship

PR #3866, opened 2026-07-14
Single File Upload/Download

narahavi:
*Issue #, if available:*

*Description of changes:*
Single File Upload/Download for TM 2.0
*Check all that applies:*

- [x] Linux
- [x] Windows
- [x] Android
- [x] MacOS
- [ ] IOS
- [ ] Other Platforms


src/aws-cpp-sdk-s3-transfer/include/aws/s3-transfer/S3TransferManager.h
```diff
@@ -11,41 +11,44 @@
...
+
+class S3TransferManagerImpl;
+
+/**
+ * Customers construct an instance directly. The manager owns the underlying CRT client; it is
+ * neither copyable nor movable.
+ */
+class AWS_S3_TRANSFER_API S3TransferManager final {
+ public:
+  explicit S3TransferManager(const S3TransferManagerConfiguration& config);
```
sbaluja: What if someone wants to construct with credentials? A specific credentials provider?

src/aws-cpp-sdk-s3-transfer/include/aws/s3-transfer/DownloadRequest.h
```diff
@@ -17,25 +18,36 @@ namespace S3 {
...
     m_destinationFilePath(std::move(destinationFilePath)),
-    m_responseStreamFactory(std::move(responseStreamFactory)),
-    m_transferListeners(std::move(transferListeners)) {}
+    m_transferListeners(std::move(transferListeners)) {
+    assert(!m_destinationFilePath.empty() && "DownloadRequest destination file path must not be empty");
+  }
+
+  // Zero-copy stream download. Each part is delivered to the receiver in object order.
+  explicit DownloadRequest(
+    Aws::S3::Model::GetObjectRequest s3Request,
```
sbaluja: Not a big fan of requiring to construct with a different request type, thats not a preferable UX imo

src/aws-cpp-sdk-s3-transfer/source/s3-transfer/internal/S3TransferManagerImpl.cpp
```diff
@@ -0,0 +1,68 @@
...
+namespace S3 {
+namespace Transfer {
+
+static const char* const S3_TRANSFER_LOG_TAG = "S3TransferManagerImpl";
+
+S3TransferManagerImpl::S3TransferManagerImpl(const S3TransferManagerConfiguration& config)
+    : m_region(config.region) {
+  Aws::Crt::Auth::CredentialsProviderChainDefaultConfig providerConfig;
+  m_credentialsProvider =
+      Aws::Crt::Auth::CredentialsProvider::CreateCredentialsProviderChainDefault(providerConfig,
```
sbaluja: same thing as earlier, we need to allow the customer to pass in credentials, credentialsprovider, endpoint provider, stuff like that instead of just default constructing them

src/aws-cpp-sdk-s3-transfer/include/aws/s3-transfer/S3TransferManager.h
sbaluja: also should be able to default construct the client

src/aws-cpp-sdk-s3-transfer/source/s3-transfer/internal/CrtOperations.cpp
```diff
@@ -4,21 +4,729 @@
...
 namespace Internal {
 
-UploadHandle CrtOperations::DispatchUpload(S3TransferManagerImpl&, const UploadRequest&) {
-  return UploadHandle();
+static const char* const CRT_OPERATIONS_LOG_TAG = "CrtOperations";
+
+namespace {
+
+Aws::Client::CoreErrors MapCrtErrorCode(int crtErrorCode) {
+  switch (crtErrorCode) {
```
sbaluja: This is pretty much what we have in s3crt right

src/aws-cpp-sdk-s3-transfer/source/s3-transfer/internal/CrtOperations.cpp
```diff
@@ -4,21 +4,729 @@
...
 
-DownloadHandle CrtOperations::DispatchDownload(S3TransferManagerImpl&, const DownloadRequest&) {
-  return DownloadHandle();
+Aws::Client::AWSError<Aws::S3::S3Errors> MapCrtError(const Aws::Crt::S3::S3MetaRequestResult& result) {
+  auto httpRequest = Aws::MakeShared<Aws::Http::Standard::StandardHttpRequest>(
+      CRT_OPERATIONS_LOG_TAG, Aws::Http::URI("/"), Aws::Http::HttpMethod::HTTP_GET);
+  // StandardHttpResponse's constructor eagerly invokes the request's response-stream factory to
+  // materialize its body stream; without this the factory is empty and aborts under -fno-exceptions.
+  httpRequest->SetResponseStreamFactory(Aws::Utils::Stream::DefaultResponseStreamFactoryMethod);
+  std::shared_ptr<Aws::Http::HttpResponse> response =
```
sbaluja: Do we have to construct a temp request and a temp response? Can we not just construct an AwsError directly since that's all this function needs to return

tests/aws-cpp-sdk-s3-transfer-integration-tests/DownloadTests.cpp
```diff
@@ -0,0 +1,300 @@
...
+ * in download-single-object.json. The SEP cases assert wire-level GETs (part GETs / ranged GETs);
+ * TM 2.0 hands a single GET_OBJECT meta request to the CRT, which parallelizes internally. These
+ * tests assert the observable SEP "outcomes": success/error, totalBytes, byte-identical content
+ * written to the destination file, checksum-validation success, progress lifecycle, and cancel.
+ *
+ * NOTE: GetObjectResult (and therefore DownloadResponse / DownloadOutcome) is move-only. Never copy
+ * the outcome; bind it by value from future::get() (which moves) and read GetResult() by reference.
+ */
+#include "RecordingProgressListener.h"
+#include "S3TransferTestFixture.h"
```
sbaluja: use includes with <>

src/aws-cpp-sdk-s3-transfer/include/aws/s3-transfer/DownloadRequest.h
```diff
@@ -17,25 +18,36 @@ namespace S3 {
...
     Aws::S3::Model::GetObjectRequest s3Request,
     Aws::String destinationFilePath,
-    Aws::IOStreamFactory responseStreamFactory,
     Aws::Vector<std::shared_ptr<DownloadProgressListener>> transferListeners = {})
   : m_s3Request(std::move(s3Request)),
     m_destinationFilePath(std::move(destinationFilePath)),
-    m_responseStreamFactory(std::move(responseStreamFactory)),
-    m_transferListeners(std::move(transferListeners)) {}
+    m_transferListeners(std::move(transferListeners)) {
+    assert(!m_destinationFilePath.empty() && "DownloadRequest destination file path must not be empty");
```
sbaluja: Asserting here is good, but we should also check this in DispatchDownload and early error out

src/aws-cpp-sdk-s3-transfer/source/s3-transfer/internal/CrtOperations.cpp
narahavi: yep I took the crt to sdk error mapping from there

src/aws-cpp-sdk-s3-transfer/include/aws/s3-transfer/internal/TransferState.h
```diff
@@ -0,0 +1,79 @@
...
+
+// Per-transfer state shared by the CRT callbacks and the customer's handle. The promise is
+// fulfilled exactly once, inside the finish callback. PublishMetaRequest and CancelMetaRequest
+// synchronize through m_metaRequestLock so a cancel that races publication is never lost.
+template <typename RequestT, typename OutcomeT>
+struct TransferStateBase {
+  std::promise<OutcomeT> promise;
+  RequestT request;
+  uint64_t totalBytes = 0;
+  bool totalBytesHasBeenSet = false;
```
sbaluja: does totalBytes not need to be atomic?

tests/aws-cpp-sdk-s3-transfer-integration-tests/RecordingProgressListener.h
```diff
@@ -0,0 +1,65 @@
...
+ */
+template <typename ListenerBase, typename RequestT, typename SnapshotT>
+class RecordingProgressListenerT : public ListenerBase {
+ public:
+  std::atomic<int> initiatedCount{0};
+  std::atomic<int> bytesTransferredCount{0};
+  std::atomic<int> completeCount{0};
+  std::atomic<int> failedCount{0};
+  std::atomic<uint64_t> lastTransferredBytes{0};
+  std::atomic<uint64_t> maxTransferredBytes{0};
```
sbaluja, line 28: nit: if the values cannot be negative and are purely for counting, use size_t

src/aws-cpp-sdk-s3-transfer/include/aws/s3-transfer/internal/TransferState.h
```diff
@@ -0,0 +1,79 @@
...
+
+namespace Aws {
+namespace S3 {
+namespace Transfer {
+
+// Per-transfer state shared by the CRT callbacks and the customer's handle. The promise is
+// fulfilled exactly once, inside the finish callback. PublishMetaRequest and CancelMetaRequest
+// synchronize through m_metaRequestLock so a cancel that races publication is never lost.
+template <typename RequestT, typename OutcomeT>
+struct TransferStateBase {
```
sbaluja: internal only classes/structs should be marked with `AWS_CORE_LOCAL` so they are not installed

src/aws-cpp-sdk-s3-transfer/include/aws/s3-transfer/internal/TransferState.h
narahavi: totalBytes doesn't need to be atomic here because its only set once per dispatch for upload/download (download is set through the SetProgressCallback from crt). The user doesn't have a direct copy and outside of dispatch nothing ever updates or writes to it.

src/aws-cpp-sdk-s3-transfer/source/s3-transfer/internal/CrtOperations.cpp
```diff
@@ -4,21 +4,729 @@
...
+  }
+  if (result.errorResponseBody.ptr != nullptr && result.errorResponseBody.len > 0) {
+    response->GetResponseBody().write(reinterpret_cast<const char*>(result.errorResponseBody.ptr),
+                                      static_cast<std::streamsize>(result.errorResponseBody.len));
+  } else if (result.errorCode != AWS_ERROR_SUCCESS) {
+    // Bodyless CRT-side failure (e.g. response checksum mismatch): supply a typed client error so
+    // the marshaller produces a precise error instead of "No response body".
+    Aws::StringStream ss;
+    ss << aws_error_str(result.errorCode) << " (" << aws_error_lib_name(result.errorCode) << ": "
+       << aws_error_name(result.errorCode) << ")";
```
sbaluja: Use the error bindings here instead of the `aws_error_lib_name` etc

src/aws-cpp-sdk-s3-transfer/source/s3-transfer/internal/CrtOperations.cpp
narahavi: We could technically construct an `AwsError`, but we'd still need to derive the actual error type which happens at the return with `Aws::Client::S3ErrorMarshaller().BuildAWSError(response); `
`BuildAwsError` takes an `HttpResponse` as input so we build a temp one and populate it with the CRT's status/body to parse the XML body into the proper `S3Error`.

src/aws-cpp-sdk-s3-transfer/include/aws/s3-transfer/internal/S3TransferManagerImpl.h
```diff
@@ -0,0 +1,61 @@
...
+
+constexpr size_t DEFAULT_EXECUTOR_POOL_SIZE = 8;
+
+// Precondition: Aws::InitAPI() must be called before construction.
+class AWS_CORE_LOCAL S3TransferManagerImpl final {
+ public:
+  // Sole real constructor; the public overloads adapt their credentials shape into the
+  // AWSCredentialsProvider taken here. A null endpointProvider means "use the default
+  // S3EndpointProvider", matching generated-client behavior.
+  S3TransferManagerImpl(const std::shared_ptr<Aws::Auth::AWSCredentialsProvider>& credentialsProvider,
```
sbiscigl: > "null endpointProvider means "use the default S3EndpointProvider"

Never rely on null for behavior, however always check for null. this sounds like a perfect case for default parameters

```c++
S3TransferManagerImpl(const S3TransferManagerConfiguration& config,
  const std::shared_ptr<Aws::Auth::AWSCredentialsProvider>& credentialsProvider = Aws::MakeShared<DeafultCredentialsChain>(...),
  const std::shared_ptr<Aws::S3::Endpoint::S3EndpointProviderBase>& endpointProvider = Aws::MakeShared<S3EndpointProvider>(...))
```

since it is a internal header we dont need to worry about `#include`'s as much because this will ultimately be compiled away for a user.

src/aws-cpp-sdk-s3-transfer/include/aws/s3-transfer/internal/S3TransferManagerImpl.h
```diff
@@ -0,0 +1,61 @@
...
+  const std::shared_ptr<Aws::Crt::Auth::ICredentialsProvider>& GetCredentialsProvider() const {
+    return m_credentialsProvider;
+  }
+  bool IsValid() const { return m_crtClient != nullptr && static_cast<bool>(*m_crtClient); }
+  Aws::Utils::Threading::Executor& GetExecutor() const { return *m_executor; }
+
+ private:
+  Aws::String m_region;
+  std::shared_ptr<Aws::Crt::Auth::ICredentialsProvider> m_credentialsProvider;
+  std::shared_ptr<Aws::S3::Endpoint::S3EndpointProviderBase> m_endpointProvider;
```
sbiscigl, line 61: any reason these need to be shared and not unique?

src/aws-cpp-sdk-s3-transfer/include/aws/s3-transfer/internal/S3TransferManagerImpl.h
```diff
@@ -0,0 +1,61 @@
...
+  Aws::S3::Endpoint::S3EndpointProviderBase& GetEndpointProvider() const { return *m_endpointProvider; }
+  const Aws::String& GetRegion() const { return m_region; }
+  const std::shared_ptr<Aws::Crt::Auth::ICredentialsProvider>& GetCredentialsProvider() const {
+    return m_credentialsProvider;
+  }
+  bool IsValid() const { return m_crtClient != nullptr && static_cast<bool>(*m_crtClient); }
+  Aws::Utils::Threading::Executor& GetExecutor() const { return *m_executor; }
+
+ private:
+  Aws::String m_region;
```
sbiscigl: if this is the region from `S3TransferManagerConfiguration` in the constructor? you should likely just make the member `S3TransferManagerConfiguration` instead of a string. Adding more to the configuration and making a decision on it is almost a guaruntee.

src/aws-cpp-sdk-s3-transfer/include/aws/s3-transfer/internal/TransferState.h
```diff
@@ -0,0 +1,80 @@
...
+
+  void CancelMetaRequest() {
+    const std::lock_guard<std::mutex> lock(m_metaRequestLock);
+    if (m_metaRequest) {
+      m_metaRequest->Cancel();
+    }
+  }
+
+ private:
+  mutable std::mutex m_metaRequestLock;
```
sbiscigl: is `m_metaRequestLock` referenced in a `const` method? if not why is it mutable?

src/aws-cpp-sdk-s3-transfer/include/aws/s3-transfer/internal/TransferState.h
```diff
@@ -0,0 +1,80 @@
...
+namespace Aws {
+namespace S3 {
+namespace Transfer {
+
+// Per-transfer state shared by the CRT callbacks and the customer's handle. The promise is
+// fulfilled exactly once, inside the finish callback. PublishMetaRequest and CancelMetaRequest
+// synchronize through m_metaRequestLock so a cancel that races publication is never lost.
+template <typename RequestT, typename OutcomeT>
+struct AWS_CORE_LOCAL TransferStateBase {
+  std::promise<OutcomeT> promise;
```
sbiscigl, line 31: public member variables means that anyone can edit them, should create setters/getter for the ones that need to be expose, and expose them properly, anything else should be scoped to the class they live in

src/aws-cpp-sdk-s3-transfer/include/aws/s3-transfer/S3DownloadBuffer.h
```diff
@@ -0,0 +1,46 @@
...
+ */
+class AWS_S3_TRANSFER_API S3DownloadBuffer final {
+ public:
+  // The cursor carries the per-part body delivered by the CRT; the ticket is an owning reference
+  // that keeps the underlying CRT-owned buffer alive for as long as this object (or a copy) is alive.
+  explicit S3DownloadBuffer(std::shared_ptr<Aws::Crt::S3::S3BufferTicket> ticket,
+                            Aws::Crt::ByteCursor bytes, uint64_t rangeStart) noexcept;
+
+  S3DownloadBuffer(const S3DownloadBuffer&) = default;
+  S3DownloadBuffer& operator=(const S3DownloadBuffer&) = default;
```
sbiscigl: why does this need to be copyable?

src/aws-cpp-sdk-s3-transfer/include/aws/s3-transfer/S3TransferManager.h
```diff
@@ -9,43 +9,70 @@
...
+ * neither copyable nor movable.
+ */
+class AWS_S3_TRANSFER_API S3TransferManager final {
+ public:
+  /**
+   * Initializes the transfer manager with the DefaultCredentialProviderChain and an optional
+   * config. If config is not specified, it will be initialized to default values.
+   */
+  S3TransferManager(const S3TransferManagerConfiguration& config = S3TransferManagerConfiguration(),
+                    std::shared_ptr<Aws::S3::Endpoint::S3EndpointProviderBase> endpointProvider = nullptr);
```
sbiscigl: same comment as above, do use null as default, use the the default endpoint provider

src/aws-cpp-sdk-s3-transfer/include/aws/s3-transfer/UploadResponse.h
```diff
@@ -15,32 +15,29 @@ namespace S3 {
...
- * Response type returned via the UploadHandle's future once the transfer completes. Wraps
- * the underlying S3 PutObjectResult, populated from either the single PutObject response
- * or the CompleteMultipartUpload response depending on the path taken.
+ * Response type returned via the UploadHandle's future once the transfer completes. Wraps the
+ * underlying S3 PutObjectResult. The S3 result is set at construction time so an UploadResponse
+ * is never in a half-populated state; access it via GetS3Result().
  */
 class AWS_S3_TRANSFER_API UploadResponse final {
  public:
+  // Default constructor exists to satisfy Aws::Utils::Outcome<R, E>, which default-constructs
```
sbiscigl, line 24: ugh i hate that, but so it goes

src/aws-cpp-sdk-s3-transfer/source/s3-transfer/internal/S3TransferManagerImpl.cpp
```diff
@@ -0,0 +1,96 @@
...
+            return Aws::MakeShared<Aws::Crt::Auth::Credentials>(
+                S3_TRANSFER_LOG_TAG,
+                Aws::Crt::ByteCursorFromCString(sdkCreds.GetAWSAccessKeyId().c_str()),
+                Aws::Crt::ByteCursorFromCString(sdkCreds.GetAWSSecretKey().c_str()),
+                Aws::Crt::ByteCursorFromCString(sdkCreds.GetSessionToken().c_str()),
+                sdkCreds.GetExpiration().Millis());
+          }},
+      Aws::Crt::ApiAllocator());
+  if (!m_credentialsProvider) {
+    AWS_LOGSTREAM_ERROR(S3_TRANSFER_LOG_TAG, "Failed to create CRT credentials provider delegate.");
```
sbiscigl: so lets think about something here, at this point th constructed object is literally unusable, period, what is the mode of error to the customer that it is unusable? it looks like we could just call it and it would try to make a call. theres a few ways around surfacing this error:

create a state object that will tell you if you class is borked or not
```c++
class widget {
public:
    widget() {
        state_ = STATE::ERROR;        
    }
    
    std::expected<...> do_something() {
        if (state_ == STATE::ERROR) {
            return std::expected<...>
        }
    }

private:
    enum class STATE {
        INITIALZED,
        ERROR,
    } state_;
};
```

make a factory method the returns a optional if it is actually useable
```c++
class widget {
public:
    static std::expected<widget, std::error_code> create_widget() {
        if (...) {
            return std::unexpected<...>
        }
        widget widget{};
    }

private:
    widget() {}
};
```

src/aws-cpp-sdk-s3-transfer/source/s3-transfer/internal/S3TransferManagerImpl.cpp
sbaluja: Is it fine that we are leaving credentialsProvider uninitialized in this path? Would like to see this path tested in the integ tests, ie constructing with a provided client

src/aws-cpp-sdk-s3-transfer/source/s3-transfer/internal/CrtOperations.cpp
sbaluja, line 304: This is what i was alluding to when talking about the m_credentialsProvider being null, because now we're passing it down to this function

src/aws-cpp-sdk-s3-transfer/include/aws/s3-transfer/DownloadDataReceiver.h
sbaluja, line 15: Is this comment true? Is it ordered?

src/aws-cpp-sdk-s3-transfer/source/s3-transfer/internal/CrtOperations.cpp
sbaluja, line 763: One thing I wanna say is both dispatch download/upload are _massive_ functions, is there functionality we can split into helpers?

src/aws-cpp-sdk-s3-transfer/source/s3-transfer/S3TransferManager.cpp
sbaluja: Here is where we should probably pass the underlying impl pointer instead of moving the entire request object. Same for download

src/aws-cpp-sdk-s3-transfer/include/aws/s3-transfer/DownloadDataReceiver.h
narahavi: yea [meta request uses a priority queue](https://github.com/awslabs/aws-c-s3/blob/1f29ef8871a27dc8b90325418780659bac534d71/include/aws/s3/private/s3_meta_request_impl.h#L205-L207) to keep them in order

src/aws-cpp-sdk-s3-transfer/include/aws/s3-transfer/internal/DownloadRequestImpl.h
sbiscigl: i dont think we should be using default constructed error as error, thats kind of weird. `Validate` has to modes, either we have detected an error or no error. a default constructed error is still and error. a `Aws::Crt::Optional<Aws::Client::AWSError<Aws::S3::S3Errors>>` here makes more sense where a empty optional signals there was no error.

sbiscigl: always skeptical of returning void, can this operation fail? if it does how can we tell its caller?

src/aws-cpp-sdk-s3-transfer/source/s3-transfer/internal/CrtOperations.cpp
sbiscigl: prefer optional of string over empty string to denote missing. a empty string is valid header value, so in this case, empty vs missing matters.

sbiscigl (approved):
fix and ship

sbaluja (approved):

sbaluja (approved):

PR #3867, opened 2026-07-14
[Bug-Fix] Fix S3 multipart copy requiring s3:GetObjectVersion on versioned buckets

kai-ion:
*Issue #, if available:*

*Description of changes:*
  Stop auto-pinning the HeadObject-derived versionId onto UploadPartCopy
  requests in S3CrtClient::PopulateCopyObjectProperties. Rely on the ETag
  conditional (x-amz-copy-source-if-match) for source consistency instead.

  GetObjectAnnotation and GetObjectTagging continue to use versionId since
  they don't support ETag conditionals.

*Check all that applies:*

- [x] Linux
- [x] Windows
- [x] Android
- [x] MacOS
- [ ] IOS
- [ ] Other Platforms

### DIFF
--- a/generated/src/aws-cpp-sdk-s3-crt/source/S3CrtClient.cpp
+++ b/generated/src/aws-cpp-sdk-s3-crt/source/S3CrtClient.cpp
@@ -875,11 +875,6 @@ S3CrtClient::CopyObjectPropertiesOutcome S3CrtClient::PopulateCopyObjectProperti
   const auto& head = headOutcome.GetResult();
   const Aws::String sourceETag = head.GetETag();
 
-  // Pin the source version so UploadPartCopy reads a stable object.
-  if (sourceVersionId.empty() && !head.GetVersionId().empty()) {
-    sourceVersionId = head.GetVersionId();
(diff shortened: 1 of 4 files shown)


generated/src/aws-cpp-sdk-s3-crt/source/S3CrtClient.cpp
```diff
@@ -875,10 +875,10 @@ S3CrtClient::CopyObjectPropertiesOutcome S3CrtClient::PopulateCopyObjectProperti
   const auto& head = headOutcome.GetResult();
   const Aws::String sourceETag = head.GetETag();
 
-  // Pin the source version so UploadPartCopy reads a stable object.
+  // Detect source mutation via ETag conditional without requiring s3:GetObjectVersion.
+  // If the user explicitly provided sourceVersionId, it is already parsed above and preserved.
   if (sourceVersionId.empty() && !head.GetVersionId().empty()) {
     sourceVersionId = head.GetVersionId();
```
sbiscigl: you should remove this entire if statement. we want to honor the versionId set in the request, however not override it by the head response if it present.

sbiscigl (approved):

PR #3870, opened 2026-07-17
Create class based on bound out BDD endpoint engine in crt

sbiscigl:
*Description of changes:*

Binds out the [CRT BDD endpoint engine into the SDK](https://github.com/awslabs/aws-crt-cpp/blob/main/include/aws/crt/endpoints/BddEngine.h). refactors common components between different CRT endpoint provider implementations behind a template.

*Check all that applies:*

- [x] Linux
- [x] Windows
- [x] Android
- [x] MacOS
- [ ] IOS
- [ ] Other Platforms


tests/aws-cpp-sdk-s3-unit-tests/S3UnitTests.cpp
```diff
@@ -4,6 +4,7 @@
 #include <aws/core/utils/HashingUtils.h>
 #include <aws/core/utils/base64/Base64.h>
 #include <aws/core/utils/crypto/CRC64.h>
+#include <aws/core/utils/memory/stl/AWSArray.h>
```
kai-ion, line 7: nit: include but not used

pulimsr (approved):

src/aws-cpp-sdk-core/include/aws/core/endpoint/BDDEndpointProvider.h
```diff
@@ -0,0 +1,53 @@
...
+
+#include <aws/core/endpoint/internal/CrtEndpointProvider.h>
+
+#include <aws/crt/endpoints/BddEngine.h>
+
+namespace Aws
+{
+    namespace Endpoint
+    {
+        static const char BDD_ENDPOINT_PROVIDER_TAG[] = "Aws::Endpoint::BDDEndpointProvider";
```
sbaluja: nit: do we ever use this or pass it anywhere?

tests/aws-cpp-sdk-core-tests/endpoint/BDDEndpointProviderTest.cpp
```diff
@@ -0,0 +1,135 @@
...
+    auto result = provider->ResolveEndpoint(parameters);
+    EXPECT_FALSE(result.IsSuccess());
+}
+
+TEST_F(BDDEndpointProviderTest, ResolveEndpointWithDifferentRegions)
+{
+    auto provider = Aws::MakeShared<Aws::Endpoint::BDDEndpointProvider<>>("BDDEndpointProviderTest",
+        reinterpret_cast<const char*>(TestBddBytecode), TestBddBytecodeSize);
+
+    const std::vector<std::pair<Aws::String, Aws::String>> testCases = {
```
sbaluja: nit: Aws::Vector

src/aws-cpp-sdk-core/include/aws/core/endpoint/internal/CrtEndpointProvider.h
```diff
@@ -0,0 +1,99 @@
...
+                                               const EndpointParameters &,
+                                               const EndpointParameters &),
+            typename ClientConfigurationT = Aws::Client::GenericClientConfiguration,
+            typename BuiltInParametersT = Aws::Endpoint::BuiltInParameters,
+            typename ClientContextParametersT = Aws::Endpoint::ClientContextParameters>
+        class CrtEndpointProvider : public EndpointProviderBase<ClientConfigurationT, BuiltInParametersT,
+                    ClientContextParametersT> {
+        public:
+            CrtEndpointProvider(const char *endpointRulesBlob, const size_t endpointRulesBlobSz)
+                : m_crtEngine(Aws::Crt::ByteCursorFromArray((const uint8_t *) endpointRulesBlob, endpointRulesBlobSz),
```
sbaluja, line 45: One thing i wanna point out here since we are wrapping the raw pointer `endpointRulesBlob` in a `ByteCursor` (non-owning). The two engines slightly differ in construction:
- RuleEngine's [constructor](https://github.com/awslabs/aws-crt-cpp/blob/2343b63d8be7cbe209918e032f6204c75cd9a84e/source/endpoints/RuleEngine.cpp#L147) calls `aws_endpoints_ruleset_new_from_string` which parses the JSON into it's internal DOM (copy)
- BddEngine's constructor calls `aws_endpoints_bdd_engine_new_from_bytecode` on the partitions blob (copy), but explicitly states that the caller is [responsible for keeping the bytecode buffer alive](https://github.com/awslabs/aws-crt-cpp/blob/2343b63d8be7cbe209918e032f6204c75cd9a84e/include/aws/crt/endpoints/BddEngine.h#L22-L23)

For the BDD path, should we ensure the lifetime of that cursor by copying it into a `Aws::Vector<uint8_t>` or `Aws::String member` member variable before constructing?  

ex.

```
BDDEndpointProvider(const char* endpointRulesBlob, const size_t endpointRulesBlobSz)
      // Copy the bytecode into a member first
      : m_ownedBytecode(reinterpret_cast<const uint8_t*>(endpointRulesBlob),
                        reinterpret_cast<const uint8_t*>(endpointRulesBlob) + endpointRulesBlobSz),
        Base(reinterpret_cast<const char*>(m_ownedBytecode.data()), m_ownedBytecode.size())
```

tests/aws-cpp-sdk-s3-unit-tests/S3UnitTests.cpp
sbiscigl: it is used, the `DefaultEndpointProvider` included this transitively, now it doesnt, so where ever we are using it, we need to to include it now, i.e. same in transfer manager

src/aws-cpp-sdk-core/include/aws/core/endpoint/internal/CrtEndpointProvider.h
sbiscigl: good callout, and we solve this through our `constexpr` ruleset blob, that puts the data firmly in `.rodata` so we never have to worry about lifetimes. Lets address this the same as CRT and add a loud warning comment. Anyone who will newly subclass this for their own endpoint provider will have to do so explicitly the only path they would get it is if they didnt hav a custom endpoint provider, so it would be a new code path. Lets add a warning to match CRT.

sbaluja (approved):

PR #3871, opened 2026-07-20
CborShapeSerializer

pulimsr:
*Check all that applies:*

- [x] Linux
- [x] Windows
- [ ] Android
- [x] MacOS
- [ ] IOS
- [ ] Other Platforms


src/aws-cpp-sdk-core/source/smithy/client/schema/CborShapeSerializer.cpp
```diff
@@ -0,0 +1,267 @@
...
+  void WriteDoubleValue(double value) {
+    m_buf += static_cast<char>(TYPE_SIMPLE | ADDITIONAL_EIGHT_BYTES);
+    uint64_t bits;
+    std::memcpy(&bits, &value, sizeof(bits));
+    for (int i = 56; i >= 0; i -= 8) {
+      m_buf += static_cast<char>((bits >> i) & 0xFF);
+    }
+  }
+
+  void WriteTagAndLength(unsigned char majorType, uint64_t value) {
```
sbiscigl: ah i think you made this WAY harder than you needed to. The DOM issue that json has, cbor does not share, and you didnt have to implement CBOR at all, you can directly use CBOR from CRT and and its more or less a one-to-one mapping. i.e. [`WriteFloat`](https://github.com/awslabs/aws-crt-cpp/blob/main/include/aws/crt/cbor/Cbor.h#L106), [`WriteText`](https://github.com/awslabs/aws-crt-cpp/blob/main/include/aws/crt/cbor/Cbor.h#L117-L124)

CRT should have a one-to-one mapping for each important API that you are writing, if they dont we can talk to them about adding one, but we definitely should lean on their impl, and not re-implement it.

src/aws-cpp-sdk-core/source/smithy/client/schema/CborShapeSerializer.cpp
```diff
@@ -0,0 +1,210 @@
...
+    WriteFieldName(schema);
+    m_encoder.WriteFloat(value);
+  }
+
+  void WriteString(const Schema& schema, const Aws::String& value) {
+    WriteFieldName(schema);
+    m_encoder.WriteText(Aws::Crt::ByteCursorFromArray(reinterpret_cast<const uint8_t*>(value.data()), value.size()));
+  }
+
+  void WriteTimestamp(const Schema& schema, const DateTime& value) {
```
sbiscigl: we should prolly keep this like the way we had it initially i.e. from the [protocol test](https://github.com/aws/aws-sdk-cpp/blob/main/generated/protocol-tests/test-clients/aws-cpp-sdk-rpcv2protocol/source/model/RpcV2CborListsRequest.cpp#L91-L92)

just so that we are not changing the wire format at all and using what we did before. so this would become

```c++
void WriteTimestamp(const Schema& schema, const DateTime& value) {
  WriteFieldName(schema);
  encoder.WriteTag(1);
  encoder.WriteUInt(item_0.Seconds());
}
```

this way it stays exactly the same

src/aws-cpp-sdk-core/source/smithy/client/schema/CborShapeSerializer.cpp
```diff
@@ -0,0 +1,210 @@
...
+    }
+    m_finalized = true;
+    auto encoded = m_encoder.GetEncodedData();
+    return Aws::String(reinterpret_cast<const char*>(encoded.ptr), encoded.len);
+  }
+
+ private:
+  Aws::Crt::Cbor::CborEncoder m_encoder;
+  int m_depth = 0;
+  Aws::Array<bool, MAX_DEPTH> m_isMap{};
```
sbiscigl: [they already bounds check for us](https://github.com/awslabs/aws-c-common/blob/main/source/cbor.c#L18-L19) so i would just remove all bounds checking logic and fall back to theirs. also then we save logic and space on the object

sbiscigl (approved):

Issue #3607, opened 2025-11-03
AddressSanitizer: heap-use-after-free during shutdown of Logger


bjosv:
### Describe the bug

During shutdown of the CRT; if a thread takes longer time to respond than expected we might get a use after free. This can be seen when running a build using the address sanitizer (`-DENABLE_ADDRESS_SANITIZER=ON`, see logs below)

The issue seems to be predicted in the following comment:
https://github.com/aws/aws-sdk-cpp/pull/2885/files#r1567933028

### Regression Issue

- [ ] Select this option if this issue appears to be a regression.

### Expected Behavior

Not intermittently having use after free issues.

### Current Behavior

Here are an output from a run when `SUMMARY: AddressSanitizer: heap-use-after-free` is hit during shutdown.

```
Aws::Utils::Logging::s_aws_logger_redirect_get_log_level(aws_logger*, unsigned int) /path/aws-sdk-cpp/src/aws-cpp-sdk-core/source/utils/logging/CRTLogging.cpp:59:42
    #1 0x7f2a17209f67 in s_destroy /path/aws-c-io/source/linux/epoll_event_loop.c:233:5
    #2 0x7f2a171e87de in s_aws_event_loop_group_shutdown_sync /path/aws-c-io/source/event_loop.c:43:13
    #3 0x7f2a171ebde5 in s_event_loop_destroy_async_thread_fn /path/aws-c-io/source/event_loop.c:55:5
    #4 0x7f2a18a53bfb in thread_fn /path/aws-c-common/source/posix/thread.c:177:5
    #5 0x7f2a1bcfdaf6  (/usr/local/lib/clang/21/lib/x86_64-unknown-linux-gnu/libclang_rt.asan.so+0x147af6)
    #6 0x7f2a19b8af6b in start_thread /usr/src/debug/glibc-2.38-150600.14.37.1.x86_64/nptl/pthread_create.c:444:8
    #7 0x7f2a19c12387 in __GI___clone3 /usr/src/debug/glibc-2.38-150600.14.37.1.x86_64/misc/../sysdeps/unix/sysv/linux/x86_64/clone3.S:78

0x7b5a153e2120 is located 16 bytes inside of 24-byte region [0x7b5a153e2110,0x7b5a153e2128)
freed by thread T0 here:
[...]

PR #3874, opened 2026-07-22
Survive static destruction order failure in AWSLogging accessors

makxenov:
*Description of changes:*
The logger is held in two namespace-scope shared_ptr statics. AWS clients owned by other translation units can be destroyed after these statics (the order of static destruction across translation units is unspecified) and may still call GetLogSystem()/Push/PopLogger while unwinding at process exit. Once the shared_ptrs have been destroyed, those calls touch freed objects.

Move the two shared_ptrs into a LoggingSet struct guarded by a trivially destructible bool flag, LoggingSetAlive, which stays readable until the very end of the process. ~LoggingSet clears the flag; afterwards the accessors behave as if no logger is installed instead of reading the destroyed members.

*Check all that applies:*
- [ ] Added proper tests to cover this PR. (If tests are not applicable, explain: **the order on destruction is unspecified**)

- [x] Linux
- [ ] Windows
- [ ] Android
- [ ] MacOS
- [ ] IOS
- [ ] Other Platforms

### DIFF
--- a/src/aws-cpp-sdk-core/source/utils/logging/AWSLogging.cpp
+++ b/src/aws-cpp-sdk-core/source/utils/logging/AWSLogging.cpp
@@ -14,8 +14,28 @@
 using namespace Aws::Utils;
 using namespace Aws::Utils::Logging;
 
-static std::shared_ptr<LogSystemInterface> AWSLogSystem(nullptr);
-static std::shared_ptr<LogSystemInterface> OldLogger(nullptr);
+namespace {
+
+// AWS clients owned by other globals can be destroyed after this translation
+// unit's statics (static destruction order across translation units is
+// unspecified) and still log through GetLogSystem() while unwinding. The flag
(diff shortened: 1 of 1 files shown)


sbaluja:
Hi @makxenov. We are aware of this use-after-free behavior (issue #3607), and there currently exists a [commented out test](https://github.com/aws/aws-sdk-cpp/blob/94d55124985f2d08788d0f7e121256360d0bef6c/tests/aws-cpp-sdk-core-tests/utils/logging/LoggingTest.cpp#L642-L698) which is there to demonstrate this exact use-after-free behavior you're describing. I've ran that test against your branch, and after about 20 runs the test failed, demonstrating that this does not fix the issue. I suggest using that test as a gate to a proper fix, and once you have the test passing for a large number of iterations, we'll be happy to review the PR

sbiscigl:
closing this PR however feel free to re-open if you can address @sbaluja 's comments

PR #3877, opened 2026-07-23
adding the Cbor Deserializer

pulimsr:
*Check all that applies:*

- [x] Linux
- [x] Windows
- [ ] Android
- [x] MacOS
- [ ] IOS
- [ ] Other Platforms

### DIFF
--- a/src/aws-cpp-sdk-core/source/smithy/client/schema/CborShapeDeserializer.cpp
+++ b/src/aws-cpp-sdk-core/source/smithy/client/schema/CborShapeDeserializer.cpp
@@ -1,0 +1,159 @@
+/**
+ * Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.
+ * SPDX-License-Identifier: Apache-2.0.
+ */
+#include <aws/crt/cbor/Cbor.h>
+#include <smithy/client/schema/CborShapeDeserializer.h>
+
+using namespace smithy::schema;
+using namespace Aws::Utils;
+using namespace Aws::Crt::Cbor;
+
+class CborShapeDeserializer::Impl {
+ public:
+  Impl(const unsigned char* data, size_t length)
+      : m_decoder(Aws::Crt::ByteCursorFromArray(data, length)) {}
+
+  bool ReadBoolean() {
+    auto val = m_decoder.PopNextBooleanVal();
+    return val.has_value() ? val.value() : false;
+  }
+
+  int ReadInteger() { return static_cast<int>(ReadLong()); }
+
+  int64_t ReadLong() {
+    auto type = m_decoder.PeekType();
+    if (!type.has_value()) {
+      return 0;
+    }
+    if (*type == CborType::UInt) {
+      auto val = m_decoder.PopNextUnsignedIntVal();
+      return val.has_value() ? static_cast<int64_t>(val.value()) : 0;
+    }
+    if (*type == CborType::NegInt) {
+      auto val = m_decoder.PopNextNegativeIntVal();
+      return val.has_value() ? static_cast<int64_t>(-(static_cast<int64_t>(val.value()) + 1)) : 0;
+    }
+    return 0;
+  }
+
+  float ReadFloat() { return static_cast<float>(ReadDouble()); }
+
+  double ReadDouble() {
+    auto type = m_decoder.PeekType();
+    if (!type.has_value()) {
+      return 0.0;
+    }
+    if (*type == CborType::Float) {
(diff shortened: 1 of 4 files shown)


src/aws-cpp-sdk-core/source/smithy/client/schema/CborShapeDeserializer.cpp
```diff
@@ -0,0 +1,159 @@
...
+    }
+    return 0;
+  }
+
+  float ReadFloat() { return static_cast<float>(ReadDouble()); }
+
+  double ReadDouble() {
+    auto type = m_decoder.PeekType();
+    if (!type.has_value()) {
+      return 0.0;
```
sbiscigl: let think about error handling here, every `Read*` has a 

```c++
if (!type.has_value()) {
  return default value
}
```

branch in it. this one specifically sticks out to me because you would not be able to tell if the value ways truly `0.0` or a error. at the `ShapeDeserializer` interface level should we be returning a optoinal, maybe a outcome with a error that shows it was a deserialization error? we definitely need some sort of signal to the called that a error occurred

src/aws-cpp-sdk-core/source/smithy/client/schema/CborShapeDeserializer.cpp
```diff
@@ -0,0 +1,159 @@
...
+  }
+
+  int ReadEnum() { return ReadInteger(); }
+
+  void BeginStruct() {
+    auto type = m_decoder.PeekType();
+    if (type.has_value() && *type == CborType::IndefMapStart) {
+      m_decoder.ConsumeNextSingleElement();
+    } else {
+      m_decoder.PopNextMapStart();
```
sbaluja: How come we're discarding the map size for definite length struct? Server can send definite length here, and then our `EndStruct` that's relying on the`IsBreak` loop wouldn't work right? Lmk if im missing something

src/aws-cpp-sdk-core/source/smithy/client/schema/CborShapeDeserializer.cpp
```diff
@@ -0,0 +1,159 @@
...
+    if (!val.has_value()) {
+      return {};
+    }
+    return Aws::String(reinterpret_cast<const char*>(val->ptr), val->len);
+  }
+
+  DateTime ReadTimestamp() {
+    auto tag = m_decoder.PopNextTagVal();
+    (void)tag;
+    return DateTime(static_cast<double>(ReadLong()));
```
sbaluja: Timestamps can also be floats. Also, `ReadLong()` can return a negative value. Negative timestamps per the [RFC](https://www.rfc-editor.org/rfc/rfc8949.html#epochdatetimesect): 
> Negative values (major type 1 and negative floating-point numbers) are interpreted as determined by the application requirements as there is no universal standard for UTC count-of-seconds time before 1970-01-01T00:00Z 

What [we currently do](https://github.com/aws/aws-sdk-cpp/blob/c9539d3ee8574d115248c73e0c435c925c7eb148/generated/src/aws-cpp-sdk-compute-optimizer-automation/source/model/AccountInfo.cpp#L117-L128) ensures that if we encounter a negative value we error out

src/aws-cpp-sdk-core/source/smithy/client/schema/CborShapeDeserializer.cpp
pulimsr: Okay, I'll either add per-field Optional or an error flag like CRT's [decoder->error_code](https://github.com/awslabs/aws-c-common/blob/main/source/cbor.c#L309):
```
d.BeginStruct();
 d.EndStruct();
 if (d.HasError()) return MakeError(d.GetError());
```

src/aws-cpp-sdk-core/source/smithy/client/schema/CborShapeDeserializer.cpp
pulimsr: You're right — the CRT doesn't synthesize a break for definite-length maps (it uses count-based iteration internally but doesn't expose it to PeekType callers). I'll change `BeginStruct` to return `size_t`, same as `BeginList`/`BeginMap` already do.

src/aws-cpp-sdk-core/source/smithy/client/schema/CborShapeDeserializer.cpp
pulimsr: I'll handle the float case and add validation for negative values.

sbiscigl (approved):

Issue #3520, opened 2025-08-19
Aws::Environment::GetEnv() ignores windows system code page which fails SDK-tests for german users


maximilian-loehr-at-ipolog-ai:
### Describe the bug

Downloaded & built the C++ sdk but the `FileSystemUtilsTest` fails.

This is because `DirectoryTreeTest::SetUp()` tries to create a directory in the user's home folder but messes up character conversion in `AWS::FileSystem::CreateDirectoryIfNotExists()`.
This conversion happens in `StringUtils::ToWString()`  which assumes that its char* input is UTF-8, c.f. `MultiByteToWideChar(CP_UTF8, ...`.

The problem is that `AWS::Environment::GetEnv()` uses `_dupenv_s()` which returns characters in the system code page. My name contains an 'ö', which is 0xF6 in Windows_1252, or 0x00F6 in UTF-16, but 0xC3 0xB6 in UTF-8.
The AWS::String returned from `Aws::FileSystem::GetHomeDirectory()` contains a single 0xF6 byte for that 'ö', which matches my Windows_1252 code page.
Naturally, the AWS::String containing Windows_1252 cannot be interpreted as UTF-8 without errors.

### Regression Issue

- [ ] Select this option if this issue appears to be a regression.

### Expected Behavior

All tests should run on a clean build of the C++ sdk.

### Current Behavior

Several tests involving the user name (e.g. home directory) fail on a windows system with windows_1252 code page if the user has special characters in their name.

### Reproduction Steps

auto homeDirectory = Aws::FileSystem::GetHomeDirectory();
ASSERT_FALSE(homeDirectory.empty());

auto dir1 = Aws::FileSystem::Join(homeDirectory, "dir1");
bool dir1Created = Aws::FileSystem::CreateDirectoryIfNotExists(dir1.c_str());
ASSERT_TRUE(dir1Created);  // will fail if user has umlauts in their name and uses win_1252 code page

### Possible Solution

The solution depends on what `AWS::String` is expected to hold:
- if `AWS::String` is expected to always hold UTF-8 characters, then `AWS::Environment::GetEnv()` needs an additional step to verify that the current code page is UTF-8 (`GetACP() == CP_UTF8`) or convert the retrieved string into UTF-8 before returning it
- if `AWS::String` is expected to always contain the "native" character set, then `StringUtils::ToWString()` must not blindly assume UTF-8 but use the actual code page, e.g. `MultiByteToWideChar( GetACP(), ...`
[...]

PR #3885, opened 2026-07-30
Fix Windows GetEnv mangling non-ASCII environment values (UTF-8)

kai-ion:
*Issue #, if available:*
https://github.com/aws/aws-sdk-cpp/issues/3520

*Description of changes:*
On Windows, `Aws::Environment::GetEnv` read variables through the narrow
  `_dupenv_s`, which returns the value in the active ANSI code page (e.g.
  CP-1252). Any non-ASCII value — an accented username in a home path, a
  region/proxy set to a localized string — was then reinterpreted elsewhere in
  the SDK as UTF-8 and came out as mojibake (e.g. `José` → `Jos\xE9` instead of
  `Jos\xC3\xA9`).

  This changes the Windows implementation to read through the wide `_wdupenv_s`
  and convert explicitly to UTF-8 via `StringUtils::FromWString`, since Windows
  stores environment variables natively as UTF-16. This mirrors the same fix
  made in the CRT (awslabs/aws-c-common#1260). `_wdupenv_s` is
  MSVC-specific, so non-MSVC compilers keep the existing `std::getenv` path.

  Added a regression test (`FileSystemUtilsTest.GetEnvReturnsNonAsciiValueAsUtf8`)
  that injects a non-ASCII value via the wide CRT API (`_wputenv_s`, so the true
  Unicode value is stored rather than a pre-mangled narrow one) and asserts
  `GetEnv` returns the correct UTF-8 bytes. The test is Windows-only and skips
  elsewhere.

*Check all that applies:*

- [x] Linux
- [x] Windows
- [x] Android
- [x] MacOS
- [ ] IOS
- [ ] Other Platforms

### DIFF
--- a/tests/aws-cpp-sdk-core-tests/utils/FileSystemUtilsTest.cpp
+++ b/tests/aws-cpp-sdk-core-tests/utils/FileSystemUtilsTest.cpp
@@ -3,11 +3,15 @@
  * SPDX-License-Identifier: Apache-2.0.
  */
 
+#include <aws/core/platform/Environment.h>
 #include <aws/core/platform/FileSystem.h>
 #include <aws/core/utils/FileSystemUtils.h>
 #include <aws/core/utils/memory/stl/AWSSet.h>
 #include <aws/testing/AwsCppSdkGTestSuite.h>
 #include <fstream>
+#ifdef _MSC_VER
+#include <stdlib.h> // _wputenv_s
+#endif
 #if defined(HAS_PATHCONF)
 #include <unistd.h>
 #include <climits>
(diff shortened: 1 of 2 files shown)


tests/aws-cpp-sdk-core-tests/utils/FileSystemUtilsTest.cpp
```diff
@@ -32,6 +36,22 @@ TEST_F(FileTest, HomeDirectory)
...
+    const char* varName = "AWS_SDK_TEST_NONASCII";
+    // Set via the wide CRT API so the true Unicode value is stored, not a pre-mangled narrow one.
+    ASSERT_EQ(0, _wputenv_s(L"AWS_SDK_TEST_NONASCII", L"José"));
+
+    const auto value = Aws::Environment::GetEnv(varName);
+    _wputenv_s(L"AWS_SDK_TEST_NONASCII", L"");
+
+    ASSERT_EQ(Aws::String("Jos\xC3\xA9"), value); // "José" in UTF-8
+#else
+    GTEST_SKIP() << "Non-ASCII environment encoding fix is Windows-specific.";
```
sbiscigl: instead of skipping, just ifdef the whole test
```c++
#ifdef _MSC_VER
TEST_F(FileTest, GetEnvReturnsNonAsciiValueAsUtf8) {
  ....
}
#endif
```
we do that in several other places, i.e. [http tests](https://github.com/aws/aws-sdk-cpp/blob/main/tests/aws-cpp-sdk-core-tests/http/HttpClientTest.cpp#L113-L148) for libcurl only tests

sbiscigl (approved):

PR #3889, opened 2026-08-03
Adding JsonShapeDeserializer

pulimsr:
*Description of changes:* This PR adds the JsonShapeDeserializer for schema-driven JSON deserialization, adds TimestampFormatTrait and fixes double/float precision  

*Check all that applies:*

- [x] Linux
- [x] Windows
- [ ] Android
- [x] MacOS
- [ ] IOS
- [ ] Other Platforms

### DIFF
--- a/src/aws-cpp-sdk-core/source/smithy/client/schema/JsonShapeDeserializer.cpp
+++ b/src/aws-cpp-sdk-core/source/smithy/client/schema/JsonShapeDeserializer.cpp
@@ -1,0 +1,418 @@
+/**
+ * Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.
+ * SPDX-License-Identifier: Apache-2.0.
+ */
+#include <aws/core/utils/HashingUtils.h>
+#include <smithy/client/schema/JsonShapeDeserializer.h>
+
+#include <cstdlib>
+#include <cstring>
+
+using namespace smithy::schema;
+using namespace Aws::Utils;
+
+namespace {
+
+enum class Token {
+  ObjectStart,
+  ObjectEnd,
+  ArrayStart,
+  ArrayEnd,
+  String,
+  Number,
+  True,
+  False,
+  Null,
+  Colon,
+  Comma,
+  EndOfInput,
+  Error
+};
+
+}  // anonymous namespace
+
+class JsonShapeDeserializer::Impl {
+ public:
+  Impl(const char* data, size_t length) : m_data(data), m_length(length), m_pos(0), m_peeked(false) {}
+
+  Aws::Crt::Optional<bool> ReadBoolean() {
+    auto tok = NextToken();
+    if (tok == Token::True) {
+      return true;
+    }
+    if (tok == Token::False) {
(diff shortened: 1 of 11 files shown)


src/aws-cpp-sdk-core/source/smithy/client/schema/JsonShapeDeserializer.cpp
```diff
@@ -0,0 +1,418 @@
...
+      return {};
+    }
+    return HashingUtils::Base64Decode(str.value());
+  }
+
+  Aws::Crt::Optional<int> ReadEnum() { return ReadInteger(); }
+
+  size_t BeginStruct() {
+    auto tok = NextToken();
+    if (tok != Token::ObjectStart) {
```
sbiscigl: if the token is not object start,  we return 0, if it is we also return 0. what is the meaning of the return type `0` as a size_t? shouldnt it be returning whether o not it could begin a struct? i think this is so perpandicular because [java has a `readStruct`](https://github.com/smithy-lang/smithy-java/blob/main/codecs/json-codec/src/main/java/software/amazon/smithy/java/json/smithy/SmithyJsonDeserializer.java#L487)

which in c++ would look something like
```c++
template<typename T>
class StructMemberConsumer {
public:
  virtual ~StructMemberConsumer() = default;
  virutal void accept(const T& state, const Schema& memberSchema, ShapeDeserializer memberDeserializer) = 0;
  virutal void unknownMember(T state, String memberName) {}
}

template<typename T>
void readStruct(const Schema& schema, 
 T& state, 
  StructMemberConsumer<T> structMemberConsumer) 
{
  ...
}
```

so my question here, is why are we doing begin and end in deserilize? there is no way to "begin" or "end" in this context. because we are writing to a object not a symbol to a buffer.

i'd push back here and try to match java on this, we should be _deserializing into a object_ not arbitrarily.

tests/aws-cpp-sdk-core-tests/smithy/client/schema/JsonShapeDeserializerTest.cpp
```diff
@@ -0,0 +1,1239 @@
...
+TEST_F(JsonShapeDeserializerTest, BooleanTrue) {
+  JsonShapeSerializer s;
+  Schema root;
+  Schema member("enabled", ShapeType::Boolean);
+  s.BeginStructure(root);
+  s.WriteBoolean(member, true);
+  s.EndStructure();
+  auto json = s.GetPayload().GetResult();
+
+  JsonShapeDeserializer d(json.c_str(), json.size());
```
sbiscigl: lets re-contextualize this test, and maybe all the tests we're writing, we want to be serializing and de-serializing to a type `T`. so imagine the class

```
class serializable {
public
  serializable(deserializer*, std::string) {
    ... construct from string using the deserialized
  }
 std::string serialize(serializer*) {
   ... write to a string
 }
}
```

we want to keep the idea in mind that we are serializing to a class, and to a string from one. the way it is currently factored we dont know what we are deserializing into, the type.

Issue #3891, opened 2026-08-07
S3 CompleteMultipartUploadRequest serializes pretty XML, causing payload to exceed 1MiB with fewer than 10,000 parts


thenamehasbeentake:
### Describe the bug

`CompleteMultipartUploadRequest::SerializePayload()` in AWS SDK for C++ serializes the multipart completion body as pretty-printed XML, including indentation and newlines for every `<Part>` entry.

For large multipart uploads close to the S3 maximum of 10,000 parts, this extra whitespace can make the `CompleteMultipartUpload` request body exceed 1 MiB. Some S3-compatible storage systems, such as Ceph RGW deployments with a 1 MiB request/XML parsing limit, reject the request even though the upload uses fewer than 10,000 parts.

In our case, an upload with 9,331 parts failed at `CompleteMultipartUpload` with HTTP 416. All `UploadPart` requests succeeded.

## Observed Behavior

The SDK generates XML like this:

```xml
<?xml version="1.0"?>
<CompleteMultipartUpload xmlns="http://s3.amazonaws.com/doc/2006-03-01/">
    <Part>
        <ETag>"4f08eef4096726bff1125e1b1b734644"</ETag>
        <PartNumber>1</PartNumber>
    </Part>
</CompleteMultipartUpload>
```
For 9,331 parts, the pretty-printed XML is approximately:
```
1,090,743 bytes ~= 1.04 MiB
```
A compact equivalent would be approximately:
```
829,472 bytes ~= 0.79 MiB
```
So the whitespace alone adds about 255 KiB and can determine whether the request succeeds.

### Regression Issue

- [ ] Select this option if this issue appears to be a regression.

### Expected Behavior

The SDK should either:
1. Serialize CompleteMultipartUpload XML in compact form by default, or
2. Provide a configuration option to use compact XML serialization for request payloads.
Compact XML is semantically equivalent and accepted by S3-compatible APIs.

### Current Behavior

`CompleteMultipartUploadRequest::SerializePayload()` calls `XmlDocument::ConvertToString()`, and `XmlDocument::ConvertToString()` uses the default tinyxml2 `XMLPrinter`, which pretty-prints XML. The resulting string is then used directly as the HTTP request body.
Relevant source paths:
- generated/src/aws-cpp-sdk-s3/source/model/CompleteMultipartUploadRequest.cpp
- src/aws-cpp-sdk-core/source/utils/xml/XmlSerializer.cpp
[...]

PR #3892, opened 2026-08-11
using compact XML serialization in XmlDocument::ConvertToString

pulimsr:
*Issue #3891*

Adds an opt-in `ClientConfiguration::compactXmlSerialization` flag (default `false`) that serializes XML request payloads without pretty-print whitespace, producing smaller request bodies. When off, output is unchanged.            
                                                                                                                                                                                                                                         
  - New `XmlDocument::ConvertToString(bool compact)` overload; no-arg version delegates with `false`.                                                                                                                                    
  - `AWSClient` reads the config flag and sets it on the request before payload serialization.                                                                                                                                           
  - XML request templates regenerated to pass `ShouldUseCompactXmlPayload()`.

*Check all that applies:*

- [x] Linux
- [x] Windows
- [ ] Android
- [x] MacOS
- [ ] IOS
- [ ] Other Platforms


### DIFF
--- a/src/aws-cpp-sdk-core/source/client/AWSClient.cpp
+++ b/src/aws-cpp-sdk-core/source/client/AWSClient.cpp
@@ -145,7 +145,8 @@ AWSClient::AWSClient(const Aws::Client::ClientConfiguration& configuration,
         configuration.awsChunkedBufferSize),
         m_userAgentInterceptor},
     m_enableNewRetries{Aws::Utils::StringUtils::ToLower(Aws::Environment::GetEnv("AWS_NEW_RETRIES_2026").c_str()) == "true"},
-    m_disableExpectHeader(configuration.disableExpectHeader)
+    m_disableExpectHeader(configuration.disableExpectHeader),
+    m_compactXmlSerialization(configuration.compactXmlSerialization)
 {
 }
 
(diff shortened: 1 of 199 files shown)


src/aws-cpp-sdk-core/source/client/AWSClient.cpp
```diff
@@ -917,6 +919,9 @@ void AWSClient::BuildHttpRequest(const Aws::AmazonWebServiceRequest& request, co
     AddHeadersToRequest(httpRequest, request.GetHeaders());
     AddHeadersToRequest(httpRequest, request.GetAdditionalCustomHeaders());
 
+    // Propagate the compact XML serialization preference to the request before its payload is serialized.
+    request.SetUseCompactXmlPayload(m_compactXmlSerialization);
```
sbiscigl, line 923: summarizing from out chat:

this is a mechanism by which we pass context from the client to serializer with information about _how to serialize_. we are horseshoeing a member variable on request objects to force a action. in reality the deserializer should be client aware. this sounds like it better fits into the upcoming schema serde work you have already been working on where a serializer will be client aware

pulimsr: Okay, closing this PR - will fold into the schema serde serializer instead.

PR #3898, opened 2026-08-24
[Feature] Add clock skew computation and unit test

kai-ion:
*Issue #, if available:*

*Description of changes:*
Add clock skew computation and unit test

*Check all that applies:*

- [x] Linux
- [x] Windows
- [x] Android
- [x] MacOS
- [ ] IOS
- [ ] Other Platforms

### DIFF
--- a/src/aws-cpp-sdk-core/include/aws/core/internal/ClockSkew.h
+++ b/src/aws-cpp-sdk-core/include/aws/core/internal/ClockSkew.h
@@ -1,0 +1,161 @@
+/**
+ * Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.
+ * SPDX-License-Identifier: Apache-2.0.
+ */
+
+#pragma once
+
+#include <aws/core/Core_EXPORTS.h>
+#include <aws/core/http/HttpRequest.h>
+#include <aws/core/utils/Chrono.h>
+#include <aws/core/utils/DateTime.h>
+#include <aws/core/utils/StringUtils.h>
+#include <aws/core/client/AWSError.h>
+#include <aws/core/client/CoreErrors.h>
+#include <aws/crt/Optional.h>
+#include <atomic>
+#include <chrono>
+
+namespace Aws
+{
+    namespace Internal
+    {
+        // AttemptSkew is owned by the client, not here.
+
+        static const std::chrono::milliseconds CLOCK_SKEW_DETECTION_THRESHOLD = std::chrono::minutes(4);
+        static const std::chrono::milliseconds CLOCK_SKEW_MAX_TRUSTED_REQUEST_DURATION = std::chrono::minutes(15);
+        static const char CLOCK_SKEW_AGE_HEADER[] = "age";
+
+        struct ClockSkewMeasurement
+        {
+            Aws::Crt::Optional<Aws::Utils::DateTime> serverTime; // empty when the response carries no usable Date header
+            Aws::Utils::DateTime timeRequestSent;
+            Aws::Utils::DateTime timeResponseReceived;
+            bool servedFromCache;
+
+            ClockSkewMeasurement(const Aws::Crt::Optional<Aws::Utils::DateTime>& serverTime,
+                                 const Aws::Utils::DateTime& timeRequestSent,
+                                 const Aws::Utils::DateTime& timeResponseReceived,
+                                 bool servedFromCache)
+                : serverTime(serverTime),
+                  timeRequestSent(timeRequestSent),
+                  timeResponseReceived(timeResponseReceived),
+                  servedFromCache(servedFromCache) {}
+        };
+
+        inline ClockSkewMeasurement MakeClockSkewMeasurement(const Aws::Http::HeaderValueCollection& headers,
+                                                             const Aws::Utils::DateTime& timeRequestSent,
+                                                             const Aws::Utils::DateTime& timeResponseReceived)
+        {
+            const auto dateIt = headers.find(Aws::Utils::StringUtils::ToLower(Aws::Http::DATE_HEADER));
+            Aws::Crt::Optional<Aws::Utils::DateTime> serverTime;
+            if (dateIt != headers.end())
+            {
+                const Aws::Utils::DateTime parsed(dateIt->second.c_str(), Aws::Utils::DateFormat::AutoDetect);
+                if (parsed.WasParseSuccessful())
+                {
+                    serverTime = parsed;
+                }
+            }
+            const bool servedFromCache = headers.find(CLOCK_SKEW_AGE_HEADER) != headers.end();
+            return ClockSkewMeasurement(serverTime, timeRequestSent, timeResponseReceived, servedFromCache);
+        }
+
+        inline Aws::Crt::Optional<std::chrono::milliseconds> ComputeClockSkewCandidate(const ClockSkewMeasurement& measurement)
+        {
+            if (!measurement.serverTime.has_value() || measurement.servedFromCache)
+            {
+                return Aws::Crt::Optional<std::chrono::milliseconds>{};
+            }
+
+            const int64_t sentMs = measurement.timeRequestSent.Millis();
+            const int64_t receivedMs = measurement.timeResponseReceived.Millis();
+            const int64_t serverMs = measurement.serverTime.value().Millis();
+
+            const std::chrono::milliseconds elapsed{receivedMs - sentMs};
+            if (elapsed > CLOCK_SKEW_MAX_TRUSTED_REQUEST_DURATION)
+            {
+                return Aws::Crt::Optional<std::chrono::milliseconds>{};
+            }
+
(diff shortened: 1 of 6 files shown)


src/aws-cpp-sdk-core/include/aws/core/internal/ClockSkew.h
```diff
@@ -0,0 +1,156 @@
...
+            Aws::Utils::DateTime timeResponseReceived{};
+            bool hasServerTime = false;
+            bool servedFromCache = false;
+        };
+
+        inline ClockSkewMeasurement MakeClockSkewMeasurement(const Aws::Http::HeaderValueCollection& headers,
+                                                             const Aws::Utils::DateTime& timeRequestSent,
+                                                             const Aws::Utils::DateTime& timeResponseReceived)
+        {
+            ClockSkewMeasurement measurement;
```
sbiscigl: you default construct a object, then assign each of its  members, why not have it as a constructor on `ClockSkewMeasurement` to avoid unnecessary copy operations?

src/aws-cpp-sdk-core/include/aws/core/internal/ClockSkew.h
```diff
@@ -0,0 +1,156 @@
...
+            measurement.timeRequestSent = timeRequestSent;
+            measurement.timeResponseReceived = timeResponseReceived;
+
+            const auto dateIt = headers.find(Aws::Utils::StringUtils::ToLower(Aws::Http::DATE_HEADER));
+            if (dateIt != headers.end())
+            {
+                measurement.serverTime = Aws::Utils::DateTime(dateIt->second.c_str(), Aws::Utils::DateFormat::AutoDetect);
+                // A default DateTime reports WasParseSuccessful() == true, so gate on the header being present too.
```
sbiscigl: this comment exactly. why are you default constructing the header, this should all happen as initialization in the constructor for `ClockSkewMeasurement`

src/aws-cpp-sdk-core/include/aws/core/internal/ClockSkew.h
```diff
@@ -0,0 +1,156 @@
...
+                measurement.hasServerTime = measurement.serverTime.WasParseSuccessful();
+            }
+            measurement.servedFromCache = headers.find(CLOCK_SKEW_AGE_HEADER) != headers.end();
+            return measurement;
+        }
+
+        inline ClockSkewMeasurement MakeClockSkewMeasurement(const Aws::Http::HttpResponse& response,
```
sbiscigl: why does this need a function override? this is a new API?

src/aws-cpp-sdk-core/include/aws/core/internal/ClockSkew.h
```diff
@@ -0,0 +1,156 @@
...
+        struct ClockSkewAdjustment
+        {
+            bool skewExceedsThreshold = false; // caller ANDs this with the error-code check before retrying
+            std::chrono::milliseconds skew{0};
+        };
+
+        class AWS_CORE_LOCAL ClientSkew
+        {
+        public:
+            ClientSkew() = default;
```
sbiscigl: why does this need to be default constructible, when is `std::chrono::milliseconds` default a valid value?

src/aws-cpp-sdk-core/include/aws/core/internal/ClockSkew.h
```diff
@@ -0,0 +1,156 @@
...
+            explicit ClientSkew(std::chrono::milliseconds initial) : m_skew(initial) {}
+
+            ClientSkew(const ClientSkew& other) : m_skew(other.m_skew.load()) {}
```
sbiscigl, line 113: [rule of 5](https://en.cppreference.com/cpp/language/rule_of_three) why are you not defining move copy and assignment, further more. why do you define copy assignment and copy equality? is it because atomics delete the copy constructor and copy assignment? if so why do we need a copy ctor and copy assignment?

src/aws-cpp-sdk-core/include/aws/core/internal/ClockSkew.h
```diff
@@ -0,0 +1,156 @@
...
+                {
+                    const std::chrono::milliseconds delta = attemptSkew - candidate.value();
+                    const std::chrono::milliseconds absDelta = delta < std::chrono::milliseconds::zero() ? -delta : delta;
+                    adjustment.skewExceedsThreshold = absDelta > CLOCK_SKEW_DETECTION_THRESHOLD;
+                }
+                return adjustment;
+            }
+
+        private:
+            std::atomic<std::chrono::milliseconds> m_skew{std::chrono::milliseconds{0}};
```
sbiscigl: why does `ClientSkew` _need_ to be thread safe? its only accessed in the call path by the request thread, why are we creating a constraint that isnt needed?

tests/aws-cpp-sdk-core-tests/utils/ClockSkewTest.cpp
```diff
@@ -0,0 +1,252 @@
...
+
+        const std::chrono::milliseconds expectedClientSkew{static_cast<int64_t>(operation.GetInteger("expectedClientSkew")) * 1000};
+        EXPECT_EQ(expectedClientSkew.count(), clientSkew.Load().count())
+            << "ClientSkew mismatch after operation " << o;
+    }
+}
+
+INSTANTIATE_TEST_SUITE_P(ClockSkew, ClockSkewTest, ::testing::Range((size_t) 0u, ClockSkewTest::TEST_CASE_COUNT));
+
+class ClockSkewMeasurementTest : public Aws::Testing::AwsCppSdkGTestSuite
```
sbiscigl: if the above test, test the json that you pass in for the tests, which is sdk standardized what is the point of these tests?

src/aws-cpp-sdk-core/include/aws/core/internal/ClockSkew.h
sbiscigl: im reopening this comment because you didnt address the underlying issue you compounded on it. `DateTime` should _NOT_ be default constructed. deafult constructed date time is a garbage value. you need to represent the optionality of datetime in you class.

missing  != default constructed

src/aws-cpp-sdk-core/include/aws/core/internal/ClockSkew.h
```diff
@@ -0,0 +1,161 @@
...
+
+        inline ClockSkewMeasurement MakeClockSkewMeasurement(const Aws::Http::HeaderValueCollection& headers,
+                                                             const Aws::Utils::DateTime& timeRequestSent,
+                                                             const Aws::Utils::DateTime& timeResponseReceived)
+        {
+            const auto dateIt = headers.find(Aws::Utils::StringUtils::ToLower(Aws::Http::DATE_HEADER));
+            const bool hasDate = dateIt != headers.end();
+            const Aws::Utils::DateTime serverTime = hasDate
+                ? Aws::Utils::DateTime(dateIt->second.c_str(), Aws::Utils::DateFormat::AutoDetect)
+                : Aws::Utils::DateTime();
```
sbiscigl: you are relying on a default construction to determine invalidity, that is wrong. default construction != invalid.

src/aws-cpp-sdk-core/include/aws/core/internal/ClockSkew.h
```diff
@@ -0,0 +1,161 @@
...
+
+            ClockSkewAdjustment EvaluateFailure(const ClockSkewMeasurement& measurement, std::chrono::milliseconds attemptSkew)
+            {
+                const Aws::Crt::Optional<std::chrono::milliseconds> candidate = RecordResponse(measurement);
+                ClockSkewAdjustment adjustment;
+                adjustment.skew = m_skew.load();
+                // Retryable when the applied skew is off from the observed skew by more than the threshold.
+                if (candidate.has_value())
+                {
+                    const std::chrono::milliseconds delta = attemptSkew - candidate.value();
```
sbiscigl: ok im sort of split on this, nice to have wont block it, the reason you are doing all of this is because [`std::chrono::abs`](https://en.cppreference.com/cpp/chrono/duration/abs) is in c++17. i sort of hate we are forced to do this the wrong way. so a suggestion. in `aws/core/utils/local` you could add `Chrono.h` and have something like
```c++
namespace Aws {
#if defined(__cpp_lib_chrono) && __cpp_lib_chrono >= 201510L
using std::chrono::abs;
using std::chrono::ceil;
using std::chrono::floor;
using std::chrono::round;
#else
// TODO: delete this branch once the minimum standard is C++17.
template <class Rep, class Period>
constexpr std::enable_if_t<std::numeric_limits<Rep>::is_signed,
                           std::chrono::duration<Rep, Period>>
abs(std::chrono::duration<Rep, Period> d) {
    return d >= d.zero() ? d : -d;
}
#endif
}
```

then here you could just call `Aws::chrono::abs`, and if you compile with 17 you get the feature, if you compile without you get the shim implementation.

sbiscigl (approved):

sbiscigl (approved):

PR #3900, opened 2026-08-25
Protocol tests updates

pulimsr:
Makes the generated protocol tests compare requests semantically instead of byte-for-byte, so they validate protocol correctness rather than incidental ordering:
  - XML bodies compared canonically (sorted elements/attributes, via tinyxml2)                                                                                                                                                          
  - form-urlencoded bodies and URI query strings compared as unordered param sets                                                                                                                                                       
  - body assertion skipped when a test vector has no body; fixes generated timestamp values

*Check all that applies:*

- [x] Linux
- [x] Windows
- [ ] Android
- [x] MacOS
- [ ] IOS
- [ ] Other Platforms


tests/testing-resources/include/aws/testing/AwsProtocolTestHelpers.h
```diff
@@ -128,6 +128,108 @@ class AwsProtocolTestSuite : public Aws::Testing::AwsCppSdkGTestSuite {
     }
   }
 
+  // Normalize an XML document so that two semantically equivalent documents compare equal:
+  // drops a leading <?xml ... ?> declaration, expands self-closing <a/> into <a></a>, and
+  // removes indentation whitespace (a whitespace run containing a newline) between elements.
+  // Attributes and significant leaf text (e.g. an all-whitespace string value) are preserved,
+  // so genuine differences such as missing xmlns attributes or omitted elements still fail.
+  static Aws::String NormalizeXml(const Aws::String& in) {
```
sbiscigl: ok this really sucks you have to do this by hand, but i get why you are doing it, its because you are testing that the serde is functioning without relying on a xml parser from the SDK itself. so i present a alternative idea lets have a direct dependency on the tinyxml specifically for protocol tests that require it using cmake fetch content declare. then compare the xml-ified objects.

i.e. in the [cmake for the xml tests](https://github.com/aws/aws-sdk-cpp/blob/main/generated/protocol-tests/tests/input/rest-xml/CMakeLists.txt) you could add something like.

```c++
FetchContent_Declare(
        tinyxml2
    GIT_REPOSITORY https://github.com/leethomason/tinyxml2.git
    GIT_TAG 11.0.0
)
FetchContent_MakeAvailable(tinyxml2)

target_link_libraries(whatever_protocol_test PRIVATE tinyxml2::tinyxml2)
```

then we can just sort and serialize nodes directly instead of having to write and parsing json too. it would look something like

```c++
auto canonical(const tinyxml2::XMLElement* el) -> std::string {
  std::vector<std::string> parts;
  for (auto* a = el->FirstAttribute(); a; a = a->Next()) {
    parts.emplace_back(std::string("@") + a->Name() + "=" + a->Value());
  }
  std::sort(parts.begin(), parts.end());

  std::vector<std::string> kids;
  for (auto* c = el->FirstChildElement(); c; c = c->NextSiblingElement()) {
    kids.push_back(canonical(c));
  }
  std::sort(kids.begin(), kids.end());

  std::string out = std::string("<") + el->Name();
  for (const auto& p : parts) out += " " + p;
  if (const char* text = el->GetText()) out += " #" + std::string(text);
  for (const auto& k : kids) out += k;
  return out + ">";
}

auto equivalent(const char* lhs, const char* rhs) -> bool {
  tinyxml2::XMLDocument a, b;
  if (a.Parse(lhs) != tinyxml2::XML_SUCCESS) return false;
  if (b.Parse(rhs) != tinyxml2::XML_SUCCESS) return false;
  return canonical(a.RootElement()) == canonical(b.RootElement());
}
```

when all is said and done. taking a direct dependency on tinyxml in testing is something that i would encourage and discourage in the SDK as a whole. our test should introduce dependencies to simplify them. the SDK should not introduce dependencies that are passed to customers. for this test, this dependency makes sense to avoid this complicated logic.

sbiscigl (approved):

Issue #843, opened 2018-04-12
Enable CPack


Bu11etmagnet:
In our project, we use the following local patch:
```
$ cat cpack.diff 
diff --git a/CMakeLists.txt b/CMakeLists.txt
index 83024b94a..1945cc145 100644
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -169,6 +169,13 @@ endif()
 
 add_sdks()
 
+SET(CPACK_PACKAGE_DESCRIPTION_SUMMARY "AWS C++ SDK")
+SET(CPACK_PACKAGE_VENDOR "Amazon")
+SET(CPACK_PACKAGE_VERSION "${PROJECT_VERSION}")
+SET(CPACK_GENERATOR "RPM;DEB")
+SET(CPACK_DEBIAN_PACKAGE_MAINTAINER "Package fairy")
+INCLUDE(CPack)
+
 # for user friendly cmake usage
 include(setup_cmake_find_module)
```
After this, I can
`$ cd <builddir> ; make; cpack -G RPM`
on a RedHat-like system or
`$ cd <builddir> ; make; cpack -G DEB`
on a Debian-like system, to create an installable package file.

It would be nice to have this by default, with proper values for the CMake variables, of course.

singku:
@Bu11etmagnet Thank you very much for sharing this. We have a similar task in our backlog, which would enable building of RPM, DEB and windows installation packages, but currently is not prioritized. It would be nicer if you can create PR for DEB and RPM, then credits is yours.

Bu11etmagnet:
Can you give suggestions for the variables? Is "Amazon" OK for CPACK_PACKAGE_VENDOR?

psalvaggio:
Has there been any movement on this? This would be a very welcome addition, as I currently have to maintain packaging scripts for this library, which break a lot because of new libraries and it seems recently that some things were split out into submodules.

Bu11etmagnet:
Ping?

richardjkendall:
This would be very useful.

jmklix:
https://github.com/aws/aws-sdk-cpp/issues/1888

yumais:
It would be really nice to have CPack targets for specific modules and their dependencies. We are now using our local patch too

kai-ion:
Converting this to a Discussion. It's an improvements we plan to pursue over time rather than one actionable change with a near-term milestone, so a Discussion better fits collecting the specifics and design input.

PR #3904, opened 2026-08-27
Add opt-in CPack support for RPM and DEB packaging

jameshickman:
*Issue #, if available:* #843 (also relevant to #1888)

*Description of changes:*

Adds opt-in CPack support for building RPM and DEB packages of the SDK, split into per-service-group components rather than one monolithic package.

Everything is behind a new `ENABLE_CPACK_PACKAGING` option that **defaults to `OFF`**.

I verified what "off" actually means rather than assuming it: I configured the `core` build twice — once with the four modified files reverted to their upstream contents, and once with them as they are on this branch and the option `OFF` — then diffed every generated `cmake_install.cmake`. They are identical except for one deliberate difference, described below. No `include(CPack)`, no changed install components, no new targets.

The one difference with the option `OFF`: the bulk `cmake/` directory install gains five `EXCLUDE` patterns. Those patterns only ever match files this PR adds (`CPackConfig.cmake`, `CPackComponents.cmake`, `ServiceGroupMapping.cmake`, `rpm-scripts/`, `deb-scripts/`). They are unconditional on purpose — without them, a packaging-disabled build would start shipping this branch's build-only scaffolding into consumers' `lib/cmake/AWSSDK`. Keeping them always-on is what makes the *installed tree* byte-for-byte identical to today's.

**What it produces**

With `-DENABLE_CPACK_PACKAGING=ON`, `cpack -G RPM` / `-G DEB` generate separate runtime and `-devel` packages per service group (e.g. `aws-sdk-cpp-core`, `aws-sdk-cpp-core-devel`, `aws-sdk-cpp-crt`, `aws-sdk-cpp-crt-devel`), with:
- version-pinned cross-package dependencies (`-devel` → matching runtime, `crt-devel` → `core-devel`)
- runtime shared-library deps resolved via `AUTOREQ` / `SHLIBDEPS` instead of hardcoded distro package names (avoids the `libcurl` vs `curl`, `libssl3` vs `libssl1.1` split across distros)
- the `-devel` dependency target switched between the runtime and `-static` package based on `BUILD_SHARED_LIBS`, so static-only builds don't depend on a runtime package that was never produced

**New files**
- `cmake/CPackConfig.cmake`, `cmake/CPackComponents.cmake` — CPack setup and per-component naming/dependency config
- `cmake/ServiceGroupMapping.cmake` — maps the ~400 generated service clients into installable groups
- `cmake/rpm-scripts/`, `cmake/deb-scripts/` — post-install/post-remove ldconfig scriptlets
- `scripts/build-packages.sh` — convenience wrapper
- `docs/PACKAGING.md` — build/install/consume documentation

**Modified files**
- `CMakeLists.txt` — the `ENABLE_CPACK_PACKAGING` option and three gated hooks
- `cmake/setup_cmake_find_module.cmake` — tag config-module installs with the `core-devel` component, and exclude build-only files from the bulk `cmake/` install so they aren't shipped to consumers
- `cmake/utilities.cmake`, `src/aws-cpp-sdk-core/CMakeLists.txt` — component routing on existing install rules

No generated client code is touched, and no changes were made under `code-generation/`.

**Scope / limitations**
- Linux only (RPM + DEB). No effect on Windows, macOS, iOS, or Android builds.
- Wired into the `LEGACY_BUILD` path (the current default). The `-DLEGACY_BUILD=OFF` modern-CMake path is **not** covered — happy to extend it if you'd prefer that first.

*Check all that applies:*
  - No automated tests added. This is a packaging/build-system change with no runtime code, and the repo has no existing harness for asserting on generated `.rpm`/`.deb` artifacts. Verified manually on Linux: (a) the configure-time A/B diff of generated install rules described above, confirming the option genuinely gates the feature; (b) with the option `ON`, `cpack -G RPM` emits the expected per-component package set, with `rpm -qp --requires` / `dpkg-deb --info` showing correct inter-package `Requires`/`Depends` and autodetected shared-library deps. Glad to add a CI job that builds packages and asserts on the output if you'd like that as part of this PR.
- [x] Checked if this PR is a breaking (APIs have been changed) change. — Not breaking; no public API or ABI change, and the feature is off by default.
- [x] Checked if this PR will _not_ introduce cross-platform inconsistent behavior. — Gated to opt-in Linux packaging; other platforms are unaffected.
- [x] Checked if this PR would require a ReadMe/Wiki update. — Added `docs/PACKAGING.md`.

- [x] Linux
- [ ] Windows
- [ ] Android
- [ ] MacOS
- [ ] IOS
- [ ] Other Platforms


sbiscigl:
Thanks for taking a look at this. CPack is something we've always wanted to look into, and the reason we've never shipped binaries ourselves is ABI: we don't set a SOVERSION on the service libraries or guarantee ABI stability between releases, so publishing packages has meant committing to something we can't hold. Appreciate you putting real work into this one.

Historically distributions have kept their own formulas for this. The [FreeBSD port](https://cgit.freebsd.org/ports/tree/devel/aws-sdk-cpp) and [Alpine's APKBUILD](https://gitlab.alpinelinux.org/alpine/aports/-/blob/master/community/aws-sdk-cpp/APKBUILD) both build from `cmake --install` and sort the files themselves, and Alpine goes as far as patching a SOVERSION in during `prepare()` since we don't set one. For this SDK the split we'd want is one package per service rather than a manually curated grouping, which is what Alpine settled on with its 443 subpackages. `ServiceGroupMapping.cmake` is a living list we'd have to maintain as services get added, and unmatched ones land silently in `misc`, so if you reworked this to derive the components from `SDK_BUILD_LIST` (already globbed from `generated/src/aws-cpp-sdk-*` in `cmake/sdksCommon.cmake`, and already respecting `BUILD_ONLY`) that's something we'd likely take.

PR #3906, opened 2026-08-31
[Feature] Adjusts request signing timestamps 

kai-ion:
*Issue #, if available:*

*Description of changes:*
Wire clock-skew correction into the legacy and smithy client pipelines

*Check all that applies:*

- [x] Linux
- [x] Windows
- [x] Android
- [x] MacOS
- [ ] IOS
- [ ] Other Platforms

### DIFF
--- a/src/aws-cpp-sdk-core/source/client/AWSClient.cpp
+++ b/src/aws-cpp-sdk-core/source/client/AWSClient.cpp
@@ -28,6 +28,7 @@
 #include <aws/core/utils/logging/LogMacros.h>
 #include <aws/core/Globals.h>
 #include <aws/core/utils/EnumParseOverflowContainer.h>
+#include <aws/core/internal/ClockSkew.h>
 #include <aws/core/utils/crypto/MD5.h>
 #include <aws/core/utils/crypto/CRC32.h>
 #include <aws/core/utils/crypto/Sha256.h>
@@ -67,11 +68,6 @@ static const int SUCCESS_RESPONSE_MAX = 299;
 static const char AWS_CLIENT_LOG_TAG[] = "AWSClient";
 static const char AWS_LAMBDA_FUNCTION_NAME[] = "AWS_LAMBDA_FUNCTION_NAME";
 static const char X_AMZN_TRACE_ID[] = "_X_AMZN_TRACE_ID";
-
-//4 Minutes
-static const std::chrono::milliseconds TIME_DIFF_MAX = std::chrono::minutes(4);
-//-4 Minutes
-static const std::chrono::milliseconds TIME_DIFF_MIN = std::chrono::minutes(-4);
 
 CoreErrors AWSClient::GuessBodylessErrorType(Aws::Http::HttpResponseCode responseCode)
 {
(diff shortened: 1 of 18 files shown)


src/aws-cpp-sdk-core/source/client/AWSClient.cpp
```diff
@@ -138,6 +134,7 @@ AWSClient::AWSClient(const Aws::Client::ClientConfiguration& configuration,
     m_hash(Aws::Utils::Crypto::CreateMD5Implementation()),
     m_requestTimeoutMs(configuration.requestTimeoutMs),
     m_enableClockSkewAdjustment(configuration.enableClockSkewAdjustment),
+    m_clientSkew(Aws::MakeShared<Aws::Internal::ClientSkew>(AWS_CLIENT_LOG_TAG, std::chrono::milliseconds(0))),
```
sbiscigl, line 137: why does it need to be a a shared pointer, why cant it just be a stack variable?

src/aws-cpp-sdk-core/source/client/AWSClient.cpp
```diff
@@ -320,6 +309,10 @@ HttpResponseOutcome AWSClient::AttemptExhaustively(const Aws::Http::URI& uri,
              {TracingUtils::SMITHY_SERVICE_DIMENSION, this->GetServiceClientName()}});
         if (outcome.IsSuccess())
         {
+            if (m_enableClockSkewAdjustment && outcome.GetResult())
```
sbiscigl, line 312: why are we only measuring clockskew if the result is success? wouldnt we also want to measure for failures?

tests/testing-resources/include/aws/testing/mocks/aws/client/MockAWSClient.h
```diff
@@ -246,3 +247,28 @@ class MockAWSClientWithStandardRetryStrategy : Aws::Client::AWSClient
         return err;
     }
 };
+
+// A MockAWSClient whose error responses carry a service error code via the x-amzn-errortype header
+// (the header a JSON error marshaller reads), so a clock-skew retry can be exercised end to end.
+class ClockSkewMockAWSClient : public MockAWSClient
```
sbiscigl: why do you need this in addition to mock aws client? why do you need to override `BuildAWSError` when supplying a expected response to a public API is how that class is supposed to be used?

src/aws-cpp-sdk-core/source/client/AWSClient.cpp
kai-ion: On a failed attempt, AdjustClockSkew runs (AWSClient.cpp:358) and calls m_clientSkew->EvaluateFailure(...), which calls RecordResponse as its first step

tests/testing-resources/include/aws/testing/mocks/aws/client/MockAWSClient.h
kai-ion: We needed a separate class because MockAWSClient::BuildAWSError doesnt return the service error code that clock skew detection checks. 

I do agree we dont have to override buildAWSError tho, dropped the override and based it on AWSJsonClient so errors now go through the real JsonErrorMarshaller

sbiscigl (approved):

PR #3910, opened 2026-09-01
Add schema-serde streaming JSON/XML deserializers, ec2Query, Codec, and ClientProtocol

pulimsr:
Adds response deserialization and the protocol-selection layer to the schema-serde runtime.                                                                                        
                                                                                                                                                                                     
  - `ShapeDeserializer` reworked to a push/consumer model (`ReadStruct`/`ReadList`/`ReadMap`), the dual of `ShapeSerializer`; scalars take const Schema&.                                      
  - `JsonShapeDeserializer` and `XmlShapeDeserializer` (new), `CborShapeDeserializer` updated to match.                                                    
  - `QueryShapeSerializer`: full nesting + `ec2Query` flavor (new `Ec2QueryNameTrait`).                                                                                                    
  - `Codec` (Json/Xml/Cbor) pairs serializer+deserializer; `ClientProtocol` picks codec + content-type per protocol (restJson1, awsJson1.0/1.1, rpcv2Cbor, restXml, awsQuery, ec2Query), 
    including the asymmetric query request/XML response case.

*Check all that applies:*

- [x] Linux
- [x] Windows
- [ ] Android
- [x] MacOS
- [ ] IOS
- [ ] Other Platforms



tests/aws-cpp-sdk-core-tests/smithy/client/schema/JsonShapeDeserializerTest.cpp
```diff
@@ -0,0 +1,360 @@
...
+        }
+      });
+    }
+  });
+  ASSERT_EQ(entries.size(), 2u);
+  EXPECT_EQ(entries["foo"], "bar");
+  EXPECT_EQ(entries["baz"], "qux");
+}
+
+TEST_F(JsonShapeDeserializerTest, NestedStructure) {
```
sbiscigl, line 261: summarizing from conversation:

so one thing that is left kind of ambigious to me in this is how we go from serializer to a hard type `T`. for instance given the class.

```
class bar {
std::string buzz;
};

class foo {
bar fizz;
};
```

how can the deserializer actually achieve constructing a object from `{fizz: {buzz: \"value\"}}`. would like to see a added test for each deserializer that accomplishes this, that can be used as the blueprint to adding it to codegen

src/aws-cpp-sdk-core/include/smithy/client/schema/Codec.h
```diff
@@ -0,0 +1,47 @@
...
+
+  virtual SerializerOutcome Serialize(const Schema& schema, const SerializableStruct& shape) const = 0;
+
+  virtual Aws::UniquePtr<ShapeDeserializer> CreateDeserializer(const unsigned char* data, size_t length) const = 0;
+};
+
+class SMITHY_API JsonCodec final : public Codec {
+ public:
+  SerializerOutcome Serialize(const Schema& schema, const SerializableStruct& shape) const override;
+  Aws::UniquePtr<ShapeDeserializer> CreateDeserializer(const unsigned char* data, size_t length) const override;
```
sbiscigl, line 35: summarizing from offline:

ok i think `CreateDeserializer` is where where we have the disconnect and where it will come together for us with "how do we actually create a object T". so looking at the [smithy java codec](https://github.com/smithy-lang/smithy-java/blob/main/codecs/json-codec/src/main/java/software/amazon/smithy/java/json/JsonCodec.java#L46-L100). we're actually borrowing a little bit more than we need. and we can actually walk this back. 

so `Serialize` looks good, lets park that lets only talk about deserialize

[in the java impl the have one called `deserializeShape`](https://github.com/smithy-lang/smithy-java/blob/main/codecs/json-codec/src/main/java/software/amazon/smithy/java/json/JsonCodec.java#L46-L100) that is closer to our needs and answers the second question which is "how do we construct T given a stream of characters".

javas looks like
```java
public <T extends SerializableShape> T deserializeShape(byte[] source, ShapeBuilder<T> builder)
```

ours should end up looking like
```c++
void deserializeShape(const unsigned char* data, 
  size_t length,
  std::function<void (const Schema& schema, const StructMemberConsumer& consumer)> consumer)
```

this is so that during coden class would implement those struct member consumers to construct self

pulimsr: Done — ported from the smithy-java `Codec` you linked. Added `DeserializeShape(data, length, SerializableStruct& shape)`, which delegates to `shape.Deserialize(*CreateDeserializer(...))` — the analogue of `deserializeShape(source, builder)` → `builder.deserialize(createDeserializer(source))`. The generated shape implements `From()` as its member consumer to construct itself.                                                                                                                                     


Two deltas from the sketch, both to match smithy-java:
- **Kept `CreateDeserializer` public** — smithy-java keeps `createDeserializer` as a core interface method and `deserializeShape` is a helper on top, so there's nothing to walk back.   
- **Passed the shape/builder** instead of `std::function<void(const Schema&, StructMemberConsumer&)>` - that literal form is circular (the callback would receive the consumer it's meant to supply).

tests/aws-cpp-sdk-core-tests/smithy/client/schema/JsonShapeDeserializerTest.cpp
pulimsr: Added nested `foo{bar{buzz}} `round-trip tests for all three deserializers in `CodecTest.cpp`

src/aws-cpp-sdk-core/source/smithy/client/schema/SerializableStruct.cpp
```diff
@@ -0,0 +1,16 @@
...
+ * Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.
+ * SPDX-License-Identifier: Apache-2.0.
+ */
+#include <smithy/client/schema/SerializableStruct.h>
+
+#include <smithy/client/schema/ShapeDeserializer.h>
+
+using namespace smithy::schema;
+
+void SerializableStruct::From(const Schema& /*memberSchema*/, ShapeDeserializer& /*deserializer*/) {}
```
sbiscigl: instead of doing nothing should this function be pure virtual (abstract) since `SerializableStruct` will never be instantiated directly? a class will always "implement (inherit)" it

tests/aws-cpp-sdk-core-tests/smithy/client/schema/CodecTest.cpp
```diff
@@ -79,3 +80,186 @@ TEST_F(CodecTest, XmlRoundTrip) {
   XmlCodec codec;
   RoundTrip(codec, root);
 }
+
+namespace {
+
+class Person : public SerializableStruct {
```
sbiscigl, line 81: at my desk i was going "yes, yes, yes, yes, yes, exactly" 💯

src/aws-cpp-sdk-core/source/smithy/client/schema/SerializableStruct.cpp
pulimsr: Yes, made it pure virtual

sbiscigl (approved):

src/aws-cpp-sdk-core/include/smithy/client/schema/ClientProtocol.h
```diff
@@ -0,0 +1,112 @@
...
+
+class SMITHY_API RestJsonProtocol final : public ClientProtocol {
+ public:
+  Aws::String GetProtocolId() const override;
+  Aws::String GetContentType() const override;
+  SerializerOutcome SerializeInput(const Schema& schema, const SerializableStruct& input) const override;
+  Aws::UniquePtr<ShapeDeserializer> CreateOutputDeserializer(const unsigned char* data, size_t length) const override;
+
+ private:
+  JsonCodec m_codec;
```
sbiscigl: nit: default init all the codec member i.e. `JsonCodec m_codec{};`

src/aws-cpp-sdk-core/include/smithy/client/schema/ClientProtocol.h
```diff
@@ -0,0 +1,112 @@
...
+ private:
+  CborCodec m_codec;
+};
+
+class SMITHY_API RestXmlProtocol final : public ClientProtocol {
+ public:
+  Aws::String GetProtocolId() const override;
+  Aws::String GetContentType() const override;
+  SerializerOutcome SerializeInput(const Schema& schema, const SerializableStruct& input) const override;
+  Aws::UniquePtr<ShapeDeserializer> CreateOutputDeserializer(const unsigned char* data, size_t length) const override;
```
sbiscigl: not absolutely needed on this, but lets talk about the API

`const unsigned char* data, size_t length`

this specific _very common and normal api_ has [caused a substantial amount of pain](https://www.open-std.org/jtc1/sc22/wg21/docs/papers/2018/p0122r6.pdf). so much so that c++ introduced a library type called [`std::span`](https://en.cppreference.com/cpp/container/span). The CRT also has the same idea as [byte_buf](https://github.com/awslabs/aws-c-common/blob/main/include/aws/common/byte_buf.h#L48-L57). it might be worth seeing if you can use [`BytBuf`](https://github.com/awslabs/aws-crt-cpp/blob/main/include/aws/crt/Types.h#L56-L62) instead of decomposition of it.

tests/aws-cpp-sdk-core-tests/smithy/client/schema/JsonShapeDeserializerTest.cpp
```diff
@@ -0,0 +1,466 @@
...
+  EXPECT_EQ(tagValues[0], "x");
+  EXPECT_EQ(tagValues[1], "y");
+  ASSERT_EQ(metaValues.size(), 1u);
+  EXPECT_EQ(metaValues["k1"], "v1");
+  EXPECT_EQ(nestedId, 7);
+}
+
+// Sketch of a generated, schema-driven shape: it owns its Schema, writes its
+// members (push), and populates itself from per-member deserialize callbacks.
+class Widget : public SerializableStruct {
```
sbiscigl: is there a test like this for XML? i didnt see one, and i think theres def value in having the full "fake codegen expierence tests" for each deserializer

src/aws-cpp-sdk-core/source/smithy/client/schema/JsonShapeDeserializer.cpp
```diff
@@ -0,0 +1,477 @@
...
+        return {};
+      }
+      return DateTime(seconds);
+    }
+    if (c == '"') {
+      auto token = ParseString();
+      if (!token.has_value()) {
+        return {};
+      }
+      DateTime parsed(*token, DateFormat::ISO_8601);
```
sbiscigl: for both read and write timestamps on all codecs does hardcoding `ISO_8601` work? i think this would break both in the read and write path where `timestampFormat` is a trait on the shape. is there protocol tests around this?

src/aws-cpp-sdk-core/include/smithy/client/schema/CborShapeDeserializer.h
```diff
@@ -12,30 +12,21 @@ class SMITHY_API CborShapeDeserializer final : public ShapeDeserializer {
...
+
+  Aws::Crt::Optional<bool> ReadBoolean(const Schema& schema) override;
+  Aws::Crt::Optional<int> ReadInteger(const Schema& schema) override;
+  Aws::Crt::Optional<int64_t> ReadLong(const Schema& schema) override;
+  Aws::Crt::Optional<float> ReadFloat(const Schema& schema) override;
+  Aws::Crt::Optional<double> ReadDouble(const Schema& schema) override;
+  Aws::Crt::Optional<Aws::String> ReadString(const Schema& schema) override;
+  Aws::Crt::Optional<Aws::Utils::DateTime> ReadTimestamp(const Schema& schema) override;
+  Aws::Crt::Optional<Aws::Utils::ByteBuffer> ReadBlob(const Schema& schema) override;
+  Aws::Crt::Optional<int> ReadEnum(const Schema& schema) override;
```
sbiscigl: I know this is late in the game, but should `ReadEnum` return a int? also should `ReadEnum` exist? [smithy java doesnt have it](https://github.com/smithy-lang/smithy-java/blob/main/codecs/json-codec/src/main/java/software/amazon/smithy/java/json/smithy/SmithyJsonDeserializer.java). since enums are represented as strings on the wire. shouldnt we keep them as string, and allow enums to interpret that value?

PR #3912, opened 2026-09-01
Migration smithy Part 4

sbaluja:
*Issue #, if available:*

*Description of changes:*
- Streaming members/Event stream generation
- Service level transforms
- QueryParameters 
- Request Specific Headers
- OperationContextParams JmesPath traversal 
- Guard windows macros
- Query compatibility 
- Static AttributeValue class for DynamoDB

*Check all that applies:*

- [x] Linux
- [x] Windows
- [x] Android
- [x] MacOS
- [ ] IOS
- [ ] Other Platforms


tools/code-generation/smithy/cpp-codegen/smithy-cpp-codegen/src/main/java/com/amazonaws/util/awsclientsmithygenerator/generators/model/ShapeClassifier.java
```diff
@@ -212,6 +271,37 @@ private static boolean hasRawStreamingPayload(StructureShape shape, Model model)
...
+        return blobPayloadMemberName(shape.asStructureShape().get(), model).isPresent();
+    }
+
+    /**
+     * Returns the member name of the single {@code @eventPayload} blob member of {@code shape}, or
+     * empty if the shape has no such member. Used by both the classifier predicate and the
+     * blob-payload event renderer so they agree on which member becomes the blob payload.
+     */
+    public static Optional<String> blobPayloadMemberName(StructureShape shape, Model model) {
+        for (MemberShape member : shape.getAllMembers().values()) {
```
sbiscigl: nit: replace with a find first stream

tools/code-generation/smithy/cpp-codegen/smithy-cpp-codegen/src/main/java/com/amazonaws/util/awsclientsmithygenerator/generators/model/ShapeClassifier.java
```diff
@@ -151,26 +163,73 @@ public static ClassifiedShapes classify(Model model, ServiceShape service, Proto
             });
         }
 
+        // Shape ids referenced as a member by any reachable shape (includes list/map element targets).
```
sbiscigl, line 163: nit: these for loops become streams

tools/code-generation/smithy/cpp-codegen/smithy-cpp-codegen/src/main/java/com/amazonaws/util/awsclientsmithygenerator/generators/model/ModelGenerator.java
```diff
@@ -64,11 +66,13 @@ public void generateAll() {
     private List<ShapeRenderer> buildRenderers(ClassifiedShapes classified, RenderContext ctx) {
         List<ShapeRenderer> renderers = new ArrayList<>();
         renderers.add(new EnumShapeRenderer(classified.enums(), ctx));
-        renderers.add(new SubObjectRenderer(classified.subObjects(), ctx));
+        renderers.add(new SubObjectRenderer(classified.subObjects(), classified.resultOutputIds(), ctx));
```
sbiscigl, line 69: if we are decomposing the `SubObjectRenderer` to take two members of the same object seems like it should become `SubObjectRenderer(classified)`

tools/code-generation/smithy/cpp-codegen/smithy-cpp-codegen/src/main/java/com/amazonaws/util/awsclientsmithygenerator/generators/model/renderers/SubObjectRenderer.java
```diff
@@ -12,23 +12,32 @@
...
     private final RenderContext ctx;
 
-    public SubObjectRenderer(List<Shape> subObjects, RenderContext ctx) {
+    public SubObjectRenderer(List<Shape> subObjects, Set<ShapeId> resultOutputIds, RenderContext ctx) {
         this.subObjects = subObjects;
+        this.resultOutputIds = resultOutputIds;
         this.ctx = ctx;
     }
 
+    /** Convenience overload for callers with no dual-role output shapes (e.g. characterization tests). */
```
sbiscigl: ah this branches from my comment earlier, do we need a overloaded constructor? if its only for tests, then it seems like we should remove it

tools/code-generation/smithy/cpp-codegen/smithy-cpp-codegen/src/main/java/com/amazonaws/util/awsclientsmithygenerator/generators/model/transforms/AccessAnalyzerTransforms.java
```diff
@@ -0,0 +1,70 @@
...
+public final class AccessAnalyzerTransforms {
+
+    private AccessAnalyzerTransforms() {}
+
+    public static ModelTransform asTransform() {
+        return AccessAnalyzerTransforms::apply;
+    }
+
+    private static Model apply(Model model, ServiceShape service) {
+        if (!"accessanalyzer".equals(ServiceNameUtil.getSmithyServiceName(service, null))) {
```
sbiscigl: OK so i see whats going on here and its basically like each transform inheriently has a "should run" and a "transform" method on it, and the pipeline evaluates them through. i think it would make sense to formalize the and actually push this idea to the interface of the pipeline i.e.

```java
interface ModelTransform {
    boolean shouldRun(ServiceShape service);
    Model transform(Model model, ServiceShape service);
}

final class AccessAnalyzerTransform implements ModelTransform {
    public boolean shouldRun(ServiceShape service) {
        return "access-analyzer".equals(ServiceNameUtil.getSmithyServiceName(service, null));
    }
    public Model transform(Model model, ServiceShape service) {
        return renameGeneratedPolicyResult(model, service);
    }
}

final class TransformPipeline {
    private final List<ModelTransform> transforms;
    TransformPipeline(List<ModelTransform> transforms) { this.transforms = List.copyOf(transforms); }

    Model apply(Model model, ServiceShape service) {
        Model current = model;
        for (ModelTransform t : transforms) {
            if (t.shouldRun(service)) {
                current = t.transform(current, service);
            }
        }
        return current;
    }
}
```

tools/code-generation/smithy/cpp-codegen/smithy-cpp-codegen/src/main/java/com/amazonaws/util/awsclientsmithygenerator/generators/model/ModelCodegenPlugin.java
```diff
@@ -42,8 +53,18 @@ public void execute(PluginContext context) {
 
         // Build transform pipeline (service-level transforms will be registered here)
         TransformPipeline pipeline = new TransformPipeline(List.of(
-            GlobalTransforms.asTransform()
-            // Future: S3Transforms.asTransform(), Ec2Transforms.asTransform(), etc.
+            GlobalTransforms.asTransform(),
+            SourceRegionTransform.asTransform(),
+            LambdaTransforms.asTransform(),
```
sbiscigl: can some transformers be run in parallel with eachother? and is there a gain for doing as such? `LambdaTransforms`, `SqsTransforms` will only ever run on one model so threading a boolean check is not worth the thread, but is there any cases where several that would take a while or could involve i/o run and can be run in parallel? are there any that need to happen in order and require a graph like pipeline?

tools/code-generation/smithy/cpp-codegen/smithy-cpp-codegen/src/main/java/com/amazonaws/util/awsclientsmithygenerator/generators/model/transforms/AdditionalRequestHeadersTrait.java
```diff
@@ -0,0 +1,52 @@
...
+ * <p>C2J emits these headers from the {@code <Prefix>Request} base class ({@code GetHeaders}) for
+ * ordinary requests, but a streaming request derives from {@code AmazonStreamingWebServiceRequest}
+ * and bypasses that base, so {@code StreamRequestSource.vm} instead emits them inside the request's
+ * own {@code GetRequestSpecificHeaders}. The base class stays C2J-generated, so only the streaming
+ * requests carry this marker; request rendering turns it into the matching {@code headers.insert(...)}
+ * lines. Kept as a data-carrying marker + generic renderer rule so the renderer stays
+ * service-agnostic.
+ */
+public final class AdditionalRequestHeadersTrait extends AbstractTrait {
+    public static final ShapeId ID = ShapeId.from("aws.cpp.internal#additionalRequestHeaders");
```
sbiscigl, line 24: so here we are creating traits internal to us.  the transforms and renderers are tied together through these synthetic aws.cpp.internal#* traits. im fine with traits being the transport, thats the normal smithy way, but the contract between the two sides is completely implicit and nothing enforces it.

we could actually register them as real smithy traits. right now they have no Provider and no TraitService entry so theyre not first class and wouldnt survive a model round trip. we could give each one a Provider and register it through the SPI.:

```
// in TopLevelHostIdTrait
public static final class Provider extends AnnotationTrait.Provider<TopLevelHostIdTrait> {
    public Provider() { super(ID, node -> new TopLevelHostIdTrait()); }
}
```

```
# src/main/resources/META-INF/services/software.amazon.smithy.model.traits.TraitService
com.amazonaws...transforms.TopLevelHostIdTrait$Provider
com.amazonaws...transforms.ChecksumMemberTrait$Provider
# one per internal trait
```

[kotlin does something very similar to this](https://github.com/smithy-lang/smithy-kotlin/blob/main/codegen/codegen/src/main/resources/META-INF/services/software.amazon.smithy.model.traits.TraitService) for their own [custom traits](https://github.com/smithy-lang/smithy-kotlin/blob/main/codegen/codegen/src/main/kotlin/aws/smithy/kotlin/codegen/model/traits/PaginationEndBehaviorTrait.kt).

tools/code-generation/smithy/cpp-codegen/smithy-cpp-codegen/src/main/java/com/amazonaws/util/awsclientsmithygenerator/generators/model/ModelCodegenPlugin.java
sbaluja: they can, but there's no real benefit to it. they're all pure in-memory model rewrites, no i/o, and each is gated to a single service so only one really does work per run. also, because of that there is no specific ordering to these transforms.

tools/code-generation/smithy/cpp-codegen/smithy-cpp-codegen/src/main/java/com/amazonaws/util/awsclientsmithygenerator/generators/model/transforms/AccessAnalyzerTransforms.java
sbaluja: good refactor suggestion, done

tools/code-generation/smithy/cpp-codegen/smithy-cpp-codegen/src/main/java/com/amazonaws/util/awsclientsmithygenerator/generators/model/renderers/SubObjectRenderer.java
sbaluja: removed it, they were only for tests but now tests just pass an empty list

tools/code-generation/smithy/cpp-codegen/smithy-cpp-codegen/src/main/java/com/amazonaws/util/awsclientsmithygenerator/generators/model/transforms/AdditionalRequestHeadersTrait.java
sbaluja: yeah agreed these would traits would not survive a round-trip but the round trip never happens here since we stamp them in the transform and read them in the renderer within the same in-memory pass in `ModelCodegenPlugin`, nothing ever serializes the model back out and re-assembles it. so there's no provider lookup to fail. the traits ride along as plain objects and Model.builder() doesn't validate applied-trait defs. however, if you feel strongly about registering them to enforce the contract + be round-trip safe, i dont mind but it will grow the PR

sbiscigl (approved):

PR #3916, opened 2026-09-02
create async http api, implement in crt

sbiscigl:
*Description of changes:*

Add a new API in the HTTP client called `MakeRequestAsync` that will return to the caller immediately after the submit to the transport layer. this will unblock clients from "true async" where the SDK thread pool builds requests in parallel however the backing CRT event loop will operate I/O operations in the SDK.

*Check all that applies:*

- [x] Linux
- [x] Windows
- [x] Android
- [x] MacOS
- [ ] IOS
- [ ] Other Platforms

### DIFF
--- a/src/aws-cpp-sdk-core/source/http/crt/CRTHttpClient.cpp
+++ b/src/aws-cpp-sdk-core/source/http/crt/CRTHttpClient.cpp
@@ -9,43 +9,18 @@
 #include <aws/core/utils/crypto/Hash.h>
 #include <aws/core/utils/logging/LogMacros.h>
 #include <aws/core/utils/ratelimiter/RateLimiterInterface.h>
+#include <aws/core/utils/threading/Semaphore.h>
+#include <aws/core/utils/threading/WaitGroup.h>
 #include <aws/crt/http/HttpConnectionManager.h>
 #include <aws/crt/http/HttpRequestResponse.h>
 
+#include <atomic>
+#include <mutex>
 #include <utility>
 
 static const char *const CRT_HTTP_CLIENT_TAG = "CRTHttpClient";
 
 namespace {
-// Just a wrapper around a Condition Variable and a mutex, which handles wait and timed waits while protecting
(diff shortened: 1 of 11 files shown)


src/aws-cpp-sdk-core/include/aws/core/utils/threading/internal/WaitGroup.h
```diff
@@ -0,0 +1,30 @@
...
+
+#include <aws/core/Core_EXPORTS.h>
+#include <aws/core/utils/memory/AWSMemory.h>
+
+#include <cstddef>
+
+namespace Aws {
+namespace Utils {
+namespace Threading {
+class AWS_CORE_API WaitGroup {
```
sbaluja: this is internal only so we should use `AWS_CORE_LOCAL`

src/aws-cpp-sdk-core/source/http/crt/CRTHttpClient.cpp
```diff
@@ -425,124 +452,86 @@ namespace Aws
...
+                if (!sync->signal.WaitOneFor(m_configuration.requestTimeoutMs))
                 {
-                    // close the connection if it's still there so we can expedite anything we're waiting on.
-                    if (connectionRef)
-                    {
-                        connectionRef->Close();
-                    }
+                    auto response = Aws::MakeShared<Standard::StandardHttpResponse>(CRT_HTTP_CLIENT_TAG, request);
+                    response->SetClientErrorType(Aws::Client::CoreErrors::REQUEST_TIMEOUT);
+                    response->SetClientErrorMessage("Request Timeout Has Expired");
```
sbaluja, line 552: For the timeout scenario, i think we aren't closing the connection, which we were doing before. this means the stream wouldn't complete, and the destructor's `m_requestLatch->Wait()` blocks indefinitely. Should we hold onto the connection/stream ref in the request state to close it on timeout?

sbaluja (approved):

PR #3921, opened 2026-09-08
JSON number parsing bug fix

pulimsr:
Fix JSON number parsing edge cases in aws-cpp-sdk-core                                                                                                                                         

  - Serialize large integer literals without the fixed-size buffer limit.
  - Saturate out-of-range double→int64 conversions instead of undefined behavior.
  - Decode scientific-notation integers correctly.

*Check all that applies:*

- [x] Linux
- [x] Windows
- [ ] Android
- [x] MacOS
- [ ] IOS
- [ ] Other Platforms


src/aws-cpp-sdk-core/source/utils/json/JsonSerializer.cpp
```diff
@@ -10,6 +10,48 @@
 #include <aws/core/utils/memory/stl/AWSStringStream.h>
 #include <aws/core/utils/StringUtils.h>
 #include <aws/core/utils/Document.h>
+#include <aws/core/utils/logging/LogMacros.h>
+#include <cstdlib>
+#include <limits>
+
+namespace
+{
+    const double LLONG_MIN_AS_DOUBLE = -9223372036854775808.0;
```
sbiscigl: lets no use actual hard coded numbers, those are fragile, use numeric limits where you can
```c++
constexpr double LLONG_MIN_AS_DOUBLE =
    static_cast<double>(std::numeric_limits<int64_t>::min());
constexpr double LLONG_MAX_PLUS_ONE =
    -static_cast<double>(std::numeric_limits<int64_t>::min());
```

src/aws-cpp-sdk-core/source/utils/Document.cpp
```diff
@@ -10,6 +10,48 @@
 #include <aws/core/utils/memory/stl/AWSStringStream.h>
 #include <aws/core/utils/StringUtils.h>
 #include <aws/core/utils/json/JsonSerializer.h>
+#include <aws/core/utils/logging/LogMacros.h>
+#include <cstdlib>
+#include <limits>
+
+namespace
+{
+    const double LLONG_MIN_AS_DOUBLE = -9223372036854775808.0;
```
sbiscigl: same about using numerical limits

src/aws-cpp-sdk-core/source/utils/Document.cpp
```diff
@@ -10,6 +10,48 @@
...
+    const double LLONG_MAX_PLUS_ONE = 9223372036854775808.0;
+
+    bool IsRepresentableAsInt64(double value)
```
sbiscigl: these functions appear copy and pasted twice lets move them into a header that is not installed/part of the public API. somewhere in `aws/core/utils`, maybe under a new directory that is build alongside the SDK but not installed. numeric utils or something of the like

src/aws-cpp-sdk-core/source/utils/json/JsonSerializer.cpp
pulimsr: Done — now derived from `std::numeric_limits<int64_t>` instead of hard-coded literals.

src/aws-cpp-sdk-core/source/utils/Document.cpp
pulimsr: Moved into a shared header at `aws/core/utils/numeric/NumericUtils.h`

sbiscigl (approved):

PR #3922, opened 2026-09-11
UB: Never expiring creds bug

okruitho:
*Issue #, if available:*

*Description of changes:*
Fix non-expiring CRT credentials being incorrectly reported as expired by `CrtCredentialsProvider`.

CRT [documents `UINT64_MAX` as “no expiration”](https://github.com/awslabs/aws-crt-cpp/blob/851d8d003c9d5150edab56807e2393013f3771de/include/aws/crt/auth/Credentials.h#L77-L80). Previously, the provider converted this sentinel into a `DateTime` as though it were an ordinary timestamp, exceeding the representable range and causing `IsExpired()` and `IsExpiredOrEmpty()` to return `true`.

For gcc this works fine; when using clang the casts from `uint64_t` to `double` to `uint64_t` give undefined behaviour, depending on compliler optimisation.

## Downstream impact

We encountered this in DuckDB’s AWS extension, which [uses this check before copying credentials into an S3 secret](https://github.com/duckdb/duckdb-aws/blob/0a30cdb5ab6bf8cc6348487a80a24b465f674989/src/aws_secret.cpp#L491). Our toolchain is clang based. The incorrect expiration status caused valid credentials to be omitted, surfacing in its [MinIO integration test](https://github.com/duckdb/duckdb-aws/blob/0a30cdb5ab6bf8cc6348487a80a24b465f674989/test/sql/aws_minio_secret.test#L46).

## Changes

- Preserve `AWSCredentials`’ default non-expiring value when CRT returns `UINT64_MAX`.
- Use the `uint64_t` seconds constructor for finite expiration timestamps, avoiding the floating-point conversion.
- Add regression coverage for non-expiring credentials, ensuring all three state checks return `false` and subsequent retrieval uses cached credentials.
- Add coverage confirming finite expiration timestamps are preserved.

*Check all that applies:*

- [ ] Linux
- [ ] Windows
- [ ] Android
- [x] MacOS
- [ ] IOS
- [ ] Other Platforms


src/aws-cpp-sdk-core/source/auth/CrtCredentialsProvider.cpp
```diff
@@ -99,8 +100,11 @@ AWSCredentials CrtCredentialsProvider::ExtractCredentialsFromCrt(const Aws::Crt:
   credentials.SetAWSAccessKeyId({reinterpret_cast<char*>(accountIdCursor.ptr), accountIdCursor.len});
   const auto secretKeyCursor = crtCredentials.GetSecretAccessKey();
   credentials.SetAWSSecretKey({reinterpret_cast<char*>(secretKeyCursor.ptr), secretKeyCursor.len});
-  const auto expiration = crtCredentials.GetExpirationTimepointInSeconds();
-  credentials.SetExpiration(DateTime{static_cast<double>(expiration)});
```
sbiscigl, line 103: so looking at the `DateTime` class

```c++
/**
   * Initializes time point to epoch time in seconds with a millis mantissa,
   *
   * i.e. 1.1 would be 1100 milliseconds
   */
DateTime(double secondsSinceEpoch);

/**
   * Initializes time point to epoch time in seconds
   */
DateTime(uint64_t secondsSinceEpoch);
```

since `GetExpirationTimepointInSeconds` returns a `uint64_t` is the double cast even needed? could have ~~just been a oversight in the initial implementation~~. removing the cast seems like a better idea than comparing numeric limits.

looking at the original implementation that was written at roughly the same time as the `uint64_t` constructor was added. so that timeline likely tangled up this issue. if you update to use the `uint64_t` constructor, i see no issue in merging this.

okruitho: Unfortunately, when assigning an `MAX ULONG` to an datetime it [also overflows](https://godbolt.org/#g:!((g:!((g:!((h:codeEditor,i:(filename:'1',fontScale:14,fontUsePx:'0',j:1,lang:c%2B%2B,selection:(endColumn:2,endLineNumber:11,positionColumn:2,positionLineNumber:11,selectionStartColumn:2,selectionStartLineNumber:11,startColumn:2,startLineNumber:11),source:'%23include+%3Ccstdint%3E%0A%23include+%3Ciostream%3E%0A%23include+%3Cchrono%3E%0A//+Type+your+code+here,+or+load+an+example.%0Aint+main()+%7B%0A++++std::chrono::duration%3Cuint64_t,+std::chrono::seconds::period%3E+timestamp((std::numeric_limits%3Cuint64_t%3E::max)())%3B%0A++++auto+m_time+%3D+std::chrono::system_clock::time_point(timestamp)%3B%0A++++std::cout+%3C%3C+m_time+%3C%3C+std::endl%3B%0A%0A++++return+1%3B%0A%7D'),l:'5',n:'0',o:'C%2B%2B+source+%231',t:'0')),k:33.333333333333336,l:'4',n:'0',o:'',s:0,t:'0'),(g:!((h:compiler,i:(compiler:g162,filters:(b:'0',binary:'1',binaryObject:'1',commentOnly:'0',debugCalls:'1',demangle:'0',directives:'0',execute:'0',intel:'0',libraryCode:'0',trim:'1',verboseDemangling:'0'),flagsViewOpen:'1',fontScale:14,fontUsePx:'0',j:1,lang:c%2B%2B,libs:!(),options:'',overrides:!(),selection:(endColumn:1,endLineNumber:1,positionColumn:1,positionLineNumber:1,selectionStartColumn:1,selectionStartLineNumber:1,startColumn:1,startLineNumber:1),source:1),l:'5',n:'0',o:'+x86-64+gcc+16.2+(Editor+%231)',t:'0')),k:33.333333333333336,l:'4',n:'0',o:'',s:0,t:'0'),(g:!((h:output,i:(compilerName:'x86-64+gcc+16.2',editorid:1,fontScale:14,fontUsePx:'0',j:1,wrap:'1'),l:'5',n:'0',o:'Output+of+x86-64+gcc+16.2+(Compiler+%231)',t:'0')),k:33.33333333333333,l:'4',n:'0',o:'',s:0,t:'0')),l:'2',n:'0',o:'',t:'0')),version:4).. 

With regards to the code, this is done [here](https://github.com/aws/aws-sdk-cpp/blob/29b4b86c819cc92270247c59bb5e0a0ad7cd0748/src/aws-cpp-sdk-core/source/utils/DateTimeCommon.cpp#L1143). Therefore, I don't see a path to fix this without overhauling more code.

sbiscigl: so from my understanding `std::chrono::system_clock::time_point` is only meant to hold at max `std::chrono::system_clock::duration:::max()`. which on platforms, as you mention is lower than `std::numeric_limits<uint64_t>::max` and gives you this overflow. We should fix this in the datetime class and not at the call site. the call site should call with a `uint64_t` no casting, then in the datetime constructor for `uint64_t` we should floor it with maximum representable time to closes represent what we are doing. creating a date time class with the longest expiration possible given the platform.

so we should update the call site to use `uint64_t` then update the constructor to do something like 

```c++
DateTime::DateTime(uint64_t secondsSinceEpoch) : m_valid(true)
{
  using sys_dur = std::chrono::system_clock::duration;
  auto max_secs = std::chrono::duration_cast<std::chrono::seconds>(sys_dur::max()).count();
  std::chrono::seconds secs{ std::min<uint64_t>(secondsSinceEpoch, max_secs) };
  m_time = std::chrono::system_clock::time_point(secs);
}
```

constructing a `DateTime` with overflowing `uint64_t` is something that we likely dont want to do, and want to avoid at construction.

okruitho: I applied your suggestion 👍

sbiscigl: thank you so much, for CI and procedure reasons i need to pull this PR onto a branch owned by the team. will merge it from there, your commit will still credit you, thanks for contributing!

sbiscigl:
Created a PR with your commit we intend on merging https://github.com/aws/aws-sdk-cpp/pull/3924

sbiscigl:
closing merged to main now and should be released under tag `1.11.896` later today

PR #3930, opened 2026-09-17
[Feature] Add shared credential-refresh lifecycle to AWSCredentialsProvider

kai-ion:
*Issue #, if available:*

*Description of changes:*
Add shared credential-refresh lifecycle to AWSCredentialsProvider

*Check all that applies:*

- [x] Linux
- [x] Windows
- [x] Android
- [x] MacOS
- [ ] IOS
- [ ] Other Platforms


src/aws-cpp-sdk-core/include/aws/core/auth/AWSCredentialsProvider.h
```diff
@@ -58,31 +60,51 @@ namespace Aws
...
              *  to aid your implementation of GetAWSCredentials.
              */
             virtual bool IsTimeToRefresh(long reloadFrequency);
             virtual void Reload();
+
+            /**
+             * One attempt against the credential source, classified fresh/recoverable/non-recoverable.
+             * Providers on the refresh lifecycle override this; the default returns a recoverable failure.
+             */
+            virtual Aws::Auth::RefreshResult<AWSCredentials> FetchCredentialsFromSource();
```
sbiscigl, line 89: so `FetchCredentialsFromSource` is a protected virtual function on `AWSCredentialsProvider`. so a inheriting class can override it. but `FetchCredentialsFromSource` is not used in the `GetAWSCredentials` as that is the method the SDK calls to fetch credentials. so how is the override actually planned on being used?

src/aws-cpp-sdk-core/source/auth/AWSCredentialsProvider.cpp
```diff
@@ -65,6 +78,46 @@ bool AWSCredentialsProvider::IsTimeToRefresh(long reloadFrequency)
     return false;
 }
 
+AWSCredentialsProvider::AWSCredentialsProvider()
+    : m_lastLoadedMs(0),
+      m_refreshState(new Aws::Internal::CredentialRefreshStateImpl(
+          [this]() { return FetchCredentialsFromSource(); },
```
sbiscigl, line 84: you've created a side-car crendeitals provider in the crednetials provider, so you're packaging two things in one, and its the incorrect layering.

you should have a caching credentials provider over the existing credentials provider that executes the caching
```c++
class CachingCredentialsProvider: public CredentialsProvider {
   CachingCredentialsProvider(CredentialsProvider* credentialsProvider) {
       ....
   }

   GetCredentials() {
      if (cache) { return cached_creds}
      cache(delegated_creds_provider->GetCreds)
   }
}
```

src/aws-cpp-sdk-core/include/aws/core/auth/CredentialsRefreshProvider.h
```diff
@@ -0,0 +1,57 @@
...
+        /**
+         * Wraps a CredentialsSource with the shared credentials-refresh lifecycle (caching,
+         * advisory/mandatory windows, jittered backoff, serve-last-good, single in-flight refresh,
+         * invalidation). The wrap is applied only when AWS_NEW_CREDENTIAL_REFRESH_2026 is on; when
+         * off the source provider is used directly, preserving legacy behavior.
+         */
+        class AWS_CORE_API CredentialsRefreshProvider : public AWSCredentialsProvider
+        {
+        public:
+            explicit CredentialsRefreshProvider(std::shared_ptr<CredentialsSource> source);
```
sbiscigl: so lets think about this constructor and the new interface `CredentialsSource`. credentials refresh provider is meant to be like other SDKs where we wrap a existing credentials provider and cache that. there are however several issues with this

_CredentialsSource is not a AWSCredentialsProvider_

in the default chain we do something like 
```
class chain(): public AWSCredentialsProvider {
public:
  chain() {
     providers.push_back(std::make_shared<SomeCredentialsProvider>);
   }
private:
  std::vector<AWSCredentialsProvider> providers
}
```

now all of our implementations of AWSCredentialsProvider specifically in this example SomeCredentialsProvider how have to implement these functions. and implement a new interface. im not strictly against this we just need to weigh whether we want to do this or not.

this approach has two problems:

__double caching/no caching__

what about wrapping credentials provider that already cache like all of the CRT credentials providers. you are more or less unwrapping those, and making a parallel caching API which is a lot of surface area. additionally if you remove the caching layer, existing users of the credentials provider now have no caching.

__direct usage of the crednetials provider has no caching_

directly constructing a credentials provider has no caching necessarily because only in the chain we wrap it.


before we do this, we need to evaluate how and where this caching will be added, and answer these questions.

Issue #909, opened 2018-07-06
withFileOffset and withFile options not present in UploadPartRequest of multipart upload in c++ sdk.


sdjunaidali:
AWS SDK for C++

### What platform/OS are you using?
Ubuntu 14.04

In multi part upload request there is no options of  withFileOffset and withFile options in UploadPartRequest of multipart upload in **C++ sdk of AWS S3** and it's present in JAVA sdk, 

**So how to split parts using C++ sdk without using the withFileOffset and withFile options ?**

singku:
I'd suggest you to use transfer manager to manage file upload/download for you.

shuzongran:
Aws::String str(data.data(), size);
auto&& input_data = Aws::MakeShared<Aws::StringStream>("UploadPartStream", str);

通过自行读取文件内容的方式可以完成分片上传,保存在buffer里转换成string就行

PR #3931, opened 2026-09-18
Wire async I/O path into AWSClient and SmithyClient for the CRT HTTP client

sbiscigl:
*Description of changes:*

Adds an opt-in asynchronous I/O path for the CRT-backed HTTP client and wires it through both client bases.

This change adds the asynchronous I/O request path for the CRT HTTP client, serving as the entry point that a future code generation change will build on to make service operations non-blocking. The path is controlled by a new CMake option that is off by default, so nothing changes for existing customers until it is explicitly enabled. Adding this plumbing now lets the follow-up code generation work wire services onto it without further changes to the client core.

*Check all that applies:*
- [x] Checked if this PR is a breaking (APIs have been changed) change. Not breaking: new opt-in API and CMake flag; default-off path is unchanged.

- [x] Linux
- [x] Windows
- [x] Android
- [x] MacOS
- [ ] IOS
- [ ] Other Platforms


src/aws-cpp-sdk-core/include/aws/core/client/AWSClient.h
```diff
@@ -224,6 +237,15 @@ namespace Aws
                                                     const char* signerRegionOverride = nullptr,
                                                     const char* signerServiceNameOverride = nullptr) const;
 
+            void AttemptExhaustivelyAsync(const Aws::Http::URI& uri,
```
sbaluja, line 240: Shouldn't this be conditionally declared? Its defintion is gated behind `#if defined(AWS_CRT_HTTP_USE_ASYNC_IO)`

src/aws-cpp-sdk-core/source/client/AWSClient.cpp
```diff
@@ -142,7 +178,8 @@ AWSClient::AWSClient(const Aws::Client::ClientConfiguration& configuration,
         configuration.awsChunkedBufferSize),
         m_userAgentInterceptor},
     m_enableNewRetries{Aws::Utils::StringUtils::ToLower(Aws::Environment::GetEnv("AWS_NEW_RETRIES_2026").c_str()) == "true"},
-    m_disableExpectHeader(configuration.disableExpectHeader)
+    m_disableExpectHeader(configuration.disableExpectHeader),
+    m_executor(configuration.executor ? configuration.executor : configuration.configFactories.executorCreateFn())
```
sbaluja, line 182: nit: we're kinda eagerly constructing this bc its only read through the async path

src/aws-cpp-sdk-core/source/http/crt/CRTHttpClient.cpp
sbaluja, line 669: I think we are setting up a double timeout on the sync path? We're waiting here, but the `SendOneAttemptAsync` also has this exact timeout mechanism in the `startSend` bit

src/aws-cpp-sdk-core/source/http/crt/CRTHttpClient.cpp
```diff
@@ -494,16 +506,134 @@ namespace Aws
...
+                                return;
+                            }
+                            if (auto connection = state->connection.lock())
+                            {
+                                connection->Close();
+                            }
+                            state->CompleteWithError(Aws::Client::CoreErrors::REQUEST_TIMEOUT, "Request Timeout Has Expired");
+                        },
+                        std::chrono::milliseconds(m_configuration.requestTimeoutMs));
+                }
```
sbaluja, line 533: one thing i wish we could do here is cancel the timeout task we're scheduling on the event loop. in succesful cases, this task stays on there for the full requestTimeoutms before it fires and no-ops. under load i could see this potentially reducing our throughput, since there would be a steady backlog of these tasks on the event loop per request. I see there's a [`aws_event_loop_cancel_task`](https://github.com/awslabs/aws-c-io/blob/54861ea52838a0d2bfc9c85288ffba4bf6cf05e4/include/aws/io/event_loop.h#L153), which we didnt bind out in [pr #909](https://github.com/awslabs/aws-crt-cpp/pull/909). Not sure if this is the right api, but might be worth looking into

src/aws-cpp-sdk-core/source/client/AWSClient.cpp
```diff
@@ -698,6 +736,320 @@ HttpResponseOutcome AWSClient::AttemptOneRequest(const std::shared_ptr<Aws::Http
...
+    context->httpRequest->SetEventStreamRequest(request.IsEventStreamRequest());
+    context->httpRequest->SetHasEventStreamResponse(request.HasEventStreamResponse());
+
+    context->attemptSentTime = DateTime::Now();
+    context->httpRequest->SetSigningTimestampOverride(context->attemptSentTime + context->attemptSkew);
+
+    const char* signerRegion = context->signerRegion ? context->signerRegion->c_str() : nullptr;
+    const char* signerServiceName = context->signerServiceName ? context->signerServiceName->c_str() : nullptr;
+
+    BuildHttpRequest(request, context->httpRequest);
```
sbaluja, line 822: missing the telemetry call here we do for the sync path

src/aws-cpp-sdk-core/source/smithy/client/AwsSmithyClientBase.cpp
```diff
@@ -659,11 +682,6 @@ void AwsSmithyClientBase::HandleAsyncReply(std::shared_ptr<AwsSmithyClientAsyncR
 
         if (!retryWithCorrectRegion && !m_clientConfig->retryStrategy->ShouldRetry(outcome.GetError(), static_cast<long>(pRequestCtx->m_retryCount)))
         {
-            const bool blockedByQuotaOnly = outcome.GetError().ShouldRetry() &&
```
kai-ion, line 662: Curious what happens to this removed logic. When a long-polling operation gives up because it's out of retry tokens, the old code slept before returning so the caller wouldn't immediately retry. I dont see a replacement logic, how does the new model handle long-polling ops that ran out of tokens?

PR #3932, opened 2026-09-21
fix operations with optional auth

sbiscigl:
*Description of changes:*

(https://github.com/aws/aws-sdk-cpp/pull/3751), merged 2026-03-25 and shipped in 1.11.777, consolidated per-operation MakeRequest calls into a shared InvokeServiceOperation helper that hardcoded SIGV4_SIGNER, which wrongly SigV4-signed operations that must be unsigned or bearer and broke IdC browser auth. along with sso, sts, cognito-identity, cognito-idp, signin, and codecatalyst. The fix threads the model-derived signerName.

*Check all that applies:*

- [x] Linux
- [x] Windows
- [x] Android
- [x] MacOS
- [ ] IOS
- [ ] Other Platforms


pulimsr (approved):

tools/code-generation/generator/src/main/resources/com/amazonaws/util/awsclientgenerator/velocity/cpp/common/ServiceInvokeOperationMethod.vm
```diff
@@ -1,6 +1,11 @@
 #set($emitGenericOperation = true)
 #parse("com/amazonaws/util/awsclientgenerator/velocity/cpp/common/ServiceClientOperationRequestRequiredMemberValidate.vm")
 #set($emitGenericOperation = false)
+#if($operation.signerName == "Aws::Auth::SIGV4_SIGNER")
+#set($signerArg = "")
+#else
+#set($signerArg = ", ${operation.signerName}")
```
kai-ion, line 7: is there a chance operation.signerName has a value of " "?

If so I think we can add an extra guard for empty value

sbiscigl: >If so I think we can add an extra guard for empty value

what do you mean extra guard? if singer is missing is in a invalid state. if signer is missing, i.e. empty value, compilation should fail, and it will because `, )` is not valid code?

the same could be said about any string variable in codegen. they all could be empty, and all of them result in compilation errors before the SDK is published

sbaluja (approved):

kai-ion (approved):

PR #3934, opened 2026-09-21
Preserving millisecond precision in DateTime parser

averyk22:
*Issue #, if available:*

*Description of changes:*
Added millisecond precision to ISO_8601, ISO_8601 Basic Date Parser, and the AutoDetect path.

*Check all that applies:*

- [ ] Linux
- [ ] Windows
- [ ] Android
- [x] MacOS
- [ ] IOS
- [ ] Other Platforms


src/aws-cpp-sdk-core/source/utils/DateTimeCommon.cpp
```diff
@@ -1529,6 +1535,7 @@ void DateTime::ConvertTimestampStringToTimePoint(const char* timestamp, DateForm
         if (IsSecondsSinceEpochRepresentable(tt))
         {
             m_time = std::chrono::system_clock::from_time_t(tt);
+            m_time += std::chrono::milliseconds(milliseconds);
```
kai-ion, line 1538: Nit/minor: Time is assigned by comparing to how long it has been since 1970, an edgecase of date before 1970 would render a negative number. if m_time is negative, % 1000 on a negative number stays negative.

Fix: floor when splitting into seconds + ms so the fraction is always 0–999

src/aws-cpp-sdk-core/source/utils/DateTimeCommon.cpp
```diff
@@ -405,6 +406,7 @@ class DateParser
     bool m_utcAssumed;
     // The size should be at least one byte greater than the maximum possible size so that we could use the last char to indicate the end of the string.
     char m_tz[7];
+    int m_milliseconds;
```
sbiscigl: nit: i know its inited to zero in the constructor however just make this `int m_milliseconds{0};` to avoid issues in the the future. `int x` [does NOT guarantee x will be 0](https://godbolt.org/z/6e7vcoMz6). the constructor value make sure now, but if another constructor is added, this will be a easy miss.

tests/aws-cpp-sdk-core-tests/utils/DateTimeTest.cpp
```diff
@@ -264,6 +264,20 @@ TEST_F(DateTimeTest, TestISO_8601ParsingNSPrecision)
     ASSERT_EQ(gmtDate, DateTime(utcOffsetVersion, DateFormat::ISO_8601));
 }
 
+TEST_F(DateTimeTest, TestISO_8601ParsingPreservesMilliseconds)
+{
+    DateTime gmtDate("2026-09-15T18:00:00.016Z", DateFormat::ISO_8601);
+    ASSERT_TRUE(gmtDate.WasParseSuccessful());
+    ASSERT_EQ(16, gmtDate.Millis() % 1000);
```
sbiscigl, line 271: ~~shouldn't `gmtDate.Millis()` be 16? and not `gmtDate.Millis() % 1000`? `2026-09-15T18:00:00.016Z` means that theres `.016` seconds on the time meaning 16 miliseconds. shouldnt we not need the modulo operation on it? let me know if im mistaken~~ Millis in this context is epoch millis not millis of datetime

src/aws-cpp-sdk-core/source/utils/DateTimeCommon.cpp
sbiscigl: not sure if this is a valid case, can you create a string such that a negative value for a millisecond exists? the state machine should take care of that.

sbiscigl (approved):

